PR #23, opened 2020-10-30
Fix C macros for arginfo parameter names

TysonAndre:
Previously, the value of ReflectionParameter->getName() were of the form
`'"parameter"'`, not `'parameter'`

Aside, out of scope: PHP 8 introduces named arguments. There's snake_case elsewhere - maybe function_name should be used instead of functionname (would also need to update docs)

### DIFF
--- a/agent/php_newrelic.c
+++ b/agent/php_newrelic.c
@@ -115,89 +115,89 @@ void nr_print_globals(FILE* fp) {
  * New Relic API function argument descriptors.
  */
 ZEND_BEGIN_ARG_INFO_EX(newrelic_add_custom_parameter_arginfo, 0, 0, 2)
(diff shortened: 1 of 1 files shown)


pyohannes:
ok jenkins

pyohannes (approved):
The second argument of `ZEND_ARG_INFO` is always used with the `#` stringizing operator, so no need for passing it as a string.

We'll merge this soon. Thanks @TysonAndre for contributing to the PHP agent.

zsistla:
ok jenkins

kneitinger:
ok jenkins

joshuabenuck:
ok jenkins

joshuabenuck (approved):

PR #27, opened 2020-11-13
Update vendor/protobuf to v3.13.0 for PHP 8 compilation compatibility

kneitinger:
In preparation for PHP 8, or more specifically, compilation of the agent on systems that only have PHP 8 libs/binaries, vendored `protobuf` dependency was updated to a version that accounts for those API changes.

This largely does not affect the agent, as we do not use the PHP protobuf extension, but for maintainability purposes, we do not heavily modify the `protobuf` directory's build system (which compiles language specific comilers ~by default~ in all cases)

Per `vendor/README.md` we remove the `benchmark/go` and `examples` directories.

Compilation tested on Ubuntu 20.04 + PHP 7.4 & 8.0, and Ubuntu 16.04 + PHP 5.6-7.3


kneitinger:
PR pending upcoming continuation of PHP 8 support work.

kneitinger:
ok jenkins

kneitinger:
Fixes #43

kneitinger:
ok jenkins

joshuabenuck (approved):

PR #28, opened 2020-11-17
PHP-2672: Reconnect on FailedPrecondition

pyohannes:
This enables the daemon to attempt a reconnect on 8T cell shifts.

How it works:
1. The `spanBatchSender` interface is extended by a `clone` method. This method returns a clone of the sender (duplicating the underlying configuration). The clone establishes its own TCP connection (which also includes resolving the host name).
2. There's an additional `statusReconnect` constant, signifying that a TCP reconnect is needed.
3. The `FailedPrecondition` gRPC return code is translated into a `statusReconnect` status.
4. In the main worker loop, if a `statusReconnect` status is received, the store sender is shut down, cloned, and replaced with its clone.

### DIFF
--- a/src/newrelic/infinite_tracing/grpc_sender.go
+++ b/src/newrelic/infinite_tracing/grpc_sender.go
@@ -102,6 +102,10 @@ func newGrpcSpanBatchSender(cfg *Config) (*grpcSpanBatchSender, error) {
 	}, nil
 }
 
+func (s *grpcSpanBatchSender) clone() (spanBatchSender, error) {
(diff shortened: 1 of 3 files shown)


pyohannes:
I also ran tests with a modified version of kitchen sink (which returns a `FailedPrecondition`). The was correctly cloned and the connection re-established.

pyohannes:
ok jenkins

pyohannes:
ok jenkins

joshuabenuck (approved):

Fahmy-Mohammed (approved):

PR #49, opened 2020-12-09
Modify agent to conform with new PHP 8.0 internal API changes

### LINKED ISSUE #46: Modify agent to conform with new PHP 8.0 internal API changes
## Summary
PHP 8.0 made quite a few [internal API changes](https://github.com/php/php-src/blob/PHP-8.0.0/UPGRADING.INTERNALS).  The PHP agent needs to accommodate these changes.

## Desired Behavior
PHP agent conforms to PHP 8.0 internal API changes.

## Additional context
See associated [MMF](https://docs.google.com/document/d/1iolwPOyr1AcOmUDqzf5ZQZ4ODm1m593onSRhFP-UhiY/edit#heading=h.uxxbook5v69l) for more details.

zsistla:
**Items addressed:**
1)  Added defines for PHP8
2) Added TSRMLS_* macros (these have been voided out in `php-src` since PHP7.0, but were officially removed in PHP8.0)
3) The Object Handlers API was changed to receive zend_object* instead of zval* and zend_string* instead of zval* for property names.  Modified agent to toggle between zval and zend_object depending on PHP8 or not.
4) The zend_fcall_info no_separation flag has been removed, and separation is never allowed. If you wish to pass (or allow passing) arguments by reference, explicitly create those arguments as references using ZEND_MAKE_REF. This function is passed as a pointer in the agent here. This removal also affects call_user_function_ex() (example in agent), which should be replaced by call_user_function().
5) Error notifications have a new API and the callback signature has changed. The current changes do not register a callback.
6) `display_disabled_function` has been removed from (and therefore `zif_display_disabled_function`) `zend_API.c.`. In PHP 8, a disabled function is indistinguishable from a function that does not exist.
7) Added `newrelic_arginfo_void` as `PHP_FE` no longer allows `0` as an argument.
8) Necessary changes to allow agent tests to compile.
9) Take into account new `libphp.so` and `libphp.a` naming scheme.
10) With PHP8, `call_user_function_ex` was removed and `call_user_function`
   * became the recommended function; however, the call handles param_values
   * differently. Prior to PHP8, we used the no_separation flag set to 0 which
   * then did some internal manipulation of param_values.  There is no longer a
   * no separation flag, so that internal manipulation is no longer done and we
   * need to do it prior to sending param_values or `call_user_function` will
   * cause a segfault.
11) `zend_get_resource_handle` now takes a different type parameter.

**Remaining items:**
1) Regarding changing `call_user_function_ex` call to the `call_user_function` call in the agent function `nr_php_call_user_func`:
With `call_user_function_ex`, we used to use the `no_separation flag = 0`, which meant PHP itself would [internally](https://github.com/php/php-src/blob/PHP-7.3.25/Zend/zend_execute_API.c#L708) manipulate ours args appropriately.

With `call_user_function` the logic has changed here:
https://github.com/php/php-src/blob/master/Zend/zend_execute_API.c#L770

Currently any UNIT TEST call to `nr_php_call` or a direct call to `nr_php_call_user_func` that has parameters will choke because the internal handling no longer wraps them as proper zval refs.

Before we pass the params to PHP function `call_user_function`, we probably need to do something like:
```
if (!Z_ISREF_P(arg)) {
 nrl_verbosedebug(NRL_AGENT, "debug22: adding ref");
 Z_TRY_ADDREF_P(arg);
 ZVAL_NEW_REF(arg, arg);
}
```
This code could either be placed directly in the `nr_php_call_user_func` function or be an inline function declared in `php_zval.h` like:
```
static inline void nr_php_zval_wrap_out_arg(zval* zv) {
#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
  if (!Z_ISREF_P(zv)) {
    Z_TRY_ADDREF_P(arg);
    ZVAL_NEW_REF(zv, zv);
  }
#endif
}
```

2) Additionally, calls to call_user_func_array have changed (as evidenced by this https://www.drupal.org/project/drupal/issues/3174022 discussion), and we still need to ensure any tests and/or wrappers handle it accordingly.

3) The API internals ticket still has a few allocated days left for someone to pick up.
4) The unit tests will need to be updated to take into account the changes.

Fixes #46

### DIFF
--- a/.github/actions/ubuntu16-build-action/build.sh
+++ b/.github/actions/ubuntu16-build-action/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 make -j $(nproc) all
-make -j $(nproc) run_tests
+make -j run_tests
 
--- a/.github/actions/ubuntu16-build-action/build.sh
+++ b/.github/actions/ubuntu16-build-action/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 make -j $(nproc) all
-make run_tests
+make -j $(nproc) run_tests
 
--- a/.github/actions/ubuntu16-build-action/build.sh
+++ b/.github/actions/ubuntu16-build-action/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 make -j $(nproc) all
-make -j run_tests
+make run_tests
 
--- a/agent/php_agent.h
+++ b/agent/php_agent.h
@@ -330,7 +330,7 @@ extern zend_function* nr_php_zval_to_function(zval* zv TSRMLS_DC);
  *           won't help you!
  */
 static inline zval* nr_php_get_return_value(NR_EXECUTE_PROTO TSRMLS_DC) {
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   if (NULL == execute_data) {
     /*
      * This function shouldn't be called from outside a function context, so
@@ -346,7 +346,7 @@ static inline zval* nr_php_get_return_value(NR_EXECUTE_PROTO TSRMLS_DC) {
   NR_UNUSED_SPECIALFN;
 
   return return_value_ptr_ptr ? *return_value_ptr_ptr : NULL;
-#endif /* PHP7 */
+#endif /* PHP7+ */
 }
 
 /*
@@ -358,7 +358,7 @@ static inline zval* nr_php_get_return_value(NR_EXECUTE_PROTO TSRMLS_DC) {
  *           other "should not happen" situation. Note that this may be an
  *           IS_REFERENCE zval in PHP 7 if the function expects a reference.
  *
- * Warning : Directly modifying the returned zval in PHP 7 may result in 
+ * Warning : Directly modifying the returned zval in PHP 7 may result in
  *           difficult issues that are difficult to debug.
  *
  * IMPORTANT : This should only be called before the call to the current user
@@ -389,14 +389,14 @@ static inline zend_function* nr_php_execute_function(
   }
 #endif
 
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   return execute_data->func;
 #elif ZEND_MODULE_API_NO >= ZEND_5_5_X_API_NO
   return execute_data->function_state.function;
 #else
   return op_array_arg->prototype ? op_array_arg->prototype
                                  : (zend_function*)op_array_arg;
-#endif /* PHP7 */
+#endif /* PHP7+ */
 }
 
 static inline zval* nr_php_execute_scope(zend_execute_data* execute_data) {
@@ -404,7 +404,7 @@ static inline zval* nr_php_execute_scope(zend_execute_data* execute_data) {
     return NULL;
   }
 
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   return &execute_data->This;
 #else
   return execute_data->object;
@@ -600,26 +600,26 @@ extern const char* nr_php_function_filename(zend_function* func);
 static inline zend_class_entry* nr_php_zend_register_internal_class_ex(
     zend_class_entry* ce,
     zend_class_entry* parent_ce TSRMLS_DC) {
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   return zend_register_internal_class_ex(ce, parent_ce);
 #else
   return zend_register_internal_class_ex(ce, parent_ce, NULL TSRMLS_CC);
-#endif /* PHP7 */
+#endif /* PHP7+ */
 }
 
 static inline char* nr_php_zend_ini_string(char* name,
                                            nr_string_len_t name_len,
                                            int orig) {
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   return zend_ini_string(name, name_len, orig);
 #else
   return zend_ini_string(name, name_len + 1, orig);
-#endif /* PHP7 */
+#endif /* PHP7+ */
 }
 
 static inline const char* NRPURE
 nr_php_class_entry_name(const zend_class_entry* ce) {
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   return (ce->name && ce->name->len) ? ce->name->val : NULL;
 #else
   return ce->name;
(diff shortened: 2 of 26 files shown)


agent/php_call.c
```diff
@@ -46,9 +48,25 @@ zval* nr_php_call_user_func(zval* object_ptr,
...
+
+  /*
+   * With PHP8, `call_user_function_ex` was removed and `call_user_function`
+   * became the recommended function; however, the call handles param_values
+   * differently. Prior to PHP8, we used the no_separation flag set to 0 which
+   * then did some internal manipulation of param_values.  There is no longer a
+   * no separation flag, so that internal manipulation is no longer done and we
+   * need to do it prior to sending param_values or `call_user_function` will
+   * cause a segfault.
+   */
```
joshuabenuck, line 61: Noting that I need to update this comment.

agent/php_compat.h
```diff
@@ -6,6 +6,10 @@
 #ifndef PHP_COMPAT_HDR
 #define PHP_COMPAT_HDR
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+#define PHP8
+#endif /* PHP 8.0+ */
+
 #if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
 #define PHP7
```
joshuabenuck, line 14: We updated our usages of `PHP7` to use the more complex `ifdef` in order to make it clear we were referring to PHP version 7.0 or greater. This define doesn't allow for us to target things just to PHP 7 (and not 8 or 5). Should we update this definition to exclude PHP 8?

agent/php_zval.h
```diff
@@ -343,13 +343,27 @@ static inline void nr_php_zval_bool(zval* zv, int b) {
...
+ *           wraps the argument.
+ */
+static inline void nr_php_zval_wrap_out_arg(zval* zv) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+  if (!Z_ISREF_P(zv)) {
+    ZVAL_NEW_REF(zv, zv);
+  }
+#endif
+}
+
```
joshuabenuck, line 366: This looks to be unused (and is causing build warnings on PHP < 8.0 which is leading to build failures). I'm going to remove it.

agent/php_compat.h
zsistla: Yes, if we are sure we've replaced all PHP7 `ifdef`s.

agent/php_newrelic.h
```diff
@@ -181,7 +181,7 @@ typedef void (*nrphpfn_t)(INTERNAL_FUNCTION_PARAMETERS);
 typedef void(ZEND_FASTCALL* nrphpfn_t)(INTERNAL_FUNCTION_PARAMETERS);
 #endif /* PHP < 7.3 */
 
-#if ZEND_MODULE_API_NO <= ZEND_8_0_X_API_NO
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
```
zsistla, line 184: Great catch!  The comment at the end of the block also needs to be updated for both of the cases: `#endif /* PHP <= 8.0 */` to `#endif /* PHP >= 8.0 */`

joshuabenuck: Fixed!

joshuabenuck:
ok jenkins

.github/actions/ubuntu16-build-action/build.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 make -j $(nproc) all
-make -j $(nproc) run_tests
+make -j run_tests
```
kneitinger, line 3: Wow, I never knew about `-j` without any flags :brain:

joshuabenuck: Makes two of us! That was a typo, but apparently `-j` without flags is a thing!

kneitinger: Oh I thought it was intentional for maximum test pedal to the metal :laughing:

agent/php_agent.h
```diff
@@ -715,11 +715,18 @@ nr_php_ini_entry_name_length(const zend_ini_entry* entry) {
...
 #define NR_PHP_USER_FN_THIS() EG(This)
-#endif
+#endif /* PHP 7.0+ */
+
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+/* PHP 8 expects zend_object not zval */
+#define ZVAL_OR_ZEND_OBJECT(x) Z_OBJ(*x)
+#else
+#define ZVAL_OR_ZEND_OBJECT(x) x
+#endif /* PHP8+ */
```
kneitinger, line 729: :+1:  Nice!

kneitinger, line 729: Nice :+1:

agent/php_includes.h
```diff
@@ -60,4 +61,17 @@
 #include "TSRM.h"
 #endif
 
+/*
+ * The TSRMLS_* functions included below have actually been voided out since PHP
+ * 7.0.  They were formally removed in PHP 8.0 but we still support in our code,
+ * so we need to add the voided out macros here.
```
kneitinger, line 67: Perfect solution!

agent/php_internal_instrument.c
```diff
@@ -231,11 +231,26 @@ void nr_php_wrap_internal_function(nrinternalfn_t* wraprec TSRMLS_DC) {
...
+   * as a security measure to disable potentially unsafe functions.
+   *
+   * This functionality still exists in PHP 8, but the disabled functions
+   * behave as if they are not declared at all.
+   * Prior to PHP 8, attempting to use a disabled function resulted in a
+   * warning.
+   * In PHP 8 and later, attempting to use a disabled function will throw a
+   * standard error for an undeclared function.”
+   * In PHP 8+, a disabled function is indistinguishable from a function that
+   * does not exist.
```
kneitinger, line 246: [nit] These two sentences could be one, and there is an unmatched `"`

agent/php_hash.h
```diff
@@ -122,13 +122,18 @@ nr_php_zend_hash_key_string_value(const zend_hash_key* hash_key) {
...
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
   zval copy;
 
   ZVAL_DUP(&copy, value);
 
-  return add_assoc_zval(arr, key, &copy);
+#ifdef PHP8
+  add_assoc_zval(arr, key, &copy);
+  return 0;
```
kneitinger, line 132: Context in case anyone else was confused by this

php/php-src/pull/5805

agent/php_compat.h
joshuabenuck: There are still a bunch of `PHP7` ifdefs left. I filed issue #56 to track the need to clean up our ifdef usage so it doesn't block this PR.

zsistla:
ok jenkins

joshuabenuck:
ok jenkins

joshuabenuck (approved):

PR #70, opened 2020-12-21
PHP 8 curl Object Fixes

### LINKED ISSUE #42: PHP 8 - Curl API changes
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

The implementation of `curl` has changed somewhat in PHP 8.  Previously, `curl` handles were resources, but as of PHP 8 are now class objects. For this ticket, correct our instrumentation to account for this change when PHP 8 is defined.

## Desired Behavior

- All occurrences of `IS_RESOURCE` are accounted for and fixed in our instrumentation
- Since `curl_close()` no longer closes the curl handle (a task left to GC), verify with valgrind that our tests do handle this situation gracefully and appropriately.
- [Our Monolog Enricher's default Handler](https://github.com/newrelic/newrelic-monolog-logenricher-php/blob/master/src/Handler.php) uses Curl to communicate with the NR API.  Verify that this behavior still works as expected!

### LINKED ISSUE #51: [PHP 8] Internals Unit Test Updates
## Description

#49 makes the necessary changes for the agent to compile with PHP 8 libs and headers, however it does not address the agent unit tests.  For this ticket, ensure all unit tests pass after #49 is merged in main.

kneitinger:
Refactors all curl resource checks to account for the fact that in PHP 8, it is an object.

The necessary changes are nicely[ summarized here](https://php.watch/versions/8.0/resource-CurlHandle).  As you can see, API compatibility was very well maintained by the upstream code in PHP, but also somewhat surprisingly, in C as well.

Unit tests (with valgrind) pass with no issues, and small manual curl test program is correctly instrumented.

Fixes #42
Fixes #51

### DIFF
--- a/agent/php_curl.c
+++ b/agent/php_curl.c
@@ -28,6 +28,14 @@ static int nr_php_curl_do_cross_process(TSRMLS_D) {
           || NRPRG(txn)->options.distributed_tracing_enabled);
 }
 
+static int nr_php_is_zval_valid_curl_handle(const zval* ch) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+  return NULL != ch && nr_php_is_zval_valid_object(curlres);
+#else
+  return NULL != ch && nr_php_is_zval_valid_resource(curlres);
+#endif /* PHP 8.0+ */
+}
+
 static void nr_php_curl_save_response_header_from_zval(const zval* ch,
                                                        const zval* zstr
                                                            TSRMLS_DC) {
@@ -113,7 +121,7 @@ static void nr_php_curl_set_default_response_header_callback(
   zval* retval = NULL;
   zval* curlopt = NULL;
 
-  if ((NULL == curlres) || (IS_RESOURCE != Z_TYPE_P(curlres))) {
+  if ((NULL == curlres)) {
     return;
   }
 
@@ -137,7 +145,7 @@ static void nr_php_curl_set_default_request_headers(zval* curlres TSRMLS_DC) {
   zval* retval = NULL;
   zval* curlopt = NULL;
 
-  if ((NULL == curlres) || (IS_RESOURCE != Z_TYPE_P(curlres))) {
+  if (!nr_php_is_zval_valid_curl_handle(curlres)) {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


kneitinger:
ok jenkins

agent/php_internal_instrument.c
```diff
@@ -2189,9 +2189,14 @@ NR_INNER_WRAPPER(curl_exec) {
...
-      != zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
-                                  ZEND_NUM_ARGS() TSRMLS_CC, "r", &curlres)) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+  int parse_status = zend_parse_parameters_ex(
+      ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "o", &curlres);
+#else
+  int parse_status = zend_parse_parameters_ex(
+      ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "r", &curlres);
+#endif /* PHP 8.0+ */
+  if (SUCCESS != parse_status) {
```
zsistla, line 2199: Great job on this! I had just a few small comments. 

Minor nit.  

Would it make sense just declare/initialize  
`int parse_status = FAILURE;` before the if/else block?

Similar to here: https://github.com/newrelic/newrelic-php-agent/pull/70/files#diff-d66d558f126dbc42fe98d5332fa13d281fe925d4e01b7608a8dc39c56dce778eR2227

?

agent/php_curl.c
```diff
@@ -113,7 +121,7 @@ static void nr_php_curl_set_default_response_header_callback(
   zval* retval = NULL;
   zval* curlopt = NULL;
 
-  if ((NULL == curlres) || (IS_RESOURCE != Z_TYPE_P(curlres))) {
+  if ((NULL == curlres)) {
```
zsistla, line 124: It seems the check became less specific.
Would it make sense to add your great new function here?
`if (!nr_php_is_zval_valid_curl_handle(curlres)) {`

agent/php_internal_instrument.c
kneitinger: Perfect, thanks!

agent/php_curl.c
kneitinger: Great catch!

kneitinger:
ok jenkins

Fahmy-Mohammed (approved):
Looks great! I think it's ready for a squash & merge

PR #72, opened 2021-01-05
Stalebot refinements.

zsistla:
Refining our stalebot and shifting the paradigm.  Issues will only become stale while awaiting additional information.


Fahmy-Mohammed (approved):

.github/stale.yml
```diff
@@ -1,25 +1,34 @@
...
+onlyLabels:
+  - "additional information requested"
+
 # Issues with these labels will never be considered stale
 exemptLabels:
   - backlog
   - epic
   - pinned
   - security
   - triage
```
joshuabenuck, line 17: Why keep these in here if the only thing we will take are issues that need more info?

.github/stale.yml
```diff
@@ -1,25 +1,34 @@
...
   things get busy.  This issue will be closed after 14 days if no 
-  further activity occurs. 
+  further activity occurs, but can always be reopened at a later date. 
   Thank you for your contributions.
+
 # Comment to post when closing a stale issue. Set to `false` to disable
 closeComment: >
   This issue has been automatically closed because it has not had
-  recent activity.
+  recent activity, but it can be reopened again.
```
joshuabenuck, line 33: nit: ... maybe change it to: `it can be reopened` (my preference) or `it can be opened again`.

`it can be reopened again` feels a bit redundant.

.github/stale.yml
zsistla: What about the case where it has both labels?

.github/stale.yml
```diff
@@ -1,25 +1,34 @@
 # Number of days of inactivity before an issue becomes stale
 daysUntilStale: 30
+
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 14
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels:
+  - "additional information requested"
```
joshuabenuck, line 9: Feels long for a tag. How about something like `needs more info` or `more info required`?

.github/stale.yml
joshuabenuck: If it needs more info to be actionable, I'd be okay with it being marked as stale even if it has any of those other labels.

zsistla:
Closed in lieu of https://github.com/newrelic/newrelic-php-agent/pull/72.
I reforked and the branch seemed orphaned.

PR #73, opened 2021-01-13
Prevent allocation error

joshuabenuck:
This PR prevents an allocation error that occurs in the Drupal instrumentation and improves our logging when allocation errors do occur.

Specifically, these changes:
* Log a stack trace whenever an allocation error occurs. This should make tracking down future allocation errors easier.
* Logs a message at `verbosedebug` when the Drupal framework instrumentation problem is detected. This is consistent with the other related log messages.
* Create a `Supportability` metric in addition to the log message.

The name of the `Supportability` metric is either: `Supportability/framework/Drupal/NegativeModuleLength` or `Supportability/framework/Drupal8/NegativeModuleLength`. This is modeled after the `MiddlewareNotCallable` supportability metric the [Guzzle 6 instrumentation](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/lib_guzzle6.c#L375-L376) sends.

The change to `config.m4` is done to fix a build error that occurs with the addition of the calls to `nr_tracer_common` in `util_memory.c` (undefined reference to `dladdr`). The reason this problem occurs now is that our test program calls `nro_new` from `util_object.c` which in turn calls functions from `util_memory.c` which now depend on `util_signals.c` and it has a reference to `dladdr`. Prior to this change, no symbols from `util_signals` were needed so the linker never had to resolve the `dladdr` reference when building the test.

The fix is to ensure that the reference to the `dl` library occurs after the reference to the `axiom` library when linking. The problem is that the intuitive places to put the reference to the `dl` library are all placed before the reference to `axiom`. By tacking on an ` -ldl` to the first argument of `PHP_CHECK_LIBRARY`, the library reference is placed in the appropriate place when building the lib axiom test program.


Fahmy-Mohammed (dismissed):

kneitinger:
ok jenkins

kneitinger (approved):

zsistla:
ok jenkins

zsistla (approved):

joshuabenuck:
ok jenkins

kneitinger:
ok jenkins

kneitinger:
ok jenkins

kneitinger (approved):
Re-approval after autoconf fix was merged in.  That PR was approved, as well as this PR prior to the merge.

PR #75, opened 2021-01-19
Added the `onlyLabels` to trigger the `stalebot`.

zsistla:


zsistla:
ok jenkins

.github/stale.yml
```diff
@@ -1,25 +1,33 @@
...
-  - epic
-  - pinned
-  - security
-  - triage
+exemptLabels:[]
+#  - backlog
+#  - epic
+#  - pinned
+#  - security
+#  - triage
```
joshuabenuck, line 17: Ideally, these should be removed from the file unless there's a reason we expect to need to re-enable them in the near term.

joshuabenuck (dismissed):

Fahmy-Mohammed (approved):

joshuabenuck (approved):

PR #82, opened 2021-01-21
Remove -ldl in agent autoconf axiom check on BSD

kneitinger:
The `-ldl` flag added in https://github.com/newrelic/newrelic-php-agent/pull/73/commits/37901aa669dcdf408154e07bb0abdabaeb6f0f6b was the culprit of the build errors on FreeBSD.  Interestingly, the flag itself is not needed for the `agent`'s `autoconf` `axiom` lib check to pass on FreeBSD.  To get around this in an easily modifiable way, I've added a case statement that when on FreeBSD, omits the the `-ldl` arg, but defaults to including it.  If we encounter any issues with this in other scenarios, adding them to the FreeBSD block is all we need to get around the error.

Tested manually by logging in to Jenkins FreeBSD worker, pulling in this change and internal repo, and running both `build.sh` and `build-pull-request.sh` successfully.

### DIFF
--- a/agent/config.m4
+++ b/agent/config.m4
@@ -117,8 +117,20 @@ if test "$PHP_NEWRELIC" = "yes"; then
   dnl Check for axiom.
   dnl The -ldl in the first argument is there to get the library check to pass.
   dnl Without it, we get an undefined reference to dladdr as the dl lib must
-  dnl come after the axiom lib when linking.
-  PHP_CHECK_LIBRARY(axiom -ldl, nro_new, [
+  dnl come after the axiom lib when linking. Note: this does not apply to
+  dnl FreeBSD builds, and if present, causes errors when building the agent.
+  dnl If other platforms are known to have the same issue, simply change the
+  dnl   *freebsd*)
+  dnl line to something like
+  dnl   *freebsd*|*darwin*)
+  dnl See the other `case $host in` block later in this file for known targets
+  case "$host" in
+    *freebsd*) AXIOM_CHECK="axiom"
+               ;;
+    *) AXIOM_CHECK="axiom -ldl"
+               ;;
+  esac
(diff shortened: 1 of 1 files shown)


kneitinger:
All tests appear to be running fine; one CentOS build hiccuped on the integration runner.  Restarting builds.

kneitinger:
ok jenkins

zsistla (approved):
Great sleuthing!

agent/config.m4
```diff
@@ -117,8 +117,20 @@ if test "$PHP_NEWRELIC" = "yes"; then
...
+  dnl   *freebsd*)
+  dnl line to something like
+  dnl   *freebsd*|*darwin*)
+  dnl See the other `case $host in` block later in this file for known targets
+  case "$host" in
+    *freebsd*) AXIOM_CHECK="axiom"
+               ;;
+    *) AXIOM_CHECK="axiom -ldl"
+               ;;
+  esac
```
joshuabenuck, line 132: The definition of `AXIOM_CHECK` could go in the FreeBSD clause of the case statement back on line 84. That would reduce the number of case statements in the file and put all platform specific definitions in one place.

kneitinger: I kept your logic where it was, because I think it is most clear there, but moved only FreeBSD's logic to the system-dependent section with a comment in the original location.

I thought setting the var in each system obscured FreeBSD's quirkiness of being the outlier here, so opted to emphasize that it stands out a little more.  Let me know if that sounds good to you

joshuabenuck (approved):

kneitinger:
ok jenkins

kneitinger:
ok jenkins

kneitinger:
ok jenkins

kneitinger:
ok jenkins

kneitinger:
Merging so that build issue can be addressed on base branch.  Note: this is getting merged into a **non-`main`** base branch.

PR #89, opened 2021-01-29
Update newrelic-install.sh to support PHP 8.0

joshuabenuck:
This PR adds the needed references to PHP 8.0 and API version 20200930 needed to be able to install the PHP8 version of the agent. It also adds the API version of PHP 7.4 to one of the sets of file checks (see line 315). I believe its absence was an oversight.

I tested this by locally modifying `make/release.mk` to add the needed targets to build the PHP 8.0 version of the agent. I manually combined those results with the contents from a previous release so I had all versions of the agent in the release directory. This was enough to get all the file existence checks in the install script to pass which allowed me to test the installation logic.

Note: Once this PR is merged, installs will fail unless the PHP 8.0 version of the agent is in the agent's package. We may want to hold off on merging it until we have the PHP 8.0 version of the agent incorporated into the main build job.

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -312,7 +312,7 @@ if [ -z "${ispkg}" ]; then
   check_file "${ilibdir}/scripts/newrelic-daemon.logrotate"
 fi
 check_file "${ilibdir}/scripts/newrelic.ini.template"
-for pmv in "20090626" "20100525" "20121212" "20131226" "20151012" "20160303" "20170718" "20180731"; do
+for pmv in "20090626" "20100525" "20121212" "20131226" "20151012" "20160303" "20170718" "20180731" "20190902" "20200930"; do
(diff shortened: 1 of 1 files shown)


agent/newrelic-install.sh
```diff
@@ -312,7 +312,7 @@ if [ -z "${ispkg}" ]; then
   check_file "${ilibdir}/scripts/newrelic-daemon.logrotate"
 fi
 check_file "${ilibdir}/scripts/newrelic.ini.template"
-for pmv in "20090626" "20100525" "20121212" "20131226" "20151012" "20160303" "20170718" "20180731"; do
+for pmv in "20090626" "20100525" "20121212" "20131226" "20151012" "20160303" "20170718" "20180731" "20190902" "20200930"; do
```
zsistla, line 315: Nice catch!

zsistla (approved):

PR #105, opened 2021-02-19
PHP: 64-bit valgrind now builds correctly.

zsistla:
* Addressed the 64-bit valgrind issue.  The PHP 8.0 build now does unit tests with valgrind.
* Changed the workflow to show 7.x builds on x64.

Issues:
1) On GHA unit tests skip for 7.2 and below because SAPI embed was not compiled into that PHP image.

2) Integration tests:
We can disable lasp on the GHA runs, but it still asks for private keys (like license key).

3) 5.5 and 5.6 on x64
Axiom builds, but agent build can’t find axiom.

4) Now that valgrind runs on 64-bit, it's surfaced a possible memory issue that needs to be investigated.

### DIFF
--- a/.github/docker/linux/pr_build.sh
+++ b/.github/docker/linux/pr_build.sh
@@ -81,9 +81,7 @@ VALGRIND_ISSUE=0
     echo "php = $PHPS"
 case $PHPS in
   *8.0*)
-    if [ $ARCH = "x86" ]; then
-      VALGRIND_ISSUE=1
-    fi
+    VALGRIND_ISSUE=1
     ;;
   *)
     VALGRIND_ISSUE=0
--- a/.github/docker/linux/pr_build.sh
+++ b/.github/docker/linux/pr_build.sh
@@ -41,6 +41,12 @@ export PATH
 
 
 #
+# Set LD_LIBRARY_PATH
+#
+LD_LIBRARY_PATH=/usr/local/lib
+export LD_LIBRARY_PATH
+
+#
 # Get PHPS from the environment.
 #
 
@@ -49,7 +55,6 @@ PHPS=${PHP_VER}
 printf \\n
 printf 'running daemon tests\n'
 make -r -s daemon daemon_integration  "ARCH=${ARCH}"
-
 
 #
 # Limit valgrind to just the Linux builders. On Alpine Linux, valgrind
@@ -76,7 +81,9 @@ VALGRIND_ISSUE=0
     echo "php = $PHPS"
 case $PHPS in
   *8.0*)
-    VALGRIND_ISSUE=1
+    if [ $ARCH = "x86" ]; then
+      VALGRIND_ISSUE=1
+    fi
     ;;
   *)
     VALGRIND_ISSUE=0
@@ -152,7 +159,8 @@ EOF
     ;;
   *)
     printf "Running agent integration tests (PHP=%s ZTS=disabled)\n" "$PHPS"
-    make integration PHPS="$PHPS" "ARCH=${ARCH}" INTEGRATION_ARGS="--retry=1"
+    printf "Temporarily disable integration tests in GHA while determining what to do with private keys.\n"
+    #make integration PHPS="$PHPS" "ARCH=${ARCH}" INTEGRATION_ARGS="--retry=1"
     ;;
   esac
   printf \\n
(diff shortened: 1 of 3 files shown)


.github/docker/linux/pr_build.sh
```diff
@@ -167,10 +173,10 @@ EOF
       *embed*)
         if [ -n "$do_valgrind" ]; then
           printf 'grinding agent unit tests\n'
-          make -r -s -j $(nproc) agent-valgrind "ARCH=${ARCH}" LDFLAGS='-Wl,--no-warn-search-mismatch -Wl,-z,muldefs'
```
joshuabenuck, line 170: Why did you remove `LDFLAGS`?

.github/workflows/agent-build.yml
```diff
@@ -38,8 +38,25 @@ jobs:
     strategy:
       matrix:
         os: [linux]
-        php_ver: ['8.0', '8.0-zts']
+        php_ver: ['7.0', '7.1-zts', '7.2', '7.3', '7.4', '8.0', '8.0-zts']
```
joshuabenuck, line 41: Why are we not building the non-zts agent for PHP 7.1? Why are we not building zts versions of the agent for all of these versions of PHP?

zsistla: The 7.x builds were to show a minimal set (to reduce PR time) of proof of concept since at this moment Jenkins still handles all those builds/tests (especially since we need more discussion around how to handle integration tests).

joshuabenuck:
> * Addressed the 64-bit valgrind issue.  The PHP 8.0 build now does unit tests with valgrind.
> * Changed the workflow to show 7.x builds on x64.
> 
> Issues:
> 
> 1. On GHA unit tests skip for 7.2 and below because SAPI embed was not compiled into that PHP image.
> 2. Integration tests:
>    We can disable lasp on the GHA runs, but it still asks for private keys (like license key).
> 3. 5.5 and 5.6 on x64
>    Axiom builds, but agent build can’t find axiom.
> 4. Now that valgrind runs on 64-bit, it's surfaced a possible memory issue that needs to be investigated.

What was the valgrind issue you resolved? I'd like to compare that to what I'm seeing in the soak tests.

1. Are we planning on adding SAPI embed to 7.2 and below? I can see disabling this temporarily, but feel we should eventually update our image so we don't have to disable the tests (that is, if all of these builds are going to be how we generate official builds for these PHP versions).
3. Not being able to find axiom is the error introduced when we added logging of stack traces on memory allocation errors. I wonder if the added dependency is located in a different lib in the OS used to build 5.5 and 5.6. For the Linux builds, we currently add `-ldl` when checking for axiom. For the FreeBSD builds we don't as `libc` contains the required function. What is the base image we use for the 5.5 and 5.6 builds?
4. Any details you can share about the memory issue?

.github/docker/linux/pr_build.sh
zsistla: It would not compile with the flags.  This, along with the 32-bit/64-bit memory issue are still being investigated for issue #108 .

zsistla:
The 64-bit issue was an issue with the LDFLAGS which are removed for now, but need more investigation of why and if/how they can be added back.
> 
> 1. Are we planning on adding SAPI embed to 7.2 and below? I can see disabling this temporarily, but feel we should eventually update our image so we don't have to disable the tests (that is, if all of these builds are going to be how we generate official builds for these PHP versions).

Yes, this definitely needs to be addressed.  Currently, it doesn't have it because PHP on those images wasn't compiled with the embed flag.

> 2. Not being able to find axiom is the error introduced when we added logging of stack traces on memory allocation errors. I wonder if the added dependency is located in a different lib in the OS used to build 5.5 and 5.6. For the Linux builds, we currently add `-ldl` when checking for axiom. For the FreeBSD builds we don't as `libc` contains the required function. What is the base image we use for the 5.5 and 5.6 builds?

We use the official `php:5.6` and `php:5.5` which is based off of the `jessie` debian image.

> 3. Any details you can share about the memory issue?

I added the details in the [MMF](https://docs.google.com/document/d/1iolwPOyr1AcOmUDqzf5ZQZ4ODm1m593onSRhFP-UhiY/edit#heading=h.v2mcfrbllohv) doc.

joshuabenuck (approved):

zsistla:
ok jenkins

kneitinger:
ok jenkins

kneitinger:
ok jenkins

kneitinger:
ok jenkins

kneitinger (approved):
✅ Looks good!

PR #118, opened 2021-03-08
PHP 8.0 and valgrind mem issue on 64bit architecture.

zsistla:
* The assumption that the new `ZEND_CHECK_UNDEF_ARGS` would always exist and therefore we needed to index back two spaces was incorrect.  We need to account for it existing and not existing.

* `pr_build.sh` does not enable valgrind yet.  This will be enabled again in the 32bit PR.

### DIFF
--- a/agent/php_vm.c
+++ b/agent/php_vm.c
@@ -1,7 +1,7 @@
 /*
-* Copyright 2020 New Relic Corporation. All rights reserved.
-* SPDX-License-Identifier: Apache-2.0
-*/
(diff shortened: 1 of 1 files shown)


zsistla:
The 32-bit PR will go up after this one is merged because the 32bit PR reenables valgrind checks (which will fail without this being merged).

zsistla:
ok jenkins

Fahmy-Mohammed (approved):

PR #120, opened 2021-03-10
Prevent build up of duplicate Distributed Tracing headers

joshuabenuck:
This PR prevents a build up of Distributed Tracing headers from being added to a request when using `file_get_contents` and reusing the stream context parameter.

It does this by:
* Modifying the test trace endpoint to detect if there is more than one value for a distributed tracing header
* Adding a scenario that reuses the stream context when calling `file_get_contents`, this triggers the modified detection logic in the trace endpoint
* Updating `nr_php_file_get_contents_add_headers_internal` to detect if one of the DT headers are present in both the existing headers string and the headers that are about to be added. This change assumes that all of the DT headers are added as a group and that detecting one of them is sufficient.

The flow goes roughly like this:
* `nr_php_file_get_contents_create_outbound_headers` calls `nr_header_create_outbound_request_create`. The resulting headers are converted to a string.
* The headers string is passed by `nr_php_file_get_contents_add_headers` to `nr_php_file_get_contents_add_headers_internal` which concats the new headers to the existing headers.

In this flow, the existing headers are already a string. This changes attempts to reparse the existing headers to reduce overhead. The simplest approach seems to be just looking for the presence of a key phrase when determining if the DT headers are already present and are about to be added.

`W3C_TRACESTATE` was somewhat arbitrarily chosen. `NEWRELIC` can be optionally excluded via configuration so it wasn't an option. `W3C_TRACEPARENT` was the other option.

A colon is included in the searched for string to avoid matching any content in the value portion of any of the headers.

### DIFF
--- a/agent/php_file_get_contents.c
+++ b/agent/php_file_get_contents.c
@@ -138,6 +138,13 @@ static void nr_php_file_get_contents_add_headers_internal(zval* context,
     return;
   }
 
+  if (nr_stridx(Z_STRVAL_P(http_header), W3C_TRACESTATE) != -1 &&
+      nr_stridx(headers, W3C_TRACESTATE) != -1) {
+    /* Distributed Tracing headers already present and we are trying to
+       add them again, don't add duplicates. */
+    return;
+  }
+
   /* There is a non-empty header string which must be preserved. */
   {
     char* all_headers = NULL;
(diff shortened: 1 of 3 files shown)


Fahmy-Mohammed (approved):

zsistla (approved):
Great find, fix, and tests!

PR #121, opened 2021-03-10
Update integration tests to run on PHP 8

### LINKED ISSUE #107: PHP 8: Ensure all integration tests run
Currently, a large number of our integration tests are failing on PHP 8.0. We need to fix these failures and ensure that all integration tests pass. Please see the [MMF doc](https://docs.google.com/document/d/1iolwPOyr1AcOmUDqzf5ZQZ4ODm1m593onSRhFP-UhiY/edit#heading=h.7evzjypgrzid) for a list of the changes that happened in PHP 8.0 that need to be accounted for in our tests. This will involve:

- Fixing all integration tests
- Ensuring the integration runner is updated for PHP 8.0 functionality. Account for any deprecated functions that have been removed in PHP 8.0
- Add the integration tests to the release build process (similar to the [Jenkins style](https://source.datanerd.us/php-agent/newrelic-php-agent-internal/blob/master/hudson/test.sh), but added to the GHA [release build script](https://github.com/newrelic/newrelic-php-agent/blob/dev/.github/docker/linux/release_build.sh))
- Re-enable the integration tests in [GHA](https://github.com/newrelic/newrelic-php-agent/blob/dev/.github/docker/linux/pr_build.sh#L146) and in the [Jenkins PR builder job](https://source.datanerd.us/php-agent/newrelic-php-agent-internal/blob/master/hudson/build-pull-request.sh#L156) (for PHP 8.0)

Fahmy-Mohammed:
This fixes #107 and includes updates to fix failing integration tests on PHP 8. There are quite a few tests that were touched, but most of them were minor changes. To make it easier to review, the commits are broken up to include the fixes for each test category. Here's a brief summary of the changes:

**Curl**
- `Curl_multi_exec` instrumentation was updated to expect a `curl` object rather than a `curl` resource.
- Any tests that performed an `IS_RESOURCE` check using a `curl` handle are updated to use another resource.

**Bad input tests**
- Any tests that performed invalid argument or type checks now throw an `ArgumentCountError` or `TypeError` in PHP 8 instead of a warning. Since that causes test execution to stop, those tests have either been rewritten for PHP 8 or have a `try/catch` block to make it backward compatible.

**Default error reporting**
- The PHP INI directive `error_reporting` controls which types of errors, warnings, and notices should be reported. In PHP 8.0, the default is changed to E_ALL. This now causes any tests using our deprecated `newrelic_accept_distributed_trace_payload()` function will show deprecation warnings. To silence these, `error_reporting` is set to `E_ALL & ~E_DEPRECATED & ~E_STRICT`.

**$_ENV[ACCOUNT_supportability]**
- Any tests that use environment variables to hide sensitive account information now need an `isset()` check. If the environment variable is not set then the test is skipped.

### Note:
- This does not include all of the changes needed for the error and drupal 6/7 tests. There are still 7 error test failures and 4 drupal test failures. Those will be added in a follow-up PR.

### DIFF
--- a/tests/integration/api/enable_params/test_enable_6.php
+++ b/tests/integration/api/enable_params/test_enable_6.php
@@ -6,7 +6,16 @@
 
 /*DESCRIPTION
 Calling newrelic_enable_params() with an argument that is neither a boolean nor
-an integer enables the recording of request parameters.
+an integer enables the recording of request parameters. This test is skipped on
+PHP 8+ because invalid arguments now throw a TypeError instead of a warning. Since
+this causes test execution to stop, it invalidates the test on PHP 8+ versions.
+*/
+
+/*SKIPIF
+<?php
+if (version_compare(PHP_VERSION, "7.4", ">")) {
+  die("skip: PHP > 7.4.0 not supported\n");
+}
 */
 
 /*ENVIRONMENT
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Build artifacts
 agent/configure.ac
 agent/newrelic.map
+axiom/tests
 bin/
 pkg/
 
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -2695,11 +2695,17 @@ NR_INNER_WRAPPER(curl_multi_exec) {
   int zcaught = 0;
   zval* curlres = NULL;
   zend_long still_running;
+  int rv = FAILURE;
 
-  if (SUCCESS
-      != zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
-                                  ZEND_NUM_ARGS() TSRMLS_CC, "rl", &curlres,
-                                  &still_running)) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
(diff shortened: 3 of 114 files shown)


Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed:
ok jenkins

tests/integration/api/enable_params/test_enable_6.php
```diff
@@ -45,7 +45,15 @@
 ]
 */
 
-newrelic_enable_params(array());
+/*
+ * This throws a TypeError on PHP 8. To simplify, lets just catch the error and continue
+ * execution to allow this test to run on any PHP version
```
joshuabenuck, line 50: Does the test description still hold when run under PHP 8?
> Calling newrelic_enable_params() with an argument that is neither a boolean nor an integer enables the recording of request parameters.

Fahmy-Mohammed: You're right. I've skipped the test on PHP 8+ since the error halts execution. Catching and continuing doesn't really do anything in PHP 8+ so it makes more sense to skip. I've added a more detailed description as well.

joshuabenuck (approved):

kneitinger:
ok jenkins

kneitinger (approved):
Thank you for going through all of these!

PR #122, opened 2021-03-11
Remove unused NR cert bundler utility

kneitinger:
### PROTOTYPE PR!
---

It seems that a utility that compiled an internal New Relic `.pem` cert bundle into a go var was included in the migration to this open source repo.

The utility grabs a URL from a VPN-only URL, and the var it creates is not used in any code, therefore, I have reasonable confidence that it can be removed to prevent further confusion.

With that said, this PR is not yet intended to be merged.  It's main purpose is to run every existing test and build config against, all the way to simulating a new release and using the generated binaries.

Historical context suggests that this file was included in this repo because at the time of the migration process, it was needed for builds (even though the configs specified to use system certs), however that dependency had been removed prior to the creation of this repo.

I'll note any findings here as I go through the tests and code.


kneitinger:
ok jenkins

PR #123, opened 2021-03-12
Fix memory leak that occurs when exceptions occur in stacked segments

joshuabenuck:
This PR fixes a couple of memory leaks introduced in v9.12 of the agent.
* A string that wasn't freed in `nr_php_error_record_exception_segment`.
* The memory pointed to by `segment->error` when an exception occurs in a stacked segment that is not converted to a heap segment.

The fix for the latter involves converting the stacked segment to a heap allocated segment prior to deallocation if `segment->error` is set. The pre-existing code does the same thing if `segment->attributes` is set.

### DIFF
--- a/agent/php_error.c
+++ b/agent/php_error.c
@@ -643,6 +643,7 @@ nr_status_t nr_php_error_record_exception_segment(nrtxn_t* txn,
   nr_segment_record_exception(nr_txn_get_current_segment(NRPRG(txn), NULL),
                               error_message, klass);
 
+  nr_free(file);
   nr_free(error_message);
   nr_free(klass);
   nr_free(message);
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1073,7 +1073,8 @@ static inline void nr_php_execute_segment_end(
 
   if (create_metric || (duration >= NR_PHP_PROCESS_GLOBALS(expensive_min))
       || nr_vector_size(stacked->metrics) || stacked->id
-      || stacked->attributes) {
+      || stacked->attributes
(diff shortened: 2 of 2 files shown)


joshuabenuck:
ok jenkins

joshuabenuck:
I was unable to create an automated test to reproduce this. Here is the manual approach I used should we want to try again in the future.

Create a php file with the following contents:
```php
<?php    
function recursiveCount(int $level): void    
{    
    if ($level <= 1) {    
        //throw new \Exception('Oops');    
        return;    
    }    
    
    recursiveCount($level - 1);    
}    
    
for($i= 0; $i<2; ++$i) {    
    try {    
        recursiveCount(1);    
    } catch(\Exception $e) {    
        // continue    
    }    
}    
```

Ensure that distributed tracing is enabled and that tracer details are set to `1`.

Run the script under `valgrind` (and remove the changes from this PR in your agent build). If successful, you should see a memory leak after running the script a couple of times as the agent must be able to successfully connect to New Relic for the memory leak to occur.

Increasing the recursion count and the number of iterations through the loop will also help to increase the chances of causing a memory leak. The leak occurs when there are short lived segments that remain on the stack and that throw an exception.

joshuabenuck:
ok jenkins

joshuabenuck:
ok jenkins

joshuabenuck:
ok jenkins

Fahmy-Mohammed (approved):

kneitinger (approved):
Well done!  Thanks for investigating and fixing this!

PR #124, opened 2021-03-17
Fix error handling related integration tests

joshuabenuck:
This is part of the fixes for #107. The changes unique to this PR fix our error handling related integration tests. This required both changes to the tests and the agent.

In `php_error.c`, we need to strip off `E_DONT_BAIL` from the error type. `E_DONT_BAIL` is a new error value [added in PHP 8](https://github.com/php/php-src/commit/db233501ff9d56765ef4a870b777a643c2136711) that is used to indicate to the runtime that a clean exit should be performed. To avoid having to sprinkle this modification of `type` in several places, this is done just once in `nr_php_error_cb` prior to any of the other related functions being called.

In PHP 8.0+, the format string for an exception is [expanded](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_exceptions.c#L905-L914) before it gets to us. In `nr_php_should_report_error`, we check both the error type (fixed with the previous fix) and the format string of the exception to ensure our handler is only called on uncaught exceptions. The simplest fix for this change is to just not look for the `%s` in the format string (or message) as that will work across PHP versions.

The other changes are all in the tests:
* `test_E_NOTICE`: replace a reference to an undefined variable (which is now an E_WARNING instead of an E_NOTICE) with two calls to `session_start()` which emits a notice.
* `test_ignore_E_WARNING`: replace a divide by zero (which is now an E_ERROR instead of an E_WARNING) with a call to `session_gc` which emits a warning when it is called outside of an active session.
* `test_E_PARSE` (PHP 8 version): the change to replace `E_PARSE` with `Error` from the previous round of fixes is no longer needed as `get_error_type_string` once again returns `E_PARSE` now that we account for `E_DONT_BAIL`. Note: The PHP 8 version of the test is still needed as the parse error text is still a bit different between the versions. Some regex work would allow the tests to be merged.

### DIFF
--- a/agent/php_error.c
+++ b/agent/php_error.c
@@ -423,12 +423,15 @@ static int nr_php_should_record_error(int type, const char* format TSRMLS_DC) {
    * conditional will help us prevent double capture once we start recording
    * more than one error per transaction.
    *
-   * Note:  The format string comparison is a fragile check:  This "Uncaught %s"
+   * Note:  The format string comparison is a fragile check:  This "Uncaught"
    * string is not guaranteed in the PHP runtime.  If there was a better way to
    * detect uncaught exceptions we would do so.  To reduce the chance that this
    * early exit triggers erroneously, we check for the exception error type.
+   *
+   * In PHP 8.0+, this is no longer a format string so we now omit the %s in the
+   * check.
    */
-  if ((E_ERROR == type) && (0 == nr_strnicmp(format, NR_PSTR("Uncaught %s")))) {
+  if ((E_ERROR == type) && (0 == nr_strnicmp(format, NR_PSTR("Uncaught")))) {
     return 0;
   }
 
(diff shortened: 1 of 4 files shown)


agent/php_error.c
```diff
@@ -423,12 +423,15 @@ static int nr_php_should_record_error(int type, const char* format TSRMLS_DC) {
...
-   * Note:  The format string comparison is a fragile check:  This "Uncaught %s"
+   * Note:  The format string comparison is a fragile check:  This "Uncaught"
    * string is not guaranteed in the PHP runtime.  If there was a better way to
    * detect uncaught exceptions we would do so.  To reduce the chance that this
    * early exit triggers erroneously, we check for the exception error type.
+   *
+   * In PHP 8.0+, this is no longer a format string so we now omit the %s in the
+   * check.
    */
-  if ((E_ERROR == type) && (0 == nr_strnicmp(format, NR_PSTR("Uncaught %s")))) {
```
Fahmy-Mohammed, line 431: Does this need a condition for < PHP8? Or is the backward compatibility not an issue here?

Fahmy-Mohammed (approved):
Looks great! Just one small question and I'll also re-approve this after the rebase!

agent/php_error.c
joshuabenuck: Backwards compatibility isn't an issue here. In PHP < 8.0, this is a format string ("Uncaught %s thrown") while in PHP 8.0+, the format string is expanded by the time we get here ("Uncaught <the full exception stacktrace> thrown"). This change just shortens the string we look for so it is the string the common across all PHP versions.

kneitinger (approved):

PR #125, opened 2021-03-18
Fix to zend_http_client which surfaced during multiverse test.

zsistla:
Defect surfaced during test cases runs.
After determining it wasn't a problem with the test case, put in the fix and retested with the failing test case which now passes.
This is for the zend_http library (not the zend framework) which was failing a zend 1.x zend_http library test.


zsistla:
Closing and reopening pull request against `dev`.

PR #36, opened 2020-11-20
Added Laminas 3.x support.

zsistla:
Duplicated the current `Zend` support with `Laminas` naming.

See https://source.datanerd.us/php-agent/multiverse/pull/56 for multiverse testing.

PR #126, opened 2021-03-18
Fix to zend_http_client which surfaced during multiverse test.

zsistla:
Defect surfaced during test cases runs.
After determining it wasn't a problem with the test case, put in the fix and retested with the failing test case which now passes.
This is for the zend_http library (not the zend framework) which was failing a zend 1.x zend_http library test.

### DIFF
--- a/agent/lib_zend_http.c
+++ b/agent/lib_zend_http.c
@@ -436,7 +436,7 @@ NR_PHP_WRAPPER_END
 void nr_zend_http_enable(TSRMLS_D) {
   if ((NR_FW_ZEND != NRPRG(current_framework))
       && (NR_FW_LAMINAS3 != NRPRG(current_framework))) {
-    nr_php_wrap_user_function(NR_PSTR(http_client_request),
+    nr_php_wrap_user_function(NR_PSTR(HTTP_CLIENT_REQUEST_Z),
                               nr_zend_http_client_request TSRMLS_CC);
   }
 }
@@ -452,7 +452,7 @@ void nr_laminas_http_enable(TSRMLS_D) {
     http_client = HTTP_CLIENT_L;
     http_client_request = HTTP_CLIENT_REQUEST_L;
 
-    nr_php_wrap_user_function(NR_PSTR(http_client_request),
+    nr_php_wrap_user_function(NR_PSTR(HTTP_CLIENT_REQUEST_L),
                               nr_zend_http_client_request TSRMLS_CC);
   }
 }


joshuabenuck:
I don't quite follow this change. Is the fix needed for all versions of PHP or just PHP 8.0? Has the defect existed for a while? Can you describe in more detail why the change is needed / what it is doing?

zsistla:
Yes, this will be needed for all versions, not just PHP.
This area of code changed when the laminas support went in.  
When this [commit](https://github.com/newrelic/newrelic-php-agent/pull/36/commits/967f28220d34d72b18b008caa7d6f46131efdfef) changed the type, and therefore changed what NR_PSTR was passed and also changed what it returned.
```
    nr_php_wrap_user_function(NR_PSTR(http_client_request),
                              nr_zend_http_client_request TSRMLS_CC);
```

Since we have the appropriate type in `HTTP_CLIENT_REQUEST_Z` and `HTTP_CLIENT_REQUEST_L`, we can just pass them straight in.

zsistla:
ok jenkins

agent/lib_zend_http.c
```diff
@@ -436,7 +436,7 @@ NR_PHP_WRAPPER_END
 void nr_zend_http_enable(TSRMLS_D) {
   if ((NR_FW_ZEND != NRPRG(current_framework))
       && (NR_FW_LAMINAS3 != NRPRG(current_framework))) {
-    nr_php_wrap_user_function(NR_PSTR(http_client_request),
+    nr_php_wrap_user_function(NR_PSTR(HTTP_CLIENT_REQUEST_Z),
```
joshuabenuck, line 439: Was the failure that led to this PR due to `http_client_request` being set to the wrong value here? If so, `http_client_request` is used elsewhere in this file and may also be the incorrect value there as well. Additionally, if `http_client_request` is the wrong value, could `http_client` be set to the wrong value as well?

If I'm reading this code correctly, `http_client_request` should never be set to the wrong value here unless both Zend HTTP and Laminas HTTP are in use in the same project at which point.

I think we should set all of these state variables (`library_name`, `curl_adapter_typename`, `uri_http_typename`, `http_client`, and `http_client_request`) in both of the `enable` functions rather than relying on them to still have their initial values here.

zsistla: It's not a case of any of the values being set wrongly as everything is being set correctly.  The value is correct, the type is wrong to be passing to [NR_PSTR](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_axiom.h#L115) which expects a const string not a pointer to a string.  These are the only call to `NR_PSTR` so should be the only place that needs a modification.

joshuabenuck: Ah. TIL. Thanks!

joshuabenuck (approved):

Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed (approved):
I disabled PHP8 integration tests again so we can hopefully get a good Jenkins run on this PR. Otherwise, everything looks great!

zsistla:
ok jenkins

PR #127, opened 2021-03-22
Fix NOTICE and WARNING tests on other PHP versions

joshuabenuck:
#124 changed a couple of error handling related tests, but the functions chosen do not exist on all versions of PHP we test against or display a different error message.

Rather than taking the time to either find an alternative that exists across all PHP versions or tweaking the expected error message so it matches across all PHP versions, this PR restores the old versions of those tests for older versions of PHP and adds new versions for PHP 8.0+.

### DIFF
--- a/tests/integration/errors/test_E_NOTICE.php8.php
+++ b/tests/integration/errors/test_E_NOTICE.php8.php
@@ -16,13 +16,13 @@ log_errors=0
 
 /*SKIPIF
 <?php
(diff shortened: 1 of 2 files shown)


joshuabenuck:
ok jenkins

joshuabenuck:
ok jenkins

ruudk:
I'm following this repository but every PR is filled with `ok jenkins`. Why?

Fahmy-Mohammed (approved):

Issue #129, opened 2021-04-05
PHP 8: Validate all applicable integration tests on PHP 8


joshuabenuck:
This is part of #34 and is a continuation of #107.

Our CI systems do not run all tests on all OS / PHP version combinations. Since this is the first release that will support PHP 8, we should run all applicable integration tests against a PHP 8 agent build.

Tasks:
* (done) Compare the lists of skipped tests on the latest libc (CentOS), alpine (musl), and FreeBSD integration PHP 8 PR runs. Come up with a list of integration tests skipped on all of these configurations.
* Ensure that all of the skipped tests are run at least once manually w/ PHP 8.0 on a dev system or an AMI.
* Fix any tests that are failing.

List of tests to run:
* predis (5 tests not run anywhere)
* memcache (8 tests)
* redis (2 known failing tests): extension was already upgraded on the 64-bit system
* lasp (5 tests)
* http (7 tests)
* memcached (10 tests) - official PHP 8 support not released, but needed changes are in their dev branch. Should we test the dev branch?
* mongo (2 tests) - N/A not supported on PHP 8.0 (mongodb extension is, but we don’t support it)
* mysql (5 tests)
* mysqli (27 tests)
* pdo (15 tests)
* pgsql (10 tests)
* sqlite (10 tests)
* sqlitedatabase (5 tests)
* uopz - N/A will be handled in a separate ticket post release
* misc
  - tests/regression/test_php_1179_pdo_options_crash.php # skip: SQLSTATE[HY000] [2002] No such file or directory
misc
  - tests/integration/events/test_database_duration.php # skip: SQLSTATE[HY000] [2002] No such file or directory
  - tests/integration/external/file_get_contents/test_cat_and_synthetics_disabled.php # skip: env vars required

zsistla:
For `redis` see PR #136 
For `mysqli` see PR #137

zsistla:
See PR #134

joshuabenuck:
Here are the results of the tests investigation:
* predis, redis, mysqli, pgsql, and uopz all had some test bit rot.
* lasp, pdo run successfully, but will need to be updated if we want to update the version of mysql on the build nodes.
* memcache (8 tests)
* http - N/A for php 8. We only support an old version of the this extension.
* memcached - all tests passed using the latest dev changes of the extension.
* sqlite - N/A for php 8.
* sqlitedatabase - all pass
* misc - one not applicable, the other two pass.

[...]

PR #132, opened 2021-04-06
PHP 8.0 agent build and install changes

joshuabenuck:
This PR:
* updates the `release` target to build for PHP 8.0, but only on 64-bit targets.
* modifies the `newrelic-install.sh` script to not look for PHP 8.0 builds when installing on either 32-bit or 64-bit targets.

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -313,8 +313,11 @@ if [ -z "${ispkg}" ]; then
 fi
 check_file "${ilibdir}/scripts/newrelic.ini.template"
 for pmv in "20090626" "20100525" "20121212" "20131226" "20151012" "20160303" "20170718" "20180731" "20190902" "20200930"; do
-  check_file "${ilibdir}/agent/${arch}/newrelic-${pmv}.so"
-  check_file "${ilibdir}/agent/${arch}/newrelic-${pmv}-zts.so"
+  # If on a 32-bit system, don't look for a PHP 8.0 build.
+  if [ "${arch}" = "x64" ] -o [ "${pmv}" != "20200930" ]; then
+    check_file "${ilibdir}/agent/${arch}/newrelic-${pmv}.so"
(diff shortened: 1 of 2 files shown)


Fahmy-Mohammed (approved):
Looks great! My assumption is that the 2 redis test failures will be fixed in #129?

joshuabenuck:
> Looks great! My assumption is that the 2 redis test failures will be fixed in #129?

Correct!

zsistla:
If we detect 32bit [here](https://github.com/newrelic/newrelic-php-agent/blob/dde0b8f121b00b24f779f84c8b5235d9dfecf2c3/agent/newrelic-install.sh#L1040) should we also exit?

agent/newrelic-install.sh
```diff
@@ -1035,18 +1038,25 @@ for this copy of PHP. We apologize for the inconvenience.
...
 
     8.0.*)
+      if [ "${arch}" = "x86" ]; then
+        UNSUPPORTED_VERSION="yes"
+      fi
       ;;
 
     *)
-      error "unsupported version '${pi_ver}' of PHP found at:
+      UNSUPPORTED_VERSION="yes"
```
zsistla, line 1047: Nice way to handle!

zsistla (approved):

joshuabenuck:
OK. I think I've handled all of the needed cases now. The latest iteration detects PHP 8.0 on 32-bit systems as well as PHP 8.0 32-bit installs on 64-bit systems. The error message in either case now clearly mentions the fact that this is an unsupported 32-bit build which should help us quickly identify this scenario in the field.

zsistla:
> OK. I think I've handled all of the needed cases now. The latest iteration detects PHP 8.0 on 32-bit systems as well as PHP 8.0 32-bit installs on 64-bit systems. The error message in either case now clearly mentions the fact that this is an unsupported 32-bit build which should help us quickly identify this scenario in the field.

Looks good.

PR #136, opened 2021-04-16
Update Redis integration tests

zsistla:
Updated Redis integration tests to account for Redis deprecation notices.


joshuabenuck:
Just an observation: In some of our other integration tests we use the pattern `testname.php8.php` and `testname.php` (or `testname.php7.php`) instead of `testname1.php` and `testname2.php` when we need to have version specific variants.

We may want to standardize on one or the other at some point.

joshuabenuck:
ok jenkins

Fahmy-Mohammed (approved):
Looks good to me! I agree with Joshua's comment on the test naming for different PHP versions, but that isn't critical here.

PR #137, opened 2021-04-18
Mysqli integration test updates

zsistla:
* Updated to work with proper mysqli command syntax.
* Verified working with mysql server 5.7 (mysql server 8.0 will return different traces since it returns more info).
* Updated traces for latest agent trace returns.
* Updated skipif.inc, if it doesn't detect a DB (i.e., running locally on a dev machine instead of a build machine) is already setup it will set one up.


joshuabenuck:
If we make the changes to look for `partitions` and `filtered` properties of the explain plan, we will also need to update some `pdo` and `lasp` tests that fail in the same way. I had assumed that our mysql integration tests ran against an older mysql release and that my dev system was just running too new of a version.

Should we add those changes to this PR or should I open a new one?

zsistla:
> If we make the changes to look for `partitions` and `filtered` properties of the explain plan, we will also need to update some `pdo` and `lasp` tests that fail in the same way. I had assumed that our mysql integration tests ran against an older mysql release and that my dev system was just running too new of a version.
> 
> Should we add those changes to this PR or should I open a new one?

This was using mysql server 5.7, so already an older release.
I'd opt to open each test type as it's own PR so they don't block each other as they each seem to need their own finessing.

joshuabenuck (approved):

zsistla:
ok jenkins

zsistla:
ok jenkins

tests/integration/mysqli/skipif.inc
```diff
@@ -21,9 +21,84 @@ if (!extension_loaded("mysqli")) {
...
+
+
+            /* commit transaction */
+            if (!mysqli_commit($link)) {
+                print("Transaction commit failed.\n");
+            } else {
+                echo "Transaction commit succeeded.\n";
+            }
+        }
+    }
```
kneitinger, line 98: This is awesome!

kneitinger (approved):
Looks good!

Fahmy-Mohammed (approved):

joshuabenuck (approved):

kneitinger (approved):

Issue #143, opened 2021-04-22
[v9.17.0.300] [workaround found] ErrorException: Warning: PDOStatement::bindParam(): Argument #2 ($var) must be passed by reference, value given


ruudk:
## Description

I just deployed the latest agent on PHP 8.0.3 and am getting this exception that I cannot explain: 
```
ErrorException: Warning: PDOStatement::bindParam(): Argument #2 ($var) must be passed by reference, value given
#16 /vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php(112): Doctrine\DBAL\Driver\PDOStatement::execute
#15 /vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1292): Doctrine\DBAL\Connection::executeQuery
#14 /vendor/doctrine/orm/lib/Doctrine/ORM/Query/Exec/SingleSelectExecutor.php(44): Doctrine\ORM\Query\Exec\SingleSelectExecutor::execute
#13 /vendor/doctrine/orm/lib/Doctrine/ORM/Query.php(325): Doctrine\ORM\Query::_doExecute
#12 /vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php(1073): Doctrine\ORM\AbstractQuery::executeIgnoreQueryCache
#11 /vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php(1027): Doctrine\ORM\AbstractQuery::execute
#10 /vendor/doctrine/orm/lib/Doctrine/ORM/AbstractQuery.php(791): Doctrine\ORM\AbstractQuery::getResult
```

My first thought was that this is related to Doctrine DBAL/ORM but then I searched for `bindParam` in the agent code and found:
https://github.com/newrelic/newrelic-php-agent/blob/adbdca64f20e999443d9aee7ff5d9596b32f976d/agent/php_pdo.c#L110-L123

Could it be related to this?

The Doctrine code passes it correctly via reference. 

## Your Environment

Ubuntu PHP 8.0.3

## Additional context

I'm using the latest Doctrine DBAL/ORM:
```
composer show | grep "doctrine/"
broadway/event-store-dbal               0.5.0              Event store implementation using doctrine/dbal
doctrine/annotations                    1.12.1             Docblock Annotations Parser
doctrine/cache                          1.10.2             PHP Doctrine Cache library is a popular cache implementation th...
[...]

PR #146, opened 2021-04-22
Pass parameters by reference when calling bindParam

joshuabenuck:
This PR fixes issue #143.

In the PHP 8 [upgrading internals](https://github.com/php/php-src/blob/PHP-8.0.0/UPGRADING.INTERNALS) document, there is this note:
>   The zend_fcall_info no_separation flag has been removed, and separation is
      never allowed. If you wish to pass (or allow passing) arguments by
      reference, explicitly create those arguments as references using
      ZEND_MAKE_REF. This removal also affects call_user_function_ex(), which
      should be replaced by call_user_function().

Here are some details about what that means in practice and how it compares to previous PHP releases.

PHP 8.0 emits a warning [when a reference is needed and wasn't provided](https://github.com/php/php-src/blob/PHP-8.0.3/Zend/zend_execute_API.c#L770-L776). The runtime still [creates a reference](https://github.com/php/php-src/blob/PHP-8.0.3/Zend/zend_execute_API.c#L799-L801), it is just [noisy](https://github.com/php/php-src/blob/PHP-8.0.0/Zend/zend_execute.c#L1955-L1968) about having to do it.

Prior to PHP 8.0, a flag could be passed in which indicated whether the runtime should [create a reference in such scenarios](https://github.com/php/php-src/blob/PHP-7.4.18/Zend/zend_execute_API.c#L762-L766). If a reference was needed and one wasn't provided, [the behavior](https://github.com/php/php-src/blob/PHP-7.4.18/Zend/zend_execute_API.c#L767-L775) is essentially what PHP 8.0 does (warning, but creates a reference anyway).

In our pdo instrumentation, we call [nr_php_call](https://github.com/newrelic/newrelic-php-agent/blob/adbdca64f20e999443d9aee7ff5d9596b32f976d/agent/php_pdo.c#L155) to invoke `bindParam`. This uses PHP's `call_user_function`. At this particular call site, a reference is needed and since we don't provide one, a warning is emitted.

The first attempt blindly created a reference for all parameters, but this breaks things as it introduces references where they are not expected. To truly copy the previous behavior, we would need to call [ARG_SHOULD_BE_SENT_BY_REF](https://github.com/php/php-src/blob/PHP-8.0.3/Zend/zend_compile.h#L995-L996) which requires that we resolve the function before invoking `call_user_function` which will have to resolve the function again. This introduces extra overhead which would be best to avoid.

We may be able to invoke a `call_user_function` variant that requires the function handle instead of the function name, but this is a more invasive change.

The approach in the latest revision chooses to require each call site make the determination about whether a reference is needed before calling `nr_php_call`. If there are places where we use `nr_php_call` and friends to make calls to arbitrary functions, such call sites will need to be reworked to programmatically determine whether a reference is needed. In practice, we usually just call a known function. We just now need to be sure to add references, where they are needed, before making such calls.

The reported issue was found using Doctrine. The call path [creates a prepared statement](https://github.com/doctrine/dbal/blob/c800380457948e65bbd30ba92cc17cda108bf8c9/lib/Doctrine/DBAL/Connection.php#L1289-L1295) and calls [bindValue](https://github.com/doctrine/dbal/blob/c800380457948e65bbd30ba92cc17cda108bf8c9/lib/Doctrine/DBAL/Connection.php#L2048) for the parameters. I was unable to reproduce the warning message using `bindParam` despite that being what we instrument. I can only assume that `bindValue` somehow depends on `bindParam`.

The added integration test runs a prepared statement and calls `bindValue` on a single parameter which enough to trigger the warning when the fix is not present.

### DIFF
--- a/agent/php_call.c
+++ b/agent/php_call.c
@@ -40,6 +40,16 @@ zval* nr_php_call_user_func(zval* object_ptr,
 
     param_values = (zval*)nr_calloc(param_count, sizeof(zval));
     for (i = 0; i < param_count; i++) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+      /*
+       * With PHP8, we need to create references in order to pass or allow
+       * passing arguments by reference when invoking call_user_function.
+       */
+      if (!Z_ISREF_P(params[i])) {
+        Z_TRY_ADDREF_P(params[i]);
+        ZVAL_NEW_REF(params[i], params[i]);
+      }
+#endif
       param_values[i] = *params[i];
     }
   }
--- a/agent/php_call.c
+++ b/agent/php_call.c
@@ -40,16 +40,6 @@ zval* nr_php_call_user_func(zval* object_ptr,
 
     param_values = (zval*)nr_calloc(param_count, sizeof(zval));
     for (i = 0; i < param_count; i++) {
-#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
-      /*
-       * With PHP8, we need to create references in order to pass or allow
-       * passing arguments by reference when invoking call_user_function.
-       */
-      if (!Z_ISREF_P(params[i])) {
-        Z_TRY_ADDREF_P(params[i]);
-        ZVAL_NEW_REF(params[i], params[i]);
-      }
-#endif
       param_values[i] = *params[i];
     }
   }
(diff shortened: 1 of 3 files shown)


joshuabenuck:
Tests that call `file_get_contents` or that touch `curl` are not passing on the 8.0 PR jenkins job. The instrumentation for these functions makes use of `nr_php_call` so they would be impacted by these changes.

zsistla (approved):

kneitinger (approved):
Looks good, thank you for tracking this down and fixing it!

zsistla (approved):

PR #148, opened 2021-04-24
Fixes `mysqli_commit` instrumentation returning NULL in some cases.

zsistla:
* Fixes `mysqli_commit` instrumentation returning NULL.
* Utilizes already existing function to more cleanly handle the datastore segment.

### DIFF
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -749,6 +749,39 @@ NR_INNER_WRAPPER(mysqli_options) {
   }
 }
 
+static void nr_php_instrument_datastore_operation_call(
+    const nrinternalfn_t* nr_wrapper,
+    nr_datastore_t datastore,
+    const char* operation,
+    nr_datastore_instance_t* instance,
+    INTERNAL_FUNCTION_PARAMETERS) {
+  int zcaught = 0;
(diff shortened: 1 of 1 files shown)


zsistla:
Fixes #144 .

joshuabenuck:
What caused the `null`? If I'm reading this properly, you switched from some custom code to instrument `mysqli_commit` to calling `nr_php_instrument_datastore_operation_call`.

zsistla:
I built 9.16 with just the mysqli commit, then narrowed it down to that area of code.  First time calling a php file would work (and 1000s of commits could happen successfully, but on the 2nd call NULL/unsuccessful commit occured). I believe the params were eventually getting set incorrectly causing the error to occur on callback after the 2nd call to a php file.  The `nr_php_instrument_datastore_operation_call` handles this more correctly for the `commit` operation.  After the 2nd call,  NULL no longer occurs, and the failure to commit no longer occurs.  Tested across 1000s of PHP calls calling 100s of commits.

joshuabenuck (approved):

Fahmy-Mohammed (approved):

Webonaute:
This make me completely nuts in the last 2 days! glad you fixed this!

PR #162, opened 2021-05-07
Initial dockerized dev environment prototype.

zsistla:
This is a prototype of a dockerized development environment that would allow contributors to both develop and test (using unit tests AND integration tests) without having to set up a specific environment on their own system.  Integration tests send actual data to the New Relic backend and therefore requires a license key (Sign up at https://newrelic.com/signup for a free account.).

Currently docker-compose spins up `mysql` and `redis` databases in separate containers.

Two environment variables to note:
`NEWRELIC_LICENSE_KEY` is required to run the integration tests and should be set to your NR license key.
If your collector isn’t the default (collector.newrelic.com), set the `NEWRELIC_COLLECTOR_HOST` to the appropriate value.

PHP_VER can also be set, but hasn’t been tested beyond `8.0`.

Set all environment variables prior to running the development environment.

To start the dev environment type `make dev-start`.  This will create a set of docker containers.
A prompt will open and you’ll be able to compile and run all `make` commands right away with no additional setup (for example: `make -j4 all` or `make -j4 valgrind` or `make -j4 run_tests`).

After compiling the agent, the integration tests can be run using the `integration_runner`.
To run all integration tests, from the prompt, run: `./bin/integration_runner -agent ./agent/.libs/newrelic.so`
To run `redis` only: `./bin/integration_runner -agent ./agent/.libs/newrelic.so -pattern tests/integration/redis/*`

To end the session type `exit`.  You can run `make dev-stop` to stop the docker-compose containers.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,48 @@
+#
+# Copyright 2021 New Relic Corporation. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+version: '3.7'
+services:
+# The Database
+  mysqldb:
+    image: mysql:5.7
+    restart: always
+    environment:
+      MYSQL_ROOT_PASSWORD: admin
+      MYSQL_DATABASE: database
+      MYSQL_USER: admin
+      MYSQL_PASSWORD: admin
+      MYSQL_HOST: mysqldb
+    ports:
+      - "33061:3306"
+    container_name: mysqldb
+  redisdb:
+    build:
+      context: .
+      dockerfile: files/Dockerfile.redis    
+    restart: always
+    ports:
+      - "6379:6379"
+    container_name: redisdb
+  agent:
+    build:
+      context: .
+      dockerfile: files/Dockerfile
+    user: ${UID}:${GID}
+    environment:
+      MYSQL_DB: database
+      MYSQL_USER: root
+      MYSQL_PASSWD: admin
+      MYSQL_HOST: mysqldb
+      REDIS_HOST: redisdb
+      PS1: "New Relic > "      
+      NEWRELIC_COLLECTOR_HOST : ${NEWRELIC_COLLECTOR_HOST:-collector.newrelic.com}
+      NEWRELIC_LICENSE_KEY: ${NEWRELIC_LICENSE_KEY}
+    volumes:
+#      - ./:/usr/src/myapp
+       - ${PWD}:/usr/src/myapp
+    working_dir: /usr/src/myapp 
+    stdin_open: true
+    tty: true
+    container_name: agent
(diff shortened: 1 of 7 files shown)


files/Dockerfile
```diff
@@ -0,0 +1,116 @@
...
+#
+# If the debian version is jessie or stretch, we need to install go manually;
+# otherwise, we just install the golang package from the repository.
+# go 1.11.6 matches the version that buster uses.
+#
+RUN if [ -z "$(grep '^10\.' /etc/debian_version)" ]; then \
+      wget --quiet https://golang.org/dl/go1.11.6.linux-amd64.tar.gz -O- | tar -C /usr/local -zxvf -; \
+      export GOROOT=/usr/local/go; \
+      exportGOPATH="${HOME}/go"; \
+      exportPATH="${GOPATH}/bin:${GOROOT}/bin:${PATH}"; \
```
kneitinger, line 95: Spaces needed here

zsistla: :) Good catch, thanks!

docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
...
+      MYSQL_DB: database
+      MYSQL_USER: root
+      MYSQL_PASSWD: admin
+      MYSQL_HOST: mysqldb
+      REDIS_HOST: redisdb
+      PS1: "New Relic > "      
+      NEWRELIC_COLLECTOR_HOST : ${NEWRELIC_COLLECTOR_HOST:-collector.newrelic.com}
+      NEWRELIC_LICENSE_KEY: ${NEWRELIC_LICENSE_KEY}
+    volumes:
+#      - ./:/usr/src/myapp
```
joshuabenuck, line 43: Delete?

files/Dockerfile
```diff
@@ -0,0 +1,116 @@
...
+ARG PHP_VER
+
+FROM php:${PHP_VER:-8.0}
+
+RUN docker-php-source extract
+
+#
+# The sql tests produce partition/filtered but tests can't currently be updated 
+# to reflect that since the sql database on CI doesn't have those values.
+# RUN docker-php-ext-install pdo pdo_mysql
```
joshuabenuck, line 19:
```suggestion
# The explain plan in the sql tests contain partition/filtered properties
# but the tests can't currently be updated  to reflect that since the sql 
# database on our CI doesn't include those values.
# RUN docker-php-ext-install pdo pdo_mysql
```

zsistla: Sounds good, thanks.

joshuabenuck (approved):

kneitinger (approved):
Awesome job Amber!

Issue #86, opened 2021-01-21
Laravel 7 - transaction names saved as UUIDs rather than class & function


darcyrailip:
## Description

Seems like transaction names are not being saved ever since upgrading to Laravel 7.

Instead, they appear like this: `generated::AhiRBYxyLAmfeypt`.

## Steps to Reproduce

Install the agent on a server running Laravel 7.

## Expected Behavior

I would expect to see the full class name / path + function being called.

## Additional context

I opened a ticket on the discussion forums here:

https://discuss.newrelic.com/t/php-transactions-no-longer-showing-correct-namespace-class-function-details/121752/6

Someone answered mentioning that Laravel 7 upgrade may be the issue, but I also see that Laravel 7 is on the list of supported frameworks:

https://docs.newrelic.com/docs/agents/php-agent/getting-started/php-agent-compatibility-requirements#frameworks

> Laravel 4.x, 5.x, 6.x, and 7.x

darcyrailip:
Hey @Fahmy-Mohammed, I see you've added the `backlog` tag.

Is this confirmed to be a bug?

I'm just a bit confused since [the release notes say support for Laravel 7 was explicitly added in version 9.13.0.270.](https://docs.newrelic.com/docs/release-notes/agent-release-notes/php-release-notes/php-agent-9130270)

Surely someone else is running into this?

zsistla:
Hi @darcyrailip, 

Thank you for your engagement!

I know a workaround was listed on the forum post you cited (did it work for you?), but we’d like to try to work through this issue with you.  Unfortunately we are not able to reproduce it on a Laravel 7 setup yet:
![image](https://user-images.githubusercontent.com/43715151/109900660-968b4380-7c4c-11eb-8f19-b0429b68f2a7.png)
  
Could you provide a few more details about your environment:
1) What version of the PHP Agent are you using?
2) What version of PHP are you using?
3) Are there any other environmental details you can provide?  (OS used, Dockerized container, etc).
4) Are all Laravel transaction names showing strangely for you, or only specific ones?  

A Laravel txn name can be [set](https://github.com/newrelic/newrelic-php-agent/blob/3f93ee47f80703d46d8fccd53be7d6b80361a594/agent/fw_laravel.c#L294) in various ways.

Would it be possible to post some agent and daemon logs (or log snippets, sanitized for your privacy) to get a better idea what is going?

To gather detailed log information for New Relic's PHP agent:
[...]

PR #174, opened 2021-06-07
Fix laravel transaction naming (issue #86) - Don't use getName() if it's a auto generated route name

stockalexander:
Route::getName() returns a random generated string (generated::*) if a route isn't named [(Named Routes)](https://laravel.com/docs/8.x/routing#named-routes) and routes are cached via `php artisan route:cache` since laravel 7.x.

This causes the problem mentioned in #86. To fix the problem, I added a check if the returned string begins with `generated::`, to skip it and use `Route::getAction()` instead.

### DIFF
--- a/agent/fw_laravel.c
+++ b/agent/fw_laravel.c
@@ -347,24 +347,37 @@ static void nr_laravel_name_transaction(zval* router, zval* request TSRMLS_DC) {
 
       route_name_zv = nr_php_call(route, "getName");
       if (nr_php_is_zval_valid_string(route_name_zv)) {
-        char* route_name
-            = nr_strndup(Z_STRVAL_P(route_name_zv), Z_STRLEN_P(route_name_zv));
+        const char generated_prefix[] = "generated::";
+        nr_string_len_t generated_prefix_len
+            = sizeof(generated_prefix) - 1;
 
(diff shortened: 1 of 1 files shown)


Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed (dismissed):
Thanks for adding this fix @stockalexander. I've tested this locally and it works well! This should be included in the upcoming PHP agent release (version 9.18).

Fahmy-Mohammed:
@stockalexander before we can merge this, would you be able to change this to the dev branch instead of main? That is where all of the work included in the upcoming release is contained.

stockalexander:
@Fahmy-Mohammed: Nice! I rebased my branch to the dev branch (+ force pushed it) and updated the PR. Can you merge it now or should I change something else?

Fahmy-Mohammed:
@stockalexander There are some internal Jenkins builds that are running currently, so I'll merge this once those tests pass. Awesome work identifying and adding the fix. We really appreciate it!

Fahmy-Mohammed:
ok jenkins

Fahmy-Mohammed (approved):

Fahmy-Mohammed:
ok jenkins

PR #176, opened 2021-06-23
Remove Drupal NegativeModuleLength Supportability metric

joshuabenuck:
This PR removes the code that attempts to create a `NegativeModuleLength` Supportability metric. The intention behind this metric was to record how often a certain edge case was encountered. Unfortunately, it appears to cause a segfault on at least some systems. The simplest approach here is to just remove the metric as it is currently causing more harm than good.


joshuabenuck:
ok jenkins

zsistla (approved):

Fahmy-Mohammed (approved):

PR #181, opened 2021-08-03
version bump to 9.18

Fahmy-Mohammed:
Bump up the agent version to 9.18.0 "yam"


axiom/nr_version.c
```diff
@@ -30,7 +30,7 @@
  *   watermelon         25Jan2021 (9.16)
  *   xigua              21Apr2021 (9.17)
```
zsistla, line 31: add it here, too?

zsistla (approved):

PR #252, opened 2021-10-08
fix broken Integration tests on the dev_environment

ZNeumann:


docs/dev_environment.md
```diff
@@ -60,12 +60,4 @@ In the shell, you can run all `make` commands as you normally would.
...
-## There are currently 5 known failing integration tests with this prototype.
-
-not ok - tests/integration/errors/test_E_WARNING.php8.php
-not ok - tests/integration/errors/test_ignore_all_except_E_WARNING.php8.php
-not ok - tests/integration/external/curl_exec/test_cat_response_header_anonymous.php
-not ok - tests/integration/external/curl_exec/test_cat_response_header_to_file.php
-not ok - tests/integration/api/internal/test_get_request_metadata_cat.php
-
-Please ignore (or fix) these errors for now.
+## There are currently 0 known failing integration tests with this prototype.
```
ZNeumann, line 63: The untouched tests passed when I ran them. Is this just outdated documentation?

tests/integration/errors/test_E_WARNING.php8.php
```diff
@@ -31,7 +31,7 @@
     [
       "?? when",
       "OtherTransaction/php__FILE__",
-      "include(): Failed opening 'abc.php' for inclusion (include_path='.:')",
+      "/include\\(\\): Failed opening 'abc.php' for inclusion \\(include_path='.:.*'\\)/",
```
ZNeumann, line 34: adding the '/' as a prefix and suffix makes the integration runner read the string as regex.

adding '\\' before the "()" escapes the "()" for the regex string. Double backslash is needed because first the string is parsed through the go json library, which will condense the escaped '\\' to just a '\'

Regex is needed to add a ".*" in the include_path, as the dev environment has an additional path on its include_path

zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla (approved):
Looks great, thanks!

Issue #113, opened 2021-02-25
Symfony5 support


b-viguier:
## Summary

The NR php agent supports [Symfony framework](https://symfony.com/) from version [1 to 4](https://docs.newrelic.com/docs/agents/php-agent/configuration/php-agent-configuration#inivar-framework). Since version 5, some breaking changes may prevent NR extension to report all expected information. Is SF5 support in your Roadmap?

## Desired Behavior

* Auto-detect SF5 framework usage
* Report exceptions

## Possible Solution

SF5 replaces the `ExceptionListener` class by `ErrorListener` (see [UPGRADE](https://github.com/symfony/symfony/blob/adbb34115c82690ab178eb9599891f4352d0ac20/UPGRADE-5.0.md#httpkernel) instructions). Unfortunately, NR extension is explicitly looking for `ExceptionListener` class, to wrap it and report exceptions.
https://github.com/newrelic/newrelic-php-agent/blob/3f93ee47f80703d46d8fccd53be7d6b80361a594/agent/fw_symfony4.c#L184-L187
By detecting SF5 framework, we could change wrapped function… if I understand the code correctly 😅 

## Additional context

None.

Thanks for your work 👍

joaojacome:
IMO the framework-specific functions should be delegated to a composer package, instead of locking the frameworks in the extension.

piotrantosik:
Hi, anyone tries setup newrelic with symfony 5? It's works when I set `newrelic.framework = "symfony4"`?
Right now I use with `newrelic.preload_framework_library_detection = false` but in this setup I don't see proper transactions.

b-viguier:
@piotrantosik We are using https://github.com/ekino/EkinoNewRelicBundle, it's a little redundant with built-in framework support, but at least transaction names are correctly handled

Ak-x:
Closing this issue per https://github.com/newrelic/newrelic-php-agent/pull/256

PR #256, opened 2021-10-19
Add Symfony 5 support

piotrantosik:
HI,

I try to introduce Symfony 5 auto-detect. I prepare this code, but totally I don't have an idea how to check locally if it's working properly, because agent built in dev-shell requires `libpcre.so.3` that isn't available on alpine that I use for docker container.
Also, dev-env don't work on latest dev branch, I remove `python-yaml` library from dockerfile as was done on GH pipelines - https://github.com/newrelic/newrelic-php-agent/commit/2d2e5f374ac08680f7242fa7a9b7b6bdefd5aa0f

This PR fix https://github.com/newrelic/newrelic-php-agent/issues/113

### DIFF
--- a/agent/fw_symfony4.c
+++ b/agent/fw_symfony4.c
@@ -42,11 +42,9 @@ NR_PHP_WRAPPER(nr_symfony4_exception) {
   /*
    * Get the exception from the event.
    * Firstly, we check if ExceptionEvent is available - if yes, that means we are using Symfony 5
-   */
-  if (nr_php_object_instanceof_class(
-        event, "Symfony\\Component\\HttpKernel\\Event\\ExceptionEvent" TSRMLS_CC)) {
-    exception = nr_php_call(event, "getThrowable");
-  } else {
+  */
+  exception = nr_php_call(event, "getThrowable");
+  if (!nr_php_is_zval_valid_object(exception)) {
     exception = nr_php_call(event, "getException");
   }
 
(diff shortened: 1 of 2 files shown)


zsistla:
Apologies regarding the `python-yaml` issue that occurred in the prototype due to a change that happened in a base image we pull from.  Our team has actually been updating the prototype to enable a wider variety of tests beyond the `redis` and `mysqli` ones it currently supports so the new change hadn't been rolled out yet.

piotrantosik:
build status is false negative - on my fork it's green https://github.com/piotrantosik/newrelic-php-agent/actions/runs/1360606524

zsistla:
ok jenkins

agent/fw_symfony4.c
```diff
@@ -39,8 +39,17 @@ NR_PHP_WRAPPER(nr_symfony4_exception) {
...
   }
 
-  /* Get the exception from the event. */
-  exception = nr_php_call(event, "getException");
+  /*
+   * Get the exception from the event.
+   * Firstly, we check if ExceptionEvent is available - if yes, that means we are using Symfony 5
+   */
+  if (nr_php_object_instanceof_class(
+        event, "Symfony\\Component\\HttpKernel\\Event\\ExceptionEvent" TSRMLS_CC)) {
```
zsistla, line 47: Thanks again for your work on this.

This might not be the best differentiator since the ExceptionEvent class was introduced in Symfony 4.3. .


What do you think about if we try something simpler like the following to cover both cases so that we handle it either way:
```
exception = nr_php_call(event, "getThrowable");
if (!nr_php_is_zval_valid_object(exception)) {
    exception = nr_php_call(event, "getException");
  }

```
That way if it is a valid throwable we get that; otherwise, we try for the exception.

piotrantosik:
Thanks, I found hax & tricky solution but it's possible to test built extension in dev shell. I'm back with updated PR at the weekend, I also try to add a named CLI transactions.

piotrantosik:
hi @zsistla, I want to check if everything from my side it's ok, all tests are passed?

ZNeumann (approved):

mfulb:
Thank you for the contribution @piotrantosik!

PR #277, opened 2021-11-08
Support for running integration tests as docker containers [DO NOT MERGE]

mfulb:
This PR supports running the integration tests in a docker container using services running in containers as well.

An 'agent' container is created which is based on a php base image and then has the following PHP extensions installed:
- redis
- predis
- memcache
- memcached
- pgsql
- pdo/pdo_mysql
- mysqli
- uopz

The following services also have containers created:
- mysql (container name 'mysqldb')
- postgres (container name 'postgres')
- redis (container name 'redisdb')
- memcached (container name 'memcached')


mfulb:
This is an old effort which has been accomplished by more recent work - was left open as a reference but it is not needed now.

PR #318, opened 2021-12-08
Curl valgrind issue

ZNeumann:
This PR addresses an issue that was left unnoticed from the transition to PHP 8.0. There was still an instance of using the curl handle as a resource instead of an object. This issue was found when exploring the upgrade to PHP 8.1.

### DIFF
--- a/agent/php_curl_md.c
+++ b/agent/php_curl_md.c
@@ -7,13 +7,6 @@
 #include "php_curl_md.h"
 #include "util_logging.h"
 
-static void nr_php_curl_md_destroy(nr_php_curl_md_t* metadata) {
-  nr_php_zval_free(&metadata->outbound_headers);
-  nr_free(metadata->method);
-  nr_free(metadata->response_header);
-  nr_free(metadata);
-}
-
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
 #define check_curl_handle(ch)                                             \
   ({                                                                      \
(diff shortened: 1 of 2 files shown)


agent/php_curl_md.c
```diff
@@ -42,76 +35,142 @@ static void nr_php_curl_md_destroy(nr_php_curl_md_t* metadata) {
...
+}
+
+static void curl_handle_vector_dtor(void* element, void* userdata NRUNUSED) {
+  zval* handle = (zval*)element;
+
+  if (!handle) {
+    return;
   }
 
-#define ensure_curl_multi_metadata_hashmap()                   \
```
ZNeumann, line 51: I couldn't find a reason why several of these macros were used instead of functions. For easier readability- as well as less code duplication when adding the ifdefs to fix the issue- I changed them to functions. I also moved some functions around so that definitions were in the correct order.

agent/php_zval.h
```diff
@@ -285,17 +285,35 @@ static inline bool nr_php_is_zval_null(const zval* z) {
...
  *
- * Warning : The type of the given zval is not checked -- if it is not a
- *           resource, you will probably get segfaults!
  */
 static inline long nr_php_zval_resource_id(const zval* zv) {
+  if (!nr_php_is_zval_valid_resource(zv))
+  {
+      return 0;
+  }
 #ifdef PHP7
```
zsistla, line 294: Can you please update the style for readability:
`#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */`

Currently PHP7 actually indicates [PHP7+](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_compat.h#L13).

agent/php_curl_md.c
```diff
@@ -42,76 +35,142 @@ static void nr_php_curl_md_destroy(nr_php_curl_md_t* metadata) {
...
+    uint64_t id = (uint64_t)nr_php_zval_object_id(mh);
+#else
+    uint64_t id = (uint64_t)nr_php_zval_resource_id(mh);
+#endif
+    if (id == 0)
+    {
+        return NULL;
+    }
+    nr_php_curl_multi_md_t* multi_metadata;
+    size_t async_index;
```
zsistla, line 129: Now that the id check is happening [here](https://github.com/newrelic/newrelic-php-agent/pull/318/files#diff-6340c4a86ab2af8769a24114fe8dc973281e36f3b5d98aab7cb2a16834eca284R124), can we move the variable definitions up to the top of the function?

agent/php_curl_md.c
```diff
@@ -42,76 +35,142 @@ static void nr_php_curl_md_destroy(nr_php_curl_md_t* metadata) {
...
-    __multi_metadata;                                                   \
-  })
+static nr_php_curl_md_t* get_curl_metadata(const zval* ch TSRMLS_DC)
+{
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 8.0+ */
+    uint64_t id = (uint64_t)nr_php_zval_object_id(ch);
+#else
+    uint64_t id = (uint64_t)nr_php_zval_resource_id(ch);
+#endif
+    if (id == 0)
```
zsistla, line 100: Flip around for `yoda` conditions ([styleguide](https://github.com/newrelic/newrelic-php-agent/blob/main/STYLEGUIDE.md)).

Same comment for [here](https://github.com/newrelic/newrelic-php-agent/pull/318/files#diff-6340c4a86ab2af8769a24114fe8dc973281e36f3b5d98aab7cb2a16834eca284R124).

zsistla:
Aside from a few nits, it looks great!  Thanks for finding/fixing this.  It'll be great to get this fix merged.

zsistla:
Looks like jenkins is showing some errors for the PHP 5.5 build which is probably why they were initially macros, but it's still fixable for the functions.  I think you can just take the TSRM value you get [here](https://github.com/newrelic/newrelic-php-agent/blob/a3f9a4bfcc72795e3f14b4b371373a6ee25e975a/agent/php_curl_md.c#L207) and pass it into your new functions that have TSRM in the signatures.

mfulb (approved):

zsistla (approved):

Issue #321, opened 2021-12-22
Compiling against PHP 8.1 with gcc 9/10/11 under Ubuntu 20.04 LTS will fail


GwynethLlewelyn:
When attempting to compile the agent in Ubuntu 20.04 LTS using `gcc` (as opposed to using `clang`) will fail, because flags such as `-Wno-typedef-redefinition` aren't supported, and since the default is to flag all warnings as errors, an 'unsupported flag' is treated as such, thus preventing a successful compilation to finish.

## Description

Recent releases of PHP 8.1 unexpectedly have introduced some slight code sloppiness, which percolates up to the macros used in New Relic's PHP agent. `clang` seems not to mind much, as it respects the `-Wno-typedef-redefinition` flag, and, as such, considers this sloppiness merely a warning, and finishes compilation.

By sheer chance, in the past, I had successfully compiled the PHP agent for 7.3, 7.4 and 8.0, without any errors whatsoever, and hadn't even noticed which compiler was being used. I _do_ have _both_ `clang` and `gcc` installed, but I hadn't noticed that I systematically used `gcc` by default — and an end-of-life version, too (9.3.0, to be more precise). Since it did the job, I didn't notice anything wrong.

However, when attempting to compile the agent for PHP 8.1 under Ubuntu 20.04.3 LTS, there were a handful of errors, which could be clearly traced to flags that the compiler didn't recognise (`-Wno-typedef-redefinition` being the one that stuck in my head, but there were a handful of others). Apparently, when `gcc` encounters a flag that it doesn't accept, **and** a warning is triggered (for whatever reason — in my case, there were some inconsistencies with the declaration of functions and/or possible lack of array boundary checks), **then** `gcc` notices that it has an unsupported flag, and, because the default is to treat all warnings as errors, throws a fatal error and stops compilation. Interestingly, the compiler seems to 'understand' that the flag _is_ there, and assumes that it _may_ be overriding some warnings, but, since the flag is _not_ being recognised, `gcc` becomes very strict and terminates compilation, assuming that an 'unknown' flag that potentially masks a warning is sufficiently problematic to deal with — letting the programmer review what exactly was meant by throwing an error.

[...]

PR #322, opened 2021-12-22
Dirty fix for #321

GwynethLlewelyn:
This is my crude attempt to fix the `gcc` compiler issue fully described under #321. I'm well aware that this is a very quick & dirty hack that could be done professionally much better. But at least I hope it gives you guys some clues on how to do a permanent fix.

I do apologise for the 'commit pollution'; I was a bit stressed out with lack of time, and I did a lot of mistakes that had to be reverted, and so forth. I suppose I could rebase things before submitting the PR, but, alas, I forgot to do that as well.

Also note that this took me an eternity to figure out _what_ was wrong, and, most importantly, _how_ to fix it easily in a way that does not get broken with subsequent versions of the NR PHP agent (or of PHP itself). Being the first time I actually delve deep in NR's code, I made dozens of futile attempts, until I settled in what is but a handful of lines being changed.

Please note that _no_ NR code was actually changed! In fact, I just needed to hack the building configuration to get this to compile under `gcc` as well as `clang`. I suspect that _most_ issues are actually upstream (e.g. in the PHP 8.1 code, or the `phpize` code, or even in the way `gcc` works, which is subtly different from `clang`) and that NR's code is fine; but a deeper analysis of NR's code _may_ allow to circumvent the warnings thrown by `gcc` — that's a bit beyond my ability (time-wise) to fix.

My 'hack' — if I may even call it _that_ — relies on a single assumption: `gcc` and `clang` behave subtly differently when forcing warnings to be thrown as compilation errors when using `-Werror`. In particular, the `-Wno-typedef-redefinition` seems to be missing under `gcc`, and, as such, causes a problem when `gcc` throws a warning and, not recognising that flag, is unsure if it's masking a known condition or not; apparently, in those cases, `gcc` throws an error instead (what was surprising to me is that `gcc` will work flawlessly if there are _no_ warnings to throw, e.g. these flags are only checked _if_ a warning is found, as opposed to what I would think of the more obvious approach, which would be to _immediately_ throw an error when parsing the flag and not recognising it — *not* attempt a full compilation first!).

So, when generating the `Makefile` for the NR PHP agent, at some step, a check has to be introduced to differentiate `clang` from `gcc`, and allow some flags only under `clang`, others under `gcc`.

This proved to be slightly harder than expected. The actual code that checks among different vendors/operating systems etc. is not under the control of the New Relic development team, but, rather, it gets automatically injected by `phpize`, which creates a temporary `build/` directory with whatever macros/scripts are necessary. By default, autoconf doesn't seem to include an _easy_ way to distinguish between `clang` from `gcc`, as _both_ claim to be 'GNU C'-compatible. Fortunately,  the [GNU Autoconf Archive contains a macro that will do the job](https://www.gnu.org/software/autoconf-archive/ax_compiler_vendor.html). As I didn't see any obvious place to put that file, for now, it resides on the root of the `agent/` subdirectory (there might be better places for that). `config.m4` was subtly changed, on the part that deals with setting New Relic-specific flags, to include that macro and check for `clang` vs `gcc`. After testing just with excluding `-Wno-typedef-redefinition` from the `gcc` flags, I noticed that a few more were not being recognised, so, to get a successful compilation under Ubuntu Linux as well as on my MacBook Pro running macOS Big Sur, I just got rid of `-Werror` for the whole `gcc` compilation chain.

This is hardly the 'right' approach, but my reasoning is that _most_ people in the New Relic development team are using `clang` exclusively, so they will catch any 'serious' errors anyway. It's just for the few of us who are still using `gcc` for some reason (in my case it's just because it's the only compiler available for the Synology NAS variety of Linux/arm64) that this becomes an issue. Of course, sensible programmers will simply cross-compile for different operating systems and vendors, or use Docker containers, or virtual machines, whatever they fancy — it's just oldtimers like me that still insist in working with whatever the bare metal gives you :-)

Although not strictly related to this issue, I've also submitted a URL change for the API documentation (it was still pointing to an outdated link, I put one that at least works, even if it might not point to the same content... I suppose I could have checked on the Wayback Machine, but I didn't). Also, because the compilation chain includes building the Go daemon, there is an issue when using modern Go versions, which will assume, by default, that modules are used. The New Relic Go daemon seems to have been written with Go 1.9 in mind, when a vendoring system was used instead. To address successful compilations under recent versions of Go, I've explicitly set `GO111MODULE` to `auto` — this will cause recent versions of Go to fall back to vendoring instead of modules, _iff_ no valid `go.mod` file is found on the root of the project. While not relevant to the PHP agent _per se_, it always annoys me when a compilation finishes with an (unrelated) error, so I fixed that as well.


GwynethLlewelyn:
Bump, bumpety, bump?

zsistla:
ok jenkins

GwynethLlewelyn:
Huh. I admit I'm stumped; thanks for re-running the tests (at least this time they finished), but now I have to figure out what the test error messages actually _mean_.

Makefile
```diff
@@ -22,6 +22,8 @@ include make/version.mk
 # Configure an isolated workspace for the Go daemon.
 export GOPATH=$(CURDIR)
 export GO15VENDOREXPERIMENT=1
+# Needed for Go > 1.11 to avoid building with modules by default (and failing!)
+export GO111MODULE=auto
```
zsistla, line 26: Great idea!

agent/config.m4
```diff
@@ -173,8 +173,20 @@ if test "$PHP_NEWRELIC" = "yes"; then
...
+    m4_include([ax_compiler_vendor.m4])
+    AX_COMPILER_VENDOR
+
+    if test "x$ax_cv_cxx_compiler_vendor" = "xclang"; then
+      dnl The flag below only works under clang, but not gcc
+      NEWRELIC_CFLAGS="${NEWRELIC_CFLAGS} -Wno-typedef-redefinition"
+      NEWRELIC_CFLAGS="${NEWRELIC_CFLAGS} -Werror"
+    else
+      dnl By contrast, the next flag only works under gcc
+      NEWRELIC_CFLAGS="${NEWRELIC_CFLAGS} -Wstringop-overflow"
```
zsistla, line 188: Unfortunately, this flag seems to be causing an[ issue on the 7.0 build](https://github.com/newrelic/newrelic-php-agent/runs/4821711049?check_suite_focus=true).

```
cc: error: unrecognized command line option '-Wstringop-overflow'; did you mean '-Wstrict-overflow'?
```

GwynethLlewelyn: Dang! Oh well. You guys can always put a note on the `README.md` instead of dealing with this PR...

bduranleau-nr:
Closing - many of the issues this PR addresses have been implemented since it was opened, including gomodule changes and gcc compiler changes for newer gcc version support. 
If you find that the latest New Relic PHP agent is still experiencing any of the issues described above, please open a GitHub issue on the repo containing a description of the problem and desired solution and a member of the team will reach out to assist.

GwynethLlewelyn:
You're quite right in closing it! Thanks @bduranleau-nr  😃

Issue #312, opened 2021-12-08
[PHP] Release builder: update hudson script to include PHP 8.1


mfulb:
Details in MMF documentation.

zsistla:
Part of #327

PR #328, opened 2022-02-07
Release build updates

zsistla:
Added support for PHP 8.1 release to `Makefile` and `make\release`.
Modified documentation.

### DIFF
--- a/make/release.mk
+++ b/make/release.mk
@@ -88,7 +88,7 @@ endif
 	$(MAKE) agent-clean; $(MAKE) release-5.4-no-zts
 	$(MAKE) agent-clean; $(MAKE) release-5.3-no-zts
 ifeq (x64,$(ARCH))
-        $(MAKE) agent-clean; $(MAKE) release-8.0-zts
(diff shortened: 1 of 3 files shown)


zsistla:
Fixes #312

make/release.mk
```diff
@@ -139,6 +141,7 @@ release-$1-zts: Makefile agent | releases/$$(RELEASE_OS)/agent/$$(RELEASE_ARCH)/
 
 endef
```
mfulb, line 142: I think this should be "endif"?

zsistla: When a define includes multiple lines, endef is used to show that the define is ending.
See here for more details: https://www.gnu.org/software/make/manual/html_node/Multi_002dLine.html

zsistla:
Dependent on a hudson PR being merged.  Rerun tests after that is merged.

make/release.mk
mfulb: Ah thanks. :)

ZNeumann (approved):

mfulb:
ok jenkins

ZNeumann:
ok jenkins

PR #329, opened 2022-02-08
test changes required to pass on a modern freebsd system

ZNeumann:
These tests failed when run on FreeBSD 13. The changes to the realloc are because the BSD provided alloc would fail when assigning to 0 size.

FreeBSD 10 is used as a test system, while reallocarray() was added to the system in FreeBSD 11. Now that we are adding more modern freebsd systems to the pipeline, using the BSD provided reallocarray() would be inconsistent with other OS's that use the New Relic realloc code. A particular inconsistency is for reallocarray() down to 0 size.


tests/integration/external/curl_multi_exec/test_malformed_url.php
```diff
@@ -30,8 +30,8 @@
     [{"name":"External/<unknown>/all"},                          [1, "??", "??", "??", "??", "??"]],
     [{"name":"External/<unknown>/all",
       "scope":"OtherTransaction/php__FILE__"},                   [1, "??", "??", "??", "??", "??"]],
-    [{"name":"External/19/all"},                                 [1, "??", "??", "??", "??", "??"]],
-    [{"name":"External/19/all",
+    [{"name":"External/(0\\.0\\.0\\.)?19/all"},                  [1, "??", "??", "??", "??", "??"]],
```
ZNeumann, line 33: double escape needed because the regex is first put through a json parser

zsistla:
Looks good, just for completeness, could you put a link in the description regarding when the FreeBSD alloc changed (and hence why it works with freebsd10 but not 13).

zsistla:
ok jenkins

mfulb:
ok jenkins

PR #331, opened 2022-02-11
Updates to support building PHP 8.1 and installer script

mfulb:
Changes:
- agent/newrelic-install.sh:
  - Remove references to 5.3 and 5.4 and add 8.1 as required.
  - Rework some logic which assumed the only possible 8.x release was 8.0 to work now we have 8.0 and 8.1.

- make/release.mk
  - Add guards so 8.1 is only built on 64 bit systems.
  - Remove building for 5.3 and 5.4.

Makefile
 - Remove 5.3 and 5.4 from lists for version to run for integration and LASP tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,7 +316,7 @@ src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go: protocol/infinite_
 
 .PHONY: integration
 integration: Makefile daemon lasp-test-all
-	for PHP in $${PHPS:-8.1 8.0 7.4 7.3 7.2 7.1 7.0 5.6 5.5 5.4 5.3}; do \
+	for PHP in $${PHPS:-8.1 8.0 7.4 7.3 7.2 7.1 7.0 5.6 5.5}; do \
           echo; echo "# PHP=$${PHP}"; \
 	  env NRLAMP_PHP=$${PHP} bin/integration_runner $(INTEGRATION_ARGS) || exit 1; \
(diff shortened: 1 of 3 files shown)


zsistla:
ok jenkins

zsistla:
Nice changes, thanks!

zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

mfulb:
ok jenkins

ZNeumann (approved):

mfulb:
I've tested the installation script (from tarball) for PHP 5.5 through 8.1 and verified that a test PHP script was able to send data to NR1.
I also tested installing agent V 9.18.1.303 on PHP 5.3 and 5.4 systems and then running the new installer.
One these systems the agent did not overwrite the existing `newrelic.so` and output a message indicating it was not going to proceed as an unsupported PHP version was present.
The `newrelic-daemon` in `/usr/bin` **was** updated however.  This appears to be the previous behavior of the script.

zsistla (approved):

mfulb:
I have also tested installing the agent on a 32 bit OS (Ubuntu 16.04) with PHP 7.0, PHP 8.0 and PHP 8.1 installed.  The installer script ignored the PHP 8.0 and 8.1 installation since there is no 32bit agent support for 8.0+.  The agent extension was installed for the PHP 7.0 installation however.

PR #335, opened 2022-02-16
Disable 32bit builds for PHP 8.x in GitHub workflow [DO NOT MERGE - IGNORE]

mfulb:
The agent does not support 32 bit environments running PHP 8.x so we can disable these test builds.


mfulb:
Ignore improperly constructed PR...

PR #336, opened 2022-02-16
Disable 32bit builds for PHP 8.x in GitHub workflow

mfulb:
The agent does not support 32 bit environments running PHP 8.x so we can disable these test builds.

This is the corrected version of PR #335 which was improperly setup targeting 'main' instead of the 'dev' branch.

### DIFF
--- a/.github/workflows/agent-build.yml
+++ b/.github/workflows/agent-build.yml
@@ -42,6 +42,7 @@ jobs:
         arch: ['x64', 'x86']
         exclude:
         # Excludes PHP 5.x, 7.x on linux 32-bit
+	# Excludes PHP 8.x as only 64bit is supported for these versions
         - os: linux
           php_ver: '7.0'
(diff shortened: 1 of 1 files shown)


.github/workflows/agent-build.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         arch: ['x64', 'x86']
         exclude:
         # Excludes PHP 5.x, 7.x on linux 32-bit
+        # Excludes PHP 8.x as only 64bit is supported for these versions
```
zsistla, line 45: Great start.  Can you please exclude `8.0-zts` and `8.1-zts` as well?

mfulb: Yes I'll fix that.

zsistla (approved):

ZNeumann (approved):

PR #337, opened 2022-02-16
Parametization of release.mk

zsistla:
If no versions are supplied, it defaults to all the PHP versions we currently support.
It automatically removes PHP versions that are not supported on the architecture.

This change was made:
* to give more flexibility as PHP versions become deprecated/EOLed on different architectures.
* to  simplifies the release.mk file
* to allow the different versions of PHP to be built depending on build node (e.g. FreeBSD10 vs FreeBSD 13).


mfulb (approved):
Nice improvement much easier to maintain and better flexibility.

ZNeumann (approved):

Issue #365, opened 2022-03-07
Test Configurable span reservoir sizes


ZNeumann:
Add tests that confirm changeability of span reservoir sizes

zsistla:
This should create test cases for the ini value created in #364 that adhere's to the agent spec's naming scheme.

There are currently 2 cases that newrelic.special.max_span_events: 
newrelic-php-agent/tests/integration/span_events/test_span_events_special_max*.php
These need to be modified for the name change and expanded.

zsistla:
Part of #206

PR #384, opened 2022-03-11
Configurable span reservoir size

ZNeumann:
Adjust span reservoir variable to conform to the agent spec and be in the normal newrelic.ini
Change default span_events_max_samples_stored to be 2000
change max span_events_max_samples_stored to be 10000
adjust tests to exercise new value

### DIFF
--- a/axiom/nr_limits.h
+++ b/axiom/nr_limits.h
@@ -33,9 +33,9 @@
 #define NR_MAX_SEGMENTS 2000
 
 /*
- * The maximum number of span events in a transaction.
+ * The default maximum number of span events in a transaction.
  */
-#define NR_MAX_SPAN_EVENTS 1000
+#define NR_SPAN_EVENTS_DEFAULT_MAX_SAMPLES_STORED 2000
 
 /*
  * The maximum number of span events in an 8T span batch.
--- a/tests/integration/span_events/test_span_events_special_max_disabled.php
+++ b/tests/integration/span_events/test_span_events_special_max_disabled.php
@@ -11,7 +11,7 @@
 newrelic.distributed_tracing_enabled = 1
 newrelic.transaction_tracer.threshold = 0
 newrelic.cross_application_tracer.enabled = false
-newrelic.special.max_span_events = 0
+newrelic.span_events.max_samples_stored = 0
 */
 
 /*EXPECT_SPAN_EVENTS
@@ -19,13 +19,13 @@ newrelic.special.max_span_events = 0
   "?? agent run id",
   {
     "reservoir_size": 1000,
-    "events_seen": 1000 
+    "events_seen": 2000
   },
   "??"
 ]
  */
 
-$NEWRELIC_SPAN_EVENTS_MAX = 1000; // The agent internal maximum of span events 
+$NEWRELIC_SPAN_EVENTS_MAX = 3000; // The agent internal maximum of span events
 				  // per transaction.
 
 newrelic_add_custom_tracer('main');
--- a/tests/integration/span_events/test_span_events_special_max_disabled.php
+++ b/tests/integration/span_events/test_span_events_special_max_disabled.php
@@ -18,7 +18,7 @@ newrelic.span_events.max_samples_stored = 0
 [
   "?? agent run id",
   {
-    "reservoir_size": 1000,
+    "reservoir_size": 10000,
     "events_seen": 2000
   },
   "??"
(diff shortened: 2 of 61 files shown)


ZNeumann:
ok jenkins

zsistla:
Great job!  
There are a lot of files you modified so still looking through all of them.
Could you please add a few more integration tests with out of bounds and invalid values?

zsistla:
Retracting my comment about expanding test cases as they are covered in their own ticket #365

zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

tests/integration/span_events/test_span_events_special_max_disabled.php
```diff
@@ -11,22 +11,22 @@
...
+    "reservoir_size": 10000,
+    "events_seen": 2000
   },
   "??"
 ]
  */
 
-$NEWRELIC_SPAN_EVENTS_MAX = 1000; // The agent internal maximum of span events 
-				  // per transaction.
+$NEWRELIC_SPAN_EVENTS_MAX = 3000; // The number of span events to send per transaction.
```
mfulb, line 28: Why is this 3000?  Seems like 2000 would be following the original example but since it is a max maybe that isn't important?

zsistla: This is just the number of spans that we will attempt to send.  For this test, we want it to be greater than the configured number.  The tests are going to be expanded and extended in issue #365

mfulb (approved):

misiekn (commented):
Wouldn't following macro names better reflect their purpose:
```
#define NR_SPAN_EVENTS_MAX_SAMPLES_STORED_DEFAULT 2000
#define NR_SPAN_EVENTS_MAX_SAMPLES_STORED_MAXIMUM 10000
```

axiom/nr_limits.h
```diff
@@ -33,9 +33,14 @@
...
- * The maximum number of span events in a transaction.
+ * The default maximum number of span events in a transaction.
  */
-#define NR_MAX_SPAN_EVENTS 1000
+#define NR_SPAN_EVENTS_DEFAULT_MAX_SAMPLES_STORED 2000
+
+/*
+ * The absolute maximum number of span events in a transaction.
+ */
+#define NR_SPAN_EVENTS_MAX_SAMPLES_STORED 10000
```
misiekn, line 43: Wouldn't following macro names better reflect their purpose:

```
#define NR_SPAN_EVENTS_MAX_SAMPLES_STORED_DEFAULT 2000
#define NR_SPAN_EVENTS_MAX_SAMPLES_STORED_MAXIMUM 10000
```

zsistla: Good point, it would make more sense at the end or beginning.
Looking at the file, the convention seems to be to move it to the beginning (after the NR_).
Maybe it should be something like the following to follow the convention:
```
#define NR_DEFAULT_SPAN_EVENTS_MAX_SAMPLES_STORED 2000 
#define NR_MAX_SPAN_EVENTS_MAX_SAMPLES_STORED 10000
```

mfulb: Agreed I am leaning towards the option following the existing convention with it at the beginning.

misiekn: Refactored with [5e663505](https://github.com/newrelic/newrelic-php-agent/pull/384/commits/5e66350575628bf34f3045a2b39f5ccd471a21d6).

axiom/tests/test_txn.c
```diff
@@ -6208,7 +6213,7 @@ static void test_txn_accept_distributed_trace_payload_w3c(void) {
   nrtime_t payload_timestamp_ms = 1529445826000;
   nrtime_t txn_timestamp_us = 15214458260000 * NR_TIME_DIVISOR_MS;
   nrtime_t delta_timestamp_us = nr_time_duration(
-    (payload_timestamp_ms * NR_TIME_DIVISOR_MS), txn_timestamp_us);
+      (payload_timestamp_ms * NR_TIME_DIVISOR_MS), txn_timestamp_us);
```
mfulb, line 6216: What is this character (characters?) at the front of the line that caused this diff?

mfulb: I see - thanks!

PR #391, opened 2022-03-18
Known issue where install incorrectly detects debian instead of alpin…

zsistla:
Install incorrectly labels alpine as debian when phpize dependencies are installed.

https://discuss.newrelic.com/t/alpine-linux-3-11-new-relic-php-agent-install-fails-to-detect-os-correctly-when-dpkg-installed/101360

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -120,11 +120,9 @@ error() {
 # Get the OS type.
 #
 ostype=
-if [ -f /proc/version ] && [ -n "$(grep 'Alpine' /proc/version)" ]; then
-  ostype=alpine
-elif [ -f /etc/redhat-release -o -f /etc/redhat_version ]; then
+if [ -f /etc/redhat-release -o -f /etc/redhat_version ]; then
   ostype=rhel
-elif [ -d /etc/dpkg ]; then
+elif [ -f /etc/os-release ] && grep -q -i 'Debian' /etc/os-release; then
   ostype=debian
 elif [ -f /etc/alpine-release ] || [ -d /etc/apk ]; then
   ostype=alpine
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -120,7 +120,9 @@ error() {
 # Get the OS type.
 #
 ostype=
-if [ -f /etc/redhat-release -o -f /etc/redhat_version ]; then
+if [ -f /proc/version ] && [ -n "$(grep 'Alpine' /proc/version)" ]; then
+  ostype=alpine
+elif [ -f /etc/redhat-release -o -f /etc/redhat_version ]; then
   ostype=rhel
 elif [ -d /etc/dpkg ]; then
   ostype=debian


agent/newrelic-install.sh
```diff
@@ -120,7 +120,9 @@ error() {
 # Get the OS type.
 #
 ostype=
-if [ -f /etc/redhat-release -o -f /etc/redhat_version ]; then
+if [ -f /proc/version ] && [ -n "$(grep 'Alpine' /proc/version)" ]; then
+  ostype=alpine
```
misiekn, line 124: I'm not sure this will work. Here's what I got on centos6:
```
docker run --platform linux/amd64 centos:6 grep 'Alpine' /proc/version
Linux version 5.10.76-linuxkit (root@buildkitsandbox) (gcc (Alpine 10.2.1_pre1) 10.2.1 20201203, GNU ld (GNU Binutils) 2.35.2) #1 SMP PREEMPT Mon Nov 8 11:22:26 UTC 2021
```
I think the check in line 129 is the correct method to detect if script is running on alpine distribution:
```
docker run --platform linux/amd64 alpine:latest cat /etc/alpine-release
3.15.1
```
The problem of incorrect detection of alpine stems from the fact that check for alpine is executed after the check if `/etc/dpkg` directory exists, which IMHO is not the correct way to determine which Linux distribution the script is running on. I think to correctly address this issue the check in line 127 has to be modified.

agent/newrelic-install.sh
```diff
@@ -120,7 +120,9 @@ error() {
...
+elif [ -f /etc/redhat-release -o -f /etc/redhat_version ]; then
   ostype=rhel
 elif [ -d /etc/dpkg ]; then
```
misiekn, line 127: How about this test to check if if's a Debian distribution:
```
  elif [ -f /etc/os-release ] && grep -q '^NAME="Debian.*' /etc/os-release; then
```
This was tested on all Debian releases:
```
for cn in jessie stretch buster bullseye; do echo "Testing on Debian $cn"; docker run --platform linux/amd64 debian:$cn grep '^NAME="Debian.*' /etc/os-release; done
Testing on Debian jessie
NAME="Debian GNU/Linux"
Testing on Debian stretch
NAME="Debian GNU/Linux"
Testing on Debian buster
NAME="Debian GNU/Linux"
Testing on Debian bullseye
NAME="Debian GNU/Linux"
```
This test is very strict and doesn't consider Ubuntu distribution to be a Debian distribution. If we want to roll up Ubuntu under Debian, the test can be relaxed to simply look for 'Debian' in /etc/os-release:
```
  elif [ -f /etc/os-release ] && grep -q  -i 'Debian' /etc/os-release; then
```

zsistla: Yes, please modify the check in 127 with the line that detects debian and remove the dpkg check!

`  elif [ -f /etc/os-release ] && grep -q  -i 'Debian' /etc/os-release; then`

thanks.

agent/newrelic-install.sh
misiekn: @mfulb lsb_requires program requires additional package to be installed. It seems like /etc/os-release becomes a new standard. [Here's](http://0pointer.de/blog/projects/os-release.html) some more background why using /etc/os-release to determine Linux distribution may be a better option than depending on external utility like lsb_release. A little more info on what is in os-release file: [man os-release](https://man7.org/linux/man-pages/man5/os-release.5.html).

agent/newrelic-install.sh
misiekn: Done [fb2273ea](https://github.com/newrelic/newrelic-php-agent/pull/391/commits/fb2273ea82f05eee4b94005adb9a41c1c8b5b2e1).

agent/newrelic-install.sh
misiekn: Done [fb2273ea](https://github.com/newrelic/newrelic-php-agent/pull/391/commits/fb2273ea82f05eee4b94005adb9a41c1c8b5b2e1).

misiekn (approved):

mfulb (approved):

ZNeumann (approved):

PR #393, opened 2022-03-21
test(integration): Add test for large error inbox

### LINKED ISSUE #362: Increase Maximum TransactionError Message Size
256 Char Limit -> 1024 Char Limit
See MMF for more details.

- [x] Code changes
- [x] New tests,
- [x] Update existing tests

ZNeumann:
This test is added to verify that our agent does not have a error message max size

Closes #362


tests/integration/errors/test_large_error_message.php
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Attempt to get rid of our exception handler.
+restore_exception_handler();
```
misiekn, line 62: Is this call to `restore_exception_handler()` needed? There aren't any exception handlers previously set with [`set_exception_handler()`](https://www.php.net/manual/en/function.set-exception-handler.php)

ZNeumann: fixed

misiekn (approved):

mfulb (approved):
Sure would be nice to have a preprocessor for these tests a shame to have to duplicate that long string.

Looks fine given our limitations.

zsistla:
Thanks for doing this :)

How many characters is the error message?
Does the full error message show up in NR1?

ZNeumann:
> Thanks for doing this :)
> 
> How many characters is the error message? Does the full error message show up in NR1?

It's around 1400 characters and the entire message shows up in NR1

zsistla (approved):

PR #402, opened 2022-04-05
tests(integration): Fixes mysql container to start properly [DO NOT MERGE]

mfulb:
The mysql container was not starting properly without having a volume defined for storage of the database.  This change adds a local volume which is mapped to the location the mysql image expects.  
Also an unneeded environment variable (MYSQL_HOST) has been removed.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -13,9 +13,10 @@ services:
       MYSQL_DATABASE: database
       MYSQL_USER: admin
       MYSQL_PASSWORD: admin
-      MYSQL_HOST: mysqldb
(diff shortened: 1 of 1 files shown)


ZNeumann (approved):

mfulb:
I investigated this further and it seems the container I was having problems with originally was having problems with the volume that had been created by Docker which was mapped to /var/lib/mysqld in the container.  I cleaned out all the containers and images and tried recreating the environment with the original docker-compose.yaml and things are working.

Closing without merging as this change does not appear to be necessary.

PR #403, opened 2022-04-05
feat(install): Only allow installer to run on supported arches and OSes

mfulb:
Adds some checking so if the installer is run on an unsupported arch then the installer exits.
On ARM64 ("aarch64" on Linux) a link to the instructions for installing from source is given.

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -221,7 +221,7 @@ arch=`(uname -m) 2> /dev/null` || arch="unknown"
 os=`(uname -s) 2> /dev/null` || os="unknown"
 
 case "${arch}" in
-  aarch64)               ;;
+  aarch64 | arm64)      arch=aarch64 ;;
   i[3456789]86)         arch=x86 ;;
   *64* | *amd*)         arch=x64 ;;
   i86pc)
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -137,8 +137,8 @@ if [ -z "${ostype}" ]; then
   case "${tus}" in
     [Dd][Aa][Rr][Ww][Ii][Nn])     ostype=unsupported_os ;;
     [Ff][Rr][Ee][Ee][Bb][Ss][Dd]) ostype=freebsd ;;
-    [Ss][Uu][Nn][Oo][Ss])         ostype=solaris ;;
-    [Ss][Mm][Aa][Rr][Tt][Oo][Ss]) ostype=solaris ;;
+    [Ss][Uu][Nn][Oo][Ss])         ostype=unsupported_os ;;
+    [Ss][Mm][Aa][Rr][Tt][Oo][Ss]) ostype=unsupported_os ;;
   esac
 fi
 : ${ostype:=generic}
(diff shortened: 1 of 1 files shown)


mfulb:
ok jenkins

mfulb:
ok jenkins

agent/newrelic-install.sh
```diff
@@ -221,6 +221,7 @@ arch=`(uname -m) 2> /dev/null` || arch="unknown"
 os=`(uname -s) 2> /dev/null` || os="unknown"
 
 case "${arch}" in
+  aarch64)               ;;
```
misiekn, line 224: aarch64 seems to be Linux specific. Should there also be a case for `arm64`? That's what's reported by `uname -m` on macOS running on Apple M1 chip. I'm not sure if we care for this case here, because there's macOS's logic later in the script. However that logic completely ignores `uname -m` output and assumes that macOS only runs on x86_64, which will cause problems if this script is executed on macOS.

misiekn (changes requested):
I like that when arm64 is detected the script points to resources how to install agent on this architecture.

However I think that OS and architecture detection may need some more refinements. Additionally, maybe it's not the time and place, but the `ostype` check in line 154 doesn't seem correct. Shouldn't `ostype` be compared against `generic` instead of `unsupported_os`? Isn't `generic` set in line 144 if OS type was not detected?

agent/newrelic-install.sh
misiekn: Since Linux is the only [supported operating system](https://docs.newrelic.com/docs/apm/agents/php-agent/getting-started/php-agent-compatibility-requirements), having aarch64 is enough.

agent/newrelic-install.sh
```diff
@@ -242,8 +243,30 @@ case "${os}" in
   *) ;;
```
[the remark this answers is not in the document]
misiekn: I addressed my concerns with [78d75343](https://github.com/newrelic/newrelic-php-agent/pull/403/commits/78d75343763b027e2afcc795b670c339b8f92bd2) - the script will bail out early on Darwin-based operating system (Mac OS X, OS X, macOS) so forcing x86_64 architecture in line 242 is now irrelevant.

misiekn (approved):
> However I think that OS and architecture detection may need some more refinements. Additionally, maybe it's not the time and place, but the `ostype` check in line 154 doesn't seem correct. Shouldn't `ostype` be compared against `generic` instead of `unsupported_os`? Isn't `generic` set in line 144 if OS type was not detected?

I think that the intention behind `ostype` check in line 154 was to gradually remove support for selected operating systems. I used this method to address all my review concerns [78d75343](https://github.com/newrelic/newrelic-php-agent/pull/403/commits/78d75343763b027e2afcc795b670c339b8f92bd2) - the script will bail out early on Darwin-based operating system (Mac OS X, OS X, macOS).

mfulb:
Thanks for the updates @misiekn.  I like using the `unsupported_os` mechanism to flag situations we should not proceed with the installation.  Also now OSes like Darwin and OpenBSD which use "arm64" will be shunted into the unsupported arch handler.

mfulb:
ok jenkins

PR #404, opened 2022-04-08
PHP: update daemon to accept and accommodate additional info from agent and collector

zsistla:
1) Adding `AgentEventLimits        collector.EventConfigs` (instead of something like SpanMaxEvents             uint64) to appInfo gives flexibility for adding additional agent configured limits in the future.
2) collector.NewHarvestLimits(* EventConfig)
Adding the parameter allows us to pass in agent configured limits. The daemon can then choose between agent limits or daemon defaults.  Making it a pointer allows us to pass in null when requesting daemon defaults only.
3) `SpanEventHarvestConfig collector.SpanEventHarvestConfig `json:"span_event_harvest_config"`` tells the daemon to start looking out for the new JSON.
4) Because span events are not included in Faster Event Harvest due to concerns
about downsampling within a distributed trace, the various report period and harvest limit are reported separately in span_event_harvest_config and event_harvest_config; however, internally, they are all still maintained inEventHarvestConfig which allows each limit to have it's own harvest period.
5) Added the rawSpanEventHarvestConfig struct for the wire representation of the span_event_harvest_config, SpanEventHarvestConfig to unmarshal the value into, and the unmarshal function.
6) Modified the agent to send the configured value.
7) modified the flat buffer to understand the additional value.
8) Added test cases.
9) Have daemon pick the lowest of the values between agent limits, daemon default limits, and collector limits


src/newrelic/collector/event_data.go
```diff
@@ -105,12 +123,70 @@ func getEventConfig(rawLimit *int, collectorRate time.Duration, defaultLimit int
 		err = fmt.Errorf("getEventConfig: event limit negative %d", rawLimit)
 		return 0, 0, err
 	} else {
-		limit = *rawLimit
+	    if *rawLimit > defaultLimit {
+	        limit = defaultLimit
+	    } else {
+	    limit = *rawLimit
```
mfulb, line 129: Formatting off here.

src/newrelic/collector/event_data.go
```diff
@@ -105,12 +123,70 @@ func getEventConfig(rawLimit *int, collectorRate time.Duration, defaultLimit int
...
+// SpanEventHarvestConfig.
+func (daemonConfig *SpanEventHarvestConfig) UnmarshalJSON(b []byte) error {
+	// Get the default report period in milliseconds.
+	reportPeriodMs, err := durationToMilliseconds(limits.DefaultReportPeriod)
+	if err != nil {
+		return err
+	}
+
+	// Build a new struct that represents what we'll get from the collector with
+	// the defaults (this is to ensure there is some value if the collector provies
```
mfulb, line 146: Typo - change to "provides"

src/newrelic/app.go
```diff
@@ -71,6 +71,7 @@ type AppInfo struct {
 	TraceObserverHost         string
 	TraceObserverPort         uint16
 	SpanQueueSize             uint64
+	AgentEventLimits        collector.EventConfigs
```
ZNeumann, line 74:
```suggestion
	AgentEventLimits          collector.EventConfigs
```

src/newrelic/collector/event_data.go
```diff
@@ -105,12 +123,70 @@ func getEventConfig(rawLimit *int, collectorRate time.Duration, defaultLimit int
...
+		log.Warnf("Unexpected report period of %d ms received; ignoring and using the default %v instead", rawConfig.ReportPeriodMS, limits.DefaultReportPeriod)
+		daemonConfig.SpanEventConfig.ReportPeriod = limits.DefaultReportPeriod
+	} else {
+		daemonConfig.SpanEventConfig.ReportPeriod = time.Duration(rawConfig.ReportPeriodMS) * time.Millisecond
+	}
+
+	var harvestConfig Event
+
+	// Check the span event value to see what the report period and limit should be.
+    // getEventConfig will do additional data verification.
```
ZNeumann, line 171:
```suggestion
        // getEventConfig will do additional data verification.
```

ZNeumann (approved):
Looks great!

mfulb (commented):
I wonder if in src/newrelic/processor_test.go it would be a good idea to have more tests run with the DoConnectConfiguredReply() and the "span_event_harvest_config" JSON present?  Currently most the tests use DoConnect() which leaves out this JSON.  As I think it is reasonable to expect the collector to send the JSON in the general case perhaps that case should be tested more?

src/newrelic/app_test.go
```diff
@@ -394,3 +321,116 @@ func TestAppKeyEquals(t *testing.T) {
...
+		`"host":"some_host",` +
+		`"display_host":"my_awesome_host",` +
+		`"settings":{"a":"1","b":true},` +
+		`"app_name":["one","two"],` +
+		`"high_security":false,` +
+		`"labels":[{"label_type":"c","label_value":"d"}],` +
+		`"environment":[["b",2]],` +
+		`"identifier":"one;two",` +
+		`"utilization":{"metadata_version":1,"logical_processors":22,"total_ram_mib":1000,"hostname":"some_host"},` +
+		`"event_harvest_config":{"report_period_ms":60000,"harvest_limits":{"error_event_data":100,"analytic_event_data":10000,"custom_event_data":10000,"span_event_data":10000}}` +
```
mfulb, line 398: Would it be possible to generate the expected "span_event_data" value using limits.MaxSpanMaxEvents here so if it changes the test is automatically updated?

zsistla:
> I wonder if in src/newrelic/processor_test.go it would be a good idea to have more tests run with the DoConnectConfiguredReply() and the "span_event_harvest_config" JSON present? Currently most the tests use DoConnect() which leaves out this JSON. As I think it is reasonable to expect the collector to send the JSON in the general case perhaps that case should be tested more?

----

They do look really similar and many fields are the same, but they both deal with different data structures specified in the agent-specs and aren't interchangeable.  The doconnect deals with the daemon's message it sends to the collector which adheres to the `connect-request-fields`
According to spec, this doesn't have a span_event_harvest_limit field, so if the daemon tried to send one to the collector, the collector wouldn't understand it so DoConnect wouldn't ever send it so the tests don't include span_event_harvest_limit.

the doconnectconfiguredreply deals with the collector's message it is sending to the daemon and has these fields: `connect-reply-fields` and which according to spec, does have a span_event_harvest_limit field that the daemon should accept and understand so the tests include that.

mfulb (approved):
Great work Amber looks good.  I think you've covered all my comments with updates.

src/newrelic/protocol/App.go
```diff
@@ -1,8 +1,3 @@
-//
-// Copyright 2020 New Relic Corporation. All rights reserved.
-// SPDX-License-Identifier: Apache-2.0
-//
-
```
misiekn, line 5: Should the file retain the Copyright and License information?

src/newrelic/commands.go
```diff
@@ -263,13 +263,16 @@ func UnmarshalAppInfo(tbl flatbuffers.Table) *AppInfo {
...
 
 	info.initSettings(app.Settings())
 
-	if app.HighSecurity() != 0 {
-		info.HighSecurity = true
-	}
+    // Of the four Event Limits (span, custom, analytic and error),
+    // only span events is configurable from the agent.
+    // If this changes in the future, the other values can be added here.
+    info.AgentEventLimits.SpanEventConfig.Limit = int(app.SpanEventsMaxSamplesStored())
```
misiekn, line 275: Increase indent.

src/newrelic/collector/event_data.go
```diff
@@ -105,12 +123,70 @@ func getEventConfig(rawLimit *int, collectorRate time.Duration, defaultLimit int
 		err = fmt.Errorf("getEventConfig: event limit negative %d", rawLimit)
 		return 0, 0, err
 	} else {
-		limit = *rawLimit
+	    if *rawLimit > defaultLimit {
+	        limit = defaultLimit
+	    } else {
+	    	limit = *rawLimit
+	    }
```
misiekn, line 130: Increase indent.

misiekn (approved):
Looks good. Just a few cosmetic formatting details.

zsistla:
ok jenkins

Issue #416, opened 2022-04-13
Send the metadata version 5 in the Utilization payload


misiekn:
As a PHP Agent I want to send the metadata version 5 in the Utilization payload sent to collector in order to better link APM and Infrastructure data. 

Depends on #414 and #415 

Part of #412

Issue #415, opened 2022-04-13
Include metadata version 5 data (Kubernetes) in the Utilization payload


misiekn:
As a PHP Agent I want to send the metadata version 5 data (detection of Kubernetes) in the Utilization payload sent to collector in order to better link APM and Infrastructure data.

Part of #412

Issue #414, opened 2022-04-13
Include metadata version 4 data (host name and ip address) in the Utilization payload


misiekn:
As a PHP Agent I want to send the metadata version 4 data (host name and ip address) in the Utilization payload sent to collector in order to better link APM and Infrastructure data

Part of #412

PR #432, opened 2022-04-29
feat(daemon): Implement Utilization Metadata Spec V5

mfulb:
This PR bumps the daemon up to reporting utilization metadata spec Level 5 which specified the inclusion of the full hostname, host ip address(es), and the Kubernetes service host (if available).  It adds a new daemon config option for "utilization.detect_kubernetes" which defaults to ON.

Much of this work is based on the Go agent (v3) commit to add [Utilization v5](https://github.com/newrelic/go-agent/commit/52464a9e9993b9dad4f904f46102debb5ea9bfc2).

This PR addresses issues #414, #415 and #416.

### DIFF
--- a/src/daemon/main.go
+++ b/src/daemon/main.go
@@ -240,7 +240,7 @@ type Config struct {
 	DetectPCF          bool           `config:"utilization.detect_pcf"`         // Whether to detect if this is running on PCF in utilization
 	DetectDocker       bool           `config:"utilization.detect_docker"`      // Whether to detect if this is in a Docker container in utilization
 	DetectKubernetes   bool		      `config:"utilization.detect_kubernetes"`  // Whether to detect if this is in a Kubernetes cluster
-    LogicalProcessors  int            `config:"utilization.logical_processors"` // Customer provided number of logical processors for pricing control.
+	LogicalProcessors  int            `config:"utilization.logical_processors"` // Customer provided number of logical processors for pricing control.
 	TotalRamMIB        int            `config:"utilization.total_ram_mib"`      // Customer provided total RAM in mebibytes for pricing control.
 	BillingHostname    string         `config:"utilization.billing_hostname"`   // Customer provided hostname for pricing control.
 	Agent              bool           `config:"-"`                              // Used to indicate if spawned by agent
--- a/src/daemon/main.go
+++ b/src/daemon/main.go
@@ -343,6 +343,7 @@ func main() {
 			DetectGCP:    true,
 			DetectPCF:    true,
 			DetectDocker: true,
+			DetectKubernetes: true,
 		})
 		str, err := json.MarshalIndent(util, "", "\t")
 		if err != nil {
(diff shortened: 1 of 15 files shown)


misiekn (approved):
Really nice work!

For future reviewers: IP address and host name utilization code changes in addresses.go, fqdn.go and provider.go modules are based on v3 go-agent code.

src/newrelic/utilization/utilization_hash_test.go
```diff
@@ -276,10 +296,6 @@ func runUtilizationCrossAgentTestcase(t *testing.T, tc utilizationCrossAgentTest
 
 	err := json.Unmarshal(tc.ExpectedOutput, &v)
 
-	if err == nil && v.MetadataVersion != 3 {
```
zsistla, line 279: Thanks for getting this up!

Why was this check removed?

mfulb: It was a filter to prevent any v5 cross agent tests from being run when the agent only supported v3.  Now that v5 is enabled we can run all the tests.

zsistla: Is the err checked for nil later?

zsistla:
It looks great.  
It looks like the new flag needs to be added to the sample payload for this test: https://github.com/newrelic/newrelic-php-agent/blob/dev/src/newrelic/collector_integration_test.go#L34

Also, for the integration runner?
https://github.com/newrelic/newrelic-php-agent/blob/main/src/integration_runner/main.go#L649

src/newrelic/utilization/utilization_hash_test.go
mfulb: The case where err != nil was not being handled previously so I broke that out first to help debugging when the test case definition JSON has an error in it.  
I also changed the metadata version check to skip any test cases requiring a version greater than is supported by the daemon to future-proof it some.

mfulb:
> It looks great. It looks like the new flag needs to be added to the sample payload for this test: https://github.com/newrelic/newrelic-php-agent/blob/dev/src/newrelic/collector_integration_test.go#L34
> 
> Also, for the integration runner? https://github.com/newrelic/newrelic-php-agent/blob/main/src/integration_runner/main.go#L649

Thanks for catching that - I found a couple other cases also which were similar that I've now fixed in the PR .

zsistla:
ok jenkins

mfulb:
ok jenkins

src/daemon/main.go
```diff
@@ -239,7 +239,8 @@ type Config struct {
 	DetectGCP          bool           `config:"utilization.detect_gcp"`         // Whether to detect if this is running on GCP in utilization
 	DetectPCF          bool           `config:"utilization.detect_pcf"`         // Whether to detect if this is running on PCF in utilization
 	DetectDocker       bool           `config:"utilization.detect_docker"`      // Whether to detect if this is in a Docker container in utilization
-	LogicalProcessors  int            `config:"utilization.logical_processors"` // Customer provided number of logical processors for pricing control.
+	DetectKubernetes   bool		      `config:"utilization.detect_kubernetes"`  // Whether to detect if this is in a Kubernetes cluster
+    LogicalProcessors  int            `config:"utilization.logical_processors"` // Customer provided number of logical processors for pricing control.
```
zsistla, line 243: Is the indentation off here?

zsistla (approved):

mfulb:
ok jenkins

mfulb:
ok jenkins

PR #434, opened 2022-05-09
P17 metadata blob test [DO NOT MERGE]

mfulb:
ONLY FOR DEMONSTRATION DO NOT MERGE

These are the changes needed to add the P17 metadata blob to the agent and thus minimally support talking to the collector.  Without the metadata blob the collector will return 400 ("Bad Request") to any requests made after the connection.

With these changes it is possible to use the newrelic-monolog-enricher to create logs in context.  For using with staging make sure to to change the return value from GetDefaultHost() in src/AbstractHandler.php to point at the staging collector.


mfulb:
Closing as prototype changes have been rolled into the main agent code base.

PR #447, opened 2022-05-11
fix(agent): Remove verbose debug log

misiekn:
Unconditional verbosedebug in nr_php_handle_cufa_fcall causes excessive
logging which impacts agent's performance when log level is set to
verbosedebug. It can additionally fill up disk space very quickly.


zsistla (approved):
Thanks!!

Issue #437, opened 2022-05-10
MUST Agent/daemon must decode and apply the NEW p17 request_headers_map (also known as metadata blobs) provided in the connect reply.


zsistla:
Please see MMF document for further details and links.

Acceptance criteria:

- [x] PR with code changes is created
- [x] There are tests that cover code changes
- [x] CI PR build passes
- [x] Code is reviewed by at least 2 team members (3 team members for more complex tickets)
- [x] Code is merged into dev branch

Note: See go repo for reference: https://github.com/newrelic/go-agent/search?q=request_headers_map

Issue #418, opened 2022-04-13
Use HTTP response status codes instead of response payload to determine agent's behavior 


lavarou:
As a PHP Agent I want to use Collector's P17: New Response Codes instead of response payload to determine my behavior in order to be compliant with Protocol 17

Part of #413

PR #449, opened 2022-05-11
feat(daemon): Support for P17 request_header_map (metadata blob)

mfulb:
This PR adds minimal support for the P17 request_header_map (metadata blob) to the daemon as defined by #437.  With this PR the agent will report it supports P17 to the collector and has been minimally tested to send data to the collector.  Also verified that the newrelic_monolog_enhancer works "out of the box" with this PR to send logs in context since entity.guid is now being returned in the collector connect response.

The PR for #418 will contain tests migrated from the Go agent which are relevant to this new functionality.

### DIFF
--- a/src/newrelic/collector/collector.go
+++ b/src/newrelic/collector/collector.go
@@ -26,7 +26,7 @@ const (
 )
 
 const (
-	protocolVersion = "16"
+	protocolVersion = "17"
 )
 
 // LicenseKey represents a license key for an account.
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -304,11 +304,6 @@ func parseConnectReply(rawConnectReply []byte) (*ConnectReply, error) {
 		return nil, errors.New("missing agent run id")
 	}
 
-	log.Debugf("request_headers_map from connect reply")
-	for k, v := range c.RequestHeadersMap {
-		log.Debugf("key=%s val=%s", k, v)
-	}
-
     // Since the collector now sends seperately, we need to internally combine the limits.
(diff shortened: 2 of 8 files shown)


mfulb:
Added fixes for 2 linking metadata integration tests which were failing.

src/newrelic/collector/collector.go
```diff
@@ -26,7 +26,7 @@ const (
 )
 
 const (
-	protocolVersion = "16"
+	protocolVersion = "17"
```
misiekn, line 29: Does the protocol version update belong to this PR? Maybe it does once it is re-based onto the code that addresses the P17 response code handling, i.e. after #418 is closed.

zsistla (approved):
Thanks for putting this up, great to see the P17 flag toggled with a stub response code section so we'll be able to move forward with other daemon tickets.

src/newrelic/collector/collector.go
mfulb: I see your point - by the book yes we probably would rather have the flag flipped in that PR.  Realistically since all dev is currently on P17 and the agent does appear to work for the purposes of P17 dev as is I think this will be OK since the #418 PR will not be too far away.  But I am fine if the consensus is to hold off on flipping this now.

misiekn: On the second thought, the earlier we have the updated the protocol version, the earlier other parts that depend on P17 being enabled can be tested (thinking labels, logs in context). Since it works, I'm ok with including protocol version update in this PR.

mfulb: That was my reasoning - I probably should have run it by everyone first.  :)

PR #451, opened 2022-05-13
feat: P17 response codes

ZNeumann:
This adds minimum P17 response codes. Work still left to do:
- ensure failed harvest limit are changed to match spec

This code also includes some refactoring to make it more similar to the go-agent. This includes splitting the generic cmd struct into 2 -> cmd and controls.

Finally, in the files touched it turns the hard tabs to soft tabs. I apologize in advance

### DIFF
--- a/axiom/cmd_appinfo_transmit.c
+++ b/axiom/cmd_appinfo_transmit.c
@@ -289,19 +289,17 @@ nr_status_t nr_cmd_appinfo_process_reply(const uint8_t* data,
       nrl_info(NRL_ACCT, "APPINFO reply disconnected app=" NRP_FMT,
                NRP_APPNAME(app->info.appname));
       return NR_SUCCESS;
-    //case APP_STATUS_INVALID_LICENSE:
-    //  app->state = NR_APP_INVALID;
-    //  nrl_error(NRL_ACCT,
-    //            "APPINFO reply invalid license app=" NRP_FMT
-    //            " please check your license "
-    //            "key and restart your web server.",
-    //            NRP_APPNAME(app->info.appname));
-    //  return NR_SUCCESS;
+    case APP_STATUS_INVALID_LICENSE:
+      app->state = NR_APP_INVALID;
+      nrl_error(NRL_ACCT,
+                "APPINFO reply invalid license app=" NRP_FMT
+                " please check your license "
+                "key and restart your web server.",
+                NRP_APPNAME(app->info.appname));
+      return NR_SUCCESS;
     case APP_STATUS_CONNECTED:
       /* Don't return here, instead break and continue below. */
       nrl_debug(NRL_ACCT, "APPINFO reply connected");
-      break;
-    case APP_STATUS_RESTART:
       break;
     case APP_STATUS_STILL_VALID:
       app->state = NR_APP_OK;
--- a/axiom/cmd_appinfo_transmit.c
+++ b/axiom/cmd_appinfo_transmit.c
@@ -289,17 +289,19 @@ nr_status_t nr_cmd_appinfo_process_reply(const uint8_t* data,
       nrl_info(NRL_ACCT, "APPINFO reply disconnected app=" NRP_FMT,
                NRP_APPNAME(app->info.appname));
       return NR_SUCCESS;
-    case APP_STATUS_INVALID_LICENSE:
-      app->state = NR_APP_INVALID;
-      nrl_error(NRL_ACCT,
-                "APPINFO reply invalid license app=" NRP_FMT
-                " please check your license "
-                "key and restart your web server.",
-                NRP_APPNAME(app->info.appname));
-      return NR_SUCCESS;
+    //case APP_STATUS_INVALID_LICENSE:
+    //  app->state = NR_APP_INVALID;
+    //  nrl_error(NRL_ACCT,
+    //            "APPINFO reply invalid license app=" NRP_FMT
+    //            " please check your license "
+    //            "key and restart your web server.",
+    //            NRP_APPNAME(app->info.appname));
+    //  return NR_SUCCESS;
     case APP_STATUS_CONNECTED:
       /* Don't return here, instead break and continue below. */
       nrl_debug(NRL_ACCT, "APPINFO reply connected");
+      break;
+    case APP_STATUS_RESTART:
       break;
     case APP_STATUS_STILL_VALID:
       app->state = NR_APP_OK;
--- a/src/integration_runner/main.go
+++ b/src/integration_runner/main.go
@@ -575,10 +575,10 @@ func discoverTests(pattern string, searchPaths []string) []string {
 	return testFiles
 }
 
-func injectIntoConnectReply(reply []byte, newRunID, crossProcessId string) []byte {
+func injectIntoConnectReply(reply collector.RPMResponse, newRunID, crossProcessId string) []byte {
 	var x map[string]interface{}
 
-	json.Unmarshal(reply, &x)
+	json.Unmarshal(reply.Body, &x)
 
 	x["agent_run_id"] = newRunID
 	x["cross_process_id"] = crossProcessId
@@ -590,7 +590,7 @@ func injectIntoConnectReply(reply []byte, newRunID, crossProcessId string) []byt
 type IntegrationDataHandler struct {
 	sync.Mutex                                       // Protects harvests
 	harvests            map[string]*newrelic.Harvest // Keyed by tc.Name (which is used as AgentRunID)
-	reply               []byte                       // Constant after creation
+	reply               collector.RPMResponse        // Constant after creation
 	rawSecurityPolicies []byte                       // policies from connection attempt, needed for AppInfo reply
 }
 
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -21,7 +21,7 @@ import (
 	"newrelic/utilization"
 )
 
-// AgentRunID is a string as of agent listener protocol version 14.
+// AgentRunID is a string as of agent listener protocol version 17.
 type AgentRunID string
 
 func (id AgentRunID) String() string {
@@ -33,8 +33,9 @@ type AppState int
 const (
 	AppStateUnknown AppState = iota
 	AppStateConnected
-	AppStateInvalidLicense
+	//AppStateInvalidLicense
 	AppStateDisconnected
+    AppStateRestart
 	AppStateInvalidSecurityPolicies
 )
 
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -33,7 +33,6 @@ type AppState int
 const (
 	AppStateUnknown AppState = iota
 	AppStateConnected
-	//AppStateInvalidLicense
 	AppStateDisconnected
     AppStateRestart
 	AppStateInvalidSecurityPolicies
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -31,11 +31,11 @@ func (id AgentRunID) String() string {
 type AppState int
 
 const (
-	AppStateUnknown AppState = iota
-	AppStateConnected
-	AppStateDisconnected
+    AppStateUnknown AppState = iota
+    AppStateConnected
+    AppStateDisconnected
     AppStateRestart
-	AppStateInvalidSecurityPolicies
+    AppStateInvalidSecurityPolicies
 )
 
 // An AppKey uniquely identifies an application.
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -21,7 +21,7 @@ import (
 	"newrelic/utilization"
 )
 
-// AgentRunID is a string as of agent listener protocol version 17.
+// AgentRunID is a string as of agent listener protocol version 14.
 type AgentRunID string
 
 func (id AgentRunID) String() string {
@@ -35,6 +35,7 @@ const (
     AppStateConnected
     AppStateDisconnected
     AppStateRestart
+    AppStateInvalidLicense
     AppStateInvalidSecurityPolicies
 )
 
--- a/src/newrelic/app_test.go
+++ b/src/newrelic/app_test.go
@@ -238,13 +238,13 @@ func TestNeedsConnectAttempt(t *testing.T) {
 		t.Fatal(now, app.lastConnectAttempt, app.state)
 	}
 
-	//app.state = AppStateInvalidLicense
-	//app.lastConnectAttempt = now.Add(-limits.AppConnectAttemptBackoff)
-	//if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
-	//	t.Fatal(now, app.lastConnectAttempt, app.state)
-	//}
-
 	app.state = AppStateDisconnected
+	app.lastConnectAttempt = now.Add(-limits.AppConnectAttemptBackoff)
+	if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
+		t.Fatal(now, app.lastConnectAttempt, app.state)
+	}
+
+	app.state = AppStateRestart
 	app.lastConnectAttempt = now.Add(-limits.AppConnectAttemptBackoff)
 	if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
 		t.Fatal(now, app.lastConnectAttempt, app.state)
--- a/src/newrelic/app_test.go
+++ b/src/newrelic/app_test.go
@@ -249,6 +249,12 @@ func TestNeedsConnectAttempt(t *testing.T) {
 	if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
 		t.Fatal(now, app.lastConnectAttempt, app.state)
 	}
+
+	app.state = AppStateInvalidLicense
+	app.lastConnectAttempt = now.Add(-limits.AppConnectAttemptBackoff)
+	if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
+		t.Fatal(now, app.lastConnectAttempt, app.state)
+	}
 }
 
 func TestAppKeyEquals(t *testing.T) {
--- a/src/newrelic/app_test.go
+++ b/src/newrelic/app_test.go
@@ -238,11 +238,11 @@ func TestNeedsConnectAttempt(t *testing.T) {
 		t.Fatal(now, app.lastConnectAttempt, app.state)
 	}
 
-	app.state = AppStateInvalidLicense
-	app.lastConnectAttempt = now.Add(-limits.AppConnectAttemptBackoff)
-	if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
-		t.Fatal(now, app.lastConnectAttempt, app.state)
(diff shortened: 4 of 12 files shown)


ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

src/integration_runner/main.go
```diff
@@ -590,7 +590,7 @@ func injectIntoConnectReply(reply []byte, newRunID, crossProcessId string) []byt
 type IntegrationDataHandler struct {
 	sync.Mutex                                       // Protects harvests
 	harvests            map[string]*newrelic.Harvest // Keyed by tc.Name (which is used as AgentRunID)
-	reply               []byte                       // Constant after creation
+	reply               collector.RPMResponse        // Constant after creation
```
ZNeumann, line 593:
```suggestion
	reply               collector.RPMResponse        // Constant
```

src/newrelic/collector/client.go
```diff
@@ -6,367 +6,335 @@
...
-			return httpURL, nil
-		}
-	}
-	return url, err
+    url, err := url.Parse(proxy)
+    if err != nil || !strings.Contains(proxy, "://") {
+    // Try again, assuming HTTP as the scheme. If this fails, return the
+    // original error so the error message matches the original URL.
+    if httpURL, err := url.Parse("http://" + proxy); err == nil {
+    return httpURL, nil
```
mfulb, line 257:
```suggestion
        return httpURL, nil
```

src/newrelic/collector/client.go
```diff
@@ -6,367 +6,335 @@
...
-	if err != nil || !strings.Contains(proxy, "://") {
-		// Try again, assuming HTTP as the scheme. If this fails, return the
-		// original error so the error message matches the original URL.
-		if httpURL, err := url.Parse("http://" + proxy); err == nil {
-			return httpURL, nil
-		}
-	}
-	return url, err
+    url, err := url.Parse(proxy)
+    if err != nil || !strings.Contains(proxy, "://") {
```
mfulb, line 253: This block and the if inside need to be indented.

src/newrelic/collector/client.go
```diff
@@ -6,367 +6,335 @@
...
-}
+    r := newRPMResponse(resp.StatusCode)
 
-func (e *rpmException) Error() string {
-	return fmt.Sprintf("%s: %s", e.ErrorType, e.Message)
+    body, err := ioutil.ReadAll(resp.Body)
+    if nil != r.Err {
+        r.Err = err
+    }
+    r.Body = body
```
mfulb, line 300: Will ```body``` have a meaningful value if the ```ReadAll``` call above has a non-nil return value?

src/newrelic/collector/client.go
```diff
@@ -6,367 +6,335 @@
...
-type rpmException struct {
-	Message   string `json:"message"`
-	ErrorType string `json:"error_type"`
-}
+    r := newRPMResponse(resp.StatusCode)
 
-func (e *rpmException) Error() string {
-	return fmt.Sprintf("%s: %s", e.ErrorType, e.Message)
+    body, err := ioutil.ReadAll(resp.Body)
+    if nil != r.Err {
```
mfulb, line 297: Should this be testing ```err``` instead of ```r.Err```?

src/newrelic/collector/client.go
```diff
@@ -6,367 +6,335 @@
...
+        timer = time.After(l.timeout)
+    }
+
+    select {
+    case <-l.semaphore:
+        defer func() { l.semaphore <- true }()
+        resp := l.orig.Execute(cmd, cs)
+        return resp
+    case <-timer:
+        return RPMResponse{}
```
mfulb, line 159: Does the need to set the ```Err``` field on the case of a timeout?  It seems the code in ```processor.go``` which calls this will check the ```Err``` field and if it is not set then it will act as if the function succeeded instead of timeing out as was the case.

zsistla:
Thanks for all your work on this!

src/newrelic/collector/collector_test.go
```diff
@@ -7,151 +7,156 @@ package collector
 
 import (
 	"net/url"
+    "net/http"
```
zsistla, line 10: Indentation is also off here.

src/newrelic/collector/collector_test.go
```diff
@@ -7,151 +7,156 @@ package collector
...
-	// collector.
-	r := `{"exception":{"message":"something","error_type":"NewRelic::Agent::ForceRestartException"}}`
-	reply, err := parseResponse([]byte(r))
-	if reply != nil {
-		t.Fatal(string(reply))
-	}
-	if !IsRestartException(err) {
-		t.Fatal(err)
-	}
+func TestResponseCodeError(t *testing.T) {
```
zsistla, line 21: This is a much cleaner and more extensible way of testing :)

src/newrelic/processor.go
```diff
@@ -6,819 +6,812 @@
...
+    }
+    switch {
+        case d.Reply.IsDisconnect() || app.state == AppStateDisconnected:
+            app.state = AppStateDisconnected
+            p.shutdownAppHarvest(d.id)
+        case d.Reply.IsRestartException() || app.state == AppStateRestart:
+            app.state = AppStateUnknown
+            p.shutdownAppHarvest(d.id)
+            p.considerConnect(app)
+    }
```
zsistla, line 680: Don't we still need the default case, just in case?

axiom/cmd_appinfo_transmit.c
```diff
@@ -289,20 +289,18 @@ nr_status_t nr_cmd_appinfo_process_reply(const uint8_t* data,
...
+                "APPINFO reply invalid license app=" NRP_FMT
+                " please check your license "
+                "key and restart your web server.",
+                NRP_APPNAME(app->info.appname));
+      return NR_SUCCESS;
     case APP_STATUS_CONNECTED:
       /* Don't return here, instead break and continue below. */
       nrl_debug(NRL_ACCT, "APPINFO reply connected");
       break;
-    case APP_STATUS_RESTART:
```
zsistla: Shouldn't we handle this case?  Does the agent need to do something when the app is in a restart state?

zsistla:
Thanks for rolling in some of the v3 changes :)

src/newrelic/processor.go
ZNeumann: The logic has changed slightly such that the default should no longer be necessary. At the top of the block, if the response code indicates saving data, we do that before any of the restart/disconnect logic. So at this point, calling failedHarvest() is no longer mutually exclusive with the cases in this switch block.

In the case where we receive an unknown response code, we will not save the harvest data, we will not disconnect, and we will not restart. It will simply discard the harvest and continue on.

zsistla: Maybe at least an error message if we hit the default saying we received unknown and are discarding?

ZNeumann: That's a little tricky, but I can add it. Basically, it cannot be the default case (because there are currently non-successful codes that neither direct for a restart or shutdown), but I can add a check against an entire list of known codes.

axiom/cmd_appinfo_transmit.c
ZNeumann: AppStateRestart is reported as AppStateUnkown. When the daemon enters AppStateRestart, [it very quickly switches over to AppStateUnkown during the restart](https://github.com/newrelic/newrelic-php-agent/pull/451/files/9796141f6fbfa6b4aabf1da58eb366e49b841bdc#diff-d34a767f316aaeded794ac3a231f867d3ffb42de19365ed338eace17f5876d94R677). While AppStateRestart is new in this PR, being in AppStateUnkown during restart is consistent with prior behavior.

zsistla: Ah!  Thanks for explaining :)

src/newrelic/app.go
```diff
@@ -21,7 +21,7 @@ import (
 	"newrelic/utilization"
 )
 
-// AgentRunID is a string as of agent listener protocol version 14.
```
mfulb, line 24: I think the comment was fine as is as it is saying AgentRunID has been a string since v14?

ZNeumann: Good catch

src/newrelic/collector/client.go
```diff
@@ -6,310 +6,302 @@
...
-	log.Audit("command='%s' url='%s', response={%s}", cmd.Name, url, resp)
-	log.Debugf("command='%s' url='%s', response={%s}", cmd.Name, cleanURL, resp)
-
-	return resp, err
+    body, err := ioutil.ReadAll(resp.Body)
+    if nil != err {
+        r.Err = err
+    } else {
+        r.Body, r.Err = parseResponse(body)
+    }
```
misiekn, line 302:
```suggestion
    body, err := ioutil.ReadAll(resp.Body)
    if nil != err {
        r.Err = err
    } else {
        r.Body = body    
    }
```
Please remove call to `parseResponse` body - we no longer want to parse collector response, we want to use HTTP Status Codes. Also with P17 `body` will be empty and parseBody calls json.Unmarshal(), which expects a valid JSON string. Without it, it'll throw "unexpected end of JSON input" error.

ZNeumann: I could use a little help on this aspect. Originally, I had removed the parseResponse() call. However, this caused issues with our LASP tests, [as seen here on Jenkins](https://phpagent-build.pdx.vm.datanerd.us/job/php-pull-request-parallel-8-1/label=centos5-64-nrcamp/149/console). Our policy map is being sent in the body of the response. I do not know of the desired way forward from here.

zsistla: I think the parsereply still needs to stay in there so we can extract security policies when they exist.  If you create an Unmarshal for the connectreply like go-agent does, it'll catch for when the json is empty and extract security policy when it is not: https://github.com/newrelic/go-agent/blob/06c801d5571056abac8ac9dfa07cf12ca869e920/v3/internal/connect_reply.go#L254

misiekn: Digging deeper through go-agent code reveals that go-agent actually does parse the reply like daemon's `parseResponse` function: 
[newrelic/go-agent/internal/collector.go:268-276](https://github.com/newrelic/go-agent/blob/4b46fc5b40e5398695413097b5c461679a74eafd/internal/collector.go#L276)
```
	resp := CollectorRequest(call, cs)
	if nil != resp.Err {
		return nil, resp
	}

	var preconnect struct {
		Preconnect PreconnectReply `json:"return_value"`
	}
	err = json.Unmarshal(resp.body, &preconnect)

```

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

src/newrelic/app_test.go
```diff
@@ -238,13 +238,19 @@ func TestNeedsConnectAttempt(t *testing.T) {
...
 	}
 
-	app.state = AppStateDisconnected
+	app.state = AppStateRestart
+	app.lastConnectAttempt = now.Add(-limits.AppConnectAttemptBackoff)
+	if app.NeedsConnectAttempt(now, limits.AppConnectAttemptBackoff) {
+		t.Fatal(now, app.lastConnectAttempt, app.state)
+	}
+
+	app.state = AppStateInvalidLicense
```
mfulb, line 253: This test seems to be testing the backoff for retries if an invalid license response is received from the collector - are we doing retries in this situation?

src/newrelic/collector/client.go
```diff
@@ -6,310 +6,331 @@
...
-	"net/http"
-	"net/url"
-	"strings"
-	"time"
-
-	"golang.org/x/net/proxy"
-
-	"newrelic/log"
-	"newrelic/version"
+    "crypto/tls"
```
mfulb, line 9: Looks like we lost the hard tabs?

src/newrelic/collector/client.go
```diff
@@ -6,310 +6,331 @@
...
+    }
+    return RPMResponse{StatusCode: StatusCode, Err: err}
+}
+// IsDisconnect indicates that the agent should disconnect.
+func (resp RPMResponse) IsDisconnect() bool {
+    return resp.StatusCode == 410 || resp.disconnectSecurityPolicy
+}
+// IsRestartException indicates that the agent should restart.
+func (resp RPMResponse) IsRestartException() bool {
+    return resp.StatusCode == 401 || resp.StatusCode == 409
```
mfulb, line 90: Should 401 still map to a restart for the PHP agent?

ZNeumann: I decided to leave it there to reduce future confusion and conform with the spec. There is a comment explaining where we differ from spec behavior.

mfulb: Maybe add a comment explaining that here as well - currently it looks like we consider a 401 a restart (following the spec) in this function while elsewhere the actual behavior differs from the spec.

src/newrelic/collector/client.go
```diff
@@ -6,310 +6,331 @@
...
+    switch resp.StatusCode {
+    case 408, 429, 500, 503:
+        return true
+    default:
+        return false
+    }
+}
+
+// Not a method of RPMResponse so that it can be called during creation
+func GetStatusCodeMessage(StatusCode int) string {
```
mfulb, line 108: Nice addition!

src/newrelic/processor.go
```diff
@@ -6,819 +6,817 @@
...
+    }  else if rep.RawReply.IsRestartException() {
+        app.state = AppStateRestart
+        // in accord with the spec, invalid license is a restart exception. Except we want
+        //    to shutdown instead of restart.
+        if rep.RawReply.IsInvalidLicense() {
+            app.state = AppStateInvalidLicense
+        }
+        log.Warnf("app '%s' connect attempt returned %s", app, rep.RawReply.Err)
+        return
+    } else if nil != rep.Err {
```
mfulb, line 409: Can we move the check for an error to the top so we don't go through several if checks that we know will fail?

ZNeumann:
ok jenkins

zsistla:
ok jenkins

src/newrelic/processor_test.go
```diff
@@ -6,985 +6,1017 @@
...
 
-	<-m.clientParams
-	m.clientReturn <- ClientReturn{nil, collector.SampleDisonnectException}
-	<-m.p.trackProgress // receive harvest error
+    <-m.clientParams
+    m.clientReturn <- ClientReturn{nil, SampleDisonnectException, 410}
+    <-m.p.trackProgress // receive harvest error
 
-	m.DoAppInfo(t, nil, AppStateDisconnected)
+    fmt.Printf("AHHHHHHHHHH\n")
```
mfulb, line 651: This debugging output is probably not indented to be in the final commit?

src/newrelic/processor_test.go
```diff
@@ -6,985 +6,1017 @@
...
+        Settings:          map[string]interface{}{},
+        Environment:       nil,
+        Labels:            nil,
+        RedirectCollector: "collector.newrelic.com",
+        HighSecurity:      true,
+        Hostname:          "agent-hostname",
+    }
+    connectClient = collector.ClientFn(func(cmd collector.RpmCmd, cs collector.RpmControls) collector.RPMResponse {
+        if cmd.Name == collector.CommandPreconnect {
+            return collector.RPMResponse{Body: []byte(`{"redirect_host":"specific_collector.com"}`), Err: nil, StatusCode: 202}
```
mfulb, line 853: I think the preconnect should return status code 200?

src/newrelic/processor_test.go
```diff
@@ -6,985 +6,1017 @@
...
+    m := NewMockedProcessor(1)
 
-	m.DoAppInfo(t, nil, AppStateUnknown)
+    m.DoAppInfo(t, nil, AppStateUnknown)
 
-	<-m.clientParams // preconnect
-	m.clientReturn <- ClientReturn{[]byte(`"`), nil}
-	<-m.p.trackProgress // receive connect reply
+    <-m.clientParams // preconnect
+    m.clientReturn <- ClientReturn{[]byte(`"`), nil, 202}
```
mfulb, line 709: Preconnect returns 200?

src/newrelic/processor_test.go
```diff
@@ -6,985 +6,1017 @@
...
 
-	<-m.clientParams // preconnect
-	m.clientReturn <- ClientReturn{[]byte(`{"redirect_host":"specific_collector.com"}`), nil}
-	<-m.clientParams // connect
-	m.clientReturn <- ClientReturn{[]byte(`{`), nil}
-	<-m.p.trackProgress // receive connect reply
+    <-m.clientParams // preconnect
+    m.clientReturn <- ClientReturn{[]byte(`{"redirect_host":"specific_collector.com"}`), nil, 202}
+    <-m.clientParams // connect
+    m.clientReturn <- ClientReturn{[]byte(`{`), nil, 202}
```
mfulb, line 695: Connect returns 200?

src/newrelic/processor_test.go
```diff
@@ -6,985 +6,1017 @@
...
-	m.DoAppInfo(t, nil, AppStateUnknown)
+    m.DoAppInfo(t, nil, AppStateUnknown)
 
-	<-m.clientParams // preconnect
-	m.clientReturn <- ClientReturn{[]byte(`{"redirect_host":"specific_collector.com"}`), nil}
-	<-m.clientParams // connect
-	m.clientReturn <- ClientReturn{[]byte(`{`), nil}
-	<-m.p.trackProgress // receive connect reply
+    <-m.clientParams // preconnect
+    m.clientReturn <- ClientReturn{[]byte(`{"redirect_host":"specific_collector.com"}`), nil, 202}
```
mfulb, line 693: Preconnect returns 200?

src/newrelic/processor_test.go
```diff
@@ -6,985 +6,1017 @@
...
-	m.DoAppInfo(t, nil, AppStateUnknown)
+    m.DoAppInfo(t, nil, AppStateUnknown)
 
-	<-m.clientParams // preconnect
-	m.clientReturn <- ClientReturn{[]byte(`{"redirect_host":"specific_collector.com"}`), nil}
-	<-m.clientParams // connect
-	m.clientReturn <- ClientReturn{nil, collector.SampleDisonnectException}
-	<-m.p.trackProgress // receive connect reply
+    <-m.clientParams // preconnect
+    m.clientReturn <- ClientReturn{[]byte(`{"redirect_host":"specific_collector.com"}`), nil, 202}
```
mfulb, line 613: Preconnect returns 200?

src/newrelic/commands.go
```diff
@@ -224,7 +224,7 @@ func MarshalAppInfoReply(reply AppInfoReply) []byte {
 		switch reply.State {
 		case AppStateUnknown:
 			protocol.AppReplyAddStatus(buf, protocol.AppStatusUnknown)
-		case AppStateInvalidLicense:
+        case AppStateInvalidLicense:
```
zsistla, line 227: Indentation?

zsistla (approved):
Looks good to merge pending a few formatting changes.  Thanks for all the work on this!!

mfulb (approved):
Great work Zach!

Issue #445, opened 2022-05-10
SHOULD Daemon creates supportability metrics based off of p17 response codes.


zsistla:
Please see MMF document for further details and links.

SHOULD Daemon creates supportability metrics based off of p17 response codes.

Acceptance criteria:
- [x] PR with code changes is created
- [x] There are tests that cover code changes
- [x] CI PR build passes
- [x] Code is reviewed by at least 2 team members (3 team members for more complex tickets)
- [x] Code is merged into dev branch

PR #453, opened 2022-05-17
feat(daemon): Implement New Relic payload size limit

### LINKED ISSUE #440: Update the daemon to decode and use the `max_payload_size_in_bytes` value.
Please see MMF document for further details and links.

Acceptance criteria.
- [x] PR with code changes is created
- [x] There are tests that cover code changes
- [x] CI PR build passes
- [x] Code is reviewed by at least 2 team members (3 team members for more complex tickets)
- [x] Code is merged into dev branch

Note: See [go repo](https://github.com/newrelic/go-agent) for reference:

Connect example: https://github.com/newrelic/go-agent/blob/06c801d5571056abac8ac9dfa07cf12ca869e920/internal/connect_reply.go#L30

Test example: https://github.com/newrelic/go-agent/blob/06c801d5571056abac8ac9dfa07cf12ca869e920/v3/newrelic/collector_test.go#L411

misiekn:
This PR implements following Agent's specification:

Agents MUST default to sending no more than 1000000 bytes in any payload
sent to New Relic for any method. The connect response contains the
max_payload_size_in_bytes field which MUST override the default value.
    
Immediately prior to transmitting a payload to New Relic (after
compression, prior to transmission), the size of the payload in bytes
should be compared against max_payload_size_in_bytes. If the payload
size is larger than max_payload_size_in_bytes, the agent should
immediately discard that payload. A payload larger than
max_payload_size_in_bytes MUST NOT be transmitted.

Following tests have been added to cover changed code:
1. Test if default payload limit is preserved if collector doesn't provide any value.
2. Test if default payload limit is overridden if collector provides limit value in connect reply.
3. Test if payload is dropped if its size is larger than max_payload_size_in_bytes.
4. Test if payload is not dropped if its size is smaller than max_payload_size_in_bytes.

Closes #440

### DIFF
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -296,7 +296,7 @@ func combineEventConfig (ehc collector.EventHarvestConfig, sehc collector.SpanEv
 }
 
 func parseConnectReply(rawConnectReply []byte) (*ConnectReply, error) {
-	var c ConnectReply
+	c := ConnectReply{MaxPayloadSizeInBytes: limits.DefaultMaxPayloadSizeInBytes}
 
 	err := json.Unmarshal(rawConnectReply, &c)
 	if nil != err {
--- a/src/newrelic/app.go
+++ b/src/newrelic/app.go
@@ -125,6 +125,7 @@ type ConnectReply struct {
 	EventHarvestConfig collector.EventHarvestConfig         `json:"event_harvest_config"`
 	SpanEventHarvestConfig collector.SpanEventHarvestConfig `json:"span_event_harvest_config"`
 	RequestHeadersMap map[string]string                     `json:"request_headers_map"`
+	MaxPayloadSizeInBytes int                               `json:"max_payload_size_in_bytes"`
 }
 
 // An App represents the state of an application.
--- a/src/newrelic/app_test.go
+++ b/src/newrelic/app_test.go
@@ -6,6 +6,7 @@
 package newrelic
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 	"strconv"
@@ -440,4 +441,30 @@ func TestConnectPayloadEncoded(t *testing.T) {
 		t.Errorf("expected: %s\nactual: %s", expected, string(b))
 	}
 
+}
+
+func TestMaxPayloadSizeInBytesFromDefault(t *testing.T) {
+	expectedMaxPayloadSizeInBytes := limits.DefaultMaxPayloadSizeInBytes
+	id := AgentRunID("1") // parseConnectReply expects at least agent_run_id in collector reply
+	cannedConnectReply, _ := json.Marshal(ConnectReply{ID: &id})
+
+	c, err := parseConnectReply(cannedConnectReply)
+	if err != nil {
+		t.Error(err)
+	} else if c.MaxPayloadSizeInBytes != expectedMaxPayloadSizeInBytes {
+		t.Errorf("expected [%v], got [%v]", expectedMaxPayloadSizeInBytes, c.MaxPayloadSizeInBytes)
+	}
+}
+
+func TestMaxPayloadSizeInBytesFromConnectReply(t *testing.T) {
+	expectedMaxPayloadSizeInBytes := 1000
+	id := AgentRunID("1") // parseConnectReply expects at least agent_run_id in collector reply
+	cannedConnectReply, _ := json.Marshal(ConnectReply{ID: &id, MaxPayloadSizeInBytes: expectedMaxPayloadSizeInBytes})
+
+	c, err := parseConnectReply(cannedConnectReply)
+	if err != nil {
+		t.Error(err)
+	} else if c.MaxPayloadSizeInBytes != expectedMaxPayloadSizeInBytes {
+		t.Errorf("expected [%v], got [%v]", expectedMaxPayloadSizeInBytes, c.MaxPayloadSizeInBytes)
+	}
 }
--- a/src/newrelic/app_test.go
+++ b/src/newrelic/app_test.go
@@ -6,6 +6,7 @@
 package newrelic
 
 import (
+	"fmt"
 	"testing"
 	"time"
 	"strconv"
@@ -456,7 +457,7 @@ func TestMaxPayloadSizeInBytesFromDefault(t *testing.T) {
 
 func TestMaxPayloadSizeInBytesFromConnectReply(t *testing.T) {
 	expectedMaxPayloadSizeInBytes := 1000
-	cannedConnectReply := []byte(`{"agent_run_id":"1", "max_payload_size_in_bytes":1000}`)
+	cannedConnectReply := []byte(`{"agent_run_id":"1", "max_payload_size_in_bytes":`+fmt.Sprint(expectedMaxPayloadSizeInBytes)+`}`)
 
 	c, err := parseConnectReply(cannedConnectReply)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


src/newrelic/app.go
```diff
@@ -294,7 +295,7 @@ func combineEventConfig (ehc collector.EventHarvestConfig, sehc collector.SpanEv
 }
 
 func parseConnectReply(rawConnectReply []byte) (*ConnectReply, error) {
-	var c ConnectReply
+	c := ConnectReply{MaxPayloadSizeInBytes: limits.DefaultMaxPayloadSizeInBytes}
```
mfulb, line 299: Do we need to worry about if the connect reply had a value > 1000*1000 or does the spec say to accept a larger value if returned?

src/newrelic/collector/client_test.go
```diff
@@ -91,3 +96,65 @@ func TestUserAgent(t *testing.T) {
...
+
+type roundTripperFunc func(*http.Request) (*http.Response, error)
+
+func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return fn(r)
+}
+
+func TestExecuteWhenMaxPayloadSizeExceeded(t *testing.T) {
+	// call is defined in collector_test.go
+	call.MaxPayloadSize = 0
```
mfulb, line 108: I feel that `call` and `actualData` should be defined in this test source file instead of pulling it in from the `var()` statement in `collector_test.go`.

Also does `call` need to be reset to a known state for each test function that access it or do changes in one test function not persist to the next?

src/newrelic/collector/client_test.go
```diff
@@ -91,3 +96,65 @@ func TestUserAgent(t *testing.T) {
...
+	} else if err == nil {
+		t.Errorf("%s, got [%v], want [%v]", testedFn, err, wantErr)
+	} else if !strings.HasPrefix(err.Error(), wantErr.Error()) {
+		t.Errorf("%s, got [%v], want [%v]", testedFn, err, wantErr)
+	}
+}
+
+func TestExecuteWhenMaxPayloadSizeNotExceeded(t *testing.T) {
+	// call is defined in collector_test.go
+	call.MaxPayloadSize = 100
```
mfulb, line 134: See above comment about `call` and `actualData`.

src/newrelic/app_test.go
```diff
@@ -452,8 +450,7 @@ func TestMaxPayloadSizeInBytesFromDefault(t *testing.T) {
 
 func TestMaxPayloadSizeInBytesFromConnectReply(t *testing.T) {
 	expectedMaxPayloadSizeInBytes := 1000
-	id := AgentRunID("1") // parseConnectReply expects at least agent_run_id in collector reply
-	cannedConnectReply, _ := json.Marshal(ConnectReply{ID: &id, MaxPayloadSizeInBytes: expectedMaxPayloadSizeInBytes})
+	cannedConnectReply := []byte(`{"agent_run_id":"1", "max_payload_size_in_bytes":1000}`)
```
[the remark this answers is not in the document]
misiekn:
```suggestion
	cannedConnectReply := []byte(`{"agent_run_id":"1", "max_payload_size_in_bytes":`+fmt.Sprint(expectedMaxPayloadSizeInBytes)+`}`)
```
Is this acceptable?

src/newrelic/collector/client_test.go
misiekn: Is this acceptable?
```suggestion
	cmdPayload := "dummy-data"
	cmd := Cmd{
		Collectible: CollectibleFunc(func(auditVersion bool) ([]byte, error) {
			if auditVersion {
				return nil, nil
			}
			return []byte(cmdPayload), nil
		}),
		MaxPayloadSize: 100,
	}
	testedFn := fmt.Sprintf("client.Execute(payload: %v, MaxPayloadSize: %v)", cmdPayload, cmd.MaxPayloadSize)

```
----
Side note: It turns out that these fields of `Command` do not need to be set for client.Execute to work.
```
Name:      "collector-method",
Collector: "the-collector.com",
License:   "12345",
RunID:     "db97531",
```

misiekn (commented):
@mfulb Thanks for your review comments!

src/newrelic/app_test.go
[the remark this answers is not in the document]
mfulb: Yes looks great.

src/newrelic/collector/client_test.go
misiekn: I plan to repeat [this](https://github.com/newrelic/newrelic-php-agent/pull/453/files#r877294219) solution.

mfulb (approved):
Great Michal!

misiekn:
ok jenkins

misiekn:
ok jenkins

src/newrelic/collector/client.go
```diff
@@ -215,12 +215,16 @@ type clientImpl struct {
...
 }
 
-func (c *clientImpl) perform(url string, data []byte, userAgent string, requestHeadersMap map[string]string) ([]byte, error) {
+func (c *clientImpl) perform(url string, data []byte, userAgent string, requestHeadersMap map[string]string, maxPayloadSize int) ([]byte, error) {
 	deflated, err := Compress(data)
 	if nil != err {
 		return nil, err
 	}
 
+	if l := deflated.Len(); l > maxPayloadSize {
```
mfulb, line 224: If the payload exceeds the maximum size we should also update the Supportability metric `Supportability/Agent/Collector/MaxPayloadSizeLimit/<endpoint>` as outlined [here](https://source.datanerd.us/agents/agent-specs/blob/68735b2884ccc836fb2a148a4814bd420bdeb647/Connect-LEGACY.md#max-payload-size).

mfulb (changes requested):
I feel we should implement the `Supportability/Agent/Collector/MaxPayloadSizeLimit/<endpoint>` metric as well it is a small addition and would be helpful for troubleshooting.

zsistla:
According to the spec, we "may" add the `Supportability/Agent/Collector/MaxPayloadSizeLimit/<endpoint>` metric and increment when hitting the payload larger than max case.  Can you see what the work is needed to add?

mfulb:
> According to the spec, we "may" add the `Supportability/Agent/Collector/MaxPayloadSizeLimit/<endpoint>` metric and increment when hitting the payload larger than max case. Can you see what the work is needed to add?

Looking at how other metrics are handled in the daemon it appears we would need to add a counter in client.go which kept up with the number of times a paylaod is created and discarded over the maximum limit as well as have a way to access this from src/newrelic/harvest.go.  Then in the function createFinalMetrics() in harvest.go we would need to add a call to h.Metrics.AddCount() for this metric.

The trickiest part would seem to be recording the "<endpoint>" the spec mentions.  Not clear what the various endpoints might be but looking at the java agent it appears to be "metrics", "analytics", and "spans".  There could be others - would need to investigate further.  Given the Execute() method in client.go doesn't have an idea of the "<endpoint>" it would require some rethinking of how to handle this.  Probably better to make adding this metric a separate effort in light of that.

mfulb (approved):
Looks good - would be good to make a issue to track adding the Supportability metric at some point.

zsistla:
Yes, those are the endpoints we are tracking and the daemon knows about.  How about we add this supportability to the other supportability ticket? https://github.com/newrelic/newrelic-php-agent/issues/445

ZNeumann (approved):

misiekn:
ok jenkins

zsistla (approved):
Looks good, thanks!

PR #462, opened 2022-06-07
fix(agent): Improve handling TT, 8T and DT configuration dependencies

### LINKED ISSUE #452: Disable infinite tracing at launch if distributed tracing is disabled
Currently if infinite tracing is enabled while distributed tracing (DT) is disabled the following warning message will be output:

```warning: cannot create a span event when a segment ID cannot be generated; is distributed tracing enabled?```

This message can be output multiple times which is not the desired solution.

The proposed solution would be to disable infinite tracing at start up if DT is disabled and output a warning message to this effect.   The message would aid troubleshooting settings and would only appear once per agent launch. 

The original message can be shifted to a DEBUG log level instead of WARN so it might further aid troubleshooting without impacting normal usage of the agent.

mfulb:
This PR improves handling TT, 8T and DT configuration dependencies. In particular:
- If 8T is enabled and DT is disabled then this PR will:
   + Output a warning that 8T will be disabled because DT is disabled.
   + Sets the trace observer host to "" if DT is disabled.
   + If for some reason the previous mitigation failed then it will batch together the previous warning message which occurred each time the creation of a span was attempted which created significant log file output.  Now the message will appear at most once every 60 seconds with a count of the number of times it occurred in the past 60 seconds.
- If DT is enabled and the transaction tracer is disabled it will also batch together the error message which occurs each time the creation of a transaction is created and outputs a summary every 60 seconds (like above).  Otherwise a log error message occurs each time the creation of a transaction is attempted which can lead to significant log file growth.

Closes #452.

### DIFF
--- a/agent/php_minit.c
+++ b/agent/php_minit.c
@@ -236,6 +236,22 @@ static char* nr_php_check_for_upgrade_license_key(void) {
   return 0;
 }
 
+static nr_status_t nr_php_check_8T_DT_config(void) {
+  /* check if infinite tracing is enabled and DT disabled */
+  if (!nr_strempty(NRINI(trace_observer_host))
+      && !NRINI(distributed_tracing_enabled)) {
+    nrl_warning(
+        NRL_INIT,
+        "Infinite tracing will be DISABLED because distributed tracing is"
+        " disabled and infinite tracing requires distributed tracing to be "
+        "enabled.  Please check the"
+        " value of 'newrelic.distributed_tracing_enabled' in the agent "
+        "configuration.");
+    return NR_FAILURE;
+  }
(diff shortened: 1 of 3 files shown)


ZNeumann (approved):

zsistla (approved):
Thanks for catching/fixing this issue.

mfulb:
ok jenkins

zsistla:
ok jenkins

bduranleau-nr (approved):

mfulb:
ok jenkins

mfulb:
ok jenkins

misiekn (approved):
Thanks for the updates!

Issue #441, opened 2022-05-10
MUST Update/verify the daemon logs an error when a failure occurs that requires a retry of an agent endpoint.


zsistla:
Please see MMF document for further details and links.

Log the error at warning level or greater each time a failure occurs that requires a retry of an agent endpoint.

Acceptance criteria
- [ ] PR with code changes is created
- [ ] There are tests that cover code changes
- [ ] CI PR build passes
- [ ] Code is reviewed by at least 2 team members (3 team members for more complex tickets)
- [ ] Code is merged into dev branch

Note: See [go repo](https://github.com/newrelic/go-agent) for reference.

PR #466, opened 2022-06-09
feat(logs): P17 log response codes

ZNeumann:
https://github.com/newrelic/newrelic-php-agent/issues/441


zsistla:
Agreed adding a test case isn't needed for this.

mfulb (approved):

PR #464, opened 2022-06-08
php feat(axiom,agent,daemon): Incorporate metadata settings

zsistla:
Agent:
* Modify the agent to understand NEW_RELIC_METADATA environmental variables
* Created global environmental variable to store the metadata
* Added test cases to test the new environment scanning functionality and the new variable
* As with all things agent, the environment variables will only be checked ONCE at agent initialization, so the variables must exist prior to starting the agent to be incorporated into the agent variables.

Axiom:
* Modify axiom to send metadata extracted from the environment via flatbuffers to the daemon.
* Add test cases to verify functionalty.

Flatbuffer:
Add a metadata variable to the flatbuffer app table.

Daemon:
* Added metadata variable to receive the new metadata info the agent extracts from environment variables.
* Added metadata JSON to connect request
* Added test cases to test the functionality of the new variable.

Tested via new test cases and manual verification.

PR #467, opened 2022-06-16
PHP feat(agent,axiom): scan environment for singular and plural label configuration

zsistla:
Currently based off of PR #464  because it needed the environment parsing logic.  After that PR is merged, this can be rebased and merged.  Only the June 15+ commits are relevant to this PR.
Axiom:
1) Updated `nr_str_append` function.  
*Now it takes an additional delimter parameter to give more flexibility.
* Updated all calls to this function.
* Added/updated tests.
Agent:
1) Added new global variable `env_labels` to keep track of the labels that were extracted from the environment.  In particular, labels are designated with the key `NEW_RELIC_LABELS` or by keys prefixed with `NEW_RELIC_LABEL_`
2) Updated the environment variable parsing logic to handle both looking for a specific key and to handle parsing keys with prefixes into strings.
3)    Integrated environment labels with the ini labels to take advantages of all the labels functionality.  By appending the environment variables to the end of the ini string, we ensure the environmental variables take precedence when it is entered into the label hash in `nr_php_txn_get_labels` which is then associated with the transaction.  Additionally, with the environmental and ini values combined, we are able to take advantage of all the previous label structures, tests, and


agent/php_environment.c
```diff
@@ -382,6 +382,134 @@ static void nr_php_gather_dispatcher_information(nrobj_t* env) {
...
+  if ((NULL == prefix) || (NULL == key)) {
+    return kv_string;
+  }
+
+  if (nr_strlen(prefix) >= nr_strlen(key)) {
+    return kv_string;
+  }
+
+  if (0 == nr_strncmp(key, prefix, nr_strlen(prefix))) {
+    kv_string = nr_str_append(kv_string, key + nr_strlen(prefix), delimiter);
```
mfulb, line 417: Looks like `nr_str_append()` will ignore `delimiter` if it is `NULL` but would it make sense to add a test to catch this case and handle it differently if needed?

zsistla: It doesn't ignore it, the design was that is null was passed in as a delimiter, nr_str_append, simply appends the strings together.  It does specifically check for NULL and if it detects it appends the strings together so that if a delimiter is passed in, it is used when appending the strings and if no delimiter is passed in, it just appends the strings.  Where were you thinking an additional test or different handling might be good?

mfulb (approved):
Very nice contribution Amber - clever storage for labels so environment would take precedence.  :)

ZNeumann (approved):

ZNeumann (approved):

axiom/util_strings.c
```diff
@@ -423,8 +423,9 @@ int nr_str_char_count(const char* s, char c) {
   return count;
 }
 
-char* nr_str_append(char* dest, const char* src) {
+char* nr_str_append(char* dest, const char* src, const char* delimiter) {
   char* tmp = NULL;
+  const char* delim = delimiter? delimiter : ""; // Treat NULL delimiter as no delimiter
```
mfulb, line 428:
```suggestion
  const char* delim = delimiter ? delimiter : ""; // Treat NULL delimiter as no delimiter
```

mfulb: This line needs to be clang formatted.

mfulb: Maybe clearer to make test (NULL == delimiter)?

axiom/util_strings.h
```diff
@@ -240,14 +240,16 @@ extern int nr_strncspn(const char* s1, int s1len, const char* s2, int s2len);
 extern int nr_str_char_count(const char* s, char c);
 
 /*
- * Purpose : Append a string to the end of another string separated by a comma.
+ * Purpose : Append a string to the end of another string separated by a
+ * delimiter.
  *
  * Params  : 1. The destination string.
  *           2. The source string.
+ *           3. The delimiter to separate the strings; NULL treated as empty string
```
mfulb, line 248:
```suggestion
 *           3. The delimiter to separate the strings; NULL treated as empty string.
```

mfulb (approved):
Thank you for handling the rebase and nice tweak of nr_str_append().

misiekn:
ok jenkins

axiom/util_strings.c
misiekn: Done with https://github.com/newrelic/newrelic-php-agent/pull/467/commits/5f8f238154b84e281b3303511f194cbda6881ef2

PR #471, opened 2022-06-21
feat(daemon): add supportability metrics for new response codes and for harvest endpoints

zsistla:
Very rough prototype for addressing issue $445

1) Adds httperror metrics for new response codes
2) Adds metrics for these endpoints: https://github.com/newrelic/newrelic-php-agent/blob/main/src/newrelic/harvest.go#L139

Needs:
1) Validation
2) Test cases
3) has not added duration, but harvest has a start time, so could theoretically be added

### DIFF
--- a/src/newrelic/analytics_events.go
+++ b/src/newrelic/analytics_events.go
@@ -61,6 +61,14 @@ func (events *analyticsEvents) Split() (*analyticsEvents, *analyticsEvents) {
 	return e1, e2
 }
 
+// NumAttempts returns the total number of attempts sent to this endpoint
+// The value is the number of times the agent attempted to call the given endpoint before it was successful.
+// This metric MUST NOT be generated if only one attempt was made.
+// Does not include the successful attempt
+func (events *analyticsEvents) NumFailedAttempts() float64 {
+	return float64(events.failedHarvests)
+}
+
 // NumSeen returns the total number of analytics events observed.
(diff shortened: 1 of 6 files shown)


ZNeumann:
Added the a test for the new supportability metrics and ran gofmt on the test file. Hiding whitespace changes is recommended to review this PR

zsistla:
* Validated correctness
* Added NumFailedAttempts function to Events and Metrics
* collect metrics for the events/metrics we support retries with  
The value is the number of times the agent attempted to call the given endpoint before it was successful. This metric MUST NOT be generated if only one attempt was made.

For PHP, only metricEvents, ErrorEvents, CustomEvents, TxnEvents, and SpanEvents are retried; therefore only those will ever have a count higher than 0 failed attemps.  For more details, see:
The Data Collection Limits section of Collector-Response-Handling.md in the agent-specs.

* Added additional test for cases that should or should not send the new Attempts metric.
* removed IncrementEndpointsAttempted as it was counting incorrectly 

The duration metric is NOT supported at this time.

zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

ZNeumann (approved):

mfulb (approved):
Nice efficient solution.

PR #487, opened 2022-07-13
docs(agent): fix `detect_kubernetes` path in template ini

bduranleau-nr:
Found what appears to be a bug in the 'path' for the detect_kubernetes INI config option while investigating for logging.


mfulb (approved):
Oops - nice catch Brian!

ZNeumann (approved):

Issue #421, opened 2022-04-25
Change the max_samples_stored ini value to no longer have 0 default


ZNeumann:
Currently, `newrelic.span_events.max_samples_stored = 0` is the default and sets the value to the agent default of 2000. This is needlessly confusing. Change the default to be 2000 and change the agent so that 0 is actually 0.

zsistla:
Tracking this via internal ticket.

PR #488, opened 2022-07-13
feat(agent): add logging configuration options

bduranleau-nr:


agent/php_nrini.c
```diff
@@ -1730,6 +1730,45 @@ static PHP_INI_MH(nr_unsigned_int_mh) {
...
+
+  /* Anything other than a valid value will result in the
+   * default value of NR_DEFAULT_LOG_EVENTS_MAX_SAMPLES_STORED.
+   */
+
+  p->where = 0;
+
+  if (0 != NEW_VALUE_LEN) {
+    val = (int)strtol(NEW_VALUE, 0, 0);
+    if ((0 >= val) || (NR_MAX_LOG_EVENTS_MAX_SAMPLES_STORED < val)) {
```
mfulb, line 1757: I believe 0 is a valid value.  This looks like 0 would set the value to the max allowed which might be a surprise to a user.

bduranleau-nr: If that's the case, then would the nr_span_events_max_samples_stored_mh also have to be modified to accept 0 values, or are 0 values considered valid there? This block was lifted directly from that mh implementation.

zsistla: Yes, it also needs to be updated, see this ticket for reference: https://github.com/newrelic/newrelic-php-agent/issues/421

bduranleau-nr: Nice, I can make both changes here + now unless we don't want to change the span event one yet

zsistla: I'd vote to do both changes if you can.

agent/php_minit.c
```diff
@@ -252,6 +252,25 @@ static nr_status_t nr_php_check_8T_DT_config(TSRMLS_D) {
...
 }
 
+/*
+ * @brief Check the INI values for 'logging_enabled' and
+ *        'log_forwarding_enabled' and log a warning on
+ *        invalid configuration state.
+ *
+ * @return nr_status_t NR_FAILURE || NR_SUCCESS
+ */
+static nr_status_t nr_php_check_logging_config(TSRMLS_D) {
```
mfulb, line 262: Does this need a return value?  Later it is ignored.

agent/php_newrelic.h
```diff
@@ -445,9 +445,10 @@ nrinibool_t
 nrinibool_t
     distributed_tracing_exclude_newrelic_header; /* newrelic.distributed_tracing_exclude_newrelic_header
                                                   */
-nrinibool_t span_events_enabled;            /* newrelic.span_events_enabled */
-nriniuint_t span_events_max_samples_stored; /* newrelic.span_events.max_samples_stored
-                                             */
+nrinibool_t span_events_enabled; /* newrelic.span_events_enabled */
```
mfulb, line 448: Is this clang formatted?

agent/php_nrini.c
```diff
@@ -1730,6 +1730,45 @@ static PHP_INI_MH(nr_unsigned_int_mh) {
...
+  NR_UNUSED_TSRMLS;
+
+  /* Anything other than a valid value will result in the
+   * default value of NR_DEFAULT_LOG_EVENTS_MAX_SAMPLES_STORED.
+   */
+
+  p->where = 0;
+
+  if (0 != NEW_VALUE_LEN) {
+    val = (int)strtol(NEW_VALUE, 0, 0);
```
mfulb, line 1756: Would be good to check strtol() error conditions.

agent/php_nrini.c
```diff
@@ -1730,6 +1730,45 @@ static PHP_INI_MH(nr_unsigned_int_mh) {
...
+#endif
+
+  p = (nriniuint_t*)(base + (size_t)mh_arg1);
+
+  (void)entry;
+  (void)mh_arg3;
+  NR_UNUSED_TSRMLS;
+
+  /* Anything other than a valid value will result in the
+   * default value of NR_DEFAULT_LOG_EVENTS_MAX_SAMPLES_STORED.
```
mfulb, line 1750: Would be good to come up with how code will handle:
- value < 0
- value >  maximum allowed log max sampled stored (prob need to figure out this value as well)
-

agent/php_nrini.c
```diff
@@ -1730,6 +1730,45 @@ static PHP_INI_MH(nr_unsigned_int_mh) {
...
+   */
+
+  p->where = 0;
+
+  if (0 != NEW_VALUE_LEN) {
+    val = (int)strtol(NEW_VALUE, 0, 0);
+    if ((0 >= val) || (NR_MAX_LOG_EVENTS_MAX_SAMPLES_STORED < val)) {
+      val = NR_DEFAULT_LOG_EVENTS_MAX_SAMPLES_STORED;
```
mfulb, line 1758: Do we want too large a requested value to result in the default value or the largest allowed value?

agent/php_minit.c
bduranleau-nr: You're correct, I'll remove the return

agent/php_nrini.c
bduranleau-nr: I have removed the changes for #421 for inclusion into a future dedicated PR.

agent/php_nrini.c
bduranleau-nr: Agreed, this change has been made in a2abf5f0ec99bd03699d7bdb4e7f92dae395a170

agent/php_nrini.c
bduranleau-nr: See https://github.com/newrelic/newrelic-php-agent/pull/488#discussion_r925868034

agent/php_nrini.c
bduranleau-nr: I agree, but am conflicted how to go about it- I could handle the error here and leave it at that, or abstract out a nr_ helper function for strtol() functionality + error handling generic use

bduranleau-nr: Currently I do not believe such a function exists in the codebase, and strtol() is used without error checking in several places in the code.

bduranleau-nr: I have attempted to address this in commit 9a4bd3ffd24564223c3fc44ee417df789fa4a95c. Currently it has only been integrated into the log event samples INI check, but I believe it should be default for all other strtol() calls in this module.

bduranleau-nr:
ok jenkins

bduranleau-nr:
ok jenkins

mfulb (approved):
Thanks for the great addition Brian!

ZNeumann (approved):

agent/php_newrelic.h
```diff
@@ -460,6 +461,18 @@ nrinitime_t
     agent_span_queue_timeout; /* newrelic.infinite_tracing.span_events.agent_queue.timeout
                                */
 
+/*
+ * configuration options for logging in context
```
misiekn, line 465: I think it's better to leave 'logging in context' out from the code. How about this:
```suggestion
 * configuration options for handling application logging
```

agent/php_nrini.c
```diff
@@ -255,6 +255,28 @@ static void check_ini_int_min_max(int* var, int minval, int maxval) {
   }
 }
 
+/*
+ * @brief error handling for strtol() functionality
+ *
+ * @param val_p   pointer to store the parsed value
```
misiekn, line 261:
```suggestion
 * @param val_p   pointer to store the parsed value; remains undefined when
 * function returns NR_FAILURE
```

agent/php_nrini.c
```diff
@@ -255,6 +255,28 @@ static void check_ini_int_min_max(int* var, int minval, int maxval) {
...
+ * @param str     string to parse
+ * @param base    base to be used for number conversion
+ * @return nr_status_t NR_SUCCESS || NR_FAILURE
+ */
+static nr_status_t nr_strtol(int* val_p, const char* str, int base) {
+  int val;
+  char* endptr_p;
+  errno = 0;
+
+  val = (int)strtol(str, &endptr_p, base);
```
misiekn, line 271: `strtol` would suggest that the function will return `long` not `int`
```suggestion
static nr_status_t nr_strtol(long* val_p, const char* str, int base) {
  long val;
  char* endptr_p;
  errno = 0;

  val = strtol(str, &endptr_p, base);
```
alternatively change the name and the C library function which may be a better option seeing how `nr_strtol` is being used in the code.
```suggestion
static nr_status_t nr_strtoi(int* val_p, const char* str, int base) {
  int val;
  char* endptr_p;
  errno = 0;

  val = strtoimax(str, &endptr_p, base);
```

agent/php_nrini.c
```diff
@@ -1730,6 +1752,55 @@ static PHP_INI_MH(nr_unsigned_int_mh) {
...
+  /*
+   * -- An invalid value will result in the default value.
+   * -- A value < 0 will result in the default value
+   * -- A value > MAX will result in MAX value
+   */
+
+  p->where = 0;
+
+  if (0 != NEW_VALUE_LEN) {
+    parse_status = nr_strtol(&val, NEW_VALUE, 0);
```
misiekn, line 1782: Depending on which way you chose this line will need a change too.

agent/php_nrini.c
```diff
@@ -1730,6 +1752,55 @@ static PHP_INI_MH(nr_unsigned_int_mh) {
...
+    if (0 > val || NR_FAILURE == parse_status) {
+      val = NR_DEFAULT_LOG_EVENTS_MAX_SAMPLES_STORED;
+      err = true;
+    } else if (NR_MAX_LOG_EVENTS_MAX_SAMPLES_STORED < val) {
+      val = NR_MAX_LOG_EVENTS_MAX_SAMPLES_STORED;
+      err = true;
+    }
+    if (err) {
+      nrl_debug(NRL_INIT,
```
misiekn, line 1791: I think it's better to issue `nrl_warning`. That's what's been done for `nr_loglevel_mh` when invalid value is detected.
```suggestion
      nrl_warning(NRL_INIT,
```

misiekn (changes requested):
Nice work! Just a few tweaks needed here and there.

agent/php_nrini.c
misiekn: Maybe create a ticket for backlog to add strtol wrapper to axiom/util_number_converter so making this improvement is not forgotten?

agent/scripts/newrelic.ini.template
```diff
@@ -1145,3 +1145,40 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
+;
+;newrelic.application_logging.forwarding.enabled = false
+
+; Setting: newrelic.application_logging.forwarding.max_samples_stored
+; Type   : integer
+; Scope  : per-directory
+; Default: 10000
+; Info   : Number of log records to send per minute to New Relic. Controls the 
+;          overall memory consumption when using log forwarding. SHOULD be sent 
+;          as part of the harvest_limits on Connect.
```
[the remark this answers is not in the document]
bduranleau-nr: Ahh I saw this and meant to cut it out but it slipped through. Fixed it now- Good catch!

agent/scripts/newrelic.ini.template
```diff
@@ -1145,3 +1145,40 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
+;
+;newrelic.application_logging.forwarding.max_samples_stored = 10000
+
+; Setting: newrelic.application_logging.metrics.enabled
+; Type   : boolean
+; Scope  : per-directory
+; Default: true
+; Info   : Toggles whether the agent gathers the the user facing Logging/lines
```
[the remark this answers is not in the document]
bduranleau-nr: fixed in 612aad50d85ef1bcd09f2c8cf461bed7f8c724a5

agent/scripts/newrelic.ini.template
```diff
@@ -1145,3 +1145,40 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ; Info   : Indicates if WordPress hooks are to be instrumented.
 ;
 ;newrelic.framework.wordpress.hooks = true
+
+; Setting: newrelic.application_logging.enabled
+; Type   : boolean
+; Scope  : per-directory
+; Default: true
+; Info   : If this is disabled, all sub-features are disabled; if it is enabled, 
+;          the individual sub-feature configurations take effect.
```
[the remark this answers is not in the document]
bduranleau-nr: fixed in 612aad50d85ef1bcd09f2c8cf461bed7f8c724a5

axiom/nr_txn.h
```diff
@@ -114,6 +114,11 @@ typedef struct _nrtxnopt_t {
                                    spans will be batched, and non-8T behaviour
                                    will be used. */
   nrtime_t span_queue_batch_timeout; /* Span queue batch timeout in us. */
+  bool logging_enabled;              /* Whether logging in context is enabled */
```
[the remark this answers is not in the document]
bduranleau-nr: fixed in 9ba9ddc02456e66d001513b8fc7d9dcde0c07475

agent/php_newrelic.h
bduranleau-nr: fixed in 8f2c5758ac268ad0ac61e8b726eabd3a10de9652

agent/php_nrini.c
bduranleau-nr: resolved in 84867ece87b76e35e2b05d1458f8706cbcc3cc40

agent/php_nrini.c
bduranleau-nr: resolved in 84867ece87b76e35e2b05d1458f8706cbcc3cc40

bduranleau-nr:
ok jenkins

agent/php_nrini.c
```diff
@@ -255,6 +255,29 @@ static void check_ini_int_min_max(int* var, int minval, int maxval) {
...
+ * @param str     string to parse
+ * @param base    base to be used for number conversion
+ * @return nr_status_t NR_SUCCESS || NR_FAILURE
+ */
+static nr_status_t nr_strtoi(int* val_p, const char* str, int base) {
+  int val;
+  char* endptr_p;
+  errno = 0;
+
+  val = (int)strtoimax(str, &endptr_p, base);
```
misiekn, line 272: I don't think the cast to `int` is needed here.
```suggestion
  val = strtoimax(str, &endptr_p, base);
```

bduranleau-nr:
ok jenkins

bduranleau-nr:
ok jenkins

misiekn (approved):
LGTM 👍

PR #489, opened 2022-07-19
feat(agent): instrument Monolog logging

misiekn:
- [x] Implement detecting [Monolog](https://github.com/Seldaek/monolog/tree/main) and create `Supportability/Logging/{language}/{framework}/{enabled|disabled}` metric
- [x] Implement detecting conflict with [newrelic-monolog-logenricher-php](https://github.com/newrelic/newrelic-monolog-logenricher-php)
- [x] Instrument methods in [Monolog](https://github.com/Seldaek/monolog/tree/main) to support log forwarding/metrics
- [x] Add ‘Logging/lines’, ‘Logging/lines/{SEVERITY}’ logging metrics
- [x] Add code to observe relevant configuration settings


agent/lib_monolog.c
```diff
@@ -0,0 +1,135 @@
...
+NR_PHP_WRAPPER(nr_monolog_logger_pushhandler) {
+  (void)wraprec;
+
+  zval* handler = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  if (!nr_php_is_zval_valid_object(handler)) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: handler is not an object", __func__);
+    goto end;
+  }
+
+  if (/*NRINI(log_forwarding_enabled) && */ nr_php_object_instanceof_class(
```
mfulb, line 117: We might consider also checking if the max log events limit is 0 here and not instrument in that case as an optimization?  It is a MAY in the agent specs to do this.

agent/lib_monolog.c
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+end:
+  NR_PHP_WRAPPER_CALL
+  nr_php_arg_release(&handler);
+}
+NR_PHP_WRAPPER_END
+
+void nr_monolog_enable(TSRMLS_D) {
+  nr_php_wrap_user_function(NR_PSTR("Monolog\\Logger::pushHandler"),
```
mfulb, line 131: Can nr_php_wrap_user_function() fail and what would be the desired handling of this case?

agent/php_execute.c
```diff
@@ -569,6 +569,12 @@ static nr_library_table_t libraries[] = {
 
 static size_t num_libraries = sizeof(libraries) / sizeof(nr_library_table_t);
 
+static nr_library_table_t logging_frameworks[] = {
+    /* Monolog - Logging for PHP */
+    {"Monolog", "monolog/logger.php", nr_monolog_enable},
+};
+
+static size_t num_logging_frameworks = sizeof(logging_frameworks) / sizeof(nr_library_table_t);
```
mfulb, line 577: Nice implementation allows for future expansion.

agent/php_execute.c
```diff
@@ -874,6 +880,30 @@ static void nr_execute_handle_library(const char* filename TSRMLS_DC) {
   nr_free(filename_lower);
 }
 
+static void nr_execute_handle_logging_framework(const char* filename TSRMLS_DC) {
+  char* filename_lower = nr_string_to_lowercase(filename);
+  size_t i;
+
+  for (i = 0; i < num_logging_frameworks; i++) {
+    if (nr_stridx(filename_lower, logging_frameworks[i].file_to_check) >= 0) {
+      char* metname = nr_formatf("Supportability/Logging/PHP/%s/enabled",
```
mfulb, line 889: I think this metric depends on if the log forward is enabled (and max log event > 0) and not just detection?

agent/php_execute.c
```diff
@@ -874,6 +880,30 @@ static void nr_execute_handle_library(const char* filename TSRMLS_DC) {
   nr_free(filename_lower);
 }
 
+static void nr_execute_handle_logging_framework(const char* filename TSRMLS_DC) {
```
mfulb, line 883: Small nit but maybe should be called "nr_execute_handle_logging_frameworks()?".

agent/lib_monolog.c
misiekn: I think I like the idea of having an API for this, say a simple inline`nr_log_forwarding_enabled(...)`. Then, apart from making this unit-testable, the code could look like this:
```suggestion
  if (nr_log_forwarding_enabled(...) && nr_php_object_instanceof_class(
```
... mean that at this time I don't know what arguments should this function take, maybe none?

mfulb: Currently I think the implementation for log config options associated a set of values for all the logging config options with each transaction.  So perhaps the nr_log_forwarding_enabled() would accept a transaction?  And then look at other settings like HSM to make a final decision.

agent/php_execute.c
misiekn: I think the intention of why this set of functions: `nr_execute_handle_framework`, `nr_execute_handle_library`,  `nr_execute_handle_logging_framework` end in singular form of the noun is because they operate on one file that is being executed and they try to handle if the file is part of a framework/library. I'd stick with singular form. I don't mind changing the `_logging_framework` part to `_logging_library` - in PHP world Monolog is called a 'package' anyway :)

agent/lib_monolog.c
misiekn: It can fail however the return value/error is ignored and the result is that the php-agent wrapper is never executed when instrumented function is executed by php. Please note that every library and framework is instrumented this way in the agent. See https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/agent/lib_predis.c#L747 or https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/agent/fw_wordpress.c#L581 or any other `nr_*_enable` method in either `fw_*.c` or `lib_*.c` modules.

agent/php_execute.c
mfulb: Ah that makes sense for the singular. Thanks!

agent/lib_monolog.c
```diff
@@ -0,0 +1,135 @@
...
+
+  if (3 <= argc) {
+    context = nr_php_arg_get(3, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+    nrl_verbosedebug(
+        NRL_INSTRUMENT, "%s: typeof(context)=[%d], len(context)=%ld", __func__,
+        Z_TYPE_P(context), nr_php_zend_hash_num_elements(Z_ARRVAL_P(context)));
+  }
+
+  /* $datetime is only available since API level 2 */
+  if (2 <= api && 4 <= argc) {
```
bduranleau-nr, line 91: Would it make sense to abstract these numbers out to macros for clarity?

agent/php_execute.c
bduranleau-nr: According to the spec, `Supportability/Logging/{language}/{framework}/{enabled|disabled}` - enabled MUST be implemented, disabled MAY be implemented, but does not depend on the state of individual features. From the spec: 

> This metric does not take the enabled/disabled state of individual features (i.e. metrics, forwarding) into account.

misiekn: Should it depend on the global `newrelic.application_logging.enabled`?

agent/lib_monolog.c
misiekn: I don't think it will improve code readability. These are positions of function arguments and are not related with arguments and arguments' names themselves. I think leaving the number explicit and not hidden behind the macro makes the code more readable here.

misiekn:
ok jenkins

misiekn:
ok jenkins

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+#include "fw_support.h"
+#include "nr_datastore_instance.h"
+#include "nr_segment_datastore.h"
+#include "util_logging.h"
+#include "util_memory.h"
+#include "util_strings.h"
+#include "util_sleep.h"
+
+/*
+ * Purpose : Return a copy of Monolog\Logger::API.
```
zsistla, line 23: The purpose seems different than what the return indicates.

agent/php_execute.c
zsistla: I agree, I think that metric should depend on the global newrelic.application_logging.enabled.

We could use Supportability/library/Monolog/detected for library detected.
And we could use Supportability/Logging/Php/Monolog/enabled if we detect  a logging library it AND if logging is enabled

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+  if (!nr_php_object_has_method(logger, "getLevelName" TSRMLS_CC)) {
+    nrl_verbosedebug(NRL_INSTRUMENT,
+                     "%s: Logger does not have getLevelName method", __func__);
+    return level_name_string;
+  }
+
+  zval* level = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  level_name = nr_php_call(logger, "getLevelName", level);
+  if (nr_php_is_zval_valid_string(level_name)) {
+    level_name_string = nr_strdup(Z_STRVAL_P(level_name));
```
zsistla, line 86: This seems like a potential memory leak since we [allocate mem via strdup](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/util_memory.c#L141)  in [line 75](https://github.com/newrelic/newrelic-php-agent/pull/489/files#diff-d48bf43b4158fad631ce65b0464841310be10cff57ac0b937061fb9d6c33bdc3R75), but then we allocate it again for this strdup.

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+  return retval;
+}
+
+/*
+ * Purpose : Convert $level argument of Monolog\Logger::addRecord to a string
+ * representation of Monolog's log level.
+ *
+ * Params  : Logger instance and Monolog\Logger::addRecord argument list
+ *
+ * Returns : A new string with Monolog's log level name
```
zsistla, line 70: Is the caller responsible for freeing?

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ */
+static char* nr_monolog_get_level_name(zval* logger,
+                                       NR_EXECUTE_PROTO TSRMLS_DC) {
+  zval* level_name = NULL;
+  char* level_name_string = nr_strdup("UNKNOWN");
+
```
zsistla, line 76: Would be good to check logger is a valid object and not NULL.

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Params  : zend_hash_key
+ * *
+ * Returns : A new string representing zval; caller must free
+ *
+ */
+static char* nr_monolog_fmt_context_key(zend_hash_key* hash_key) {
+  char* key_str = NULL;
+  zval* key = nr_php_zval_alloc();
+  if (nr_php_zend_hash_key_is_string(hash_key)) {
```
zsistla, line 136: Is it safer to use `nr_php_is_zval_non_empty_string(hash_key)` ?

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Params  : zval value
+ * *
+ * Returns : A new string representing zval; caller must free
+ *
+ */
+static char* nr_monolog_fmt_context_value(zval* zv) {
+  char* val_str = NULL;
+  zval* zv_str = nr_php_zval_alloc();
+
```
zsistla, line 164: check for NULL?

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Params  : zval value, pointer to string buffer to store formatted output
+ * and hash key
+ *
+ * Side effect : string buffer is reallocated with each call.
+ *
+ * Returns : ZEND_HASH_APPLY_KEEP to keep iteration
+ *
+ */
+static int nr_monolog_fmt_context_item(zval* value,
```
zsistla, line 184: Global comment: for the zvals that are passed in that we know we don't want to modify, consider passing them in like: `const zval* value`

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+                     __func__, Z_TYPE_P(context_arg));
+    goto return_context;
+  }
+
+  if (0 == nr_php_zend_hash_num_elements(Z_ARRVAL_P(context_arg))) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: $context has no elements", __func__);
+    goto return_context;
+  }
+
+  context = nr_monolog_fmt_context(context, Z_ARRVAL_P(context_arg) TSRMLS_CC);
```
zsistla, line 252: Possible mem leak since context had allocated mem [here](https://github.com/newrelic/newrelic-php-agent/pull/489/files#diff-d48bf43b4158fad631ce65b0464841310be10cff57ac0b937061fb9d6c33bdc3R230), but then gets set to another allocated string here.

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Params  : # of Monolog\Logger::addRecord arguments, and
+ * Monolog\Logger::addRecord argument list
+ *
+ * Returns : A new string with a log record message; caller must free
+ */
+static char* nr_monolog_build_message(size_t argc, NR_EXECUTE_PROTO TSRMLS_DC) {
+  char* message_and_context = nr_strdup("");
+
+  char* message = nr_monolog_get_message(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
zsistla, line 272: What does the spec say to do if the message is NULL or empty?

zsistla:
Fantastic job instrumenting this library and amazing job digging into all the details!

Would it be feasible to add some unit tests for your helper functions, similar to how it was done for drupal: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/tests/test_fw_drupal.c

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Returns : A new string with Monolog's log context
+ */
+static char* nr_monolog_get_context(size_t argc, NR_EXECUTE_PROTO TSRMLS_DC) {
+  char* context = nr_strdup("");
+  zval* context_arg = NULL;
+
+  if (3 > argc) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: $context not available", __func__);
+    goto return_context;
```
ZNeumann, line 235: I know `goto`s are present in other parts of our codebase, but is this something that we want to avoid adding more of? Or are we fine with it in simple contexts like this?

agent/lib_monolog.c
mfulb: "A non-null, non-empty message. If the message is nil, empty, or missing, or simply whitespace, the log record MUST NOT be reported."

agent/lib_monolog.c
```diff
@@ -0,0 +1,387 @@
...
+ * current time otherwise.
+ */
+static nrtime_t nr_monolog_get_timestamp(int api,
+                                         size_t argc,
+                                         NR_EXECUTE_PROTO TSRMLS_DC) {
+  nrtime_t timestamp = nr_get_time();
+
+  zval* datetime = NULL;
+  /* $datetime is only available since API level 2 */
+  if (2 <= api && 4 <= argc) {
```
mfulb, line 300: Maybe make the 2 and 4 something like MONOLOG_API_2 and MONOLOG_API_4 just to give some meaning to the values?

agent/lib_monolog.c
misiekn: `nr_php_object_has_method` will check if logger is a valid object

https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/agent/php_agent.c#L111

 with the help of `nr_php_is_zval_valid_object`

https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/agent/php_zval.h#L186

Is checking if `logger` is NULL with `nr_php_object_has_method` enough?

agent/lib_monolog.c
misiekn: This spec requirement will be handled at the time when the log event is added to the transaction.

agent/lib_monolog.c
misiekn: I don't think it will improve code readability. It should be clear from the context that 2 is the API level and 4 is the number of arguments. I think leaving the explicit numbers and not hidden behind the macros makes the code more readable here. Maybe changing the `int api` to `const int monolog_api` and `size_t argc` to `const size_t add_record_argc` might. Then the code would read:
```
  if (2 <= monolog_api && 4 <= add_record_argc) {
```
What do you think?

agent/lib_monolog.c
misiekn: I wanted to avoid multiple return points from this function and avoid re-allocating memory. But since this approach led to memory leak, I'll refactor it.

agent/lib_monolog.c
misiekn: `hash_key` is not of `zval` type so it's not possible to use `nr_php_is_zval_non_empty_string`. However `nr_php_zend_hash_key_is_string` does check if the `hash_key` is NULL and returns false in such case

https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/agent/php_hash.h#L28

This behavior ensures that this function returns `unsuported-key-type` generated in line 147.

agent/lib_monolog.c
misiekn: It is a leftover from initial version of this function. I'll change the description.

agent/lib_monolog.c
misiekn: Yes.

agent/lib_monolog.c
misiekn: Nice catch - thank you! Good 👁️

agent/lib_monolog.c
misiekn: Nice catch - thank you! Good 👁️

misiekn (commented):
@zsistla @mfulb @ZNeumann Thank you for thorough review and comments! I responded to a few of the comments and will change the code to address issues that were pointed out.

agent/lib_monolog.c
misiekn: Upon further inspection this actually is not a memory leak. `nr_monolog_fmt_context` takes the `context` as an argument and then memory pointed to by `context` gets reallocated by `nr_str_append` in `nr_monolog_fmt_context`.

agent/lib_monolog.c
misiekn: @mfulb This has been addressed in 387fd61011084a537d93620bfea32f23d608f94e

agent/php_execute.c
misiekn: @zsistla @mfulb @bduranleau-nr This has been addressed in aaa733c9d8335c06031cbe0370eb5d8fa367d8d0 and refactored in 03b1abd51e6853f5e2c5ebef5053420ac10b2d82.

agent/lib_monolog.c
misiekn: @zsistla This has been addressed in 7cbdff18a4ac99dc5e165d7c1f000140f1b851b6

agent/lib_monolog.c
misiekn: @zsistla This has been addressed in 39bd9db2ec30ad4422cb92a17ce09e73c84b62ca

agent/lib_monolog.c
misiekn: @zsistla This has been addressed in 7cbdff18a4ac99dc5e165d7c1f000140f1b851b6

agent/lib_monolog.c
misiekn: @zsistla This has been addressed in 15a434214eb78fd96f6d723cba1c4308147c1a93. 

**Note**
It was not possible to use `const` for `nr_monolog_get_level_name` `logger` argument, because it is used in `nr_php_object_has_method` and `nr_php_call`. The later functions signature has first argument's type as `zval*`. Casting from `const zval*` down to `zval*`, when using `logger` in those function calls, was not an option, because of `-Werror=cast-qual` compile switch, which gives the following compilation error:
```
php-agent/agent/lib_monolog.c:77:33: error: cast discards ‘const’ qualifier from pointer target type [-Werror=cast-qual]
   77 |   if (!nr_php_object_has_method((zval*) logger, "getLevelName" TSRMLS_CC)) {
      |                                 ^
```

misiekn:
> Would it be feasible to add some unit tests for your helper functions, similar to how it was done for drupal:

@zsistla I like this idea very much but I need to investigate how feasible is that because `nr_monolog_get_*` set of helper functions work with either `zend_execute_data` (through `NR_EXECUTE_PROTO`) or with instance of Monolog\Logger object so the unit tests would have to mock these.

agent/lib_monolog.c
misiekn: @mfulb I hope c570e48d08a7ba55fb48c0adbfa44f0403b41066 increases code clarity and maintainability

misiekn:
ok jenkins

misiekn:
ok jenkins

misiekn:
ok jenkins

agent/config.m4
```diff
@@ -183,7 +183,7 @@ if test "$PHP_NEWRELIC" = "yes"; then
...
...codeigniter.c fw_drupal8.c fw_drupal.c fw_drupal_common.c fw_joomla.c fw_kohana.c fw_laminas3.c fw_laravel.c fw_laravel_queue.c fw_lumen.c fw_magento1.c fw_magento2.c fw_magento_common.c fw_mediawiki.c fw_silex.c fw_slim.c fw_support.c fw_symfony4.c fw_symfony2.c fw_symfony.c fw_symfony_common.c fw_wordpress.c fw_yii.c fw_zend2.c fw_zend.c lib_doctrine2.c lib_guzzle3.c lib_guzzle4.c lib_guzzle6.c lib_guzzle_common.c lib_mongodb.c lib_phpunit.c lib_predis.c lib_zend_http.c php_agent.c php_api.c php_api_datastore.c php_api_distributed_trace.c php_api_internal.c php_autorum.c php_call.c php_curl.c php_curl_md.c php_datastore.c php_environment.c php_error.c php_execute.c php_explain.c php_explain_mysqli.c php_explain_pdo_mysql.c php_extension.c php_file_get_contents.c php_globals.c php_hash.c php_header.c php_httprequest_send.c php_internal_instrument.c php_minit.c php_mshutdown.c php_mysql.c php_mysqli.c php_newrelic.c php_nrini.c php_output.c php_pdo.c php_pdo_mysql.c php_pdo_pgsql.c php_pgsql.c php_psr7.c php_redis.c php_rinit.c php_rshutdown.c php_samplers.c php_stack.c php_stacked_segment.c php_txn.c php_user_instrument.c php_vm.c php_wrapper.c, $ext_shared,, \\$(NEWRELIC_CFLAGS))
```
bduranleau-nr, line 186: This is a real nitpick, but it looks like this list is organized into fw_ -> lib_ -> php_, which doesn't have an impact on the code itself, but likely makes it easier to find things in this long list. I would suggest putting lib_monolog.c in the lib_ "section" for consistency, or even better in my opinion, breaking this out into separate calls to PHP_NEW_EXTENSION with each type, to try and make this more readable overall.

agent/lib_monolog.c
```diff
@@ -0,0 +1,397 @@
...
+static char* nr_monolog_get_level_name(zval* logger,
+                                       NR_EXECUTE_PROTO TSRMLS_DC) {
+  zval* level_name = NULL;
+  char* level_name_string = NULL;
+
+  if (!nr_php_object_has_method(logger, "getLevelName" TSRMLS_CC)) {
+    nrl_verbosedebug(NRL_INSTRUMENT,
+                     "%s: Logger does not have getLevelName method", __func__);
+  } else {
+    zval* level = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
mfulb, line 81: What will level be if the number of args are not what are expected?  Do we need a check to verify the return value is valid for the use case?

agent/config.m4
bduranleau-nr: or for example, unique macros for FW types, LIB types, and PHP types that contain the respective files and are passed to the file list argument in PHP_NEW_EXTENSION

axiom/nr_txn.c
```diff
@@ -3247,3 +3247,74 @@ char* nr_txn_get_current_span_id(nrtxn_t* txn) {
...
+  nrm_force_add(txn->unscoped_metrics, "Logging/lines", 0);
+  metric_name = nr_formatf("Logging/lines/%s", level_name);
+  nrm_force_add(txn->unscoped_metrics, metric_name, 0);
+  nr_free(metric_name);
+}
+
+void nr_txn_record_log_event(NRUNUSED nrtxn_t* txn,
+                             NRUNUSED const char* log_level_name,
+                             NRUNUSED const char* log_message,
+                             NRUNUSED nrtime_t timestamp) {
```
bduranleau-nr, line 3312: it doesn't seem like these arguments are needed in this function

axiom/tests/test_txn.c
```diff
@@ -8068,6 +8068,122 @@ static void test_segment_record_error(void) {
...
 
+static void test_record_log_event(void) {
+  /* This is the plan for this test:
+      1. Setup state (create transaction with logging options)
+      2. Call nr_txn_record_log_event();
+      3. Check for state:
+        - log_event is added to transaction log events
+        - check if metrics are created
+  */
+}
```
bduranleau-nr, line 8079: Is this just a placeholder for future work?

axiom/nr_txn.c
misiekn: @bduranleau-nr This is a stub of an API. This API will be implemented in a different PR. The implementation will make use of all the parameters as these are the attributes of a log event. Please see the comment in line 3313 as well as https://github.com/newrelic/newrelic-php-agent/pull/489/commits/6e5f790ad0939ab6cb03399d80661f73ca6270af commit message.

agent/lib_monolog.c
misiekn: @mfulb It will be `NULL` returned by lower level functions here:
https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/agent/php_agent.c#L430
Using `NULL` in `nr_php_call(logger, "getLevelName", level);` will crash. I need to fix this. Good :eye:!

axiom/nr_txn.c
bduranleau-nr: My only concern with stub functions is that someone might assume that the function is already completely implemented if they only see it being called in lib_monolog. Since this seems like a chicken-and-egg situation though I can see the merit to this choice.

axiom/tests/test_txn.c
misiekn: @bduranleau-nr Correct.

agent/lib_monolog.c
misiekn: @mfulb It has been addressed in b19d60caf1b40634f05c0d227dbe36f0106f7989.

bduranleau-nr (approved):
👍

agent/config.m4
misiekn: @bduranleau-nr Will 7d981925208ec24a12098cf92a43479e2fca2c8d work?

bduranleau-nr: Yes! this is exactly what I was thinking- Thanks!

agent/lib_monolog.c
```diff
@@ -0,0 +1,416 @@
...
+ *
+ * Returns : A new string with Monolog's log message; caller must free
+ */
+
+static char* nr_monolog_get_message(NR_EXECUTE_PROTO TSRMLS_DC) {
+  char* message = NULL;
+  zval* message_arg = NULL;
+
+  message_arg = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  if (message_arg) {
```
mfulb, line 119: To be consistent with other checks in this file for the return value of `nr_php_arg_get()`.

```suggestion
  if (NULL == message_arg) {
```

misiekn: Of course! 🤦

agent/lib_monolog.c
misiekn: @ZNeumann Upon further inspection it turns out this function does not have mem leak issue. See [this comment](https://github.com/newrelic/newrelic-php-agent/pull/489/files#r935845159). There I do not plan to refactor it at this time.

agent/lib_monolog.c
```diff
@@ -0,0 +1,416 @@
...
+static char* nr_monolog_build_message(const size_t argc,
+                                      NR_EXECUTE_PROTO TSRMLS_DC) {
+  char* message_and_context = nr_strdup("");
+
+  char* message = nr_monolog_get_message(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  message_and_context = nr_str_append(message_and_context, message, "");
+  nr_free(message);
+
+  char* context = nr_monolog_get_context(argc, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  message_and_context = nr_str_append(message_and_context, context, " ");
```
mfulb, line 305:
```suggestion
    if (0 < nr_strlen(context))
        message_and_context = nr_str_append(message_and_context, context, " ");
```

This avoids appending a blank space on every message which has an empty context.

misiekn: @mfulb Thanks for great suggestion! I addressed this in bcbcaefad9e1380eb03bcf9c86e292090328cb63. However I opted to use `nr_strempty` instead of `nr_strlen` as the former function exists for this particular use case.

mfulb (approved):
Thank you Michal!

PR #492, opened 2022-07-20
chore(agent/daemon): Update flatbuffers for flatc v2.0.6

mfulb:
Moving to the latest flatc compiler (v2.0.6) generates different output (though functionally equivalent) to what the current autogenerated files (from an old flatc) look like.
    
This commit includes these new autogenerated files as well as some minor changes in related Go files to support changes to how booleans are handed in the autogenerated files.  Since a byte and bool type in Go both use 1 byte of storage it should be functionally equivalent.


misiekn (approved):
Thanks for taking care of this! 👍

zsistla (approved):
Nice work, thanks!

PR #493, opened 2022-07-26
ci: add release workflow

kentonshade:
Add a new workflow to release newrelic-php-agent. This workflow uses custom container images for building newrelic-php-agent artifacts for release. 

Note: In this initial PR a private Docker Hub registry is used as source of container images. This is to be replaced with New Relic Docker Hub.

### DIFF
--- a/.github/workflows/NEW-release-build.yml
+++ b/.github/workflows/NEW-release-build.yml
@@ -1,0 +1,368 @@
+#
+# This is the workflow to do a release build.
+# GHA: 04/07: This is actually replacing the php-release-agent job in jenkins.
+# We do not need the php-release-agent-checkout job.
+# https://issues.newrelic.com/browse/NR-6477
+#
+
+---
+
+name: 'NEW release_CI (php-release-agent)'
+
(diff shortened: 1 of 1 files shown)


kentonshade:
@zsistla This most likely will fail.  We'll need to transfer/create secrets, change refs on checkout actions, etc.

kentonshade:
@zsistla So already I see my Docker Hub creds will not be in the upstream secrets.  I prefer not to add those.  We're still waiting on the NR Docker Hub request.

.github/workflows/NEW-release-build.yml
```diff
@@ -0,0 +1,368 @@
+#
+# This is the workflow to do a release build.
+# GHA: 04/07: This is actually replacing the php-release-agent job in jenkins.
+# We do not need the php-release-agent-checkout job.
+# https://issues.newrelic.com/browse/NR-6477
```
misiekn, line 5:
```suggestion
```

misiekn:
Superseded by #623.

PR #494, opened 2022-07-26
fix(agent): fix path text for logging INIs

bduranleau-nr:


mfulb (approved):
Thanks for fixing this.

PR #499, opened 2022-08-01
style(daemon): gofmt daemon code

misiekn:
This PR contains only the changes after running `gofmt` against daemon source code.


mfulb (approved):
Thanks Michal!

ZNeumann (approved):

src/newrelic/collector/certs_system.go
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+//go:build use_system_certs
 // +build use_system_certs
```
bduranleau-nr, line 7: I believe that // +build can be removed since // go:build is now the preferred directive

(stackoverflow for context: https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives)

not sure if this is something we want to address in this PR or its own, also repeated in certs_test and compile_certs from what I see so far.

misiekn: For compatibility with different build environments both forms of this directives should be kept.

bduranleau-nr (approved):

PR #500, opened 2022-08-05
feat(agent): record log events in transaction

misiekn:
This PR adds log events pool to transaction as well as associated API to add log events to the pool with unit tests. Log events pool is implemented with analytics event pool API to reduce code duplication and reuse tested code.


axiom/nr_log_events.h
```diff
@@ -0,0 +1,42 @@
...
+#define NR_LOG_EVENTS_HDR
+
+#include "nr_analytics_events.h"
+#include "nr_log_event.h"
+#include "util_random.h"
+
+/*
+ * Purpose : Convert a log event to an analytic event and add it to
+ *           an event pool.
+ *
```
mfulb, line 16: Mention the return value and its meaning?

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+#include "util_memory.h"
+
+bool nr_log_events_add_event(nr_analytics_events_t* events,
+                             const nr_log_event_t* event,
+                             nr_random_t* rnd) {
+  bool events_sampled = false;
+  char* log_event_json = NULL;
+  nr_analytics_event_t* log_event;
+
+  if (0 == events) {
```
mfulb, line 22:
```suggestion
  if (NULL == events) {
```

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+    return false;
+  }
+  if (0 == event) {
```
mfulb, line 25:
```suggestion
  if (NULL == event) {
```

axiom/tests/test_log_events.c
```diff
@@ -0,0 +1,169 @@
...
+  json = nr_log_events_get_event_json(events,
+                                      nr_log_events_number_saved(events));
+  tlib_pass_if_null("event from out of bounds (higher)", json);
+
+  nr_random_destroy(&rnd);
+  nr_log_events_destroy(&events);
+  tlib_pass_if_null("events destroyed", events);
+}
+
+static void test_events_sample(void) {
```
mfulb, line 102: Maybe exercise the ```nr_analytics_events_is_sampling()``` function as well as the return value from ```nr_log_events_add_event()``` indicating sampling occurred?

axiom/tests/test_txn.c
```diff
@@ -130,6 +130,22 @@ nrapp_t* nr_app_verify_id(nrapplist_t* applist NRUNUSED,
   return p->txns_app;
 }
 
+const char* nr_app_get_host_name(const nrapp_t* app) {
```
mfulb, line 133: Isnt this already defined in nr_app.c?

axiom/tests/test_txn.c
```diff
@@ -130,6 +130,22 @@ nrapp_t* nr_app_verify_id(nrapplist_t* applist NRUNUSED,
...
+  if (NULL == app) {
+    return NULL;
+  }
+
+  return app->host_name;
+}
+
+const char* nr_app_get_entity_guid(const nrapp_t* app) {
```
mfulb, line 141: Isnt this already defined in nr_app.c?

axiom/tests/test_txn.c
```diff
@@ -8068,6 +8084,256 @@ static void test_segment_record_error(void) {
...
+  /* Setup app state */
+  nr_memset(&app, 0, sizeof(app));
+  app.state = NR_APP_OK;
+  app.entity_name = entity_name;
+
+  /* Setup log feature options */
+  nr_memset(&opts, 0, sizeof(opts));
+  opts.logging_enabled = true;
+  opts.log_forwarding_enabled = true;
+  opts.log_events_max_samples_stored = true;
```
mfulb, line 8102: This takes an integer value doesn't it?

axiom/tests/test_txn.c
misiekn: @mfulb Please see 87a2a7bc533b489f784c68e8fd655f00f16f0abf commit message for an explanation why this is required to be defined here too.

axiom/tests/test_txn.c
misiekn: @mfulb Please see 87a2a7bc533b489f784c68e8fd655f00f16f0abf commit message for an explanation why this is required to be defined here too.

axiom/nr_log_events.h
misiekn: Addressed in d5ff416e2f497e9e5651cd3b5d2667914e422f20.

axiom/tests/test_log_events.c
misiekn: Addressed in https://github.com/newrelic/newrelic-php-agent/commit/d5ff416e2f497e9e5651cd3b5d2667914e422f20.

axiom/tests/test_txn.c
misiekn: Addressed in 6683408e0823c8746b73219a130f9cd9e56edcd2

misiekn:
ok jenkins

axiom/nr_analytics_events.c
```diff
@@ -201,3 +201,10 @@ const char* nr_analytics_events_get_event_json(nr_analytics_events_t* events,
 
   return nr_analytics_event_json(events->events[i]);
 }
+
+bool nr_analytics_events_is_sampling(nr_analytics_events_t* events) {
```
mfulb, line 205: Check for events being valid (NULL, etc?).

axiom/nr_txn.c
```diff
@@ -3293,6 +3297,90 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
   return false;
 }
 
+#define ENSURE_LOG_LEVEL_NAME(level_name) \
```
mfulb, line 3300: Should a zero length string be converted to "UNKNOWN" as well?  Not clear if this is the same as not having a level defined at all (ie NULL)?

axiom/nr_txn.c
```diff
@@ -3293,6 +3297,90 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+    nr_log_event_set_guid(e, nr_app_get_entity_guid(app));
+  }
+}
+
+static nr_log_event_t* log_event_create(const char* log_level_name,
+                                        const char* log_message,
+                                        nrtime_t timestamp,
+                                        nrtxn_t* txn,
+                                        nrapp_t* app) {
+  nr_log_event_t* e = nr_log_event_create();
```
mfulb, line 3336: Verify event was created successfully?

axiom/nr_txn.c
```diff
@@ -3293,6 +3297,90 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+  nr_log_event_set_log_level(e, ENSURE_LOG_LEVEL_NAME(log_level_name));
+  nr_log_event_set_message(e, log_message);
+  nr_log_event_set_timestamp(e, timestamp);
+
+  log_event_set_linking_metadata(e, txn, app);
```
mfulb, line 3341: What is some required metadata wasn't available - do we still want to create the event if it does not satisfy what is expected?

axiom/nr_txn.c
```diff
@@ -3293,6 +3297,90 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+    nrl_debug(NRL_TXN, "%s: failed to create log event", __func__);
+    event_dropped = true;
+  } else {
+    event_dropped = nr_log_events_add_event(txn->log_events, e, rnd);
+  }
+  nr_log_event_destroy(&e);
+
+  nr_random_destroy(&rnd);
+
+  if (event_dropped) {
```
mfulb, line 3379: Does this metrics creation depend on if log metrics are enabled?

mfulb, line 3379:
```suggestion
  if (event_dropped && !nr_txn_log_metrics_enabled(txn)) {
```

mfulb: This patch fixes my test case for disabled metrics where log sampling occurs.

axiom/tests/test_cmd_txndata.c
```diff
@@ -55,10 +55,6 @@ int nr_get_daemon_fd(void) {
   return 0;
 }
 
-nrapp_t* nr_app_verify_id(nrapplist_t* applist NRUNUSED,
```
mfulb, line 58: Why was this removed?

axiom/nr_txn.c
```diff
@@ -3293,6 +3297,90 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+static void nr_txn_add_log_event(nrtxn_t* txn,
+                                 const char* log_level_name,
+                                 const char* log_message,
+                                 nrtime_t timestamp,
+                                 nrapp_t* app) {
+  nr_random_t* rnd = NULL;
+  nr_log_event_t* e = NULL;
+  bool event_dropped = false;
+
+  if (!nr_txn_log_forwarding_enabled(txn)) {
```
mfulb, line 3355: This causes an interesting problem - since nr_txn_log_forwarding_enabled() will return false if the max log samples is set to 0 then we have a case where the "dropped" metric isn't updated (which happens at the end of this function).
If the forwarding flag is enabled but max samples set to 0 it seems like we are dropping log lines and it should be reflected.

axiom/nr_txn.c
misiekn: I don't know. The spec is not clear about it.

axiom/nr_txn.c
misiekn: I don't know if linking metadata presence is required by backend. The spec is not clear about it.

axiom/nr_txn.c
misiekn: It does not. The spec clearly states it is part of log forwarding feature.

misiekn: I think the test needs to be changed because the spec clearly states `Logging/Forwarding/Dropped` is part of log forwarding feature and not part of log metrics feature.

axiom/tests/test_cmd_txndata.c
misiekn: Please see https://github.com/newrelic/newrelic-php-agent/commit/fbd8b2ff08e7e8ce26ff63afbc66c2871467cb7b commit message for an explanation why this is required to be defined here too.

axiom/tests/test_log_events.c
```diff
@@ -0,0 +1,182 @@
...
+static void test_events_success(void) {
+  /* Normal operation */
+  nr_analytics_events_t* events = NULL;
+  nr_log_event_t* e = NULL;
+  nr_random_t* rnd = nr_random_create_from_seed(12345);
+  const char* json;
+  const char* expected;
+
+  events = nr_log_events_create(10);
+  tlib_fail_if_null("events created", events);
```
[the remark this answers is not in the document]
misiekn: I do not think this is needed because:
1. `nr_analytics_events_t` is an opaque type that does not allow accessing individual fields.
2. There already exist [axiom/tests/test_analytics_events.c](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/tests/test_analytics_events.c) that tests analytic events API, which is the underlying and re-used implementation of log events API.

axiom/nr_txn.c
mfulb: I'll ask for clarification on this behavior - I agree as written it does seem like this would currently be created even if the metrics flag is set to disabled (which seems odd/wrong).

mfulb: Whether this is appropriate pending some further clarification on the spec.

axiom/nr_txn.c
misiekn: @mfulb Nice find! I agree. I think the solution to this would be to remove this check:
```
  if (0 == txn->options.log_events_max_samples_stored) {
    return false;
  }
```
from `nr_txn_log_forwarding_enabled` function and allow `txn->log_events` pool to have the size of 0. This way `nr_analytics_events_add_event` would correctly return information that an event is dropped. Thoughts?

axiom/nr_analytics_events.c
misiekn: What should be the return value if `events` is NULL?

axiom/nr_txn.c
mfulb: I really like having the pool size of 0 return false from `nr_txn_log_forwarding_enabled` as it will guarantee nothing gets added even if the pool was created accidentally with a non-zero size.
But it does complicate the implementation here.

axiom/nr_txn.c
mfulb: This is correct as implemented to resolving this conversation.

mfulb: This patch is not needed existing code working as desired.

axiom/nr_txn.c
```diff
@@ -531,6 +532,8 @@ nrtxn_t* nr_txn_begin(nrapp_t* app,
   nt->intrinsics = nro_new_hash();
 
   nt->custom_events = nr_analytics_events_create(app->limits.custom_events);
+  nt->log_events
```
mfulb, line 535: This needs to support creating a pool size of 0 - currently `nr_analytics_events_create()` returns a NULL ptr not a pool of size 0.

axiom/nr_txn.c
```diff
@@ -3293,6 +3297,90 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+                                 nrapp_t* app) {
+  nr_random_t* rnd = NULL;
+  nr_log_event_t* e = NULL;
+  bool event_dropped = false;
+
+  if (!nr_txn_log_forwarding_enabled(txn)) {
+    return;
+  }
+
+  if (nr_strempty(log_message)) {
```
mfulb, line 3359: I'm concerned if an app for some reason was generating lots of empty messages this could become a agent log file growth problem.  It is marked for DEBUG so hopefully wouldn't happen in production.  Is this message really needed once we ship the agent however?

axiom/tests/test_log_events.c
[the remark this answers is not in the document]
mfulb: That seems reasonable.

axiom/tests/test_txn.c
```diff
@@ -8068,14 +8084,165 @@ static void test_segment_record_error(void) {
...
+  /* Setup app state */
+  nr_memset(&app, 0, sizeof(app));
+  app.state = NR_APP_OK;
+  app.entity_name = entity_name;
+
+  /* Setup log feature options */
+  nr_memset(&opts, 0, sizeof(opts));
+  opts.logging_enabled = true;
+  opts.log_forwarding_enabled = true;
+  opts.log_events_max_samples_stored = 10;
```
mfulb, line 8102: I'm not sure this will have the intended effect as `nr_txn_begin()` will eventually (if not already) use `app->limits.log_events` for the pool size.

axiom/tests/test_txn.c
```diff
@@ -8068,14 +8084,165 @@ static void test_segment_record_error(void) {
...
+                     0, 0, 0, 0);
+  test_txn_metric_is("happy path, event recorded, metric created",
+                     txn->unscoped_metrics, MET_FORCED,
+                     "Logging/lines/" LOG_LEVEL, 1, 0, 0, 0, 0, 0);
+  tlib_pass_if_null(
+      "Logging/Forwarding/Dropped not created",
+      nrm_find(txn->unscoped_metrics, "Logging/Forwarding/Dropped"));
+  nr_txn_destroy(&txn);
+
+  /* Happy path with sampling */
```
mfulb, line 8233: Could you add a case with a pool size of 0 for log events and test sampling/events not being created?

axiom/tests/test_txn.c
```diff
@@ -8068,14 +8084,165 @@ static void test_segment_record_error(void) {
...
+  txn = new_txn_for_record_log_event_test(APP_ENTITY_NAME);
+  /* fill up events pool to force sampling */
+  while (!nr_analytics_events_is_sampling(txn->log_events)) {
```
mfulb, line 8236: References to `nr_analytics_events_is_sampling()` and other `nr_analytics_events_*` funcs might be best to be wrapped when used in the log events context so when we change the implementation to do priority sampling we won't have to find all of them?

axiom/nr_txn.c
misiekn: @mfulb Look at [axiom/nr_txn.c#L495](https://github.com/misiekn/newrelic-php-agent/blob/e7f4dcb489b4b7659732b83994157838aff1a876/axiom/nr_txn.c#L495) This is how it's done for other types of events - the `<event_type>_enabled` flag is set based on agent setting and app limits. This seems to be in line with what `nr_txn_log_forwarding_enabled` is currently doing (with `if (0 == txn->options.log_events_max_samples_stored) return false;` check).

axiom/nr_txn.c
misiekn: @mfulb Addressed in 68f066d23fff6bff33e682d04d0c9b0f54e25c40

axiom/nr_txn.c
misiekn: @mfulb Addressed in 211ffdc4b4ded6768ecbc2ad5122e3da8d1d75fb

axiom/tests/test_txn.c
misiekn: @mfulb Addressed in 486ce76a699939e65c40d50fdcc3d9affa6063ec

axiom/nr_txn.c
misiekn: Addressed in 68f066d23fff6bff33e682d04d0c9b0f54e25c40

axiom/nr_txn.c
misiekn: Addressed in 2f54982c76d39280b5c946062eb050dd25e12f8f

axiom/nr_txn.c
misiekn: Addressed in 211ffdc4b4ded6768ecbc2ad5122e3da8d1d75fb

axiom/tests/test_txn.c
misiekn: Addressed in 486ce76a699939e65c40d50fdcc3d9affa6063ec

axiom/tests/test_txn.c
misiekn: I refactored the code and exposed the size of events pool in 6fe185615622f11c9fef187a154acd718fde0c9c

axiom/nr_analytics_events.c
```diff
@@ -100,6 +100,13 @@ struct _nr_analytics_events_t {
   nr_analytics_event_t** events; /* Array of events */
 };
 
+int nr_analytics_events_max_events(const nr_analytics_events_t* events) {
+  if (0 == events) {
```
ZNeumann, line 104:
```suggestion
  if (NULL == events) {
```

ZNeumann (approved):
Looks great! And thanks Michael for such a thorough first pass, the conversations ya'll had helped a lot!

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+#include "nr_log_events.h"
+#include "nr_analytics_events_private.h"
+#include "util_memory.h"
+
+bool nr_log_events_add_event(nr_analytics_events_t* events,
+                             const nr_log_event_t* event,
+                             nr_random_t* rnd) {
+  bool events_sampled = false;
+  char* log_event_json = NULL;
+  nr_analytics_event_t* log_event;
```
zsistla, line 20: Set `log_event` to NULL to initialize?

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+    return false;
+  }
+  if (NULL == event) {
+    return false;
+  }
+
+  events_sampled = nr_analytics_events_is_sampling(events);
+
+  log_event_json = nr_log_event_to_json(event);
+  log_event = nr_analytics_event_create_from_string(log_event_json);
```
zsistla, line 32: Is it possible this returns a null or uninitialized pointer?  Should we check for null before adding the event?

axiom/nr_log_events.h
```diff
@@ -0,0 +1,48 @@
...
+ * Returns : true if sampling occured (see Notes below)
+ *           false otherwise
```
zsistla, line 18: Also false if the event was not added for some reason.

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+
+bool nr_log_events_add_event(nr_analytics_events_t* events,
+                             const nr_log_event_t* event,
+                             nr_random_t* rnd) {
+  bool events_sampled = false;
+  char* log_event_json = NULL;
+  nr_analytics_event_t* log_event;
+
+  if (NULL == events) {
+    return false;
```
zsistla, line 23: the return value seems a little overloaded.  it both indicates if sampling occurred and if an error adding the event occurred. Should we clarify that here: https://github.com/newrelic/newrelic-php-agent/pull/500/files#diff-e5f4823dae53cfaeb71612eec0542c18feef8bc10115a585516b1fe487bd2a35R18

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+  }
+  if (NULL == event) {
+    return false;
+  }
+
+  events_sampled = nr_analytics_events_is_sampling(events);
+
+  log_event_json = nr_log_event_to_json(event);
```
zsistla, line 31: Check this value for NULL?

axiom/nr_txn.c
```diff
@@ -3293,6 +3296,91 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+
+static void nr_txn_add_log_event(nrtxn_t* txn,
+                                 const char* log_level_name,
+                                 const char* log_message,
+                                 nrtime_t timestamp,
+                                 nrapp_t* app) {
+  nr_random_t* rnd = NULL;
+  nr_log_event_t* e = NULL;
+  bool event_dropped = false;
+
```
zsistla, line 3356: check txn, log_level_name, log_message, app for null before going on?
Even if it is currently called from a function that already checks those, we want to future proof it in case it is called from another function in the future.  it can be an nrunlikely wrap, but should be checked just in case.

axiom/nr_txn.c
zsistla: The logs will still be associated with the app/account.

axiom/nr_txn.c
```diff
@@ -3293,6 +3296,91 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+    nr_log_event_set_guid(e, nr_app_get_entity_guid(app));
+  }
+}
+
+static nr_log_event_t* log_event_create(const char* log_level_name,
+                                        const char* log_message,
+                                        nrtime_t timestamp,
+                                        nrtxn_t* txn,
+                                        nrapp_t* app) {
+  nr_log_event_t* e = nr_log_event_create();
```
zsistla, line 3335: Global comment: check txn, app for NULL (even if nrunlikely) before proceeding.

zsistla:
Awesome job on a complicated piece!

axiom/nr_log_events.c
misiekn: `nr_analytics_event_create_from_string` is basically an alias of `nr_strdup`. `nr_strdup` is `NULL` safe (handles `NULL` input gracefully) and will [exit](https://github.com/newrelic/newrelic-php-agent/blob/9fd43b56a296e761156db9aa08a1919a975cd56c/axiom/util_memory.c#L161) if it's not able to allocate memory rather than return `NULL`.

axiom/nr_log_events.c
misiekn: @zsistla These `NULL` checks are more of an asserts (check if the arguments are valid pointers) rather than business logic checks (i.e. has sampling occurred?). If an event was not added because invalid arguments were passed to this function (event != NULL is ensured by the caller [here](https://github.com/misiekn/newrelic-php-agent/blob/53d5617ed81910c5e007485fd821b60cc0356f04/axiom/nr_txn.c#L3369)), sampling could not have occurred and therefore `false` is returned. `true` is returned if and only if sampling occurs. It's impossible not to overload `false` return value with this design of the function (failing check for `NULL` arguments needs to return something) however the interesting value (`true`) is returned only if sampling occurs and this is the intention of this function: return `true` if sampling occurred, `false` otherwise (sampling did not occur or event was not added, which implies sampling must not have occurred).

axiom/nr_txn.c
misiekn: Empty log level will be treated as NULL. It's addressed in 3da93909c2cb3520fd2ef16c6340fd58c26a53f5

axiom/nr_txn.c
misiekn: Support for log event's pool of size 0 is added with 2f54982c76d39280b5c946062eb050dd25e12f8f, 486ce76a699939e65c40d50fdcc3d9affa6063ec, 4e17dc09758b5455fcf6a8e8dc37b48fea497886. 09c96740de6b1e674508510b55e85178215a4c58 removes dependency of log forwarding feature on the log event's pool size. This way when event will be added to a pool of size 0 it will be marked as **seen** and **dropped**.

Note: I decided not to 'short-circuit' `nr_txn_add_log_event` when log event pool has the size 0. This way the log event pool has the **seen** counter updated too.

axiom/nr_log_events.c
misiekn: Adressed in b339a7ace1c663c9fd0b3539ad6da613e87c7ad3

axiom/nr_log_events.h
misiekn: Addressed in f426ac1f41bf07ec0bd24851078ce944edafd680

axiom/nr_log_events.c
misiekn: Function's doc string updated in f426ac1f41bf07ec0bd24851078ce944edafd680

axiom/nr_txn.c
misiekn: `nrunlikely(NULL == txn)` checks added in 00da593846c2ab78c36c271da3b5a21225fac560. `log_event_set_linking_metadata` already checked if `NULL != app` and there's no need to abort if `app == NULL` earlier.

axiom/nr_txn.c
misiekn: nrunlikely(NULL == txn) checks added in 00da593846c2ab78c36c271da3b5a21225fac560.

axiom/nr_analytics_events.c
mfulb: I would imagine FALSE as it isn't sampling?

mfulb (approved):
Thank you Michal for this important addition and speedy feedback on my comments.

axiom/nr_analytics_events.c
misiekn: Addressed in a52cca63b84671b84c650ecda18481d38f4defa8

misiekn:
ok jenkins

zsistla (approved):
Looks good, thanks!

axiom/nr_analytics_events.c
```diff
@@ -117,11 +124,16 @@ int nr_analytics_events_number_saved(const nr_analytics_events_t* events) {
...
-
   if (max_events <= 0) {
-    return 0;
+    return NULL;
   }
+
+  return nr_analytics_events_create_ex(max_events);
+}
+
+nr_analytics_events_t* nr_analytics_events_create_ex(int max_events) {
```
bduranleau-nr, line 134: 1. Is there a reason not to make this function static and only accessible via nr_analytics_events_create? Is there a case where we want to access this directly without checking for NULL?
2. I would suggest a more descriptive name, like "_init" rather than "_ex", just for clarity-at-a-glance sake

axiom/nr_log_events.c
```diff
@@ -0,0 +1,39 @@
...
+  log_event = nr_analytics_event_create_from_string(log_event_json);
+  nr_free(log_event_json);
+
+  nr_analytics_events_add_event(events, log_event, rnd);
+  nr_free(log_event);
+
+  return events_sampled;
+}
```
bduranleau-nr, line 39: newline
```suggestion
}

```

misiekn:
ok jenkins

axiom/nr_log_events.h
```diff
@@ -0,0 +1,50 @@
...
+  return nr_analytics_events_number_seen(events);
+}
+NRINLINE int nr_log_events_number_saved(const nr_analytics_events_t* events) {
+  return nr_analytics_events_number_saved(events);
+}
+NRINLINE const char* nr_log_events_get_event_json(nr_analytics_events_t* events,
+                                                  int i) {
+  return nr_analytics_events_get_event_json(events, i);
+}
+#endif
```
bduranleau-nr, line 50: newline
```suggestion
#endif

```

axiom/nr_txn.c
```diff
@@ -3293,28 +3292,123 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
   return false;
 }
 
+#define ENSURE_LOG_LEVEL_NAME(level_name) \
+  (nr_strempty(level_name) ? "UNKNOWN" : level_name)
```
bduranleau-nr, line 3296: Is it possible to get a non-NULL non-empty level_name here that does not correspond to a valid log level?

axiom/nr_txn.c
```diff
@@ -531,6 +532,7 @@ nrtxn_t* nr_txn_begin(nrapp_t* app,
   nt->intrinsics = nro_new_hash();
 
   nt->custom_events = nr_analytics_events_create(app->limits.custom_events);
+  nt->log_events = nr_log_events_create(1000 /* app->limits.log_events */);
```
bduranleau-nr, line 535: Is this just a placeholder for a dependency on another PR?

axiom/nr_txn.c
misiekn: It should not be possible under normal circumstances. But whatever the agent gets from instrumented method, it honors.

axiom/nr_txn.c
misiekn: Correct. And actually since #497 went in, I might as well address it.

axiom/nr_analytics_events.c
misiekn: 1. The `_ex` method is called by `nr_log_events_create`. That's why it was added in the first place - to allow log_events pool have the size of 0.
2. `_ex` is a standard suffix added when an interface of existing function is **__ex__**tendted.

axiom/nr_txn.c
```diff
@@ -3293,28 +3292,123 @@ bool nr_txn_log_decorating_enabled(nrtxn_t* txn) {
...
+                             nrtime_t timestamp,
+                             nrapp_t* app) {
+  if (nrunlikely(NULL == txn)) {
+    return;
+  }
+
+  nr_txn_add_log_event(txn, log_level_name, log_message, timestamp, app);
 
   nr_txn_add_logging_metrics(txn, log_level_name);
 }
```
bduranleau-nr, line 3414: newline
```suggestion
}

```

axiom/tests/test_log_events.c
```diff
@@ -0,0 +1,182 @@
...
+#include <stddef.h>
+
+#include "nr_log_events.h"
+#include "util_random.h"
+#include "util_strings.h"
+
+#include "tlib_main.h"
+
+#define LOG_LEVEL "INFO"
+#define LOG_MESSAGE "this is a test log error message"
```
bduranleau-nr, line 17: I'm curious what happens if there are things like formatting characters in a log message? Does it get sanitized out somewhere down the line? 
For example, if someone has a log message with a `'\n'` character, or `'\0'`, or `'%s'`. 
Maybe here would be a good place to test those kinds of messages?

axiom/nr_analytics_events.c
bduranleau-nr: I missed the NRINLINE implementation by only searching .c files. Oops!

If we are going to allow the bypass of nr_analytics_events_create, do we need a check here for max_events < 0, or are we confident that will never happen?

axiom/tests/test_log_events.c
```diff
@@ -0,0 +1,182 @@
...
+  tlib_pass_if_null("events destroyed", events);
+}
+
+tlib_parallel_info_t parallel_info = {.suggested_nthreads = 2, .state_size = 0};
+
+void test_main(void* p NRUNUSED) {
+  test_events_success();
+  test_events_sample();
+  test_events_null();
+}
```
bduranleau-nr, line 182: newline
```suggestion
}

```

axiom/tests/test_txn.c
```diff
@@ -8284,4 +8474,4 @@ void test_main(void* p NRUNUSED) {
   test_segment_record_error();
   test_record_log_event();
   test_txn_log_configuration();
-}
+}
```
bduranleau-nr, line 8477: newline
```suggestion
}

```

axiom/tests/test_log_events.c
misiekn: @bduranleau-nr It's a good idea. However I think a better place for this type of test would be either log_event `set_message` and `get_event_json` API, which log_events API rely on, or integration test which exercise all agent's interfaces that could have been impacted by special characters.

axiom/nr_analytics_events.c
misiekn: Good catch! I will add a check that `max_events` is not negative to `nr_log_events_create_ex`.

misiekn: @bduranleau-nr Addressed in 4e74eec94f98f5e1ed59c3f60eb92751c5411608.

bduranleau-nr (approved):
Looks great!

misiekn:
ok jenkins

misiekn:
ok jenkins

misiekn:
ok jenkins

misiekn:
ok jenkins

misiekn:
ok jenkins

PR #503, opened 2022-08-05
WIP: feat(agent,daemon): Prototype agent with log forwarding [DO NOT MERGE]

mfulb:
I combined the open PRs with some additional code needed to make harvest work correctly in the daemon.
I was able to send log events which were viewable on the NR1 UI!


mfulb:
A final version of this work was released in agent 10.1.0.313 and this branch is only useful for historical reference.

PR #505, opened 2022-08-15
feat(agent): Add/Update functions to utilize the OAPI zend_execute_data.

zsistla:
1) Added 4 new functions for use with OAPI instrumentation:
extern const char* nr_php_zend_execute_data_function_name(
    const zend_execute_data* execute_data);

extern const char* nr_php_zend_execute_data_filename(
    const zend_execute_data* execute_data);

extern const char* nr_php_zend_execute_data_scope_name(
    const zend_execute_data* execute_data);

extern uint32_t nr_php_zend_execute_data_lineno(
    const zend_execute_data* execute_data);
2) Added test cases to test new functionality.
3) Updated the following to reflect new OAPI paradigm and add additional checks for robustness:
* #define NR_PHP_USER_FN_THIS()
* nr_php_execute_scope
3) Added unit test cases to test the 4 new functions.  Currently existing functions are being called and used in current tests and will automatically be tested with php8+.
4) Also verified functionality via integration tests.

### DIFF
--- a/agent/php_agent.c
+++ b/agent/php_agent.c
@@ -1142,7 +1142,7 @@ bool nr_php_function_is_static_method(const zend_function* func) {
   return (func->common.fn_flags & ZEND_ACC_STATIC);
 }
 
-#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP8+ */
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP7+ */
 extern const char* nr_php_zend_execute_data_function_name(
     const zend_execute_data* execute_data) {
   zend_string* function_name = NULL;
@@ -1204,4 +1204,4 @@ extern uint32_t nr_php_zend_execute_data_lineno(
   return 0;
 }
 
-#endif /* PHP 8+ */
+#endif /* PHP 7+ */
--- a/agent/php_agent.c
+++ b/agent/php_agent.c
@@ -1141,3 +1141,67 @@ bool nr_php_function_is_static_method(const zend_function* func) {
 
   return (func->common.fn_flags & ZEND_ACC_STATIC);
 }
+
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP8+ */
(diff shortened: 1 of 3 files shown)


zsistla:
ok jenkins

zsistla:
ok jenkins

mfulb (approved):

ZNeumann (approved):

bduranleau-nr (approved):
👍

misiekn:
Why `nr_php_zend_execute_data_filename`, `nr_php_zend_execute_data_scope_name` and `nr_php_zend_execute_data_lineno` look through `execute_data->prev`?

ZNeumann (approved):

zsistla:
If the zend_func is NULL OR if the zend_execute_data is NOT a user function, we can use `prev` to backtrack the call stack until we find what we need in order to return a good value for those functions.

zsistla:
Make nr_php_zend_execute_data_scope_name a wrapper to call zend_get_called_scope  (https://github.com/php/php-src/blob/master/Zend/zend_execute_API.c#L1209) then return the string.

bduranleau-nr:
@misiekn @zsistla 
Regarding wrapping the php function`zend_get_called_scope` inside of `nr_php_zend_execute_data_scope_name`, the two are not actually compatible.
Looking at our custom implementation in this PR, this block:
```c
    if (execute_data->func
        && (ZEND_USER_CODE(execute_data->func->type)
            || execute_data->func->common.scope)) {
      ce = execute_data->func->common.scope;
      execute_data = NULL;
    }
```
Will return the `zend_class_entry`
Whereas If we look at the php [implementation](https://github.com/php/php-src/blob/master/Zend/zend_execute_API.c#L1209):
```c
else if (ex->func) {
	if (ex->func->type != ZEND_INTERNAL_FUNCTION || ex->func->common.scope) {
		return NULL;
	}
}
```
We can see that the same logic is being applied as our implementation, but instead will result in returning `NULL`, rather than the `zend_class_entry` that we want.

misiekn (approved):
Nice job figuring all this out! 👍

PR #506, opened 2022-08-16
feat(agent): Add code level metrics functionality.

zsistla:
1) Add INI value to disable/enable code level metrics.
2) Create new function `nr_php_execute_metadata_add_code_level_metrics` that uses nr_php_zend_execute_data_* family of functions to extract CLM and add it to the metadata struct.  The metadata is especially important for PHP 7x functionality since the zend_execute_data cannot be depended on to be persistent.
3) Create new function `nr_php_txn_add_code_level_metrics` that uses metadata to add CLM as agent attributes.
3.5) Added new NR_TXN_ATTRIBUTES for `code.namespace`, `code.lineno`, `code.filepath`, and `code.function`. 
4) Added call to php_execute_enabled to call the new CLM function
5) Added new integration tests to exercise the new functionality.
6) Do NOT automatically initialize attributes for segments; otherwise we end up recording super short segments that we should ignore. In the case of a super short segment that would have been ignored, we do NOT add CLM (otherwise it wouldn't be ignored).  We wait until after we decide to ignore or not to add the CLM data.
7) Ensure CLM are propagated properly to spans and transaction traces.
Note:  CLM functionality is only compatible with PHP 7+.
8) updated `nr_php_execute_metadata_t` to hold additional information and moved the definition of the struct to php_execute.h.
9) `nr_php_execute_enabled` uses the metadata at the beginning so now we can simply release at the end.

Later additions include:
1) discard any CLM attributes that will be truncated.
2) NULL check hardening
3) helper macros


agent/php_nrini.c
```diff
@@ -3045,10 +3057,11 @@ void zm_info_newrelic(void); /* ctags landing pad only */
 PHP_MINFO_FUNCTION(newrelic) {
   php_info_print_table_start();
   php_info_print_table_header(2, "New Relic RPM Monitoring",
-                              NR_PHP_PROCESS_GLOBALS(enabled) ? "enabled"
-                              : NR_PHP_PROCESS_GLOBALS(mpm_bad)
-                                  ? "disabled due to threaded MPM"
-                                  : "disabled");
+                              NR_PHP_PROCESS_GLOBALS(enabled)
```
mfulb, line 3060: These lines need to be clang formatted I believe.

ZNeumann (approved):

zsistla:
ok jenkins

agent/php_nrini.c
mfulb: After discussing the formatting I believe it is fine.

mfulb (approved):

agent/php_agent.c
```diff
@@ -1141,3 +1141,72 @@ bool nr_php_function_is_static_method(const zend_function* func) {
...
+  }
+  return function_name ? ZSTR_VAL(function_name) : NULL;
+}
+
+extern const char* nr_php_zend_execute_data_filename(
+    const zend_execute_data* execute_data) {
+  zend_string* filename = NULL;
+  while (
+      execute_data
+      && (!execute_data->func || !ZEND_USER_CODE(execute_data->func->type))) {
```
bduranleau-nr, line 1173: Extracting this statement
```c
     execute_data
      && (!execute_data->func || !ZEND_USER_CODE(execute_data->func->type))
```
out to it's own helper function could improve readability here, as well as in `nr_php_zend_execute_data_lineno`, where it seems to be repeated

agent/php_agent.c
```diff
@@ -1141,3 +1141,72 @@ bool nr_php_function_is_static_method(const zend_function* func) {
...
+    execute_data = execute_data->prev_execute_data;
+  }
+  if (execute_data) {
+    return execute_data->opline ? execute_data->opline->lineno : 0;
+  }
+  return 0;
+}
+
+#endif /* PHP 7+ */
```
bduranleau-nr, line 1212: Missing newline character at EOF
```suggestion
#endif /* PHP 7+ */

```

agent/php_execute.c
```diff
@@ -71,9 +71,9 @@
  * constructions and teardown as reasons to avoid excessive function calls.
  * This too is erroneous. The cost of calling a function is about 4 assembler
  * instructions. This is negligible. Therefore, as a means of reducing stack
- * usage, if you need stack space it is better to put that usage into a static
- * function and call it from the main function, because then that stack space
- * in genuinely only allocated when needed.
+ * usage, if you need stack space it is better to put that usage into a static*
+ * function and call it from the main function, because then that stack space in
```
bduranleau-nr, line 75:
```suggestion
 * usage, if you need stack space it is better to put that usage into a static
 * function and call it from the main function, because then that stack space is
```

agent/php_execute.c
```diff
@@ -919,40 +919,97 @@ static void nr_php_execute_file(const zend_op_array* op_array,
...
-typedef struct {
-#ifdef PHP7
-  zend_string* scope;
-  zend_string* function;
+static void nr_php_execute_metadata_add_code_level_metrics(
+    nr_php_execute_metadata_t* metadata,
+    NR_EXECUTE_PROTO) {
+#if ZEND_MODULE_API_NO < ZEND_7_0_X_API_NO /* PHP7+ */
+  (void)metadata;
+  NR_UNUSED_SPECIALFN;
```
bduranleau-nr, line 937: Do we need these lines if we're just going to call return without doing anything to them?

zsistla: Yes, it's because we aren't doing anything with metadata and execute_data when PHP is less than 7 that we need to void both of them explicitly for metadata and with the [macro for execute_data ](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_newrelic.h#L61)since it can change depending on PHP version; otherwise, the compiler would complain that the vars were not used.

bduranleau-nr: That makes sense, thanks for clarifying 👍

agent/php_txn.c
```diff
@@ -1072,3 +1072,47 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
...
+                                           metadata->function_filepath);
+  }
+  if (!nr_strempty(metadata->function_namespace)) {
+    nr_txn_attributes_set_string_attribute(
+        attributes, nr_txn_clm_code_namespace, metadata->function_namespace);
+  }
+  nr_txn_attributes_set_long_attribute(attributes, nr_txn_clm_code_lineno,
+                                       metadata->function_lineno);
+}
+#endif
```
bduranleau-nr, line 1118:
```suggestion
#endif

```

bduranleau-nr (approved):
👍

zsistla:
https://github.com/newrelic/newrelic-php-agent/pull/506/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R996-R999
https://github.com/newrelic/newrelic-php-agent/pull/506/files#diff-f5f8c5da8db458c336e8505b331bcb42a571175d89cdd0e65fbc39c1121a5960R1098-R1111
nr_strempty supposedly checks for NULL, but was still getting some segfaults when calling it straight up.  Safer to do something like:
if (NULL == function || nr_strempty(function))
or
  if (NULL != metadata->function_filepath && !nr_strempty(metadata->function_filepath))

agent/php_execute.c
```diff
@@ -919,40 +919,97 @@ static void nr_php_execute_file(const zend_op_array* op_array,
...
+   *
+   * If we don't have the minimum requirements, exit and don't add any
+   * attributes.
+   */
+
+  filepath = nr_php_zend_execute_data_filename(execute_data);
+  namespace = nr_php_zend_execute_data_scope_name(execute_data);
+  function = nr_php_zend_execute_data_function_name(execute_data);
+  lineno = nr_php_zend_execute_data_lineno(execute_data);
+
```
zsistla, line 979: check all the string lengths with NRSAFELEN.  If any exceed 255 exit without adding here.

bduranleau-nr: implemented in 125fa65bc36a89858fcadcff05d5aacf447ee533

bduranleau-nr:
ok jenkins

PR #507, opened 2022-08-19
feat(daemon): add data usage metrics

ZNeumann:
Adds the collection of data usage metrics into a separate channel in the processor. The JSON payload is now extracted from the http client (by passing the command structure by pointer), so we are able to get its size.

There is a maximum amount of data usage metrics that will be stored in between metrics harvests. If many other harvests happen in between a metric harvest, or if the metrics harvest stalls for any number of reasons, usage data will be dropped.

Clarifies, reformats, and updates comments lots of the tests and infrastructure for processor.go/processor_test.go

### DIFF
--- a/src/newrelic/collector/client.go
+++ b/src/newrelic/collector/client.go
@@ -184,12 +184,12 @@ func (control *RpmControls) userAgent() string {
 }
 
 type Client interface {
-	Execute(cmd RpmCmd, cs RpmControls) RPMResponse
+	Execute(cmd *RpmCmd, cs RpmControls) RPMResponse
 }
 
-type ClientFn func(cmd RpmCmd, cs RpmControls) RPMResponse
+type ClientFn func(cmd *RpmCmd, cs RpmControls) RPMResponse
 
-func (fn ClientFn) Execute(cmd RpmCmd, cs RpmControls) RPMResponse {
+func (fn ClientFn) Execute(cmd *RpmCmd, cs RpmControls) RPMResponse {
 	return fn(cmd, cs)
 }
 
@@ -199,7 +199,7 @@ type limitClient struct {
 	semaphore chan bool
 }
 
-func (l *limitClient) Execute(cmd RpmCmd, cs RpmControls) RPMResponse {
+func (l *limitClient) Execute(cmd *RpmCmd, cs RpmControls) RPMResponse {
 	var timer <-chan time.Time
 
 	if 0 != l.timeout {
@@ -398,7 +398,8 @@ func parseResponse(b []byte) ([]byte, error) {
 	return r.ReturnValue, nil
 }
 
-func (c *clientImpl) Execute(cmd RpmCmd, cs RpmControls) RPMResponse {
+func (c *clientImpl) Execute(cmd *RpmCmd, cs RpmControls) RPMResponse {
+	// Create the JSON payload
 	data, err := cs.Collectible.CollectorJSON(false)
 	if nil != err {
 		return RPMResponse{Err: err}
@@ -406,15 +407,14 @@ func (c *clientImpl) Execute(cmd RpmCmd, cs RpmControls) RPMResponse {
 	cmd.Data = data
 
 	var audit []byte
-
 	if log.Auditing() {
 		audit, err = cs.Collectible.CollectorJSON(true)
 		if nil != err {
 			log.Errorf("unable to create audit json payload for '%s': %s", cmd.Name, err)
-			audit = data
+			audit = cmd.Data
 		}
 		if nil == audit {
-			audit = data
+			audit = cmd.Data
 		}
 	}
 
@@ -422,12 +422,12 @@ func (c *clientImpl) Execute(cmd RpmCmd, cs RpmControls) RPMResponse {
 	cleanURL := cmd.url(true)
 
 	log.Audit("command='%s' url='%s' payload={%s}", cmd.Name, url, audit)
-	log.Debugf("command='%s' url='%s' max_payload_size_in_bytes='%d' payload={%s}", cmd.Name, cleanURL, cmd.MaxPayloadSize, data)
+	log.Debugf("command='%s' url='%s' max_payload_size_in_bytes='%d' payload={%s}", cmd.Name, cleanURL, cmd.MaxPayloadSize, cmd.Data)
 
-	resp := c.perform(url, cmd, cs)
-	if err != nil {
+	resp := c.perform(url, *cmd, cs)
+	if nil != resp.Err {
 		log.Debugf("attempt to perform %s failed: %q, url=%s",
-			cmd.Name, err.Error(), cleanURL)
+			cmd.Name, resp.Err.Error(), cleanURL)
 	}
 
 	log.Audit("command='%s' url='%s', status=%d, response={%s}", cmd.Name, url, resp.StatusCode, string(resp.Body))
--- a/src/newrelic/infinite_tracing/trace_observer.go
+++ b/src/newrelic/infinite_tracing/trace_observer.go
@@ -23,6 +23,7 @@ const (
 	supportabilitySent        = "Supportability/InfiniteTracing/Span/Sent"
 	supportabilityResponseErr = "Supportability/InfiniteTracing/Span/Response/Error"
 	supportabilityQueueDumped = "Supportability/InfiniteTracing/Span/AgentQueueDumped"
+	supportabilityDataUsage   = "Supportability/PHP/InfiniteTracing/Output/Bytes"
 )
 
 type spanBatch struct {
@@ -112,7 +113,8 @@ type metricIncrement struct {
 type traceObserverSupportability struct {
 	incrementSent chan float64
 	increment     chan metricIncrement
-	dump          chan map[string]float64
+	dataUsage     chan float64
+	dump          chan map[string][6]float64
 }
 
 func newTraceObserverWithWorker(cfg *Config) (*TraceObserver, func()) {
@@ -303,12 +305,15 @@ func (to *TraceObserver) doStreaming() spanBatchSenderStatus {
 				msg.count, to.messagesRemainingCapacity, to.QueueSize)
 			if err, status := to.sender.send(encodedSpanBatch(msg.batch)); err != nil {
 				to.messagesSent <- msg.count
+				// Add 0 to dataUsage channel so that we successfully count the send attempt
+				to.supportability.dataUsage <- 0.0
 				to.supportabilityError(status)
 				log.Errorf("trace observer error while sending span batch:  %v", err)
 				return status
 			} else {
 				to.messagesSent <- msg.count
 				to.supportability.incrementSent <- float64(msg.count)
+				to.supportability.dataUsage <- float64(len(msg.batch))
 			}
 		case status := <-to.responseError:
 			to.supportabilityError(status)
@@ -358,10 +363,29 @@ func (to *TraceObserver) handleSupportability() {
 			// alive when a shutdown was triggered from the sender.
 			return
 		case inc := <-to.supportability.increment:
-			metrics[inc.name] += inc.count
+			v, ok := metrics[inc.name]
+			if !ok {
+				v = [6]float64{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}
+			}
+			v[0] += inc.count
+			metrics[inc.name] = v
 		case batchCount := <-to.supportability.incrementSent:
+			v, ok := metrics[supportabilitySent]
+			if !ok {
+				v = [6]float64{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}
+			}
 			// Since we're sending span batches, we increment by the number of spans in the batch.
-			metrics[supportabilitySent] += batchCount
+			v[0] += batchCount
+			metrics[supportabilitySent] = v
+		case dataUsage := <-to.supportability.dataUsage:
+			v, ok := metrics[supportabilityDataUsage]
+			if !ok {
+				v = [6]float64{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}
+			}
+			v[0] += 1
+			v[1] += dataUsage
+			// There is never a body returned so no need to increment v[2]
+			metrics[supportabilityDataUsage] = v
 		case to.supportability.dump <- metrics:
 			// reset the metrics map
 			metrics = newSupportabilityMetrics()
@@ -369,11 +393,13 @@ func (to *TraceObserver) handleSupportability() {
 	}
 }
 
-func newSupportabilityMetrics() map[string]float64 {
-	// grpc codes, plus 1 for sent, and 1 for response errs
-	metrics := make(map[string]float64, numCodes+2)
+func newSupportabilityMetrics() map[string][6]float64 {
+	// grpc codes, plus 1 for sent, 1 for data usage,  and 1 for response errs
+	// Sending full metrics for AddRaw() (6 floats instead of just for AddCount())
+	// because dataUsage metrics requires multiple fields
+	metrics := make(map[string][6]float64, numCodes+3)
 	// supportabilitySent metric must always be sent
-	metrics[supportabilitySent] = 0
+	metrics[supportabilitySent] = [6]float64{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}
 	return metrics
 }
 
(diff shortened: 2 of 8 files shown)


src/newrelic/collector/client.go
```diff
@@ -316,8 +335,78 @@ func parseProxy(proxy string) (*url.URL, error) {
...
+func (c *clientImpl) WaitOnDataUsage(id string) {
+	v, ok := c.harvestDataUsageMaps.Load(id)
+	if ok {
+		t := v.(dataUsageMap)
+		t.wg.Wait()
+	}
+}
+
+func (c *clientImpl) CollectDataUsage(id string) map[string]DataUsage {
+	v, ok := c.harvestDataUsageMaps.LoadAndDelete(id)
```
ZNeumann, line 378: LoadAndDelete() was introduced in go 1.15 which is beyond our current minimum go version

ZNeumann:
Once the PHP team is happy with this PR, I will add the Golang team as reviewers for a final pass

src/newrelic/processor.go
```diff
@@ -524,7 +526,40 @@ type harvestArgs struct {
...
+
+func addDataUsage(duc chan dataUsageInfo, endpoint string, data_stored int, data_received int) {
+	select {
+		case duc <- dataUsageInfo{
+			         endpoint_name: endpoint,
+			         payloadSize: data_stored,
+			         responseSize: data_received,
+			        }:
+			// data stored
+		default:
```
mfulb, line 556: I think this will make the select non-blocking and could cause performance problems - see[this](https://github.com/golang/go/issues/1590) for example.

src/newrelic/processor.go
```diff
@@ -544,14 +579,17 @@ func harvestPayload(p PayloadCreator, args *harvestArgs) {
...
 
 	// We don't need to process the response to a harvest command unless an
 	// error happened.  (Note that this may change if we have to support metric
 	// cache ids).
 	if nil == reply.Err {
+		addDataUsage(duc.duc, cmd.Name, len(cmd.Data), len(reply.Body))
 		return
 	}
+	// If we receive an error, the data was not stored into the collector
+	addDataUsage(duc.duc, cmd.Name, 0, len(reply.Body))
```
mfulb, line 592: Does the usage metric care about accepted transfers out to collector or just the amount of data being sent w/o regard for if it was accepted?

src/newrelic/processor.go
```diff
@@ -560,48 +598,53 @@ func harvestPayload(p PayloadCreator, args *harvestArgs) {
...
 	}
 }
 
-func considerHarvestPayloadTxnEvents(txnEvents *TxnEvents, args *harvestArgs) {
+func considerHarvestPayloadTxnEvents(txnEvents *TxnEvents, args *harvestArgs, duc dataUsageController) {
 	if args.splitLargePayloads && (txnEvents.events.Len() >= (limits.MaxTxnEvents / 2)) {
 		events1, events2 := txnEvents.Split()
-		considerHarvestPayload(&TxnEvents{events1}, args)
-		considerHarvestPayload(&TxnEvents{events2}, args)
+		considerHarvestPayload(&TxnEvents{events1}, args, duc)
```
mfulb, line 618: Have you been able to test if this works if a payload is split?  Maybe test by setting the max payload size small enough to trigger it with a normal payload.

src/newrelic/processor.go
ZNeumann: That example hangs because the `select` is in a `for` loop, causing it to endlessly hit the `default` case. Since this is not in a loop, the non-blocking empty default will simply drop the data and move on.

src/newrelic/processor.go
ZNeumann: After asking around, the answer I received is the former. It makes more sense for it to care about billable data, which is only that which is accepted.

src/newrelic/processor.go
mfulb: I understand  - this is a non-blocking write operation in Go.

src/newrelic/processor.go
```diff
@@ -663,42 +707,93 @@ func harvestByType(ah *AppHarvest, args *harvestArgs, ht HarvestType) {
...
+		if args.blocking {
+			harvestDataUsage(args, duc)
+		} else {
+			go harvestDataUsage(args, duc)
+		}
+	}
+}
+
+func harvestDataUsage(args *harvestArgs, duc dataUsageController) {
+	duc.wg.Wait()
```
mfulb, line 756: Could this ever block forever?  Is there a wait to have a timeout?

src/newrelic/processor.go
```diff
@@ -663,42 +707,93 @@ func harvestByType(ah *AppHarvest, args *harvestArgs, ht HarvestType) {
...
 
+	metrics := NewMetricTable(limits.MaxMetrics, time.Now())
+	for name, data := range(dataUsageMap) {
+		metrics.AddRaw([]byte("Supportability/" + args.agentLanguage + "/" + args.collector + "/" + name + "/Output/Bytes"),
+					   "", "", [6]float64{float64(data.attempts), float64(data.payloadSize), float64(data.responseSize), 0.0, 0.0, 0.0}, Forced)
 	}
+	metrics.AddRaw([]byte("Supportability/" + args.agentLanguage + "/" + args.collector + "/Output/Bytes"),
+				   "", "", [6]float64{float64(sumAttempts), float64(sumPayload), float64(sumResponse), 0.0, 0.0, 0.0}, Forced)
+	metrics = metrics.ApplyRules(args.rules)
+	considerHarvestPayload(metrics, args, duc)
```
mfulb, line 796: What keeps this from creating another cycle ending with sending data usage metrics again into an infinite recurse?

ZNeumann: `considerHarvestPayload` only calls the `Exectue()` to send the data and adds the resulting usage to the data usage channel. It does not trigger another harvest nor call `harvestDataUsage()`. The next harvest cycle that includes metrics will then call `harvestDataUsage()` again, pulling any usage since the previous call to be included in this one.

src/newrelic/processor.go
ZNeumann: This theoretically could potentially block forever, resulting in a hung child thread. This could **only** happen in the instance that there is already another hung child thread, otherwise the wait group would be empty. And because the other threads that are blocking the wait group are just http calls that do have a timeout, in practice, this should never be hung.

src/newrelic/processor.go
ZNeumann: Added a test for this in [test(daemon): add usage metrics test for when data usage channel is f…](https://github.com/newrelic/newrelic-php-agent/pull/507/commits/cb7fdc88902f5387b30b2386f8dcc0241d70acf8)

src/newrelic/processor.go
ZNeumann: Added a test for this in [test(daemon): add usage metrics test for multiple same-endpoint metrics](https://github.com/newrelic/newrelic-php-agent/pull/507/commits/68408bd2fa7a2ecd4df2fd02bc68b9f95f385ac2)

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

misiekn (approved):
Really neatly implemented feature! 👍

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

src/newrelic/processor_test.go
```diff
@@ -295,16 +333,134 @@ func TestProcessorHarvestCleanExit(t *testing.T) {
...
 
 	expected := `["one",{"reservoir_size":5,"events_seen":1},[half birthday]]`
 	if string(cp.data) != expected {
 		t.Fatalf("expected: %s \ngot: %s", expected, string(cp.data))
 	}
+
+	time := strings.Split(string(cp2.data), ",")[1]
+	usageMetrics := `["one",` + time + `,` + time + `,` +
+		`[[{"name":"Supportability/c/specific_collector.com/Output/Bytes"},[2,1315,0,0,0,0]],` +
+		`[{"name":"Supportability/c/specific_collector.com/custom_event_data/Output/Bytes"},[1,60,0,0,0,0]],` +
```
zsistla, line 352: I thought the spec didn't indicate it needed the specific collector.  Thought this was mainly `Collector`, `OTLP`, or `InfiniteTracing`

Here's example output:
```
[
  {
    “name”:”Supportability/DotNET/Collector/Output/Bytes”
  },
  [10, 1000, 550, 0, 0, 0]
]
```

zsistla:
Thanks for getting this up!  It's definitely the most technical `go` PR from our team.

src/newrelic/processor_test.go
```diff
@@ -295,16 +333,134 @@ func TestProcessorHarvestCleanExit(t *testing.T) {
...
+	cp2 := <-m.clientParams /* usage metrics */
 
 	expected := `["one",{"reservoir_size":5,"events_seen":1},[half birthday]]`
 	if string(cp.data) != expected {
 		t.Fatalf("expected: %s \ngot: %s", expected, string(cp.data))
 	}
+
+	time := strings.Split(string(cp2.data), ",")[1]
+	usageMetrics := `["one",` + time + `,` + time + `,` +
+		`[[{"name":"Supportability/c/specific_collector.com/Output/Bytes"},[2,1315,0,0,0,0]],` +
```
zsistla, line 351: Can you please add an InfiniteTracing destination test case?

src/newrelic/processor_test.go
ZNeumann: Nice catch! I've addressed the collector name and added 8T usage metrics here [feat(daemon): add 8T data usage metrics](https://github.com/newrelic/newrelic-php-agent/pull/507/commits/69b03e1abc731d339dd809952cd1ee149bc27f05)

src/newrelic/processor_test.go
ZNeumann: Added 8T data usage and tests here [feat(daemon): add 8T data usage metrics](https://github.com/newrelic/newrelic-php-agent/pull/507/commits/69b03e1abc731d339dd809952cd1ee149bc27f05)

ZNeumann:
ok jenkins

src/newrelic/infinite_tracing/trace_observer.go
```diff
@@ -358,36 +363,58 @@ func (to *TraceObserver) handleSupportability() {
 			// alive when a shutdown was triggered from the sender.
 			return
 		case inc := <-to.supportability.increment:
-			metrics[inc.name] += inc.count
+			v, ok := metrics[inc.name]
+			if !ok {
+				v = [6]float64{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}
```
zsistla, line 368: Technically, v[0]/Count is a long, but looks like this is how we've been doing it all along?

It'd be useful to either have helper functions to update the the supportability usage, or since elements are being modified directly, a comment clarifying all the parts of v ( like interaction count, bytes sent, bytes received, unused, unused, unused).

src/newrelic/processor_test.go
zsistla: Nice addition and impressive turnaround time :)

src/newrelic/processor_test.go
zsistla: Thanks again for adding this.

src/newrelic/infinite_tracing/trace_observer.go
ZNeumann: Yes, all of our metrics are stored as float64's.

ZNeumann: Addressed the helper and comment here [chore(daemon): add clarifying comment and refactor code duplication](https://github.com/newrelic/newrelic-php-agent/pull/507/commits/03576673bebf12c2a2835c985936a9a5fe5ecd76)

mfulb (approved):
Thank you Zach for the persistence and getting this all pulled together.

src/newrelic/processor.go
```diff
@@ -777,6 +872,9 @@ func NewProcessor(cfg ProcessorConfig) *Processor {
 		harvestErrorChannel:   make(chan HarvestError),
 		quitChan:              make(chan struct{}),
 		processorHarvestChan:  make(chan ProcessorHarvest),
+		// We don't want data usage collection to ever block, so we create a
+		// sized buffer that will drop any excess data once filled
+		dataUsageChannel:      make(chan dataUsageInfo, 25),
```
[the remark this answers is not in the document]
zsistla: Nice idea.

zsistla (approved):

src/newrelic/processor.go
```diff
@@ -777,8 +874,11 @@ func NewProcessor(cfg ProcessorConfig) *Processor {
 		harvestErrorChannel:   make(chan HarvestError),
 		quitChan:              make(chan struct{}),
 		processorHarvestChan:  make(chan ProcessorHarvest),
-		appConnectBackoff:     limits.AppConnectAttemptBackoff,
-		cfg:                   cfg,
+		// We don't want data usage collection to ever block, so we create a
+		// sized buffer that will drop any excess data once filled
+		dataUsageChannel:  make(chan dataUsageInfo, 25),
```
iamemilio, line 879: out of curiosity, why are you buffering 25 dataUsageInfo objects in the chanel?

ZNeumann: I just chose 25 for little reason. I wanted it to be big enough that it could miss a few harvests and still be fine, but small enough to not have any sizeable footprint. It would probably be good to put it into a settings file.

iamemilio: Yeah, I don't think it's actively harmful, unless the memory footprint matters that much. How many potential writers are there to the channel? There is only 1 reader, right? Does the reader drain the channel on read, or just read one object?

ZNeumann: https://github.com/newrelic/newrelic-php-agent/pull/507/files#diff-d34a767f316aaeded794ac3a231f867d3ffb42de19365ed338eace17f5876d94R755 drains the channel.

There is only one reader, though it is technically possible for multiple if the thread is being starved and another harvest begins. A harvestAll directive creates 9-ish writers.

iamemilio: Gotcha, makes sense to me. Thanks for answering my questions, it's so hard to cold read this kind of code. LGTM

iamemilio (approved):
Looks good

src/newrelic/infinite_tracing/trace_observer.go
```diff
@@ -112,7 +113,8 @@ type metricIncrement struct {
 type traceObserverSupportability struct {
 	incrementSent chan float64
 	increment     chan metricIncrement
-	dump          chan map[string]float64
+	dataUsage     chan float64
+	dump          chan map[string][6]float64
```
bduranleau-nr, line 117: Is there a special reason we want to use `6` here?

ZNeumann: Metrics are stored in the agent (and NR) as a [6]float. Previously, we were just creating "count" metrics in the trace_observer, and the interface we have in the daemon to handle that just took the single count (but then stored it in a [6]float). Now that data usage metrics are added, there are non-count metrics being created, which need multiple slots in the [6]float.

The options to address this are 1) create a new dump chan that is [6]float that only takes the data usage metrics and therefore use 2 channels for metrics on top of some other overhead or 2) (what I did) increase the width of the channel to account for data usage metrics coming through, even though most of the metrics will only be using 1 slot of the 6

src/newrelic/infinite_tracing/trace_observer.go
```diff
@@ -358,36 +363,66 @@ func (to *TraceObserver) handleSupportability() {
...
 
-func newSupportabilityMetrics() map[string]float64 {
-	// grpc codes, plus 1 for sent, and 1 for response errs
-	metrics := make(map[string]float64, numCodes+2)
+func newEmptyMetric() [6]float64 {
+	// [0] count
+	// [1] bytes sent
+	// [2] bytes received
+	// [3:] unused
+	return [6]float64{0.0, 0.0, 0.0, 0.0, 0.0, 0.0}
```
bduranleau-nr, line 401: can we abstract out `6` to a constant value here and reference that instead?

ZNeumann: Potentially, though the usage of [6]float for metrics (and other hard coded assumptions about the structure of metrics) is prevalent throughout all of our codebase. Unless the protocol changes the structure that NR expects for metrics, this will never change. I don't believe that setting the 6 to a constant would necessarily help in that regard.

bduranleau-nr:
Nice job with a lot of complex Go code!
My only suggestion would be to abstract out the `6` to it's own constant for consistency and to avoid 'magic numbers'

bduranleau-nr (approved):

PR #508, opened 2022-08-22
ci: allow multiple symbol defs at agent test link

misiekn:
PHP embed SAPI (libphp) embeds PCRE library. So does axiom library.
Agent unit tests require both: PHP embed SAPI and axiom library. PHPs
5.5, 5.6, 7.0, 7.1, and 7.2 embed older version of PCRE and this causes
linker to find PCRE symbols in two places: PHP embed SAPI (libphp) and
axiom library. This causes the linker to report a fatal error and abort.
This change allows for multiple definitions and the first definition
found will be used.

Note: PHP embed SAPIs 7.3+ come with PCRE2, which uses different symbol
names and does not create conflict with PCRE embedded in axiom library.


zsistla (approved):
Looks great.  Thanks so much for investigating/fixing this issue :)

PR #511, opened 2022-08-24
fix(agent): Datastore segments should not have children segments.

zsistla:
Datastore segments are not allowed to have children segments since it would scramble the exclusive time calculation.

It is enforced here where we use the API to create a datastore segment: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_api_datastore.c#L195

However, it was not enforced when we automatically create a datastore segment.

There is already a test case that verifies it here: https://github.com/newrelic/newrelic-php-agent/blob/main/tests/integration/api/datastore/test_all_parameters_no_children.php
During CLM debugging, this case would fail, but after the fix it succeeds.

### DIFF
--- a/axiom/nr_segment_datastore.c
+++ b/axiom/nr_segment_datastore.c
@@ -77,7 +77,7 @@ static char* create_metrics(nr_segment_t* segment,
 
 bool nr_segment_datastore_end(nr_segment_t** segment_ptr,
                               nr_segment_datastore_params_t* params) {
-  nrtxn_t* txn;
+  nrtxn_t* txn = NULL;
   bool is_sql = false;
   const char* datastore_string = NULL;
   const char* collection = NULL;
@@ -89,7 +89,7 @@ bool nr_segment_datastore_end(nr_segment_t** segment_ptr,
   nr_slowsqls_labelled_query_t input_query_allocated = {NULL, NULL};
   char* input_query_query = NULL;
   nr_segment_datastore_t datastore = {0};
-  nr_segment_t* segment;
+  nr_segment_t* segment = NULL;
   bool rv = false;
 
   if (NULL == segment_ptr) {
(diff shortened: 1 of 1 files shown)


zsistla:
ok jenkins

ZNeumann (approved):

axiom/nr_segment_datastore.c
```diff
@@ -107,6 +107,19 @@ bool nr_segment_datastore_end(nr_segment_t** segment_ptr,
...
 
+  /*
+   * We don't want datastore segments to have any children, as
+   * this would scramble the exclusive time calculation.
+   *
+   * Therefore, we delete all children of the segment.
+   */
+  if (segment) {
+    for (size_t i = 0; i < nr_segment_children_size(&segment->children); i++) {
+      nr_segment_t* child = nr_segment_children_get(&segment->children, i);
```
mfulb, line 118: Is this loop going to work while you remove segments?  Seems like the index will be off as it is relative to the original size and it appears this code is removing elements?

zsistla: In this case no, we didn't actually remove the child from `children`, we just turned it NULL when we discarded the child segment so it is thereafter ignored. 
if we had instead used `nr_segment_children_remove` that would have changed the value of `nr_segment_children_size` each time through the loop and the incrementing `i` would have caused the index to be off.  However, `nr_segment_children_remove` doesn't discard the segment, it just goes through the expensive task of removing from the children vector after which we still have to discard.  So doing the above loop saves us the expense of removing the child and then discarding it because once we discard the segment, it is ignored.

mfulb (approved):

misiekn (commented):
Since `newrelic_record_datastore_segment` calls `nr_segment_datastore_end` [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_api_datastore.c#L211), right after deleting all children of the datastore segment, and this fix adds deleting all children of the datastore segment to `nr_segment_datastore_end`, wouldn't it be prudent to remove the logic of deleting all children of the datastore segment from `newrelic_record_datastore_segment` to reduce code duplication? 

The only way I see `nr_segment_datastore_end` wouldn't delete all children of the datastore segment when called from `newrelic_record_datastore_segment` is if it were called with improper arguments, i.e. NULL segment and NULL params but both arguments are set correctly in newrelic_record_datastore_segment`.

**EDIT:** Ah! I see the subtle difference between the logic in `nr_segment_datastore_end` and `newrelic_record_datastore_segment` - the latter forces the newly created data store segment as the current transaction's segment (`NRTXN(force_current_segment)`). Therefore deleting all children of the datastore segment cannot be simply moved from `newrelic_record_datastore_segment` to `nr_segment_datastore_end`.

bduranleau-nr (approved):
👍

misiekn (approved):

PR #514, opened 2022-08-27
feat(agent, daemon): Add priority sampled log event harvest

mfulb:
This work adds priority sampled log event harvest to the agent.  It moves the representation of log events on the agent side from a JSON string stored in an `nr_analytics_event_t` to storing `nr_log_event_t` elements in a minmax heap.  This is similar to how span events (via segments) are stored using priority sampling.  

Log event priorities are set to match that of the linked segment.  When considering whether an event will stay in the heap as a tie breaker when elements have the same priority the age of a log event is considered.  Older events are dropped in favor of newer events with the same priority.   

On the daemon side the transaction sampling priority (DT, if it exists, otherwise a random number is chosen on transaction creation) is used for sampling log events.  

This work also adds the needed code to allow the agent to send log events to the daemon via the flatbuffer.

Unit tests have been added to exercise the new code added.

Additionally a new suite of integration tests for monolog are included to extend coverage.


mfulb:
Created using wrong root branch - ignore...

PR #515, opened 2022-08-27
 feat(agent, daemon): Add priority sampled log event harvest 

mfulb:
This work adds priority sampled log event harvest to the agent.  It moves the representation of log events on the agent side from a JSON string stored in an `nr_analytics_event_t` to storing `nr_log_event_t` elements in a minmax heap.  This is similar to how span events (via segments) are stored using priority sampling.  

Log event priorities are set to match that of the linked segment.  When considering whether an event will stay in the heap as a tie breaker when elements have the same priority the age of a log event is considered.  Older events are dropped in favor of newer events with the same priority.   

On the daemon side the transaction sampling priority (DT, if it exists, otherwise a random number is chosen on transaction creation) is used for sampling log events.  

This work also adds the needed code to allow the agent to send log events to the daemon via the flatbuffer.

Unit tests have been added to exercise the new code added.

Additionally a new suite of integration tests for monolog are included to extend coverage.


mfulb:
ok jenkins

mfulb:
ok jenkins

axiom/cmd_txndata_transmit.c
```diff
@@ -106,6 +106,89 @@ static uint32_t nr_txndata_prepend_custom_events(nr_flatbuffer_t* fb,
...
+  offsets = (uint32_t*)nr_calloc(event_count, sizeof(uint32_t));
+  offset = &offsets[0];
+
+  /* convert log events from heap into a vector */
+  events_vec = nr_vector_create(event_count, NULL, NULL);
+  nr_log_events_to_vector(txn->log_events, events_vec);
+
+  /* Using a buffer here means we can write the encoded span events into it,
+   * saving a couple of allocations (including at least one full string
+   * duplication) per span event. */
```
zsistla, line 139:
```suggestion
  /* Using a buffer here means we can write the encoded log events into it,
   * saving a couple of allocations (including at least one full string
   * duplication) per log event. */
```

axiom/nr_log_event.c
```diff
@@ -37,11 +37,83 @@ void nr_log_event_destroy(nr_log_event_t** ptr) {
...
+  nr_buffer_add(buf, NR_PSTR("\""));
+  nr_buffer_add(buf, final_value, nr_strlen(final_value));
+  nr_buffer_add(buf, NR_PSTR("\""));
+
+  return true;
+}
+
 char* nr_log_event_to_json(const nr_log_event_t* event) {
   nrbuf_t* buf = nr_buffer_create(0, 0);
   char* json = NULL;
```
zsistla, line 115: check for NULL event?

axiom/nr_log_events.c
```diff
@@ -8,32 +8,214 @@
...
+      || 0 == events->events_allocated) {
+    return true;
+  }
 
-  nr_analytics_events_add_event(events, log_event, rnd);
-  nr_free(log_event);
+  events_sampled = nr_log_events_is_sampling(events);
+
+  // make a copy of the event so the heap can control its entire life cycle
+  dup_event = nr_log_event_clone(event);
```
zsistla, line 178: It seems expensive to clone every log event.  What is the advantage?
Why not just pass control of the original event to the heap and not destroy the event [here](https://github.com/newrelic/newrelic-php-agent/pull/515/files#diff-aff431af9f447c5431ea6466c518a37c44b2491884a829f6c9e4675082ca8186R3385)?

axiom/nr_log_events.c
```diff
@@ -8,32 +8,214 @@
...
+    events->events_seen++;
 
-  log_event_json = nr_log_event_to_json(event);
-  log_event = nr_analytics_event_create_from_string(log_event_json);
-  nr_free(log_event_json);
+  // if no event queue exists or size is 0 then event will be dropped
+  if (NULL == events || NULL == events->events
+      || 0 == events->events_allocated) {
+    return true;
+  }
```
zsistla, line 173: Style nit global comment.
Inconsistent use of `{}` after one line if statements.
I'd vote for using the `{}` with all the if statements because it makes it clear and makes it easier to add additional lines to the if statement later.

axiom/nr_log_events.h
```diff
@@ -6,13 +6,17 @@
 #ifndef NR_LOG_EVENTS_HDR
 #define NR_LOG_EVENTS_HDR
 
-#include "nr_analytics_events.h"
+//#include "nr_analytics_events.h"
```
zsistla, line 9:
```suggestion
```

axiom/nr_log_events.h
```diff
@@ -23,28 +27,83 @@
...
+ *
+ */
+extern void nr_log_events_destroy(nr_log_events_t** events);
+
+/*
+ * Purpose : Get the maximum number of events held by a event pool.
+ *
+ * Params  : 1. Log event pool
+ *
+ * Returns : Maximum number of events held by a event pool.
```
zsistla, line 57:
```suggestion
 * Returns : Maximum number of events held by an event pool.
```

axiom/nr_log_events.h
```diff
@@ -23,28 +27,83 @@
...
+/*
+ * Purpose : Destrory a log event pool.
+ *
+ * Params  : 1. Log event pool
+ *
+ */
+extern void nr_log_events_destroy(nr_log_events_t** events);
+
+/*
+ * Purpose : Get the maximum number of events held by a event pool.
```
zsistla, line 53:
```suggestion
 * Purpose : Get the maximum number of events held by an event pool.
```

axiom/nr_log_events.h
```diff
@@ -23,28 +27,83 @@
...
+ *
+ */
+extern size_t nr_log_events_max_events(const nr_log_events_t* events);
+
+/*
+ * Purpose : Get the number of log events seen by a event pool.
+ *
+ * Params  : 1. Log event pool
+ *
+ * Returns : Number of log events seen by a event pool.
```
zsistla, line 67:
```suggestion
 * Purpose : Get the number of log events seen by an event pool.
 *
 * Params  : 1. Log event pool
 *
 * Returns : Number of log events seen by an event pool.
```

axiom/nr_log_events.h
```diff
@@ -23,28 +27,83 @@
...
+ *
+ */
+extern size_t nr_log_events_number_seen(const nr_log_events_t* events);
+
+/*
+ * Purpose : Get the number of log events saved by a event pool.
+ *
+ * Params  : 1. Log event pool
+ *
+ * Returns : Number of log events saved by a event pool.
```
zsistla, line 77:
```suggestion
 * Purpose : Get the number of log events saved by an event pool.
 *
 * Params  : 1. Log event pool
 *
 * Returns : Number of log events saved by an event pool.
```

axiom/nr_txn.c
```diff
@@ -3300,12 +3302,24 @@ static void log_event_set_linking_metadata(nr_log_event_t* e,
...
+  nr_log_event_set_priority(e, 0);
+
   if (nrlikely(txn)) {
+    segment = nr_txn_get_current_segment(txn, NULL);
+    if (NULL != segment) {
+      // bump segment priority to increase chance it is saved
+      // if sampling occurs
+      nr_segment_set_priority_flag(segment, NR_SEGMENT_PRIORITY_LOG);
+      nr_log_event_set_priority(e, nr_segment_get_priority_flag(segment));
+    }
```
zsistla, line 3321:
```suggestion
  /* default priority to lowest value */
  nr_log_event_set_priority(e, 0);

  if (nrlikely(txn)) {
    segment = nr_txn_get_current_segment(txn, NULL);
    if (NULL != segment) {
      /*
      * bump segment priority to increase chance it is saved
      * if sampling occurs
      */
      nr_segment_set_priority_flag(segment, NR_SEGMENT_PRIORITY_LOG);
      nr_log_event_set_priority(e, nr_segment_get_priority_flag(segment));
    }
```
Another nit, but when we scrubbed for open source, we tried to move all comments to the `/* */` format.

axiom/nr_log_events.c
```diff
@@ -8,32 +8,214 @@
...
+  bool events_sampled = false;
+  nr_log_event_t* dup_event;
+
   if (NULL == event) {
     return false;
   }
 
-  events_sampled = nr_analytics_events_is_sampling(events);
+  // need to increment this if a valid event was sent
+  // so take care if it here then sanity check other args
```
zsistla, line 165:
```suggestion
  /*
   * need to increment this if a valid event was sent
   *  so take care if it here then sanity check other args
   */
```

axiom/nr_log_events.c
```diff
@@ -8,32 +8,214 @@
...
+        (nr_minmax_heap_dtor_t)nr_log_event_discard_wrapper, NULL);
+  } else {
+    events->events = NULL;
+  }
+
+  return events;
+}
 
+void nr_log_events_destroy(nr_log_events_t** events_ptr) {
+  nr_log_events_t* events;
```
zsistla, line 117:
```suggestion
  nr_log_events_t* events = NULL;
```

axiom/nr_log_events_private.h
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * This file contains log events data internals.
+ */
+#ifndef NR_LOG_EVENTS_PRIVATE_HDR
+#define NR_LOG_EVENTS_PRIVATE_HDR
+
+//#include "nr_log_event.h"
```
zsistla, line 12:
```suggestion
```

zsistla:
Thanks for working on this complex task and adding really solid test cases!
The integration tests were especially satisfying to see with all the log events being generated :)

axiom/cmd_txndata_transmit.c
mfulb: Thanks!

axiom/nr_log_events.c
mfulb: Thanks for catching this - I noticed some inconsistencies but didn't catch them all.

Should be fixed in commit [a4b26a8](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/a4b26a8d2106e7d06b164a34f5e7c3081779bf46).

axiom/nr_txn.c
mfulb: Thanks for the heads up on this.

axiom/nr_log_events.c
zsistla: Thanks!  One more here, I think: https://github.com/newrelic/newrelic-php-agent/pull/515/files#diff-80d8ae697d75559f2c45cc1e90dd0fa8b210833ed35750748f156f8c651f1cb4R143

axiom/nr_log_event.c
mfulb: This is checked in the function which is used to actually format the json (`nr_log_event_to_json_buffer_ex()`) but I will add a check here see it is more obvious it is handled.

Fixed in commit [741a19b](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/741a19b0bfc79066d8ffad066cc30a58238ee90b).

axiom/nr_log_events.c
mfulb: Doh - fixed in commit [b24a5cf](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/b24a5cfb856bdba70b50583a3c9e47b5a7d89e76).

axiom/nr_log_events_private.h
mfulb: Missed something else - thanks for taking the time to catch these.

axiom/nr_log_events.c
mfulb: The initial implementation (prior to switching to minmax heap) worked where the caller was freeing the event so I modeled this to follow that flow.
Let me see how invasive it would be to change the existing code to not free the event.

axiom/nr_log_events.h
```diff
@@ -23,28 +27,83 @@
...
-#endif
+extern bool nr_log_events_add_event(nr_log_events_t* events,
+                                    const nr_log_event_t* event);
+
+/*
+ * Purpose : Create a log event pool of specified size.
+ *
+ * Params  : 1. Log event pool
+ *
+ *  * Returns : Allocated log event pool or NULL on failure.
```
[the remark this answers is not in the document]
mfulb: Yes the caller is responsible for calling nr_log_events_destroy() to free the item.

I updated the comment in commit [cec0caf](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/cec0caf86a09582905793efadabed91a7393520a).

axiom/nr_log_events.c
mfulb: Ah good idea thanks.

axiom/nr_log_events.c
mfulb: I believe commit [08a36da](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/08a36dae321abfef2d4b572edfc5ab6af5a07d7d) implements this functionality without the clone.
Please check this commit for correctness.  It does pass axiom-valgrind but is a departure from how the code was originally written so extra review is appreciated.

axiom/tests/test_log_events.c
```diff
@@ -19,120 +20,138 @@
...
-  nr_log_event_destroy(&e);
+  nr_log_event_set_priority(e, 1);
+  nr_log_events_add_event(events, e);
 
   e = create_sample_event(LOG_MESSAGE_3);
-  nr_log_events_add_event(events, e, rnd);
-  nr_log_event_destroy(&e);
+  nr_log_event_set_priority(e, 0);
+  nr_log_events_add_event(events, e);
+
```
misiekn, line 63:
```suggestion
// clang-format off
```
To ensure that clang-format will never touch the formatting of the min-max heap presentation. I've seen clang-format destroy neatly formatted explanation comments and would really like to avoid it here.

axiom/tests/test_log_events.c
```diff
@@ -19,120 +20,138 @@
...
+  //         0
+  //       3   2
+  //      1
+  //
+  // so when walking the heap and converting to a vector we expect
+  // the following order:
+  // LOG_MESSAGE_3 (pri 0)
+  // LOG_MESSAGE_0 (pri 3)
+  // LOG_MESSAGE_1 (pri 2)
+  // LOG_MESSAGE_2 (pri 1)
```
misiekn, line 74:
```suggestion
  // LOG_MESSAGE_2 (pri 1)
  // clang-format on
```
Restore clang-format operation

axiom/tests/test_log_events.c
```diff
@@ -19,120 +20,138 @@
...
+  void* test_e;
+  nr_vector_t* vector = NULL;
+  bool pass;
+  char* json;
   const char* expected;
 
   events = nr_log_events_create(10);
   tlib_fail_if_null("events created", events);
 
   e = create_sample_event(LOG_MESSAGE_0);
```
misiekn, line 48: Would it make sense to extend `create_sample_event` to become `create_sample_event_with_priority` that would take priority as an additional argument? This way the code would read like this:
```
e = create_sample_event_with_priority(LOG_MESSAGE_0, 3);
nr_log_events_add_event(events, e);
```
and almost match the explaining comment below.

axiom/tests/test_log_events.c
```diff
@@ -19,120 +20,138 @@
...
-                 "\"hostname\":\"null\"" \
-              "}]";                                    \
-    tlib_pass_if_str_equal("add event", expected, json); \
+  // convert log event pool to a vector and test
+  vector = nr_vector_create(10, NULL, NULL);
+  nr_log_events_to_vector(events, vector);
+
+#define TEST_EVENTS_GET(MESSAGE, IDX)                              \
+  do {                                                             \
+    pass = nr_vector_get_element(vector, IDX, &test_e);            \
```
misiekn, line 87: Same comment about `pass` name as in `cmd_txndata_transmit.c:nr_txndata_prepend_log_events` - `found` reads better.

axiom/tests/test_log_events.c
```diff
@@ -19,120 +20,138 @@
...
+          "\","                                                    \
+          "\"timestamp\":" NR_STR2(LOG_TIMESTAMP) "}]";            \
+    tlib_pass_if_str_equal("add event", expected, json);           \
+    nr_free(json);                                                 \
   } while (0)
 
-  /* Test events are stored in order */
-  TEST_EVENTS_GET(LOG_MESSAGE_0, 0);
-  TEST_EVENTS_GET(LOG_MESSAGE_3, 3);
+  /* Test events are stored in order as explained above for minmax heap*/
```
misiekn, line 103: I'm not sure these tests should exists. `nr_minmax_heap_iterate` (underlying implementation of `nr_log_events_to_vector`) documentation says:
> Note : Ordering is not guaranteed; calling code should be prepared to
receive values in any order.

I'm afraid this may make this test unpredictable.

axiom/tests/test_log_events.c
```diff
@@ -142,33 +161,125 @@ static void test_events_sample(void) {
   tlib_pass_if_int_equal("events number saved updated", 2,
                          nr_log_events_number_saved(events));
 
-  nr_random_destroy(&rnd);
+  // verify two highest priority events were retained
+  // convert log event pool to a vector and test
```
misiekn, line 165: I'm not sure these tests should exists. nr_minmax_heap_iterate (underlying implementation of nr_log_events_to_vector) documentation says:

> Note : Ordering is not guaranteed; calling code should be prepared to
receive values in any order.

I'm afraid this may make this test unpredictable. I don't know how to make this test independent of the order of elements in the vector.

axiom/tests/test_log_events.c
```diff
@@ -142,33 +161,125 @@ static void test_events_sample(void) {
...
 
+static void test_log_event_comparator(void) {
+  nr_log_event_t log0 = {.message = "LOG0", .priority = 30, .timestamp = 0};
+  nr_log_event_t log1 = {.message = "LOG1", .priority = 20, .timestamp = 0};
+  nr_log_event_t log2 = {.message = "LOG2", .priority = 10, .timestamp = 0};
+  nr_log_event_t log3 = {.message = "LOG3", .priority = 0, .timestamp = 0};
+  nr_log_event_t log4 = {.message = "LOG4", .priority = 0, .timestamp = 100};
+
+  nr_vector_t* events = nr_vector_create(12, NULL, NULL);
+
```
misiekn, line 227:
```suggestion
  // clang-format off
```

axiom/tests/test_log_events.c
```diff
@@ -142,33 +161,125 @@ static void test_events_sample(void) {
...
+   * that are considered by the comparator function.
+   *
+   * Position | Message    | Priority  | Timestamp
+   * ---------+------------+-----------+----------
+   * 4        | "LOG0"     | 30        | 0
+   * 3        | "LOG1"     | 20        | 0
+   * 2        | "LOG2"     | 10        | 0
+   * 1        | "LOG4"     | 0         | 100
+   * 0        | "LOG3"     | 0         | 0
+   */
```
misiekn, line 250:
```suggestion
   */
   // clang-format on
```

tests/integration/logging/monolog2/test_monolog_basic.php
```diff
@@ -0,0 +1,198 @@
...
+    $stdoutHandler = new StreamHandler('php://stdout', Logger::DEBUG);
+    $stdoutHandler->setFormatter($formatter);
+
+    $logger->pushHandler($stdoutHandler);
+    
+    // insert delays between log messages to allow priority sampling
+    // to resolve that later messages have higher precedence
+    // since timestamps are only millisecond resolution
+    // without delays sometimes order in output will reflect
+    // all having the same timestamp.
```
misiekn, line 180: Does the integration_runner care about the order of elements in EXPECT_LOG_EVENTS? If so, is the delay between events enough to guarantee that they will be sent by daemon in this order?

tests/integration/logging/monolog2/test_monolog_basic.php
```diff
@@ -0,0 +1,198 @@
...
+require_once(realpath(dirname(__FILE__)) . '/../../../include/config.php');
+require_once(realpath(dirname(__FILE__)) . '/../../../include/monolog.php');
+require_monolog(2);
+
+use Monolog\Logger;
+use Monolog\Handler\StreamHandler;
+use Monolog\Formatter\LineFormatter;
+
+
+function test_logging() {
```
misiekn, line 165: I wonder if it is possible to have this function take a monolog_version parameter, and put it in a reusable module in tests/integration/logging/test_monolog.php. It might reduce some code duplication but I'm not entirely convinced this will be beneficial though.

misiekn (commented):
Great work! And pleasant to read too! 👍 It's so nice to see the support for logging features in integration tests that you created so meticulously! I have a few questions to be clarified and a few refactoring suggestions to consider.

axiom/nr_log_event.h
```diff
@@ -51,7 +67,23 @@ extern char* nr_log_event_to_json(const nr_log_event_t* event);
...
+ *           3. Whether or not to wrap JSON with starting and
+ *              ending "[" "]".  If enabled the resulting
+ *              JSON is not valid unless enclosed in another
+ *              JSON string.
+ *
+ * Returns : True on success; false on error.
+ */
+extern bool nr_log_event_to_json_buffer_ex(const nr_log_event_t* event,
+                                           nrbuf_t* buf,
+                                           bool partial);
```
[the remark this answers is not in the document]
mfulb: The naming follows the naming scheme used for span events - I would prefer to retain the name as it is specific to the type of serialization occurring.   If we supported multiple serialization methods I could see using a more general name.

The API also follows that used for span events (and other events) in that all other functions which generate JSON for events makes valid JSON strings including the wrapping '[' ']'.   To be consistent we provide similarly named functions for log events which provide valid JSON.  However when constructing a payload for the log event endpoint on the collector we cannot string together these valid JSON strings (like apparently works for other event types).  So for efficiency I created a special version of the function which could generate an invalid JSON string (but a valid JSON substring) if requested.

It might be the non-"ex" versions of these functions are not needed at all and perhaps I could leave just the "ex" version.  I'd like to designate it differently as there is precedence with the span event versions of these functions that a valid JSON string is returned.

ZNeumann (approved):

axiom/tests/test_log_events.c
mfulb: I don't think this will be reformatted I kept the line lengths short enough.

axiom/tests/test_log_events.c
mfulb: I don't think this will be reformatted I kept the line lengths short enough.

tests/integration/logging/monolog2/test_monolog_basic.php
mfulb: Currently the runner does care.  If I did not give enough time between events that they had unique timestamps (log messages only stored with millisecond precision) the events would occasionally come out ordered differently.  This is because the tie breaker for event with the same priority is to favor newer events.  Since all had the same age normally this worked.  But sometimes the test ran and one (or more) events would have a different timestamp (ran the test right as millisecond flipped over to a new one) and it seemed to cause a different ordering.

For this test we don't really care about the order - just that the log events listed actually are created.  But the runner doesn't have a way to specify this currently - it does a exact match.  So this is probably the best we can do at the moment.  I don't believe as written it would ever produce a false positive which is good.

tests/integration/logging/monolog2/test_monolog_basic.php
mfulb: There are strings in the expected metrics which depend on a file path which is different for monolog 2 and monolog 3.  I suppose we could wildcard that?
I would suggest this is a good idea for a future enhancement.

axiom/nr_log_event.h
```diff
@@ -33,6 +36,19 @@ extern nr_log_event_t* nr_log_event_create(void);
...
+ *
+ * Params  : A pointer that points at the allocated
+ *           nr_log_event_t (created with nr_log_event_create) that
+ *           is the source log event.
+ *
+ * Returns : Pointer to a new log event with all fields copied
+ *           from the source with newly allocated memory so
+ *           the original log event can be destroyed.
+ */
+extern nr_log_event_t* nr_log_event_clone(const nr_log_event_t* event);
```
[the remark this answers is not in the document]
mfulb: Removed in commit [f03ad30](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/f03ad30334590a12d4fa28af828f12a46329a8e7).

axiom/nr_log_event.h
```diff
@@ -61,15 +93,19 @@ extern bool nr_log_event_to_json_buffer(const nr_log_event_t* event,
  *
  * Returns : Nothing.
  */
-extern void nr_log_event_set_message(nr_log_event_t* event, const char* message);
-extern void nr_log_event_set_log_level(nr_log_event_t* event, const char* log_level);
+extern void nr_log_event_set_message(nr_log_event_t* event,
```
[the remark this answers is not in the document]
mfulb: Good idea - fixed in commit [268a4fe](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/268a4fe0a297df41ff3cbfb5daea03086e13625b).

axiom/nr_log_events.c
```diff
@@ -8,32 +8,217 @@
...
-  nr_analytics_event_t* log_event = NULL;
+bool nr_log_events_is_sampling(nr_log_events_t* events) {
+  if (events->events_used < events->events_allocated) {
+    return false;
+  }
+  return true;
+}
+
+/*
+ * Safety check for comparator functions
```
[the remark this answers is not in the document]
mfulb: Got it from the segment units tests.

axiom/nr_log_events_private.h
```diff
@@ -0,0 +1,22 @@
...
+
+/*
+ * This file contains log events data internals.
+ */
+#ifndef NR_LOG_EVENTS_PRIVATE_HDR
+#define NR_LOG_EVENTS_PRIVATE_HDR
+
+#include "util_minmax_heap.h"
+
+struct _nr_log_events_t {
```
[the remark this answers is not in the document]
mfulb: I reworked this code so I was able to remove the "nr_log_events_private.h" as suggested as well as not needing to include "nr_log_event.h" in "nr_log_events.c" as I moved the comparator function to "nr_log_event.c" where it really belonged.

Please review these changes in commit [2ff66e5](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/2ff66e55a96f95555ba4f28ba463e88a57841721).

axiom/cmd_txndata_transmit.c
```diff
@@ -106,6 +106,89 @@ static uint32_t nr_txndata_prepend_custom_events(nr_flatbuffer_t* fb,
...
+  nr_vector_t* events_vec;
+  nrbuf_t* buf;
+
+  const size_t event_size = sizeof(uint32_t);
+  const size_t event_align = sizeof(uint32_t);
+
+  event_count = nr_log_events_number_saved(txn->log_events);
+  if (0 == event_count) {
+    return 0;
+  } else if (event_count > log_event_limit) {
```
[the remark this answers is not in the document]
mfulb: This might be just defensive programming on my part to make sure there is sanity between how the txn->log_events was created and what is later being requested sent to the daemon.  And the desire to make sure if the limit is set to 0 then nothing is sent?

axiom/nr_log_event.h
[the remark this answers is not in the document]
misiekn: Ok. I am convinced about the reason for `nr_log_event_to_json_buffer`. But I'd still suggest reducing unused code by dropping the `_ex` suffix and `bool partial` arguments and remove the original `nr_log_event_to_json_buffer` that creates an array of one element.

axiom/nr_log_events.c
```diff
@@ -8,32 +8,217 @@
...
+#include "util_minmax_heap.h"
+#include "util_vector.h"
 
-bool nr_log_events_add_event(nr_analytics_events_t* events,
-                             const nr_log_event_t* event,
-                             nr_random_t* rnd) {
-  bool events_sampled = false;
-  char* log_event_json = NULL;
-  nr_analytics_event_t* log_event = NULL;
+bool nr_log_events_is_sampling(nr_log_events_t* events) {
```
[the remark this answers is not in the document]
mfulb: Thanks for catching this.  I fixed it and added a unit test to test this function with NULL as an argument.

Changes in commit [46ff32c](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/46ff32cc4805e3af10d4ee6b86a6a32b6994a93e).

axiom/nr_log_events.c
```diff
@@ -8,32 +8,217 @@
...
+ * Purpose : Place an nr_log_event_t pointer in a heap into a nr_set_t,
+ *             or "heap to set".
+ *
+ * Params  : 1. The log event pointer in the heap.
+ *           2. A void* pointer to be recast as the pointer to the set.
+ *
+ * Note    : This is the callback function supplied to nr_minmax_heap_iterate
+ *           used for iterating over a heap of log events and placing each
+ *           log events into a set.
+ */
```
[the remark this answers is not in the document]
mfulb: Thanks for catching these.

src/newrelic/log_events.go
```diff
@@ -30,3 +35,48 @@ func (events *LogEvents) AddEventFromData(data []byte, priority SamplingPriority
...
+// by the collector.
+func (events *LogEvents) CollectorJSON(id AgentRunID) ([]byte, error) {
+	buf := &bytes.Buffer{}
+
+	es := *events.analyticsEvents.events
+
+	estimate := len(es) * 128
+	buf.Grow(estimate)
+	buf.WriteString(`[{` +
+		`"common": {"attributes": {}},` +
```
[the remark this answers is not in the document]
mfulb: The collector did not seem to be happy with it otherwise.

axiom/cmd_txndata_transmit.c
```diff
@@ -106,6 +106,89 @@ static uint32_t nr_txndata_prepend_custom_events(nr_flatbuffer_t* fb,
...
+   * duplication) per log event. */
+  buf = nr_buffer_create(0, 0);
+
+  for (i = 0; i < event_count; i++, offset++) {
+    void* event;
+    uint32_t data;
+    bool pass;
+
+    nr_buffer_reset(buf);
+    pass = nr_vector_get_element(events_vec, i, &event);
```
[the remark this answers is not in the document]
mfulb: I went through and addressed this in commit [57d0c41](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/57d0c41d523951200221cc4a1dc83e877af15b0c).

axiom/tests/test_log_events.c
mfulb: Fixed in commit [57d0c41d](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/57d0c41d523951200221cc4a1dc83e877af15b0c).

axiom/tests/test_log_events.c
mfulb: I'd rather leave this as is - I think the code legibility is fine.

axiom/tests/test_log_events.c
mfulb: The internals of the heap maybe aren't what we care about.
What if I change the test to keep popping the min value and make sure these follow the expected pattern?

axiom/tests/test_log_events.c
mfulb: I've changed the test to sort the vector using the same comparator as the heap uses.  This should put the event with the minimum priority in position 0 and the other event at position 1 reliably.

Fix in commit [284f848](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/284f848717cb9c0eef21f25b75899379077f9ceb).

axiom/nr_log_events_private.h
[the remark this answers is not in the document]
misiekn: Looks good. Thank you!

axiom/tests/test_log_events.c
misiekn: > ...change the test to keep popping...
Yes or do what you did in 284f848 :) Both solutions will improve test stability.

mfulb:
ok jenkins

zsistla (approved):
Looks good.

axiom/nr_log_event.h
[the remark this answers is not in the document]
mfulb: Fixed in commit [2622c75](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/2622c757d16d109329f6bcc4b7b571cadbfcd4cf).

axiom/tests/test_log_events.c
mfulb: Fixed in commit [7532fdf](https://github.com/newrelic/newrelic-php-agent/pull/515/commits/7532fdf7755f39e1ca31cd646a3192f133f37322).

axiom/nr_log_event.h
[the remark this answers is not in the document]
misiekn: Looks good. Thank you for bearing with me.

misiekn (approved):
Thank you for patiently addressing my feedback. LGTM 👍

mfulb:
ok jenkins

mfulb:
ok jenkins

mfulb:
ok jenkins

PR #516, opened 2022-08-29
feat(agent): Propagate OAPI return values and update return value functions.

zsistla:
1) updated macros to pass the OAPI given return value throughout the userland system.
2) changed nr_php_get_return_value to return oapi given pointer
3) Added the OVERWRITE_ZEND_EXECUTE_DATA to allow us to toggle off during feature addition, but toggle off to maintain CI as long as possible.  It also gives the flexibility to revert to the instrumentation prior to OAPI.
4) macro to toggle between just using the OAPI return value when OAPI is enabled or calling nr_php_get_return_value_ptr when it is not OAPI.

### DIFF
--- a/agent/fw_drupal.c
+++ b/agent/fw_drupal.c
@@ -216,7 +216,7 @@ NR_PHP_WRAPPER(nr_drupal_http_request_exec) {
     goto end;
   }
 
-  return_value = nr_php_get_return_value_ptr(TSRMLS_C);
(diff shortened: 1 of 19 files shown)


mfulb (approved):
Thanks Amber!

zsistla:
ok jenkins

ZNeumann (approved):

misiekn (approved):

PR #520, opened 2022-08-31
feat(agent): send logging configuration metrics

misiekn:
Send `Supportability/Logging/Metrics/PHP/{enabled|disabled}` and
`Supportability/Logging/Forwarding/PHP/{enabled|disabled}` with
value for {enabled|disabled} based on the INI settings. Metrics
will only be sent once.

### DIFF
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -624,6 +624,35 @@ static void nr_php_txn_log_error_dt_on_tt_off(void) {
   }
 }
 
+static void nr_php_txn_send_metrics_once(nrtxn_t* txn) {
+  static unsigned int sent = 0;
+  char* metname = NULL;
+
+  if (nrunlikely(NULL == NRPRG(txn))) {
+    return;
+  }
+
+  if (nrlikely(0 != sent)) {
+    return;
+  }
+
+#define FMT_BOOL(v) (v) ? "enabled" : "disabled"
+
+  metname = nr_formatf("Supportability/Logging/Forwarding/PHP/%s",
+                       FMT_BOOL(nr_txn_log_forwarding_enabled(txn)));
+  nrm_force_add(NRTXN(unscoped_metrics), metname, 0);
+  nr_free(metname);
+
+  metname = nr_formatf("Supportability/Logging/Metrics/PHP/%s",
+                       FMT_BOOL(nr_txn_log_metrics_enabled(txn)));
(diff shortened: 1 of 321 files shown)


ZNeumann:
Looks good!

agent/php_txn.c
```diff
@@ -624,6 +624,35 @@ static void nr_php_txn_log_error_dt_on_tt_off(void) {
...
+
+  if (nrunlikely(NULL == NRPRG(txn))) {
+    return;
+  }
+
+  if (nrlikely(0 != sent)) {
+    return;
+  }
+
+#define FMT_BOOL(v) (v) ? "enabled" : "disabled"
```
zsistla, line 639: Small nit, not enough to hold up merging.

This is a nice macro, but if it's only going to be used here twice and then undefed, not sure why it was used vs just something like:
```
metname = nr_formatf("Supportability/Logging/Forwarding/PHP/%s",
                       FMT_BOOL(nr_txn_log_forwarding_enabled(txn) ? "enabled" : "disabled"));
```
Declaring it, using it twice, then undefing it seems a bit extra work.

Maybe make it a macro we can use going forward, declare it with our other macros, and don't undef it.  It is a handy little macro and adds clarity, but

zsistla (approved):
Great job, thanks.

agent/php_txn.c
misiekn: Here `true` needs to be formatted as `enabled` and `false` as `disabled`. In other contexts it may be necessary to format `bool` differently, maybe `true` and `false`. Hence I wouldn't make this macro globally available. The reasons this short lived macro was introduced were:
1. Don't repeat "enabled" and "disabled" string literals in the code in case bool representation will have to be changed
2. To make the lines shorter and have `metname` with `"Supportability/Logging/Forwarding/PHP/%s"` on the same line :) Without it, clang-format generates this code, which IMHO reads bad:
```
  metname
      = nr_formatf("Supportability/Logging/Forwarding/PHP/%s",
                   nr_txn_log_forwarding_enabled(txn) ? "enabled" : "disabled");
```

zsistla: Sounds good.

ZNeumann (approved):

zsistla:
ok jenkins

misiekn:
ok jenkins

misiekn:
ok jenkins

mfulb (approved):

PR #521, opened 2022-08-31
fix(agent): don't instrument Monolog's $context

misiekn:
This PR disables `$context` argument of `Monolog\Logger::addRecord` instrumentation. 
Context is no longer appended to the message in the log record. The code instrumenting
context can be re-enabled at compile time based on the  `HAVE_CONTEXT_IN_MESSAGE` 
macro value:
* 0 - `$context` argument will not be instrumented: its existence and value are ignored 
* 1 - the message of the log record forwarded by the agent will include the value of `$context`, which is appended to the value of `$message`.

### DIFF
--- a/agent/lib_monolog.c
+++ b/agent/lib_monolog.c
@@ -19,6 +19,19 @@
 #include "util_strings.h"
 #include "util_sleep.h"
 
+// clang-format off
+/*
+ * This macro affects how instrumentation $context argument of
+ * Monolog\Logger::addRecord works:
+ *
+ * 0 - $context argument will not be instrumented: its existance and value
+ *     are ignored 
+ * 1 - the message of the log record forwarded by the agent will have the value
+ *     of $context appended to the value of $message.
+ */
+// clang-format on
+#define HAVE_CONTEXT_IN_MESSAGE 0
+
 /*
  * Purpose : Convert Monolog\Logger::API to integer
  *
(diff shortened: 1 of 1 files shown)


agent/lib_monolog.c
```diff
@@ -295,17 +310,23 @@ static char* nr_monolog_get_context(const size_t argc,
  */
 static char* nr_monolog_build_message(const size_t argc,
                                       NR_EXECUTE_PROTO TSRMLS_DC) {
+#if !HAVE_CONTEXT_IN_MESSAGE
+  /* Make the compiler happy - argc is not used when $context is ignored */
+  (void)argc;
+#endif
```
ZNeumann, line 316:
```suggestion
static char* nr_monolog_build_message(const size_t argc NRUNUSED, /* unused when $context is ignored */
                                      NR_EXECUTE_PROTO TSRMLS_DC) {
```

ZNeumann: I've never used NRUNUSED, but I think this should work?

misiekn: I was thinking about `NRUNUSED` attribute in argument list option but IMHO 
```
#if !HAVE_CONTEXT_IN_MESSAGE
  /* Make the compiler happy - argc is not used when $context is ignored */
  (void)argc;
#endif
```
is more explicit than having `NRUNUSED` attribute on an argument that can be used.

ZNeumann: I'm okay with either. NRUNUSED is cleaner but the preprocessor is more explicit.

zsistla (approved):

PR #523, opened 2022-09-01
feat(agent): Remove function hooks that overwrite zend_execute_ex for USER functions.

zsistla:
1) Remove the function hooks from php_minit when we are using OAPI.
2) Made orig_func a no_op when we are using OAPI.
3) Made nr_zend_call_orig_execute_* functions no_op when using OAPI.
4) Updated the nr_php_observer module.

Testing: Utilized the OVERWRITE_ZEND_EXECUTE_DATA setting to toggle between current and new mode. With the overwrite setting all valgrind/integration tests passed. Without the setting, 121 tests fail (374 pass) since we aren't doing USER instrumentation anymore, but the internal instrumentation still functioned.

### DIFF
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -47,7 +47,32 @@
  *
  * Many functions here call zend_bailout to continue handling fatal PHP errors,
  * Since zend_bailout calls longjmp it never returns.
+ *
+ * Note: The agent ONLY needs to call this if it has overwritten the original
+ * zend_execute_ex.
  */
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA /* PHP 8.0+ and OAPI */
+/*
+ * There are a few various places, aside from the php_execute_* family that will
+ * call nr_zend_call_orig_execute_* so make it a noop to handle all cases when
+ * using OAPI.
+ */
+int nr_zend_call_orig_execute(NR_EXECUTE_PROTO TSRMLS_DC) {
+  NR_UNUSED_EXECUTE_DATA;
+  NR_UNUSED_FUNC_RETURN_VALUE;
+  return 0;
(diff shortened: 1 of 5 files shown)


zsistla:
Depends on functionality in https://github.com/newrelic/newrelic-php-agent/pull/516/

agent/php_user_instrument.c
```diff
@@ -47,6 +47,9 @@
  *
  * Many functions here call zend_bailout to continue handling fatal PHP errors,
  * Since zend_bailout calls longjmp it never returns.
+ *
+ * Note: The agent ONLY needs to call this if it has overwritten the original
+ * zend_execute_ex.
```
mfulb, line 52: Can we add a guard to this function to have it exit harmlessly if OAPI is enabled - or is this even a good idea?

zsistla: It will exit harmlessly since orig_execute points to a no_op.  This way we maintain backward compatibility when frameworks call it from within instrumentation.

agent/php_observer.c
```diff
@@ -86,12 +86,21 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
    */
   zend_observer_fcall_register(nr_php_fcall_register_handlers);
   zend_observer_error_register(nr_php_error_cb);
+
+  /*
+   * For Observer API with PHP 8+, we no longer need to ovewrwrite the zend
+   * execute hook.  orig_execute is called various ways in various places, so
+   * turn it into a no_op when using OAPI.
+   */
+  NR_PHP_PROCESS_GLOBALS(orig_execute) = nr_php_observer_no_op;
```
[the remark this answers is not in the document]
zsistla: Yep, it's a no_op so zcaught should never get toggled to a bad state.

mfulb (approved):

misiekn:
Closing as the changes from this PR are are included in #551.

PR #524, opened 2022-09-01
fix(axiom/test): increase time window on a time-dependent test

ZNeumann:
This is a test I've noticed failing on Jenkins multiple times. This failure is expected if the test process is not receiving much CPU time. Increasing the time window on the test should reduce the number of failures we see due to resource sharing.


misiekn (approved):
Nice find! Thanks for figuring it out!

zsistla (approved):
Nice fix, thank you!

ZNeumann:
ok jenkins

PR #529, opened 2022-09-12
fix(agent, daemon): Bump max log samples to 20k

mfulb:
This PR bumps the max log samples to 20000 per minute or 1666 per 5 second fast harvest.
The previous max was 10000 and based on OATS results the new value was selected.
The default value remains 10000.

The new maximum was verified by sending 1666 log events to the collector in a single harvest cycle and confirming this many event were recorded.

Several logging integration tests were updated in accordance with the new max value.

This PR also adds the HTTP response status to the debugging output as newer communications rely on a response code and not the response body to indicate status and errors.  This will help troubleshooting issues.

### DIFF
--- a/axiom/nr_limits.h
+++ b/axiom/nr_limits.h
@@ -55,6 +55,6 @@
 /*
  * The maximum number of log events in a transaction.
  */
-#define NR_MAX_LOG_EVENTS_MAX_SAMPLES_STORED 10000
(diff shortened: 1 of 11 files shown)


ZNeumann (approved):

zsistla (approved):

misiekn (approved):
Nice addition of the HTTP status code in the daemon logs! 👍

PR #534, opened 2022-09-21
feat(daemon): Add notification on # of apps

mfulb:
Simple addition to the daemon to output a notification when the number of applications is approaching the maximum limit.  Also another notification when the number of apps has dropped to a low value.

The intention is to allow for detecting when a new daemon container needs to be spun up to service more applications and also when the number of applications on that daemon container has dropped to a lower threshold.

Also added a debug only message that outputs the current number of applications which might be useful for troubleshooting .

### DIFF
--- a/src/newrelic/processor.go
+++ b/src/newrelic/processor.go
@@ -366,7 +366,6 @@ func (p *Processor) processAppInfo(m AppInfoMessage) {
 	}
 
 	numapps := len(p.apps)
-	log.Debugf("number of apps is %d", numapps)
 	if numapps > limits.AppLimit {
 		log.Errorf("unable to add app '%s', limit of %d applications reached",
 			m.Info, limits.AppLimit)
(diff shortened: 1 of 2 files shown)


misiekn (approved):
Nice little addition! 👍

bduranleau-nr (approved):
:+1:

src/newrelic/processor.go
```diff
@@ -365,10 +365,15 @@ func (p *Processor) processAppInfo(m AppInfoMessage) {
 		return
 	}
 
-	if len(p.apps) > limits.AppLimit {
+	numapps := len(p.apps)
+	log.Debugf("number of apps is %d", numapps)
```
zsistla, line 369: This would generate a LOT of debug messages.  It doesn't seem necessary with the high low messages already being generated. Same comment for https://github.com/newrelic/newrelic-php-agent/pull/534/files#diff-d34a767f316aaeded794ac3a231f867d3ffb42de19365ed338eace17f5876d94R716

mfulb: It only occurs if a new app connects to the daemon.  Do customers have lots of new apps connecting constantly?

mfulb: Removed in commit [182a501](https://github.com/newrelic/newrelic-php-agent/pull/534/commits/182a501d2d825404652e790fbefb8042128208ec).

zsistla (approved):
Nicely done, thanks!

mfulb:
ok jenkins

mfulb:
ok jenkins

PR #535, opened 2022-09-21
fix(agent): Correct message handler for log options

mfulb:
The various "enabled" INI options for agent log handling (forwarding) currently do not specify a handler for displaying the value in "phpinfo()" output so it is handles as a string pointer.  This causes the disabled state (integer 0) to be interpreted as a NULL string so "<no value>" is output instead.
This PR assigned the boolean display handler to these options.
Some previously non-clant-formatted code is corrected as well.


misiekn:
ok jenkins

misiekn (approved):
Nice little addition! 👍

bduranleau-nr (approved):
👍

zsistla (approved):
Thanks, great catch and fix!!

ZNeumann (approved):

PR #538, opened 2022-09-27
fix(agent): fix nr_drupal_http_request_exec

misiekn:
Fix memory corruption leading to seg fault caused by retrieving the name
of HTTP request method from execute_data after the call to drupal_http_request.

Accessing execute_data after NR_PHP_WRAPPER_CALL leads to memory 
corruption, because execute_data (which can be thought of as a function 
call stack frame) is destroyed after returning from function call and 
therefore it no longer holds a valid information about function args.

This change makes `drupal_http_request` instrumentation compliant with
[Checklist for writing a user function wrapper](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L64):
> 1. Any call to nr_php_arg_get, nr_php_scope_get, or
>    nr_php_get_return_value_ptr follows any NR_PHP_WRAPPER_REQUIRE_FRAMEWORK_*
>    calls and precedes any NR_PHP_WRAPPER_CALL.

More information can be found [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L48).

### DIFF
--- a/agent/fw_drupal.c
+++ b/agent/fw_drupal.c
@@ -230,20 +230,6 @@ NR_PHP_WRAPPER(nr_drupal_http_request_exec) {
         = {.library = "Drupal",
            .uri = nr_strndup(Z_STRVAL_P(arg1), Z_STRLEN_P(arg1))};
 
-    segment = nr_segment_start(NRPRG(txn), NULL, NULL);
-
-    /*
-     * Our wrapper for drupal_http_request (which we installed in
-     * nr_drupal_replace_http_request()) will take care of adding the request
-     * headers, so let's just go ahead and call the function.
-     */
-    NR_PHP_WRAPPER_CALL;
-
-    external_params.encoded_response_header
-        = nr_drupal_http_request_get_response_header(return_value TSRMLS_CC);
-
-    external_params.status
-        = nr_drupal_http_request_get_response_code(return_value TSRMLS_CC);
     /*
      * Drupal 6 will have a third argument with the method, Drupal 7 will not
      * have a third argument it must be parsed from the second.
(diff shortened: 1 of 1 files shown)


ZNeumann (approved):

misiekn:
ok jenkins

mfulb (approved):
Awesome work on this Michal!

misiekn:
ok jenkins

misiekn:
ok jenkins

misiekn:
ok jenkins

bduranleau-nr (approved):
👍 👍

zsistla:
Does having already called NR_PHP_WRAPPER_CALL
affect freeing the args here:
https://github.com/newrelic/newrelic-php-agent/pull/538/files#diff-d7eb1ac10e6c5271226788572bc376cf3eb6e38f6feb431178675e2ebb00233eR289-R291

misiekn:
> Does having already called NR_PHP_WRAPPER_CALL affect freeing the args [here](https://github.com/misiekn/newrelic-php-agent/blob/fix/nr_drupal_http_request_exec/agent/fw_drupal.c#L290)?

No it does not. `nr_php_arg_get` used [here](https://github.com/misiekn/newrelic-php-agent/blob/fix/nr_drupal_http_request_exec/agent/fw_drupal.c#L237) and [here](https://github.com/misiekn/newrelic-php-agent/blob/fix/nr_drupal_http_request_exec/agent/fw_drupal.c#L240) creates (allocates) new `zvals`. They do not refer to anything in `execute_data` because `nr_php_arg_get` makes a deep copy of arguments in `execute_data` via [ZVAL_DUP](https://github.com/php/php-src/blob/master/Zend/zend_types.h#L1299). Moreover `nr_php_arg_release` is NULL-safe so if `arg2` or `arg3` is NULL (which is how they are initialized [here](https://github.com/misiekn/newrelic-php-agent/blob/fix/nr_drupal_http_request_exec/agent/fw_drupal.c#L143)), it'll still work.

zsistla (approved):
Great job finding and fixing this extremely hard to find bug!

PR #541, opened 2022-09-29
refactor(agent): Wraprec functionality no longer changes or relies on the op_array reserved fields.

zsistla:
Changes:
1) Added new function to get wraprec by name (vs by status of reserved field) 
2) Updated calling functions to use the new functionality. 
3) Used existing list to continue to keep all wraprecs. 

Testing:
1) made this the main wraprec used by our current instrumentation and ran all integration and unit tests with it.  Didn't check the performance, but it instrumented as expected.
2) Using this w/our current instrumentation, ran laravel 8 + php 8.1 and all transactions were properly named. 
3) Updated unit tests to verify the new get_wraprec_by_name functionality.


zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

agent/tests/test_user_instrument.c
```diff
@@ -52,14 +52,124 @@ static void test_op_array_wraprec(TSRMLS_D) {
   tlib_php_request_end();
 }
 
+static void test_get_wraprec_by_name() {
+#if ZEND_MODULE_API_NO < ZEND_7_3_X_API_NO
```
mfulb, line 56: Elsewhere it seems `nr_php_get_wraprec_by_name()` is used for PHP 8.0 and newer - is there a reason this test is run for PHP 7.4?

agent/tests/test_user_instrument.c
```diff
@@ -52,14 +52,124 @@ static void test_op_array_wraprec(TSRMLS_D) {
...
+  /*
+   * NULL if function name is NULL.
+   */
+  tlib_pass_if_ptr_equal("obtain instrumented function",
+                         nr_php_get_wraprec_by_name(&zend_func), NULL);
+
+  /*
+   * NULL if name is correct but type is wrong.
+   */
+  zend_func.common.function_name
```
mfulb, line 96: Maybe set type explicitly for zend_func (like below when set to user function) for clarity and to improve test robustness?

agent/tests/test_user_instrument.c
zsistla: We will only use this for oapi (so php 8.0+), but the functionality is valid for other PHPs, so the logic was if it works for other PHPs, why not test for it.

mfulb:
In fw_drupal8.c in the function `nr_drupal8_add_method_callback()` there is a call to `nr_php_op_array_get_wraprec()` - does this need to be handled differently with OAPI?

agent/php_user_instrument.c
```diff
@@ -435,6 +552,10 @@ void nr_php_destroy_user_wrap_records(void) {
  */
 nruserfn_t* nr_wrapped_user_functions = 0;
 
+/*
+ * nr_php_user_function_add_declared_callback is ONLY called from drupal for
+ * PHP < 7.3.  It does not need to be adjusted for OAPI.
+ */
```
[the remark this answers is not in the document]
zsistla: Theoretically yes, but we still need to determine how we will handle changing the drupal header args for OAPI so we can leave this functionality in now, pending that PR.  If needed it can be changed in that PR as the change is not really relevant to this PR.  I'll add it as a note to that ticket.

zsistla:
> In fw_drupal8.c in the function `nr_drupal8_add_method_callback()` there is a call to `nr_php_op_array_get_wraprec()` - does this need to be handled differently with OAPI?

Not currently because OAPI is PHP 8+ so it currently won't be called at all.

agent/tests/test_user_instrument.c
zsistla: See commit [2b4a42a](https://github.com/newrelic/newrelic-php-agent/pull/541/commits/2b4a42aaa23d1c19063cdbeae013c62131518337)

zsistla:
ok jenkins

mfulb (approved):

ZNeumann (approved):

agent/php_user_instrument.c
```diff
@@ -47,11 +47,15 @@
...
+ *
+ * Note: The agent ONLY needs to call this if it has overwritten the original
+ * zend_execute_ex.
  */
 int nr_zend_call_orig_execute(NR_EXECUTE_PROTO TSRMLS_DC) {
   volatile int zcaught = 0;
   zend_try {
-    NR_PHP_PROCESS_GLOBALS(orig_execute)(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+    NR_PHP_PROCESS_GLOBALS(orig_execute)
+    (NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
misiekn, line 58: Why was this reformatted? The line wasn't exceeding `ColumnLimit:     80` defined in [.clang-format](https://github.com/newrelic/newrelic-php-agent/blob/main/.clang-format#L39). I'm sorry for the nit but all formatting changes outside logic changes are very distracting.

agent/php_user_instrument.c
```diff
@@ -88,10 +93,64 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
  * (or can't) match by name and have the zend_function available. (The main use
  * case for this is to allow instrumenting closures, but it's useful anywhere
  * we're dealing with a callable rather than a name.)
+ *
+ * There are two main structures containing `wraprecs`.
+ * `nr_php_op_array_set_wraprec` is a list of pointers to `wraprecs` that will
```
misiekn, line 98: Is the name `nr_php_op_array_set_wraprec` correct? Isn't it the name of the function that stores an index into `user_function_wrappers` vector defined here: [agent/php_newrelic.h:487](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_newrelic.h#L487)? Aren't `user_function_wrappers` vector and `nr_wrapped_user_functions` list (defined in [agent/php_user_instrument.c:436](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_user_instrument.c#L436)) the two structures that hold information about `wraprecs`?

zsistla: Good catch, thanks for reading!

agent/php_user_instrument.c
```diff
@@ -67,7 +70,8 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
       wraprec->special_instrumentation(wraprec, segment,
                                        NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
     } else {
-      NR_PHP_PROCESS_GLOBALS(orig_execute)(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+      NR_PHP_PROCESS_GLOBALS(orig_execute)
+      (NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
misiekn, line 74: [This comment](https://github.com/newrelic/newrelic-php-agent/pull/541#discussion_r988323293) applies here too.

agent/php_user_instrument.c
zsistla: Oh, I see now, the parameter modified to be longer so (in another PR) so clang split, but the change didn't exist yet in this PR so it was removed, but clang didn't put it back on the same line.  While resolving conflicts, the longer param went back in so this will be split again.

misiekn: Ok. This makes sense. Thanks for taking the time to explain this!

zsistla:
ok jenkins

PR #551, opened 2022-10-07
feat(agent): Add instrumentation via PHP's Observer API for PHP 8+.

zsistla:
Changes:
1) added `nr_php_wraprec_matches` helper function
2) added `txn_start_time` and `wraprec` to segments to help oapi pass and verify the segment for starting/ending stacked segments and so we only make the call to get_wraprec_by_name once (in the begin function handling and not again in the end function handling)
3) added `reportedclass` to `nruserfn_t` to account for functions existing in one class table while reporting they belong to another class (details commented in code).
5) added instrumentation functionality to oapi function handlers
6) updated `test_php_stacked_segment` to handle the changes to the stacked segments.
7) added `special_instrumentation_before` to `nruserfn_t` to account for functions having special instrumentation before and after a zend_function ends.
8) added `nr_php_wrap_user_function_before_after` to set the special_instrumentation_before callback and `nr_zend_call_oapi_special_before
` to call it.
9) ensure the allocated stack segments get freed.
10) added `nr_php_instrument_func_end` and `nr_php_instrument_func_begin` functions to encapsulate the instrumentation functionalities.
11) Updated unit tests to incorporate changes.
12) Changes from #523 (https://github.com/newrelic/newrelic-php-agent/pull/523/commits/48570dd14ba1d8b3d37d326882a45c975a93d2f9 in https://github.com/newrelic/newrelic-php-agent/pull/551/commits/48570dd14ba1d8b3d37d326882a45c975a93d2f9, https://github.com/newrelic/newrelic-php-agent/pull/523/commits/8a32fbcac6fad4a65a5b7c1f486cd75d24b05dd4 in https://github.com/newrelic/newrelic-php-agent/pull/551/commits/c16a8a9f20b8c99207bf353f52431e45d84345b0 and https://github.com/newrelic/newrelic-php-agent/pull/523/commits/c16a8a9f20b8c99207bf353f52431e45d84345b0 in https://github.com/newrelic/newrelic-php-agent/pull/551/commits/c16a8a9f20b8c99207bf353f52431e45d84345b0; this change https://github.com/newrelic/newrelic-php-agent/pull/523/commits/abee421bb9bfab2f7798befbbee4a4d978d57778 is in https://github.com/newrelic/newrelic-php-agent/pull/551/commits/b7a27d9e4437ccf29b56b5839d3876769e73dc17).

Things not done:
1) fix to make drupal_http_request work - handled in another ticket.  The Drupal unit tests are disabled for now so that all the other unit tests can run.  test_fw_drupal needs to be re-enabled!
2) remaining integration tests

Notes:
1) all unit and valgrind tests (aside from test_fw_drupal) pass.
2) Some integration tests do not pass.
3) Contains the same basic logic flow as php_execute + php_execute_enabled but split between the begin/end functions.
4) Based off of oapi branch and all other open PRs (aside from the cufa one) so please look at the oct 7 commit.


agent/php_agent.h
```diff
@@ -849,6 +849,10 @@ extern void nr_php_txn_add_code_level_metrics(
     const nr_php_execute_metadata_t* metadata);
 
 #if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP7+ */
+
+#define NR_ZEND_USER_FUNC_EXISTS(x) \
```
mfulb, line 853: Maybe add parentheses around 'x' in this macro to avoid any surprises when expanding?

agent/php_execute.c
```diff
@@ -1017,9 +1019,22 @@ static void nr_php_execute_metadata_add_code_level_metrics(
...
+
   namespace = nr_php_zend_execute_data_scope_name(execute_data);
+  CHK_CLM_STRLEN(namespace)
+
   function = nr_php_zend_execute_data_function_name(execute_data);
+  CHK_CLM_STRLEN(function)
+
+#undef CHK_CLM_STRLEN
+
   lineno = nr_php_zend_execute_data_lineno(execute_data);
```
mfulb, line 1038: Is this line required - it seems in the conditional below this is called if necessary?

agent/php_execute.c
```diff
@@ -1036,18 +1051,31 @@ static void nr_php_execute_metadata_add_code_level_metrics(
...
+      /*
+       * If the metadata was already set (in the case of OAPI where we need to
+       * get it preemptively, use that value for lineno instead.
+       */
+      lineno = metadata->function_lineno;
+    }
   }
-  if (nr_strempty(function)) {
+
+#define CHK_CLM_EMPTY(s) ((NULL == s || nr_strempty(s)) ? true : false)
```
mfulb, line 1065: Seems like a good utility function to have if nr_strempty() isn't reliable for this test.

agent/php_execute.c
```diff
@@ -1654,10 +1904,36 @@ void nr_php_observer_fcall_begin(zend_execute_data* execute_data) {
...
    */
+  zval* func_return_value = NULL;
+  if (nrunlikely(NULL == execute_data)) {
+    return;
+  }
+
+  NRPRG(php_cur_stack_depth) += 1;
 
-  if (NULL == execute_data) {
+  if (((int)NRINI(max_nesting_level) > 0)
```
mfulb, line 1914: Need to reverse comparison per style guide.
```suggestion
  if (0 < ((int)NRINI(max_nesting_level))
```

agent/tests/test_api_internal.c
```diff
@@ -56,13 +56,14 @@ static void test_invalid_parameters(TSRMLS_D) {
 
   /* Literally any parameter should cause this to bail. */
 #ifdef PHP8
```
mfulb, line 58: Maybe a good opportunity to replace a "PHP8" macro with the preferred API number macro(s)?

agent/tests/test_fw_drupal.c
```diff
@@ -370,6 +370,16 @@ static void test_drupal_http_request_drupal_6(TSRMLS_D) {
...
 
 void test_main(void* p NRUNUSED) {
+/*
+DO NOT LEAVE THIS.
+When drupal_http_request header functionality is refactored,
+please put the tests back in!
+*/
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA /* PHP 8.0+ and OAPI */
+  return;
```
mfulb, line 380: Can this generate a skipped message to reinforce it needs to be updated later?

agent/php_txn.c
```diff
@@ -911,7 +911,8 @@ nr_status_t nr_php_txn_begin(const char* appnames,
     nr_php_txn_log_error_dt_on_tt_off();
   }
 
-#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO \
```
bduranleau-nr, line 914: Is this functionality unique to 8.1 and not applicable to 8.0?

agent/php_user_instrument.c
```diff
@@ -138,17 +155,11 @@ static void nr_php_wrap_zend_function(zend_function* func,
...
-     * Prevent future wrap attempts for performance and to prevent spamming the
-     * logs with this message.
-     */
-    wraprec->is_disabled = 1;
-    return;
+  if ((NULL != func) && (NULL != func->common.scope)) {
+    if ((NULL != func->common.scope->name)
+        && (NULL == wraprec->reportedclass)) {
+      wraprec->reportedclass = nr_strdup(ZSTR_VAL(func->common.scope->name));
+    }
```
bduranleau-nr, line 162: suggestion for abstraction to improve readability:
```
static bool chk_reported_class(zend_function* func, nruserfn_t* wraprec) {
  if ((NULL == func) || (NULL == func->common.scope)) {
    return false;
  }

  if ((NULL == func->common.scope->name) || (NULL != wraprec->reportedclass)) {
    return false;
  }
  return true;
}
```
```suggestion
  if (chk_reported_class(func, wraprec)) {
    wraprec->reportedclass = nr_strdup(ZSTR_VAL(func->common.scope->name));
  }
```

agent/php_user_instrument.c
```diff
@@ -62,7 +62,24 @@ int nr_zend_call_orig_execute(NR_EXECUTE_PROTO TSRMLS_DC) {
   zend_end_try();
   return zcaught;
 }
-
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP8+ */
+int nr_zend_call_oapi_special_before(nruserfn_t* wraprec,
+                                     nr_segment_t* segment,
+                                     NR_EXECUTE_PROTO TSRMLS_DC) {
```
bduranleau-nr, line 68: `TSRMLS_*` no longer necessary in PHP7+
```suggestion
                                     NR_EXECUTE_PROTO) {
```

agent/php_user_instrument.h
```diff
@@ -171,7 +208,11 @@ extern int nr_zend_call_orig_execute(NR_EXECUTE_PROTO TSRMLS_DC);
 extern int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
                                              nr_segment_t* segment,
                                              NR_EXECUTE_PROTO TSRMLS_DC);
-
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP8+ */
+extern int nr_zend_call_oapi_special_before(nruserfn_t* wraprec,
+                                            nr_segment_t* segment,
+                                            NR_EXECUTE_PROTO TSRMLS_DC);
```
bduranleau-nr, line 214: `TSRMLS_*` no longer necessary in PHP7+
```suggestion
                                            NR_EXECUTE_PROTO);
```

agent/php_wrapper.c
```diff
@@ -8,6 +8,43 @@
...
+        nrl_verbosedebug(NRL_INSTRUMENT,
+                         "%s: attempting to set special_instrumentation_before "
+                         "for %.*s, but "
+                         "it is already set",
+                         __func__, NRSAFELEN(namelen), NRBLANKSTR(name));
+      } else {
+        wraprec->special_instrumentation_before = before_callback;
+      }
+    }
+  }
```
bduranleau-nr, line 43: Suggestion to improve readability:
```
static bool chk_special_instrumentation_not_set(nruserfn_t* wraprec,
                                                nrspecialfn_t callback) {
  if ((NULL != wraprec->special_instrumentation)
      && (callback != wraprec->special_instrumentation)) {
    return false;
  }
  return true;
}
```

```suggestion
  if (NULL == wraprec) {
    return wraprec;
  }
  if (after_callback) {
    if (!chk_special_instrumentation_not_set(wraprec, after_callback)) {
      nrl_verbosedebug(
          NRL_INSTRUMENT,
          "%s: attempting to set special_instrumentation for %.*s, but "
          "it is already set",
          __func__, NRSAFELEN(namelen), NRBLANKSTR(name));
    } else {
      wraprec->special_instrumentation = after_callback;
    }
  }
  if (before_callback) {
    if (!chk_special_instrumentation_not_set(wraprec, before_callback)) {
      nrl_verbosedebug(NRL_INSTRUMENT,
                       "%s: attempting to set special_instrumentation_before "
                       "for %.*s, but "
                       "it is already set",
                       __func__, NRSAFELEN(namelen), NRBLANKSTR(name));
    } else {
      wraprec->special_instrumentation_before = before_callback;
    }
  }
```
abstract out special instrumentation check, unwrap nested if statement and return immediately if `NULL == wraprec`

agent/tests/test_api_internal.c
```diff
@@ -56,13 +56,14 @@ static void test_invalid_parameters(TSRMLS_D) {
...
-                        "    $exception = true;"
-                        "}" TSRMLS_CC);
+  tlib_php_request_eval(
+      "$exception = false;"
+      "try {"
+      "    $value = newrelic_get_trace_json('invalid');"
+      "    echo \"No exception, returned \" . $value . \".\\n\";"
+      "} catch(ArgumentCountError $_e) {"
+      "    $exception = true;"
+      "}" TSRMLS_CC);
```
bduranleau-nr, line 66:
```suggestion
      "}");
```

agent/php_txn.c
zsistla: Yes, it was special for 8.1

agent/php_user_instrument.c
zsistla: Yes, more readable...but going back to your point about every additional function call adds to performance overhead since this gets called/executed very often.

agent/php_wrapper.c
zsistla: Thanks for pointing this out since there's a bug there (fixed in [324dbba](https://github.com/newrelic/newrelic-php-agent/pull/551/commits/324dbbaec646d5ada5c4fe79eacf30e7de8915a5)), both calls shouldn't be checking the same wraprec value. Both calls would need to check different values the function wouldn't work for both.  I think in this case it's okay since function itself is basically a check for if it is set.  It's a pretty small function, and as you mentioned, we're wanting to consider reducing function calls not add more.

agent/tests/test_fw_drupal.c
zsistla: It's noted in the ticket already.

agent/php_wrapper.c
bduranleau-nr: we can either inline it or turn it into a macro in that case

agent/php_user_instrument.c
zsistla: Sounds good, can you add this?

agent/php_wrapper.c
zsistla: inline sounds fine.  Can you add that?

bduranleau-nr: added in 136249c70c332aeac014bcccf65b3a003e5ed973

agent/php_user_instrument.c
```diff
@@ -309,6 +321,45 @@ static void nr_php_add_custom_tracer_common(nruserfn_t* wraprec) {
...
+
+  if (0 == nr_stricmp(p->funcnameLC, ZSTR_VAL(func->common.function_name))) {
+    if (NULL != func->common.scope && NULL != func->common.scope->name) {
+      klass = ZSTR_VAL(func->common.scope->name);
+    }
+    if ((0 == nr_strcmp(p->reportedclass, klass))
+        || (0 == nr_stricmp(p->classname, klass))) {
+      return true;
+    }
+  }
```
bduranleau-nr, line 358: un-nest `if`s
```suggestion
  if (0 != nr_stricmp(p->funcnameLC, ZSTR_VAL(func->common.function_name))) {
    return false;
  }
  if (NULL != func->common.scope && NULL != func->common.scope->name) {
    klass = ZSTR_VAL(func->common.scope->name);
  }
  if ((0 == nr_strcmp(p->reportedclass, klass))
      || (0 == nr_stricmp(p->classname, klass))) {
    return true;
  }
```

agent/php_user_instrument.c
```diff
@@ -309,6 +321,47 @@ static void nr_php_add_custom_tracer_common(nruserfn_t* wraprec) {
...
+    return false;
+  }
+
+  if (0 != nr_stricmp(p->funcnameLC, ZSTR_VAL(func->common.function_name))) {
+    return false;
+  }
+  if (NULL == func->common.scope || NULL == func->common.scope->name) {
+    return false;
+  }
+  klass = ZSTR_VAL(func->common.scope->name);
```
bduranleau-nr, line 356: fix: apparently `NULL` and `NULL` are valid comparisons in nr_strcmp
```suggestion
  if (NULL != func->common.scope && NULL != func->common.scope->name) {
    klass = ZSTR_VAL(func->common.scope->name);
  }
```

agent/php_execute.c
misiekn: `nr_strempty` is NULL-safe - the first thing it checks is if its argument is NULL. I'd also like to understand why this extra `NULL` check is done before calling NULL-safe function.

misiekn (approved):
Great job! Thanks putting this together! 👍 

An issue with with `NRPRG(php_cur_stack_depth)` was identified and addressed with [4629438](https://github.com/zsistla/newrelic-php-agent/commit/46294383a876e66b085933c43ae0d24bf0ddeb5f).

The below remarks are minor/cosmetic - they don't need to be addressed/resolved in this PR before it is merged. Subsequent PRs (if needed) can be created to address/resolve them.

1. It would be good to update function's doc strings in `php_stacked_segment.h` to indicate that segment is allocated on the heap. It would also be good to update the comment in `php_stacked_segment.h` to explain why stacked segments (stacked as in segment struct is allocated on stack and segments are stacked one on top of another to represent call stac) became stacked heap segments (still stacked but no longer allocated on stack) when instrumenting with PHP's Observer API.
2. It seems like a good idea to keep the toggle to switch between PHP's Observer API and legacy instrumentation method. However to make the code more readable the
```
```
could become something like
```
#if OAPI_ENABLED
```
3. It would be good to explain why `nr_php_max_nesting_level_reached` is still called/used  when instrumenting via PHP's Observer API. PHP's Observer API eliminates the problem of call stack becoming too deep.
4. Some code could be refactored for performance and maintainability, e.g. call to `nr_zend_call_orig_execute_special` [here](https://github.com/zsistla/newrelic-php-agent/blob/oapi_instrumentation/agent/php_execute.c#L1853) is basically a two call stack frames resulting in no-op or return value of [this](https://github.com/zsistla/newrelic-php-agent/blob/oapi_instrumentation/agent/php_execute.c#L1038) call to `nr_php_zend_execute_data_lineno` is ignored - `lineno` is set in between line [1043](https://github.com/zsistla/newrelic-php-agent/blob/oapi_instrumentation/agent/php_execute.c#L1043) and line [1063](https://github.com/zsistla/newrelic-php-agent/blob/oapi_instrumentation/agent/php_execute.c#L1063) (lines 1043 through 1063 could become a standalone, testable function).

ZNeumann (approved):

bduranleau-nr (approved):

PR #552, opened 2022-10-12
fix(agent): instrument `drupal_http_request`

misiekn:
Fix `drupal_http_request` instrumentation for PHP 8.0+ by using Observer
APIs `before` callback to add New Relic headers and `after` callback
to finalize external segment with metrics. Observer API instrumentation
requires the external request segment to be created in `before` callback
but be available in `after` callback. Therefore it is made a NRPRG global.

### DIFF
--- a/agent/fw_drupal.c
+++ b/agent/fw_drupal.c
@@ -67,6 +67,54 @@ NR_PHP_WRAPPER(nr_drupal_qdrupal_name_the_wt) {
 }
 NR_PHP_WRAPPER_END
 
+#if ZEND_MODULE_API_NO >= ZEND_7_3_X_API_NO
+static void nr_drupal_http_request_ensure_second_arg(NR_EXECUTE_PROTO) {
+  zval* arg = NULL;
+
+  /*
+   * If only one argument is given, an empty list is inserted as second
+   * argument. NR headers are added during a later step.
+   */
+  if (ZEND_NUM_ARGS() == 1) {
+    arg = nr_php_zval_alloc();
+    array_init(arg);
+
+    nr_php_arg_add(NR_EXECUTE_ORIG_ARGS, arg);
+
+    nr_php_zval_free(&arg);
+  }
+}
+
+static zval* nr_drupal_http_request_add_headers(NR_EXECUTE_PROTO TSRMLS_DC) {
+  bool is_drupal_7;
+  zval* second_arg = NULL;
+  zend_execute_data* ex = nr_get_zend_execute_data(NR_EXECUTE_ORIG_ARGS);
+
+  if (nrunlikely(NULL == ex)) {
+    return NULL;
+  }
+
+  /* Ensure second argument exists in the call frame */
+  nr_drupal_http_request_ensure_second_arg(NR_EXECUTE_ORIG_ARGS);
+
+  is_drupal_7 = (ex->func->common.num_args == 2);
+
+  /*
+   * nr_php_get_user_func_arg is used, as nr_php_arg_get calls ZVAL_DUP
+   * on the argument zval and thus doesn't allow us to change the
+   * original argument.
+   */
+  second_arg = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
+
+  /*
+   * Add NR headers.
+   */
+  nr_drupal_headers_add(second_arg, is_drupal_7 TSRMLS_CC);
+
+  return second_arg;
+}
+#endif
+
 static uint64_t nr_drupal_http_request_get_response_code(
     zval** return_value TSRMLS_DC) {
   zval* code = NULL;
@@ -143,40 +191,7 @@ NR_PHP_WRAPPER(nr_drupal_http_request_before) {
    * (nr_php_swap_user_functions), which breaks as since PHP 7.3 user
    * functions are stored in shared memory.
    */
-  bool is_drupal_7;
-  zval* arg = NULL;
-  zend_execute_data* ex = nr_get_zend_execute_data(NR_EXECUTE_ORIG_ARGS);
-
-  if (NULL == ex) {
-    NR_PHP_WRAPPER_LEAVE
-  }
-
-  is_drupal_7 = (ex->func->common.num_args == 2);
-
-  /*
-   * If only one argument is given, an empty list is inserted as second
-   * argument. NR headers are added during a later step.
-   */
-  if (ZEND_NUM_ARGS() == 1) {
-    arg = nr_php_zval_alloc();
-    array_init(arg);
-
-    nr_php_arg_add(NR_EXECUTE_ORIG_ARGS, arg);
-
-    nr_php_zval_free(&arg);
-  }
-
-  /*
-   * nr_php_get_user_func_arg is used, as nr_php_arg_get calls ZVAL_DUP
-   * on the argument zval and thus doesn't allow us to change the
-   * original argument.
-   */
-  arg = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
-
-  /*
-   * Add NR headers.
-   */
-  nr_drupal_headers_add(arg, is_drupal_7 TSRMLS_CC);
+  nr_drupal_http_request_add_headers(NR_EXECUTE_ORIG_ARGS);
 
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL);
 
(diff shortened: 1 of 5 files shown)


agent/fw_drupal.c
```diff
@@ -67,6 +67,90 @@ NR_PHP_WRAPPER(nr_drupal_qdrupal_name_the_wt) {
...
+  zval* method = NULL;
+  char* http_request_method = NULL;
+
+  /*
+   * Drupal 6 will have a third argument with the method, Drupal 7 will not
+   * have a third argument it must be parsed from the second.
+   */
+  arg3 = nr_php_arg_get(3, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  // There is no third arg, this is drupal 7
+  if (0 == arg3) {
```
mfulb, line 130: Use NULL for comparison?
```suggestion
  if (NULL == arg3) {
```

agent/fw_drupal.c
```diff
@@ -155,40 +227,108 @@ NR_PHP_WRAPPER(nr_drupal_http_request_exec) {
...
-  zend_execute_data* ex = nr_get_zend_execute_data(NR_EXECUTE_ORIG_ARGS);
+  nr_drupal_http_request_add_headers(NR_EXECUTE_ORIG_ARGS);
 
-  if (NULL == ex) {
-    NR_PHP_WRAPPER_LEAVE
-  }
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL);
 
-  is_drupal_7 = (ex->func->common.num_args == 2);
+  NRPRG(drupal_http_request_depth) += 1;
```
mfulb, line 234: Is this something we might want to flag with a comment that will be affected by fibers support?

agent/php_newrelic.h
```diff
@@ -409,7 +409,10 @@ size_t drupal_module_invoke_all_hook_len; /* The length of the current Drupal
                                              hook */
 size_t drupal_http_request_depth; /* The current depth of drupal_http_request()
                                      calls */
-
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+nr_segment_t* drupal_http_request_segment;
```
mfulb, line 414: Does this need to be initialized anywhere or is it assumed to be set to zero by the compiler because of the global scope?

agent/fw_drupal.c
misiekn: Will do - I'll wait for remaining feedback to commit in a batch. Just a word of explanation - this refactoring was achieved literally with nothing more than cut and paste. No line of code got modified between cut and paste :)

agent/fw_drupal.c
misiekn: Will a comment `not thread/fiber safe` [here](https://github.com/misiekn/newrelic-php-agent/blob/oapi/drupal_http_request/agent/php_newrelic.h#L410) be ok?

agent/php_newrelic.h
misiekn: It doesn't need to be initialized to anything other than `NULL` which is handled by the complier because of global scope. The segment assigned to `drupal_http_request_segment` is created [here](https://github.com/misiekn/newrelic-php-agent/blob/oapi/drupal_http_request/agent/fw_drupal.c#L243) and unassigned [here](https://github.com/misiekn/newrelic-php-agent/blob/oapi/drupal_http_request/agent/fw_drupal.c#L291).

agent/fw_drupal.c
mfulb: Sure - although on second thought maybe just the NRPRG() macro is enough of a signature for us to l look for in the future?

misiekn: I agree with this statement - NRPRG() macro is enough of a signature for us to l look for when adding fibers support. Any global state for that matter.

agent/fw_drupal.c
```diff
@@ -67,6 +67,90 @@ NR_PHP_WRAPPER(nr_drupal_qdrupal_name_the_wt) {
 }
 NR_PHP_WRAPPER_END
 
+#if ZEND_MODULE_API_NO >= ZEND_7_3_X_API_NO
```
bduranleau-nr, line 70: Is this not applicable to PHP versions 7.0 -> 7.2?

agent/fw_drupal.c
```diff
@@ -67,6 +67,90 @@ NR_PHP_WRAPPER(nr_drupal_qdrupal_name_the_wt) {
...
+  nr_drupal_http_request_ensure_second_arg(NR_EXECUTE_ORIG_ARGS);
+
+  is_drupal_7 = (ex->func->common.num_args == 2);
+
+  /*
+   * nr_php_get_user_func_arg is used, as nr_php_arg_get calls ZVAL_DUP
+   * on the argument zval and thus doesn't allow us to change the
+   * original argument.
+   */
+  second_arg = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
```
bduranleau-nr, line 107: Could check if this value is `NULL` and return immediately without having to call `nr_drupal_headers_add`

bduranleau-nr (approved):
👍

agent/fw_drupal.c
```diff
@@ -155,40 +227,108 @@ NR_PHP_WRAPPER(nr_drupal_http_request_exec) {
...
+  arg1 = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  if (0 == nr_php_is_zval_non_empty_string(arg1)) {
+    goto end;
   }
 
   /*
-   * nr_php_get_user_func_arg is used, as nr_php_arg_get calls ZVAL_DUP
-   * on the argument zval and thus doesn't allow us to change the
-   * original argument.
+   * We only want to create a metric here if this isn't a recursive call to
```
zsistla, line 266: What happens if it is a recursive call based on a redirect?

agent/php_newrelic.h
zsistla: Should we doublecheck that this got freed in rshutdown?
Maybe something here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_rshutdown.c#L92
Since the segment start and stop are in different functions now we have to be prepared for something happening in between.

zsistla:
Thanks for debugging this issue and getting the fix up :)

agent/fw_drupal.c
misiekn: I didn't change the original implementation which didn't check for `NULL` before calling `nr_drupal_headers_add`. This was pure copy/paste re-factoring to re-use code in pre-oapi and oapi instrumentation. It is highly unlikely that `second_arg` will be `NULL` so I think for code clarity this can be left as is.

agent/fw_drupal.c
misiekn: The global `NRPRG(drupal_http_request_segment)` segment remains unchanged if it is a recursive call based on a redirect. `NRPRG(drupal_http_request_segment)` segment is only created by the first call to `drupal_http_request`  - see [agent/fw_drupal.c:242](https://github.com/misiekn/newrelic-php-agent/blob/oapi/drupal_http_request/agent/fw_drupal.c#L242). It is only destroyed when the first call to `drupal_http_request` returns - see [agent/fw_drupal.c:291](https://github.com/misiekn/newrelic-php-agent/blob/oapi/drupal_http_request/agent/fw_drupal.c#L291). It was better visible in pre-oapi instrumentation, because everything happened within the scope of a single function - `drupal_http_request_exec` - and the segment was local to that function. There was no need to maintain state before and after call to `drupal_http_request`. With oapi the management of external segment must be split between before and after callbacks and it's easier to lose track of what is going on. For now `drupal_http_request` oapi instrumentation uses a simple solution of maintaining external segment as a global variable. This will probably have to be revisited when fibers support is added to the agent.

zsistla: Got it, thanks for the explanation :)

agent/php_newrelic.h
misiekn: `drupal_http_request_segment` points to memory chunk allocated by slab allocator (`nr_segment_start` calls [nr_txn_allocate_segment](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_segment.c#L179) which in turn calls [nr_slab_next](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_txn.h#L1138)). It is guaranteed to be freed by slab allocator [here](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_txn.c#L1237) (as part of  `nr_txn_destroy_fields` which is called from within `nr_txn_destroy` which is called by `nr_php_txn_end` called in `nr_php_post_deactivate`). I can add setting `drupal_http_request_segment` to NULL in `nr_php_post_deactivate`.

ZNeumann (approved):

agent/php_newrelic.h
misiekn: @mfulb @zsistla I tried to address these concerns with c30a987486a5bc2ea8e6c28fd470be5d593874d7 and e4128608b6df37f1892e6e5ef505fcaff8c0d1f4. Is this acceptable?

mfulb: The "exception path" in e412860 would be if the segment was not kicked from the segment minheap and persists to the end of the request?

misiekn: > The "exception path" in https://github.com/newrelic/newrelic-php-agent/commit/e4128608b6df37f1892e6e5ef505fcaff8c0d1f4 would be if the segment was not kicked from the segment minheap and persists to the end of the request?

The 'exception path' covers anything that does not fall under the 'happy path' category, i.e. segment ended in the _after_ callback. The change in e4128608b6df37f1892e6e5ef505fcaff8c0d1f4 does not ensure that the memory pointed by `drupal_http_request_segment` is freed - this is guaranteed to be freed by slab allocator [here](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_txn.c#L1237). The change in e4128608b6df37f1892e6e5ef505fcaff8c0d1f4 only ensures that at the end of the request, the `drupal_http_request_segment` pointer will no longer point to anything. In my tests `drupal_http_request_segment` pointer had non-NULL value regardless of the outcome of `drupal_http_request` (successful execution or execution ended by exception), and regardless if external segment was or was not moved to min-max heap (moving the segment to the min-max heap should invalidate the pointer pointing to the segment but it is not happening). This change is last resort failsafe to ensure accessing `drupal_http_request_segment` outside the scope of `nr_drupal_http_request_before` and `nr_drupal_http_request_after` callbacks will not cause unexpected memory corruption.

FYI: I tested `drupal_http_request` execution with and without exception as well as nested calls to `drupal_http_request` caused by redirect.

zsistla (approved):
Thanks, LGTM!

PR #554, opened 2022-10-17
fix(agent): detect framework in oapi's fcall_begin

misiekn:
Framework detection (`nr_execute_handle_framework`) not only adds `wraprecs`
but also sets `NRPRG(current_framework)`. It is important to set it before
function is called, because its value affects other instrumentation, e.g.
when datastore segment for SQL operation is ended, a table name modify
function (`nr_php_modify_table_name_fn`) to shorten table name is selected
based on `NRPRG(current_framework)`.

This fixes `frameworks/magento/test_temp_tables.php` and
`frameworks/wordpress/test_site_specific_tables.php` integration tests.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1692,11 +1692,10 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
 
   NRTXNGLOBAL(execute_count) += 1;
 
-  /*
-   * Wait to do this handling in the end function handler when all the files
-   * have been loaded; otherwise, the classes might not be loaded yet.
-   */
   if (nrunlikely(OP_ARRAY_IS_A_FILE(NR_OP_ARRAY))) {
+    const char* filename = nr_php_op_array_file_name(NR_OP_ARRAY);
+    nr_execute_handle_framework(all_frameworks, num_all_frameworks,
+                              filename TSRMLS_CC);
     return;
   }
   wraprec = nr_php_get_wraprec_by_name(execute_data->func);


agent/php_execute.c
```diff
@@ -1692,11 +1692,10 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
...
   NRTXNGLOBAL(execute_count) += 1;
 
-  /*
-   * Wait to do this handling in the end function handler when all the files
-   * have been loaded; otherwise, the classes might not be loaded yet.
-   */
   if (nrunlikely(OP_ARRAY_IS_A_FILE(NR_OP_ARRAY))) {
+    const char* filename = nr_php_op_array_file_name(NR_OP_ARRAY);
+    nr_execute_handle_framework(all_frameworks, num_all_frameworks,
+                              filename TSRMLS_CC);
```
zsistla, line 1698: This means handle_framework is now getting called twice.  Here in the begin_func and [here in the end_func](https://github.com/newrelic/newrelic-php-agent/pull/554/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1749) when it calls `nr_php_execute_file` https://github.com/misiekn/newrelic-php-agent/blob/oapi/integration-tests/agent/php_execute.c#L925.

I had initially moved from begin_func to end_func because some wraprecs weren't getting properly set until the file and any included classes were fully loaded.
This is probably okay to call twice because it depends on the file being called for detection and it looks like some files need earlier handling and some prefer later handling.  For instance, if you took out the call in end_func, other integration tests would fail.

However, the libraries and logging detections are still only being loaded in the end_func.  Does only framework detection need to be called in begin_func because of the special way we detect magento and wordpress?

bduranleau-nr (approved):

agent/php_execute.c
zsistla: The reported class is because:
```
 /*
   * Internally, there are cases where the zend_function reports it is one
   * class; however, the zend_function is also contained in another class_entry
   * table.
   * For an example, see tests/integration/laravel
   * A lookup for the class = Illuminate\Console\Application returns the class
   * entry named classname = Illuminate\Console\Application
   * So far so good!
   * Lookup Illuminate\Console\Application method doRun and a zend_function is
   * returned.  Ask that zend_func what its classname is and it says:
   * Symfony\Component\Console\Application instead of Illuminate\Console\Application
   * Okay.
   * Track both class infos for any wraprecs.
   */
```
Additionally now we need to lookup the function to get the lineno so we still need to have the functionality of the function lookup.

zsistla (approved):

ZNeumann (approved):

PR #555, opened 2022-10-17
feat(agent): implement php_execute_show functionality for OAPI

bduranleau-nr:
#### Add support for nr_php_show_exec* family of functions for Observer API.
- `begin` fcall will only show `execute`, since it does not have access to return data.
- `end` fcall will only show `return`, since `execute` will be handled solely by `begin`

Several blocks of functions have been wrapped in 
```c
#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
```
statements in order to placate the compiler, since they are defined but never called under the OAPI paradigm.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1928,7 +1928,7 @@ void nr_php_observer_fcall_begin(zend_execute_data* execute_data) {
   int show_executes = NR_PHP_PROCESS_GLOBALS(special_flags).show_executes;
 
   if (nrunlikely(show_executes)) {
-    nr_php_show_exec(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+    nr_php_show_exec(NR_EXECUTE_ORIG_ARGS);
   }
   nr_php_instrument_func_begin(NR_EXECUTE_ORIG_ARGS);
 
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1928,10 +1928,6 @@ void nr_php_observer_fcall_begin(zend_execute_data* execute_data) {
   int show_executes = NR_PHP_PROCESS_GLOBALS(special_flags).show_executes;
 
   if (nrunlikely(show_executes)) {
-    /*
-     * For OAPI don't call nr_php_execute_enabled from execute_show.
-     * Can show execute but CANNOT show returns here.
-     */
     nr_php_show_exec(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
   }
   nr_php_instrument_func_begin(NR_EXECUTE_ORIG_ARGS);
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1227,6 +1227,8 @@ static inline void nr_php_execute_segment_end(
   }
 }
 
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
+    || defined OVERWRITE_ZEND_EXECUTE_DATA
 /*
  * This is the user function execution hook. Hook the user-defined (PHP)
  * function execution. For speed, we have a pointer that we've installed in the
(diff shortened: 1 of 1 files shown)


agent/php_execute.c
```diff
@@ -1913,26 +1918,24 @@ void nr_php_observer_fcall_begin(zend_execute_data* execute_data) {
...
-      // nr_php_execute_show(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
-    }
-    nr_php_instrument_func_begin(NR_EXECUTE_ORIG_ARGS);
+  int show_executes = NR_PHP_PROCESS_GLOBALS(special_flags).show_executes;
+
+  if (nrunlikely(show_executes)) {
+    /*
+     * For OAPI don't call nr_php_execute_enabled from execute_show.
+     * Can show execute but CANNOT show returns here.
+     */
```
misiekn, line 1934: I don't think this comment is needed anymore.

agent/php_execute.c
```diff
@@ -1950,26 +1953,21 @@ void nr_php_observer_fcall_end(zend_execute_data* execute_data,
...
-        || NR_PHP_PROCESS_GLOBALS(special_flags).show_execute_returns;
-
-  if (nrunlikely(show_executes)) {
+  if (nrlikely(1 == nr_php_recording())) {
+    int show_executes_return
+        = NR_PHP_PROCESS_GLOBALS(special_flags).show_execute_returns;
     /*
      * For OAPI don't call nr_php_execute_enabled from execute_show.
-     * nr_php_execute_show updated in another ticket.
      * Can show execute and returns here.
```
misiekn, line 1961: I don't think this comment is needed anymore.

misiekn (approved):
LGTM! Just a small nit about comments before calls to `nr_php_show_exec` and `nr_php_show_exec_return`. I don't think they're needed anymore.

agent/php_execute.c
bduranleau-nr: I liked the idea of keeping them in place as an explanation for why it was done this way as opposed to the old paradigm

misiekn: IMHO it only adds confusion as to why a search for `nr_php_execute_enabled` would find the oapi code where `nr_php_execute_enabled` is not used and expected to appear. `nr_php_execute_enabled` is not oapi way of instrumenting that's been replaced with `nr_php_observer_fcall_begin` and `nr_php_observer_fcall_end`.  Things are done differently in oapi. This comment made sense when `nr_php_observer_fcall_begin` and `nr_php_observer_fcall_end` were just a stubs to indicate what should be implemented in them. Now that the implementation is in place, there's no longer need for this comment.

bduranleau-nr: Can you add a suggestion comment to both so I can commit that change here directly?

misiekn: Unfortunately GH's code review suggestions cannot be made to code that was not modified in PR. So I cannot create suggestion for this comment.

agent/php_execute.c
misiekn:
```suggestion
```

agent/php_execute.c
bduranleau-nr: Updated and pushed

agent/php_execute.c
```diff
@@ -1913,26 +1918,20 @@ void nr_php_observer_fcall_begin(zend_execute_data* execute_data) {
...
-       * nr_php_execute_show updated in another ticket.
-       * Can show execute but CANNOT show returns here.
-       */
-      // nr_php_execute_show(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
-    }
-    nr_php_instrument_func_begin(NR_EXECUTE_ORIG_ARGS);
+  int show_executes = NR_PHP_PROCESS_GLOBALS(special_flags).show_executes;
+
+  if (nrunlikely(show_executes)) {
+    nr_php_show_exec(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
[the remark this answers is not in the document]
bduranleau-nr: Yup, I'll remove it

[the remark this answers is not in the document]
bduranleau-nr:
```suggestion
    nr_php_show_exec(NR_EXECUTE_ORIG_ARGS);
```

zsistla (approved):
Great job, thanks.

mfulb (approved):

PR #557, opened 2022-10-18
fix(agent): name exception handler's wraprec

misiekn:
Add a name to a wraprec for a user function used as exception handler
so that the instrumentation of user's exception handler can be found
with get_wraprec_by_name and an error event can be recorded.

Use `NAME_EXCEPTION_HANDLER` macro to enable/disable
this fix to test things out.


misiekn:
Superseded with #568

PR #558, opened 2022-10-19
refactor(agent): Wraprec refactoring to take advantage of lineno.

zsistla:
1) Added nr_php_zend_function_lineno(zend_function func); 
2) removed lineno from segment (not needed anymore since nr_php_zend_function_lineno(execute_data->func); gives the same lineno whether entering or exiting function 
3) added filename, lineno to wraprec
4) updated/added test cases
5) renamed get_wraprec_by_name to get_wraprec_by_func 6) updated  nr_php_wraprec_matches
 To first check for lineno/filename match.  If lineno/filename are not set (this happens when we are wrapping before all funcs are loaded), defaults to original check, but adds lineno/filename as soon as they are known so we can pop back into the optimized comparison.

### DIFF
--- a/agent/php_agent.h
+++ b/agent/php_agent.h
@@ -912,7 +912,7 @@ extern uint32_t nr_php_zend_execute_data_lineno(
  *
  */
 static inline uint32_t nr_php_zend_function_lineno(const zend_function* func) {
-  if (NULL != func) {
+  if ((NULL != func) && (func->type == ZEND_USER_FUNCTION)) {
     return func->op_array.line_start;
   }
   return 0;
--- a/agent/php_agent.h
+++ b/agent/php_agent.h
@@ -902,6 +902,22 @@ extern const char* nr_php_zend_execute_data_scope_name(
  */
 extern uint32_t nr_php_zend_execute_data_lineno(
     const zend_execute_data* execute_data);
+
+/*
+ * Purpose : Return a uint32_t (zend_uint) line number value of zend_function.
+ *
+ * Params  : 1. zend_function.
+ *
+ * Returns : uint32_t lineno value
+ *
+ */
+static inline uint32_t nr_php_zend_function_lineno(const zend_function* func) {
+  if (NULL != func) {
+    return func->op_array.line_start;
+  }
+  return 0;
+}
+
 #endif /* PHP 7+ */
 
 #endif /* PHP_AGENT_HDR */
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -640,7 +640,7 @@ static void nr_php_show_exec(NR_EXECUTE_PROTO TSRMLS_DC) {
     || defined OVERWRITE_ZEND_EXECUTE_DATA
         nr_php_op_array_get_wraprec(NR_OP_ARRAY TSRMLS_CC) ? " *" : "",
 #else
-        nr_php_get_wraprec_by_name(execute_data->func) ? " *" : "",
+        nr_php_get_wraprec_by_func(execute_data->func) ? " *" : "",
 #endif
         NRP_FILENAME(filename), NR_OP_ARRAY->line_start);
   } else if (NR_OP_ARRAY->function_name) {
@@ -662,7 +662,7 @@ static void nr_php_show_exec(NR_EXECUTE_PROTO TSRMLS_DC) {
     || defined OVERWRITE_ZEND_EXECUTE_DATA
         nr_php_op_array_get_wraprec(NR_OP_ARRAY TSRMLS_CC) ? " *" : "",
 #else
-        nr_php_get_wraprec_by_name(execute_data->func) ? " *" : "",
+        nr_php_get_wraprec_by_func(execute_data->func) ? " *" : "",
 #endif
         NRP_FILENAME(filename), NR_OP_ARRAY->line_start);
   } else if (NR_OP_ARRAY->filename) {
@@ -1035,7 +1035,8 @@ static void nr_php_execute_metadata_add_code_level_metrics(
 
 #undef CHK_CLM_STRLEN
 
-  lineno = nr_php_zend_execute_data_lineno(execute_data);
+  // lineno = nr_php_zend_execute_data_lineno(execute_data);
+  lineno = nr_php_zend_function_lineno(execute_data->func);
 
   /*
    * Check if we are getting CLM for a file.
@@ -1051,15 +1052,7 @@ static void nr_php_execute_metadata_add_code_level_metrics(
     /*
      * We are getting CLM for a function.
      */
-    if (0 == metadata->function_lineno) {
-      lineno = nr_php_zend_execute_data_lineno(execute_data);
-    } else {
-      /*
-       * If the metadata was already set (in the case of OAPI where we need to
-       * get it preemptively, use that value for lineno instead.
-       */
-      lineno = metadata->function_lineno;
-    }
+    lineno = nr_php_zend_function_lineno(execute_data->func);
   }
 
 #define CHK_CLM_EMPTY(s) ((NULL == s || nr_strempty(s)) ? true : false)
@@ -1264,7 +1257,7 @@ static void nr_php_execute_enabled(NR_EXECUTE_PROTO TSRMLS_DC) {
     || defined OVERWRITE_ZEND_EXECUTE_DATA
   wraprec = nr_php_op_array_get_wraprec(NR_OP_ARRAY TSRMLS_CC);
 #else
-  wraprec = nr_php_get_wraprec_by_name(execute_data->func);
+  wraprec = nr_php_get_wraprec_by_func(execute_data->func);
 #endif
 
   if (NULL != wraprec) {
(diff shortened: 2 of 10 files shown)


agent/php_agent.h
```diff
@@ -902,6 +902,22 @@ extern const char* nr_php_zend_execute_data_scope_name(
...
+ * Purpose : Return a uint32_t (zend_uint) line number value of zend_function.
+ *
+ * Params  : 1. zend_function.
+ *
+ * Returns : uint32_t lineno value
+ *
+ */
+static inline uint32_t nr_php_zend_function_lineno(const zend_function* func) {
+  if (NULL != func) {
+    return func->op_array.line_start;
```
mfulb, line 916: Could `op_array` be NULL?

agent/php_execute.c
```diff
@@ -640,7 +640,7 @@ static void nr_php_show_exec(NR_EXECUTE_PROTO TSRMLS_DC) {
     || defined OVERWRITE_ZEND_EXECUTE_DATA
         nr_php_op_array_get_wraprec(NR_OP_ARRAY TSRMLS_CC) ? " *" : "",
 #else
-        nr_php_get_wraprec_by_name(execute_data->func) ? " *" : "",
+        nr_php_get_wraprec_by_func(execute_data->func) ? " *" : "",
```
mfulb, line 643: Is it generally safe to assume `execute_data` is not NULL?

agent/php_user_instrument.c
```diff
@@ -343,6 +363,34 @@ bool nr_php_wraprec_matches(nruserfn_t* p, zend_function* func) {
   if ((NULL == func) || (ZEND_USER_FUNCTION != func->type)) {
     return false;
   }
+
+  if (0 != p->lineno) {
+    /*
+     * We have lineno/filename pair.  If this doesn't match, we can exit without
```
mfulb, line 369: Maybe "If this does match"?

agent/php_agent.h
zsistla: It's not a pointer.
https://github.com/infusion/PHP/blob/master/Zend/zend_compile.h#L282

agent/php_execute.c
zsistla: This just got formatted here and isn't related to this PR.
I think this comment should be directed to the https://github.com/newrelic/newrelic-php-agent/pull/555 ticket which is modifying the function.
Doesn't hurt to check execute_data and func for NULL.  NR_OP_ARRAY  also uses execut_data.

agent/php_user_instrument.c
zsistla: Clarified comment here 
8ec508e

agent/php_user_instrument.c
```diff
@@ -58,9 +58,7 @@ int nr_zend_call_orig_execute(NR_EXECUTE_PROTO TSRMLS_DC) {
     NR_PHP_PROCESS_GLOBALS(orig_execute)
     (NR_EXECUTE_ORIG_ARGS_OVERWRITE TSRMLS_CC);
   }
-  zend_catch {
-    zcaught = 1;
-  }
+  zend_catch { zcaught = 1; }
```
bduranleau-nr, line 61: Is this the result of clang-formatting? 
My clang-format sets this line to the original format, which uses the .clang-format file at the top level directory.

agent/php_execute.c
```diff
@@ -1035,7 +1035,8 @@ static void nr_php_execute_metadata_add_code_level_metrics(
 
 #undef CHK_CLM_STRLEN
 
-  lineno = nr_php_zend_execute_data_lineno(execute_data);
+  // lineno = nr_php_zend_execute_data_lineno(execute_data);
```
bduranleau-nr, line 1038:
```suggestion
```

zsistla: Actually, that line needs to be totally removed.
[c24df84](https://github.com/newrelic/newrelic-php-agent/pull/558/commits/c24df8444ac9b0a9c43d49b85225ce09a7ea1322)

agent/php_user_instrument.c
zsistla: Yes, we can revert this change, not sure why clang-format changed it.

agent/php_user_instrument.c
```diff
@@ -343,6 +363,38 @@ bool nr_php_wraprec_matches(nruserfn_t* p, zend_function* func) {
...
+       * could be a cli call or evaluated expression that has no filename.
+       */
+      if (NULL != p->filename) {
+        if (0 == nr_strcmp(p->filename, filename)) {
+          return true;
+        }
+        return false;
+      }
+    }
+  }
```
bduranleau-nr, line 396: we can remove the `else` since the `if` statement returns
```suggestion
    } 
    /*
     * lineno matched, let's check the filename
     */
    filename = nr_php_function_filename(func);

    /*
     * If p->filename isn't NULL, we know the comparison is accurate;
     * otherwise, it's inconclusive even if we have a lineno because it
     * could be a cli call or evaluated expression that has no filename.
     */
    if (NULL != p->filename) {
      if (0 == nr_strcmp(p->filename, filename)) {
        return true;
      }
      return false;
    }
  }
```

agent/php_user_instrument.c
```diff
@@ -353,15 +405,32 @@ bool nr_php_wraprec_matches(nruserfn_t* p, zend_function* func) {
...
+      if ((NULL != filename) && (0 != nr_strcmp("-", filename))) {
+        p->filename = nr_strdup(filename);
+      }
+    }
+    if (0 == p->lineno) {
+      p->lineno = nr_php_zend_function_lineno(func);
+    }
     return true;
   }
   return false;
```
bduranleau-nr, line 428: opportunity to refactor to un-nest conditionals by explicitly returning `false` if `reportedclass` && `classname` != `klass`

```suggestion
  if ((0 != nr_strcmp(p->reportedclass, klass))
      && (0 != nr_stricmp(p->classname, klass))) {
        return false;
  }
  /*
    * If we get here it means lineno/filename weren't initially set.
    * Set it now so we can do the optimized compare next time.
    * lineno/filename is usually not set if the func wasn't loaded when we
    * created the initial wraprec and we had to use the more difficult way to
    * set, update it with lineno/filename now.
    */
  if (NULL == p->filename) {
    filename = nr_php_function_filename(func);
    if ((NULL != filename) && (0 != nr_strcmp("-", filename))) {
      p->filename = nr_strdup(filename);
    }
  }
  if (0 == p->lineno) {
    p->lineno = nr_php_zend_function_lineno(func);
  }
  return true;
```

bduranleau-nr (approved):
👍 Looks good!

agent/php_execute.c
misiekn: The old instrumentation didn't have any NULL checks for `execute_data` received from Zend engine: no NULL checks for `execute_data` when agent receives control in `nr_php_execute ` ([agent/php_execute.c:1336](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1336)); no NULL checks [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1287) or [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1126). I think the reason for this approach might have been that if the agent receives NULL `execute_data`, then something is fundamentally wrong within the Zend engine and maybe it's better to crash.

agent/php_execute.c
```diff
@@ -1035,7 +1035,7 @@ static void nr_php_execute_metadata_add_code_level_metrics(
 
 #undef CHK_CLM_STRLEN
 
-  lineno = nr_php_zend_execute_data_lineno(execute_data);
+  lineno = nr_php_zend_function_lineno(execute_data->func);
```
misiekn, line 1038: Is this line needed? Isn't `lineno` set either on line 1049 or line 1054?

zsistla: I refactored this quite a bit with the latest PR I'm working on.  I've changed this function to populate the observer metadata and moved all the CLM logic into the php_txn function.
The new PR will do the following regarding this logic:
1. Renamed nr_php_execute_metadata_add_code_level_metrics to  nr_php_observer_metadata_init as this will be the way we populate the metadata.  We don't need to do duplicate effort with nr_php_execute_metadata_init
anymore.

2. Since nr_php_execute_metadata_init
 Is now for populating metadata, moved all CLM checks out of nr_php_observer_metadata_init
 And into nr_php_txn_add_code_level_metrics
 Where it is more appropriate.

agent/php_agent.h
misiekn, line 916: Should `func` be checked if it is a user function? Do all types of `zend_function` that may be passed to `nr_php_zend_function_lineno` have `line_start` defined/initialized? Is `nr_php_zend_execute_data_lineno` still needed?

zsistla: Thanks, addressed in [3a3a44f](https://github.com/newrelic/newrelic-php-agent/pull/558/commits/3a3a44f38797a2900577f65ef385ab46cbd35e56)

zsistla:
Closing in lieu of rebase against dev with most of these changes.
After oapi is rebased against dev, the following changes will be added to oapi via another PR.

[one](https://github.com/newrelic/newrelic-php-agent/pull/558/commits/3a3a44f38797a2900577f65ef385ab46cbd35e56) 
[two](https://github.com/newrelic/newrelic-php-agent/pull/558/commits/021fb44cdb8fa56cfbfa9feee7e773e13cc7d02d)

PR #559, opened 2022-10-20
feat(agent): Remove php_vm.c overwriting opcodes when using OAPI

ZNeumann:
Occasionally, we want to instrument functions called with call_user_func_array, when we otherwise wouldn't instrument that function. To do this, we instrument the internal function call_user_func_array and check for contexts in which we want to begin instrumentation. However, call_user_func_array can be inlined by the zend compiler. Previously, we were detecting inlined calls by overwriting the DO_FCALL opcode. With OAPI, we no longer want to touch the opcodes and are instead checking for the inlined calls during observer_do_fcall.

Additionally, this PR fixes instrumentation impacted by this change, as well as adding much needed testing for section of code.


agent/fw_drupal.c
```diff
@@ -588,6 +603,47 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke) {
...
+    goto leave;
+  }
+
+  ZVAL_DUP(hook_copy, hook);
+  if (!nr_php_is_zval_non_empty_string(hook_copy)) {
+    nrl_verbosedebug(NRL_FRAMEWORK,
+                     "%s: could not copy hook zval",
+                     __func__);
+    goto leave;
+  }
```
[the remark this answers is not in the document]
ZNeumann: I called `nr_php_arg_release` because, while the after function doesn't directly access the args, the value on the stack that it is freeing is the direct result from `nr_php_arg_get`

agent/fw_drupal.c
```diff
@@ -588,6 +603,30 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke) {
...
+  if (nr_php_is_zval_non_empty_string(hook_copy)) {
+    nr_stack_push(&NRPRG(drupal_module_invoke_all_hooks),
+                  hook_copy);
+  }
+}
+NR_PHP_WRAPPER_END
+
+NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_after) {
+    (void)wraprec;
+    zval* hook_copy = nr_stack_pop(&NRPRG(drupal_module_invoke_all_hooks));
```
bduranleau-nr, line 624: Do we just want to peek the top element of the stack, or pop it? The description for this function `nr_stack_pop` states that it's only performing a 'peek' operation.

bduranleau-nr: upon further investigation it looks like the description for `nr_stack_pop` is just wrong and was taken directly from `nr_stack_get_top` without being updated

agent/fw_drupal.c
```diff
@@ -588,6 +603,30 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke) {
...
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_before) {
+  (void)wraprec;
+  zval* hook_copy = NULL;
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL);
+
+  hook_copy = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
bduranleau-nr, line 614:
```suggestion
  hook_copy = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
```

agent/fw_drupal.c
ZNeumann: Looks like the function description is incorrect. It has the same description as nr_stack_get_top(), which actually does a peek. I'll add a fix to that with my next push.

agent/fw_wordpress.c
```diff
@@ -361,16 +379,27 @@ NR_PHP_WRAPPER(nr_wordpress_wrap_hook) {
    */
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+  if ((0 == NRINI(wordpress_hooks)) || (NULL == nr_stack_get_top(&NRPRG(wordpress_tags)))) {
```
bduranleau-nr, line 384: storing the result of `nr_stack_get_top(&NRPRG(wordpress_tags))` in a variable and referencing the variable instead could optimize 3 calls to `nr_stack_get_top()` down to 1 in this function.

agent/fw_wordpress.c
```diff
@@ -569,16 +615,60 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+                            nr_wordpress_exec_handle_tag,
+                            nr_wordpress_handle_tag_stack_after TSRMLS_CC);
+
+  nr_php_wrap_user_function_before_after(NR_PSTR("do_action"),
+                            nr_wordpress_exec_handle_tag,
+                            nr_wordpress_handle_tag_stack_after TSRMLS_CC);
+
+  nr_php_wrap_user_function_before_after(NR_PSTR("do_action_ref_array"),
+                            nr_wordpress_exec_handle_tag,
+                            nr_wordpress_handle_tag_stack_after TSRMLS_CC);
```
bduranleau-nr, line 669:
```suggestion
                            nr_wordpress_apply_filters_after);

  nr_php_wrap_user_function_before_after(NR_PSTR("apply_filters_ref_array"),
                            nr_wordpress_exec_handle_tag,
                            nr_wordpress_handle_tag_stack_after);

  nr_php_wrap_user_function_before_after(NR_PSTR("do_action"),
                            nr_wordpress_exec_handle_tag,
                            nr_wordpress_handle_tag_stack_after);

  nr_php_wrap_user_function_before_after(NR_PSTR("do_action_ref_array"),
                            nr_wordpress_exec_handle_tag,
                            nr_wordpress_handle_tag_stack_after);
```

agent/php_execute.c
```diff
@@ -1685,6 +1685,68 @@ void nr_php_user_instrumentation_from_opcache(TSRMLS_D) {
  */
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP8+ */
 
+static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
+  NR_UNUSED_FUNC_RETURN_VALUE;
+  if (NULL == execute_data->prev_execute_data) {
```
bduranleau-nr, line 1690: is `execute_data` itself always guaranteed to be non-null here?

agent/php_rinit.c
```diff
@@ -91,6 +91,16 @@ PHP_RINIT_FUNCTION(newrelic) {
       "(^([a-z_-]+[_-])([0-9a-f_.]+[0-9][0-9a-f.]+)(_{0,1}.*)$|(.*))",
       NR_REGEX_CASELESS, 0);
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+  /*
+   * Pre-OAPI, this variables were kept on the call stack and
+   * therefore had no need to be in an nr_stack
+   */
+  nr_stack_init(&NRPRG(wordpress_tags), 32);
```
bduranleau-nr, line 100: Is there a reason the value `32` was chosen?

agent/fw_drupal.c
mfulb, line 624: What happens if on line 615 above it is decided to not push due to `hook_copy` being an empty string.  Will a value get popped in the "after" call that shouldn't?

agent/fw_wordpress.c
```diff
@@ -285,7 +285,13 @@ static char* nr_wordpress_plugin_from_function(zend_function* func TSRMLS_DC) {
     nrl_verbosedebug(NRL_FRAMEWORK,
```
mfulb, line 285: `nrl_verbosedebug()` appears to be based on `vsnprintf()` under the hood and I am concerned (depending on the libc in use) that passing `NRP_PHP((char *)nr_stack_get_top(&NRPRG(wordpress_tags))` without checking it for NULL might be passing NULL to "%s" which could be a problem.  Are we certain this can't happen at this point in the code?

agent/fw_wordpress.c
bduranleau-nr: This block is wrapped in PHP 8 + OAPI checks, which is impossible to have ZTS enabled for, so these can be removed from this block safely. TSRM functionality was removed completely for PHP 8: https://github.com/php/php-src/blob/PHP-8.0/UPGRADING.INTERNALS#L50.

mfulb: Gotcha thanks for the clarificaiton.

agent/fw_drupal.c
ZNeumann: Yeah, good find. I'm trying to think of ways to solve this and can only think of 2 viable solutions. 1) keep a second stack of bools that indicates whether or not the current call has pushed to the drupal_module_invoke_all_hooks stack 2) push to the stack even when the string is null.empty and then when we are accessing the drupal_module_invoke_all_hooks, walk up the stack until we find a non-empty string.

I am currently leaning towards the first.

agent/fw_wordpress.c
ZNeumann: fixed in [de23c51](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/de23c51e5e08e3661282d595b47809ec91d25ded)

agent/php_rinit.c
ZNeumann: I accidentally chose the `NR_STACK_DEFAULT_CAPACITY` value instead of using the macro. fixed in [de23c51](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/de23c51e5e08e3661282d595b47809ec91d25ded)

agent/fw_drupal.c
ZNeumann: I suppose the other option is 3) copy the hook from the args again in the after function to check if it's empty. This would create an unnecessary copy for the sake of code simplicity and readability.

I am actually pretty compelled to go with option 3 if nobody has issue with its tradeoffs.

ZNeumann: I actually opted for 3) but using the nr_php_get_user_func_arg to prevent the additional copy.

Walking up the stack would be necessary to maintain functionality with previous instrumentation. The module_invoke_all instrumentation was what maintains which value is set to the global variable (and refuses to add NULL/empty string) but this value is used in other instrumentations.

agent/fw_wordpress.c
```diff
@@ -569,16 +615,60 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+  nr_wordpress_handle_tag_stack_after(NR_SPECIALFNPTR_ORIG_ARGS);
 }
 NR_PHP_WRAPPER_END
 
+#endif //OAPI
+
 void nr_wordpress_enable(TSRMLS_D) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+  nr_php_wrap_user_function_before_after(NR_PSTR("apply_filters"),
```
misiekn, line 655: Would it be possible to add integration tests for `apply_filters` to exercise `nr_wordpress_apply_filters` and `nr_wordpress_apply_filters_after`?

agent/fw_wordpress.c
```diff
@@ -569,16 +615,60 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+                            nr_wordpress_apply_filters,
+                            nr_wordpress_apply_filters_after);
+
+  nr_php_wrap_user_function_before_after(NR_PSTR("apply_filters_ref_array"),
+                            nr_wordpress_exec_handle_tag,
+                            nr_wordpress_handle_tag_stack_after);
+
+  nr_php_wrap_user_function_before_after(NR_PSTR("do_action"),
```
misiekn, line 663: Would it be possible to add integration tests for `do_action` to exercise `nr_wordpress_exec_handle_tag` and `nr_wordpress_handle_tag_stack_after`?

tests/integration/frameworks/drupal/test_module_invoke_all.php
```diff
@@ -0,0 +1,66 @@
...
+    echo "f\n";
+}
+
+function module_g() {
+    echo "g\n";
+    module_f();
+}
+
+module_invoke_all("g");
+module_invoke_all("f");
```
misiekn, line 66: Maybe add a test for a hook with arguments? Something like:
```
function module_hook_with_arg($arg) {
    echo "module_hook_with_arg(arg=[$arg])\n";
}
module_invoke_all("hook_with_arg", "arg_value");
```
This would require a slight modification of `module_invoke_all` mock to actually pass the arguments. This should do the trick:
```
function module_invoke_all($f) {
  $args = func_get_args();
  unset($args[0]);
  call_user_func_array("module_" . $f, $args);
}
```
Additionally consider testing negative cases: 
* `module_invoke_all` is called with non-existing function
* `module_invoke_all` is called with a function expecting arguments but didn't receive any
* `module_invoke_all` is called with empty argument
I'm not sure these scenarios deserve tests in this PR, but would be good to know how the agent will handle these scenarios. Maybe these cases belong to #574 or to a subsequent PR?

misiekn (approved):
Really neat implementation! 👍

agent/fw_drupal.c
ZNeumann: wordpress check added here [1a451a7](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/1a451a7af25aba4298dbc56c78f7341920d0a8b1)

agent/fw_wordpress.c
```diff
@@ -471,12 +506,17 @@ NR_PHP_WRAPPER(nr_wordpress_exec_handle_tag) {
   tag = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
 
   if (1 == nr_php_is_zval_non_empty_string(tag)
-      || (0 != NRINI(wordpress_hooks))) {
+      && (0 != NRINI(wordpress_hooks))) {
```
ZNeumann, line 509: This check was changed to match that of `nr_wordpress_apply_filters`. I believe this to be the correct functionality, as we likely don't ever want to be setting the hook tag if the INI is false and also want to prevent potential NULL tags if the INI is true.

The downstream code is safe for NULL tags, and just skips instrumentation if NULL. So this change could instrument more, as it prevents NULL skip from ever happening (while within a `nr_wordpress_exec_handle_tag` call stack).

agent/fw_wordpress.c
ZNeumann: NRP_PHP checks for NULL and passes an empty string if so

tests/integration/frameworks/drupal/test_module_invoke_all.php
ZNeumann: Added the first part in [d6412ef](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/d6412ef0cfa8c82bfa93a8a263f691071ed741f6)

The three negative cases cannot really test agent functionality because each one throws a fatal PHP error.

bduranleau-nr (dismissed):

agent/fw_wordpress.c
ZNeumann: added in [ac44475](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/ac4447541e259e4c6ef19c3bd7e5e45241bb6d34)

agent/fw_wordpress.c
ZNeumann: added in [ac44475](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/ac4447541e259e4c6ef19c3bd7e5e45241bb6d34)

agent/php_execute.c
ZNeumann: Can I get @zsistla 's input here?

ZNeumann: There is a check in the oapi branch further up the call stack. So unless it will be modified in between it should be fine.

ZNeumann:
ok jenkins

agent/fw_drupal.c
```diff
@@ -717,6 +732,35 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke) {
...
+                  hook_copy);
+  }
+}
+NR_PHP_WRAPPER_END
+
+NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_after) {
+  (void)wraprec;
+  /* using nr_php_get_user_func_arg() so that we don't perform another copy
+   * when all we want to do is check the string length */
+  zval* orig = nr_php_get_user_func_arg((zend_uint)1, NR_EXECUTE_ORIG_ARGS);
```
zsistla, line 755: This function is pointed to by the clean callback, but when cleaning up, we don't have access to the zend_execute_data.  We can only clean up NR globals.

agent/fw_wordpress.c
```diff
@@ -569,16 +615,71 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+      char* cleaned_tag = nr_stack_pop(&NRPRG(wordpress_tags));
+      nr_free(cleaned_tag);
+  }
+}
+NR_PHP_WRAPPER_END
+
+NR_PHP_WRAPPER(nr_wordpress_apply_filters_after) {
+  /* using nr_php_get_user_func_arg() so that we don't perform another copy
+   * when all we want to do is check the string length */
+  zval* tag = nr_php_get_user_func_arg(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
zsistla, line 647: When called from the clean_callback, the NR_EXECUTE_ORIG_ARGS will be null, make sure dependent calls are okay.

agent/fw_wordpress.c
```diff
@@ -569,16 +615,71 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+  if (1 == nr_php_is_zval_non_empty_string(tag)) {
+    zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
+    nr_wordpress_name_the_wt(tag, retval_ptr TSRMLS_CC);
+  }
+
+  nr_wordpress_handle_tag_stack_after(NR_SPECIALFNPTR_ORIG_ARGS);
```
zsistla, line 653: NR_SPECIALFNPTR_ORIG_ARGS will resolve to:
```
#define NR_SPECIALFNPTR_ORIG_ARGS \
  wraprec, auto_segment, execute_data, func_return_value
```
execute_data and func_return_value will be NULL.  Make sure `nr_wordpress_handle_tag_stack_after` is okay with that.

agent/fw_drupal.c
ZNeumann: Gotcha, that makes a lot of sense. Unfortunately, that means that we are back to needing another global stack to determine whether or not we have pushed to this stack. Because an empty string/null will not be pushed to this global stack and the execute_data was previously how I was determining this.

agent/php_rinit.c
```diff
@@ -104,9 +108,18 @@ PHP_RINIT_FUNCTION(newrelic) {
     char* str = (char*)e;
     nr_free(str);
   }
+  void zval_stack_dtor(void* e, NRUNUSED void* d) {
+    zval* zv = (zval*)e;
+    nr_php_zval_free(&zv);
+  }
   nr_stack_init(&NRPRG(predis_ctxs), NR_STACK_DEFAULT_CAPACITY);
+  nr_stack_init(&NRPRG(wordpress_tags), NR_STACK_DEFAULT_CAPACITY);
```
zsistla, line 116: Does nr_stack_destroy_fields deinit this?
Or does it also need to be deinited in rshutdown?

tests/integration/frameworks/drupal/test_module_invoke_all.php
```diff
@@ -0,0 +1,75 @@
...
+    module_f();
+}
+
+function module_hook_with_arg($arg) {
+        echo "module_hook_with_arg(arg=[$arg])\n";
+}
+
+module_invoke_all("hook_with_arg", "arg_value");
+module_invoke_all("g");
+module_invoke_all("f");
```
zsistla, line 75: Can you add a test case where you have the nested invoke alls and then one generates an exception to verify your clean callback behaves as expected?

agent/fw_drupal.c
```diff
@@ -454,25 +454,40 @@ static void nr_drupal_wrap_hook_within_module_invoke_all(
    * module_invoke_all(), the drupal_module_invoke_all_hook global should be
    * available.
    */
-  if (NULL == NRPRG(drupal_module_invoke_all_hook)) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+  zval* curr_hook = (zval*)nr_stack_get_top(&NRPRG(drupal_module_invoke_all_hooks));
```
zsistla, line 459: Is it possible to add unit tests that exercise the functionality of the new hooks stack?
https://github.com/newrelic/newrelic-php-agent/blob/main/agent/tests/test_fw_drupal.c

agent/fw_wordpress.c
```diff
@@ -285,7 +285,13 @@ static char* nr_wordpress_plugin_from_function(zend_function* func TSRMLS_DC) {
     nrl_verbosedebug(NRL_FRAMEWORK,
                      "Wordpress: cannot determine plugin name:"
                      " missing filename, tag=" NRP_FMT,
-                     NRP_PHP(NRPRG(wordpress_tag)));
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+                     NRP_PHP((char *)nr_stack_get_top(&NRPRG(wordpress_tags)))
```
zsistla, line 290: Is it possible to add unit tests that exercise the functionality of the new tags stack?
https://github.com/newrelic/newrelic-php-agent/blob/main/agent/tests/test_fw_wordpress.c

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
```diff
@@ -0,0 +1,63 @@
...
+    echo "\n";
+}
+
+function f($str) {
+    echo "f: ";
+    echo $str;
+    echo "\n";
+    apply_filters("g", "string2");
+}
+
```
zsistla, line 62: Is it possible to nest it somehow so we now we are building the internal stack then throw an exception?

tests/integration/frameworks/wordpress/test_wordpress_do_action.php
```diff
@@ -0,0 +1,59 @@
...
+function g() {
+    echo "g\n";
+}
+
+function f() {
+    echo "f\n";
+    do_action("g");
+}
+
+do_action("f");
```
zsistla, line 59: possible to nest and throw exception?

agent/php_rinit.c
ZNeumann: The stack itself is a vector wrapper with the actual vector being a member `elements` that is a pointer to the stack. `nr_stack_init` allocates this memory and sets `elements` accordingly. `nr_stack_destroy_fields` then iterates over all elements in the `elements` vector, calling the dtor on each, before finally freeing the actual vector. No additional cleanup should be needed beyond this.

agent/fw_drupal.c
```diff
@@ -778,8 +823,16 @@ void nr_drupal_enable(TSRMLS_D) {
   if (NRINI(drupal_modules)) {
     nr_php_wrap_user_function(NR_PSTR("module_invoke"),
                               nr_drupal_wrap_module_invoke TSRMLS_CC);
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+     && !defined OVERWRITE_ZEND_EXECUTE_DATA
+    nr_php_wrap_user_function_before_after_clean(NR_PSTR("module_invoke_all"),
+                                                 nr_drupal_wrap_module_invoke_all_before,
+                                                 nr_drupal_wrap_module_invoke_all_after,
+                                                 nr_drupal_wrap_module_invoke_all_after);
```
misiekn, line 831: I'd really prefer to see a separate function `nr_drupal_wrap_module_invoke_all_clean` or `nr_drupal_wrap_module_invoke_all_exception_handler`, and if possible extract common code for `nr_drupal_wrap_module_invoke_all_after` and this new function to a helper method which can operate independent of the after/cleanup context. Let's try to follow to tried and tested [SOLID](https://en.wikipedia.org/wiki/SOLID) design principles that make the code maintainable. Having the `_after` function do two things (handle the _after user function call and handle cleanup after exception) smells like a violation of [Single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

agent/fw_wordpress.c
ZNeumann: I don't think we should use unit testing here. The manipulation of the stack over the lifetime of a transaction is what we want to be testing. For a unit test to do this, it would need to call the instrumentation functions in order, and at that point it is doing the same thing as the integration tests.

There are already unit tests to verify generic stack functionality.

agent/fw_drupal.c
```diff
@@ -748,6 +792,7 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all) {
   nr_php_arg_release(&hook);
 }
 NR_PHP_WRAPPER_END
+#endif //OAPI
```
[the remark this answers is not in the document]
ZNeumann: fixed in [79c7ee8](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/79c7ee84f4d33549eb345c94afaea0a58874d33d)

agent/fw_wordpress.c
ZNeumann: fixed in [79c7ee8](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/79c7ee84f4d33549eb345c94afaea0a58874d33d)

tests/integration/frameworks/drupal/test_module_invoke_all.php
ZNeumann: added in [79c7ee8](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/79c7ee84f4d33549eb345c94afaea0a58874d33d)

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
ZNeumann: added in [79c7ee8](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/79c7ee84f4d33549eb345c94afaea0a58874d33d)

tests/integration/frameworks/wordpress/test_wordpress_do_action.php
ZNeumann: added in [79c7ee8](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/79c7ee84f4d33549eb345c94afaea0a58874d33d)

agent/fw_drupal.c
ZNeumann: extracted core functionality to a helper and added a "clean" function in [79c7ee8](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/79c7ee84f4d33549eb345c94afaea0a58874d33d)

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
ZNeumann: added in [6cd6234](https://github.com/newrelic/newrelic-php-agent/pull/559/commits/6cd62348111bb4a5a27ac059a782f3e8961f13e2)

zsistla:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

ZNeumann:
ok jenkins

zsistla (dismissed):
Great job, thanks!

PR #560, opened 2022-10-20
fix(agent): fix exception and error handling

misiekn:
This PR addresses issues with exception and error attributes in spans.


misiekn:
Superseded with #574.

PR #562, opened 2022-10-21
tests(integration): update generator tests for oapi

bduranleau-nr:
* Update 7.1 test_generator for PHP version range 7.1 - 7.4
* Add new test for PHP version 8.0+

### DIFF
--- a/tests/integration/lang/test_generator_7.1-7.4.php
+++ b/tests/integration/lang/test_generator_7.1-7.4.php
@@ -11,7 +11,7 @@ The agent should correctly instrument generators.
 /*SKIPIF
 <?php
 
-if (version_compare(PHP_VERSION, '7.1', '<')) {
+if (version_compare(PHP_VERSION, '7.1', '<') || version_compare(PHP_VERSION, '7.4', '>')) {
   die("skip: generators either not available or with different behaviour");
 }
 */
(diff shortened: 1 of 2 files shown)


mfulb (approved):
Good job on a brain twister.

ZNeumann (approved):

misiekn (approved):
### Explanation of the value of 11 for a count in “Custom/xrange” scoped and unscoped metrics.
 It turns out that pre-observer api instrumentation got notified when a generator object was created and destroyed. PHP’s `zend_execute` function is called not only to execute the body of `xrange` function, but also to create and destroy a generator object. `execute_data` had `xrange` as a name of a function being executed in all 12 cases (10 x actual execution of the body + 1 execution when `xrange` generator was created + 1 execution when `xrange` generator was destroyed). Because the pre-observer api agent overwrites `zend_execute` function, it therefore recorded that `xrange` was called 12 times! Now with observer api the situation is different: 1) the agent no longer gets notified for every `zend_execute` but only for the calls that PHP thinks may be interesting to observers; 2) PHP doesn’t think that generator creation is worth notifying the observers; 3) PHP thinks that letting the generator complete its execution before it is destroyed is worth notifying the observers. Adding 1), 2) and 3) we get the final value of 11 as number of calls to `xrange` because the agent knows about 10x the `xrange` is called to yield something useful to the caller + 1x when it finishes execution before it gets destroyed.

PR #563, opened 2022-10-21
fix(agent): fix check for max strlen when generating clm attributes

bduranleau-nr:
* fixed the check for the length of the function/namespace/filepath attribute when generating CLM metadata
* added integration tests to cover cases where string length > 255

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1020,7 +1020,7 @@ static void nr_php_execute_metadata_add_code_level_metrics(
    */
 
 #define CHK_CLM_STRLEN(s)                          \
-  if (CLM_STRLEN_MAX < NRSAFELEN(sizeof(s) - 1)) { \
+  if (CLM_STRLEN_MAX < NRSAFELEN(nr_strlen(s))) {  \
     s = NULL;                                      \
   }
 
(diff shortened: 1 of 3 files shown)


ZNeumann (approved):

agent/php_execute.c
```diff
@@ -1020,7 +1020,7 @@ static void nr_php_execute_metadata_add_code_level_metrics(
    */
 
 #define CHK_CLM_STRLEN(s)                          \
-  if (CLM_STRLEN_MAX < NRSAFELEN(sizeof(s) - 1)) { \
+  if (CLM_STRLEN_MAX < NRSAFELEN(nr_strlen(s))) {  \
```
mfulb, line 1023: Looking at `nr_strlen()` it returns an `int` but `strlen()` (which it uses) returns a `size_t` which is an unsigned integer type.
This code uses `NRSAFELEN()`to handle converting a `size_t` to a `signed int` but it seems that the `nr_strlen()` inline function also needs it?

bduranleau-nr: I can remove the `NRSAFELEN` wrapper from this block if we think it's unnecessary/harmful, but I'm definitely not comfortable modifying `nr_strlen`'s functionality in the scope of this PR

mfulb: Tough call but if what I wrote above it true `nr_strlen()` is fundamentally flawed so added a macro around it won't do much.

PR #566, opened 2022-10-25
tests(agent): Update integration tests for logging on by default

mfulb:


misiekn (approved):
👍

bduranleau-nr:
Did we leave any checks for logging off? I would think we'd also want coverage for that configuration

zsistla (approved):

PR #568, opened 2022-10-26
fix(agent): re-work user instrumentation lookup

ZNeumann:
This PR adds a new method of looking up user instrumentation for PHP 7.4+. Using `reserved` field of `zend_func`'s `op_array` is no longer reliable when opcache is enabled. This is mostly visible in php-fpm configurations where an agent from one php-fpm worker processes creates `wraprec` and sets `reserved` field of `zend_func`'s `op_array` and an agent from a different php-fpm worker processes tries to access the instrumentation. This could happen because PHP uses shared memory for opcache. Another manifestation of the problem caused by using `reserved` field of `zend_func`'s `op_array` is when opcache got corrupted. The new method no longer relies on the agent storing any information in `zend_func` that might got stored and/or corrupted in opcache. The agent is self-sufficient to store and lookup user functions instrumentation by filename/line number in conjunction with {scope::}(method|function) name. With this approach the agent is always able to find the instrumentation regardless if opcache is or is not enabled and if agent runs in php-fpm context.


misiekn (changes requested):
@ZNeumann This PR includes observer api and code level metrics changes that are not relevant to this fix. I've opened https://github.com/ZNeumann/newrelic-php-agent/pull/1 to clean this up.

misiekn:
Shouldn't [agent/fw_drupal8.c:57](https://github.com/ZNeumann/newrelic-php-agent/blob/wraprec_changes/agent/fw_drupal8.c#L57) be updated to use new method of getting wraprec?

ZNeumann:
> @ZNeumann This PR includes observer api and code level metrics changes that are not relevant to this fix. I've opened [ZNeumann#1](https://github.com/ZNeumann/newrelic-php-agent/pull/1) to clean this up. This PR should be focused on wraprec fix only.

Thanks for catching those! Git was overly happy to silently include code changes from other commits during some of the cherry-picks, and we also had to manually pull code from some OAPI commits.

agent/php_user_instrument.h
```diff
@@ -82,6 +106,31 @@ typedef struct _nruserfn_t {
...
+/*
+ * Purpose : Determine if a func matches a wraprec.
+ *
+ * Params  : 1. The wraprec to match to a zend function
+ *           2. The zend function to match to a wraprec
+ *
+ * Returns : True if the class/function of a wraprec match the class function
+ *           of a zend function.
+ */
+extern bool nr_php_wraprec_matches(nruserfn_t* p, zend_function* func);
```
zsistla, line 118: can you please make this static inline?
We aren't using it externally anymore, it can just exist in the .c file to hopefully add a little more speed.

misiekn: @ZNeumann I made suggestion not to export this function in https://github.com/ZNeumann/newrelic-php-agent/pull/1/commits/6eacc32388de762c301b35b1bab72d185afe07ba so keep it in mind if you plan to accept https://github.com/ZNeumann/newrelic-php-agent/pull/1 before applying `inline` keyword.

agent/php_user_instrument.c
```diff
@@ -88,10 +93,69 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
+   * an evaluated expression or called directly from the CLI there is no
+   * filename, but the function says the filename is "-". Avoid setting in this
+   * case; otherwise, all the evaluated/cli calls would match.
+   */
+  if ((NULL == wraprec->filename) && (NULL != filename)
+      && (0 != nr_strcmp("-", filename))) {
+    wraprec->filename = nr_strdup(filename);
+  }
+
+  wraprec->lineno = nr_php_zend_function_lineno(func);
```
zsistla, line 151: What is we move this as the first check?
If it is zero, I believe we can skip the filename lookup/strdups.

misiekn: You mean that if filename is '-' then lineno is 0? Is reverse true: if lineno is 0 will the filename be '-' or can it be something different, useful, meaningful?

agent/php_user_instrument.h
ZNeumann: Made inline in [93baea2](https://github.com/newrelic/newrelic-php-agent/pull/568/commits/93baea2610bc2b2f373a0a04a36a27f31313c62b)

agent/php_user_instrument.c
ZNeumann: This function is infrequently called and the strcomp is trivial because it is only 1 character.

misiekn (approved):
LGTM :+1:

mfulb:
Can we wrap the function definition for `nr_php_op_array_get_wraprec()` with PHP api number < 7.4 so this function is not present when compiling as a safeguard it is not being used by a compile time check?

zsistla (approved):

ZNeumann:
ok jenkins

misiekn:
> Can we wrap the function definition for nr_php_op_array_get_wraprec() with PHP api number < 7.4 so this function is not present when compiling as a safeguard it is not being used when it should be looking up by line number/filename by a compile time check?

This is a good idea. That's how I caught [this](https://github.com/newrelic/newrelic-php-agent/pull/568#issuecomment-1293420955).

ZNeumann:
> > Can we wrap the function definition for nr_php_op_array_get_wraprec() with PHP api number < 7.4 so this function is not present when compiling as a safeguard it is not being used when it should be looking up by line number/filename by a compile time check?
> 
> This is a good idea. That's how I caught [this](https://github.com/newrelic/newrelic-php-agent/pull/568#issuecomment-1293420955).

Thanks for reminding me! I tend to forget about the global comments that aren't tied to the code. Added in [0186b5f](https://github.com/newrelic/newrelic-php-agent/pull/568/commits/0186b5f93a4a5e5223620f44187157b6175c4ca4)

ZNeumann:
ok jenkins

PR #574, opened 2022-11-01
feat(agent): OAPI exception handling

zsistla:
1. Renamed nr_php_execute_metadata_add_code_level_metrics to  nr_php_observer_metadata_init as this will be the way we populate the metadata.  We don't need to do duplicate effort with `nr_php_execute_metadata_init` anymore.
2. Since nr_php_execute_metadata_init is now for populating metadata, moved all CLM checks out nr_php_observer_metadata_init and into nr_php_txn_add_code_level_metrics where it is more appropriate.
3. Modified stacked segments to have a pointer to the metadata so if they are closed off by an exception, they will properly propogate the information.  Added initializations/deninitializations.
4. For php_execute_enabled, only call nr_php_observer_metadata_init if we need to.
5. Added nr_php_observer_exception_segment_end
 To handled closing off observer segments if an exception occurs.
6. Now use zend_throw_exception_hook (more info:
https://www.phpinternalsbook.com/php7/extensions_design/hooks.html
Note: This ONLY notifies when an exception is thrown.  It gives no indication if that exception was subsequently caught or not.
8. OAPI leaves dangling segments in the case of exceptions.  These need to be cleaned up for functions that rely on the current segment (includes begin/end functions, stacked segment unwinding, and API calls) 
9. New inline function re`nr_php_api_ensure_current_segment` to account for dangling segments when calling our API.
10. TXN globals to keep track of the exception
11. Change in php_txn turn off recording after unwind the segments to give timer to attach exception to dangling segment(s).
12. Modify stacked segments init/denit to handle additional segment metadata variable.
13. Functions to clear/set TXN uncaught_exception variables.
14. Update metadata struct to retain more context of the segment.
15. Removed legacy exception code that wasn't getting called anymore.
16. Added tests.

### DIFF
--- a/agent/php_agent.h
+++ b/agent/php_agent.h
@@ -39,6 +39,7 @@
 #include "php_compat.h"
 #include "php_newrelic.h"
 #include "php_zval.h"
+#include "util_logging.h"
 #include "util_memory.h"
 #include "util_strings.h"
 #include "php_execute.h"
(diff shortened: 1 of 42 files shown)


zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
Close in lieu of #580

zsistla:
ok jenkins

zsistla:
closed in lieu of #580

PR #575, opened 2022-11-03
Fix creating some logging metrics

mfulb:
The logging metrics "Supportability/Logging/Forwarding/PHP/<state>" and "Supportability/Logging/Metrics/PHP/<state>"} are supposed to be created once per request.  The previous mechanism did not properly create these metrics if a transaction was started and restarted during a request, or if newrelic_set_appname() was called.

This fix uses a flag within the transaction structure to track if the metrics have been created for the transaction which fixes the previous issue.

Affected integration tests were updated to reflect the metrics are now being created properly.

### DIFF
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -635,7 +635,6 @@ static void nr_php_txn_send_metrics_once(nrtxn_t* txn TSRMLS_DC) {
     return;
   }
 
-  nrl_verbosedebug(NRL_TXN, "creating one time logging metrics");
 #define FMT_BOOL(v) (v) ? "enabled" : "disabled"
 
   metname = nr_formatf("Supportability/Logging/Forwarding/PHP/%s",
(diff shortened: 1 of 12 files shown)


agent/php_txn.c
```diff
@@ -625,17 +625,17 @@ static void nr_php_txn_log_error_dt_on_tt_off(void) {
...
   if (nrunlikely(NULL == NRPRG(txn))) {
     return;
   }
 
-  if (nrlikely(0 != sent)) {
+  if (nrlikely(0 != txn->created_logging_onetime_metrics)) {
     return;
   }
 
+  nrl_verbosedebug(NRL_TXN, "creating one time logging metrics");
```
bduranleau-nr, line 638: Can probably remove this debug line

zsistla (approved):

agent/php_txn.c
mfulb: Yeah it should only happen once or twice during a request and is only when debugging level is active but I'll remove it.

bduranleau-nr (approved):

mfulb:
ok jenkins

mfulb:
ok jenkins

ZNeumann (approved):

PR #576, opened 2022-11-03
fix(agent): move globals that utilized the call stack into a global stack

ZNeumann:
Symfony1 instrumentation utilizes the call stack, but should never be run with PHP8+


ZNeumann:
Duplicate of https://github.com/newrelic/newrelic-php-agent/pull/582

PR #579, opened 2022-11-07
docs(agent): clean up `newrelic.framework` INI info

mfulb:
Update comment in template newrelic.ini to reflect supported frameworks.

### DIFF
--- a/agent/scripts/newrelic.ini.template
+++ b/agent/scripts/newrelic.ini.template
@@ -638,10 +638,11 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;
 ;          Must be one of the following values:
 ;          cakephp, codeigniter, drupal, drupal8, joomla, kohana, laravel,
-;          magento, magento2, mediawiki, silex, slim, symfony1, symfony2,
+;          magento, magento2, mediawiki, slim, symfony2, symfony4,
 ;          wordpress, yii, zend, zend2, no_framework
 ;
-;          Note that "drupal" covers only Drupal 6 and 7.
+;          Note that "drupal" covers only Drupal 6 and 7 and "symfony2"
+;          now only supports Symfony 3.x.
 ;
 ;newrelic.framework = ""
 


bduranleau-nr:
Saying that "symfony2 only supports Symfony 3.x" seems pretty confusing- I think we would be better off removing `symfony2` as an option altogether and replacing it with `symfony3`

mfulb:
ok jenkins

bduranleau-nr (approved):

agent/scripts/newrelic.ini.template
```diff
@@ -638,10 +638,11 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;
 ;          Must be one of the following values:
 ;          cakephp, codeigniter, drupal, drupal8, joomla, kohana, laravel,
-;          magento, magento2, mediawiki, silex, slim, symfony1, symfony2,
+;          magento, magento2, mediawiki, slim, symfony2, symfony4,
 ;          wordpress, yii, zend, zend2, no_framework
 ;
-;          Note that "drupal" covers only Drupal 6 and 7.
+;          Note that "drupal" covers only Drupal 6 and 7 and "symfony2"
+;          now only supports Symfony 3.x.
```
misiekn, line 645: Maybe add a note that symfony4 covers Symfony 5 (see https://github.com/newrelic/newrelic-php-agent/pull/256)?

misiekn (approved):
Nice little tweak! Thank you! 👍

agent/scripts/newrelic.ini.template
mfulb: I was more just emphasizing the "symfony2" doesnt actually support Symfony 2.x.

agent/scripts/newrelic.ini.template
```diff
@@ -638,10 +638,11 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;
 ;          Must be one of the following values:
 ;          cakephp, codeigniter, drupal, drupal8, joomla, kohana, laravel,
-;          magento, magento2, mediawiki, silex, slim, symfony1, symfony2,
+;          magento, magento2, mediawiki, slim, symfony2, symfony4,
 ;          wordpress, yii, zend, zend2, no_framework
 ;
-;          Note that "drupal" covers only Drupal 6 and 7.
+;          Note that "drupal" covers only Drupal 6 and 7 and "symfony2"
```
ZNeumann, line 644: This is so confusing even to me as an engineer on the code base. I know it's not in the scope of this PR, but can we add a backlog item to change the way we handle this?

ZNeumann (approved):

agent/scripts/newrelic.ini.template
misiekn: I'm fine with merging it as is and I think [Zach's suggestion](https://github.com/newrelic/newrelic-php-agent/pull/579#discussion_r1024456730) makes more sense.

PR #580, opened 2022-11-08
feat(agent): OAPI exception handling

zsistla:
1. Renamed nr_php_execute_metadata_add_code_level_metrics to  nr_php_observer_metadata_init as this will be the way we populate the metadata.  We don't need to do duplicate effort with `nr_php_execute_metadata_init` anymore.
2. Since nr_php_execute_metadata_init is now for populating metadata, moved all CLM checks out nr_php_observer_metadata_init and into nr_php_txn_add_code_level_metrics where it is more appropriate.
3. Modified stacked segments to have a pointer to the metadata so if they are closed off by an exception, they will properly propogate the information.  Added initializations/deninitializations.
4. For php_execute_enabled, only call nr_php_observer_metadata_init if we need to.
5. Added nr_php_observer_exception_segment_end
 To handled closing off observer segments if an exception occurs.
6. Now use zend_throw_exception_hook (more info:
https://www.phpinternalsbook.com/php7/extensions_design/hooks.html
Note: This ONLY notifies when an exception is thrown.  It gives no indication if that exception was subsequently caught or not.
8. OAPI leaves dangling segments in the case of exceptions.  These need to be cleaned up for functions that rely on the current segment (includes begin/end functions, stacked segment unwinding, and API calls) 
9. New inline function re`nr_php_api_ensure_current_segment` to account for dangling segments when calling our API.
10. TXN globals to keep track of the exception
11. Change in php_txn turn off recording after unwind the segments to give timer to attach exception to dangling segment(s).
12. Modify stacked segments init/denit to handle additional segment metadata variable.
13. Functions to clear/set TXN uncaught_exception variables.
14. Update metadata struct to retain more context of the segment.
15. Removed legacy exception code that wasn't getting called anymore.
16. Added tests.
17. Added a `clean`callback to the wraprec functionality and associated unit tests.


zsistla:
ok jenkins

mfulb:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

agent/php_execute.c
```diff
@@ -992,16 +991,16 @@ static void nr_php_execute_metadata_add_code_level_metrics(
...
   const char* function = NULL;
-  uint32_t lineno = 1;
-
   /*
    * Check if code level metrics are enabled in the ini.
    * If they aren't, exit and don't update CLM.
    */
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
+    || defined OVERWRITE_ZEND_EXECUTE_DATA
   if (!NRINI(code_level_metrics_enabled)) {
```
bduranleau-nr, line 1000: `code_level_metrics_enabled` does not seem to have a section in the ini template yet- is this being added in a separate PR?

agent/php_execute.c
```diff
@@ -1221,20 +1181,33 @@ static inline void nr_php_execute_segment_end(
...
+
+    nr_php_execute_segment_add_metric(stacked, metadata, create_metric);
+    if (NULL == stacked->attributes) {
+      stacked->attributes = nr_attributes_create(NRPRG(txn)->attribute_config);
+    }
+
+    nr_php_txn_add_code_level_metrics(stacked->attributes, metadata);
+    nr_segment_t* s = nr_php_stacked_segment_move_to_heap(stacked TSRMLS_CC);
+
+#else
```
bduranleau-nr, line 1195: A brief comment describing why this is done this way for OAPI & PHP8+ vs other configurations would be valuable in the future

agent/php_execute.c
zsistla: Yes, it was added in this PR for oapi:https://github.com/newrelic/newrelic-php-agent/pull/506/files#diff-07c39860ed983532dfdc2a9d3093475beb7bd749067607c324a1e7718cffd92fR2870

It's also added in this PR on dev: https://github.com/newrelic/newrelic-php-agent/pull/585/files#diff-07c39860ed983532dfdc2a9d3093475beb7bd749067607c324a1e7718cffd92fR3108

agent/php_execute.c
```diff
@@ -992,16 +991,16 @@ static void nr_php_execute_metadata_add_code_level_metrics(
...
    * Check if code level metrics are enabled in the ini.
    * If they aren't, exit and don't update CLM.
    */
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
+    || defined OVERWRITE_ZEND_EXECUTE_DATA
   if (!NRINI(code_level_metrics_enabled)) {
     return;
   }
-
+#endif
```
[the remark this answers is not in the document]
zsistla: PHP 8+ was using this function to also populate regular metadata(currently strdups) which nonoapi populates differently(via zrefs).  However, if we can reuse the zref methodology from this PR: https://github.com/newrelic/newrelic-php-agent/pull/585 which enhances the nonoapi metadata logic  to populate clm values, then the logic will change for all.

agent/php_execute.c
```diff
@@ -1688,12 +1658,135 @@ void nr_php_user_instrumentation_from_opcache(TSRMLS_D) {
...
+     * the currently executing segment.  If the pointers match, then the
+     * execute_data is still executing and could theoretically still catch it.
+     * If the pointers don't match, then the previous exception caused the
+     * fcall_end to be skipped, so we need to close those stacked segments
+     * manually until we arrive at the correct stacked segment that corresponds
+     * to exception we just recieved.  This will close all necessary stacked
+     * segments. If the previous exception had been caught anywhere along the
+     * calling chain (by an fcall_end happening for a function) the segments
+     * would have been closed and the exception cleared.
+     */
```
bduranleau-nr, line 1775: great comments explaining this issue and solution!

agent/php_execute.c
zsistla: Added comment: [dcd676d](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/dcd676dcb43e52f7d69a6334e6afb4aa8b537271)

bduranleau-nr: Awesome!

agent/php_execute.c
```diff
@@ -1715,48 +1823,143 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
    */
   if ((NULL != wraprec) || (NRINI(tt_detail) && NR_OP_ARRAY->function_name)) {
     /*
```
bduranleau-nr, line 1825: Since this case results in just a `return` if false, we can refactor it to return immediately and alleviate some of the `if` nesting inside it.
This should do the trick:
```c
  if ((NULL == wraprec)
      && (!NRINI(tt_detail) || NULL == NR_OP_ARRAY->function_name)) {
    return;
  }
```
Along with removing the `}` at the end of the function.

agent/php_execute.c
```diff
@@ -1715,48 +1823,143 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
...
+
+      nr_txn_force_single_count(NRPRG(txn), wraprec->supportability_metric);
+      /*
+       * Check for, and handle, frameworks.
+       */
+      if (wraprec->is_names_wt_simple) {
+        nr_txn_name_from_function(NRPRG(txn), wraprec->funcname,
+                                  wraprec->classname);
+      }
     }
```
bduranleau-nr, line 1933: This `if` can test for the inverse case and return if `true` to unnest conditionals.

```c
  if (NULL == wraprec) {
    return;
  }
  /*
   * If a function needs to have arguments modified this is the place.
   */
  segment->wraprec = wraprec;
  zcaught = nr_zend_call_oapi_special_before(wraprec, segment,
                                             NR_EXECUTE_ORIG_ARGS);
  if (nrunlikely(zcaught)) {
    zend_bailout();
  }

  nr_txn_force_single_count(NRPRG(txn), wraprec->supportability_metric);
  /*
   * Check for, and handle, frameworks.
   */
  if (wraprec->is_names_wt_simple) {
    nr_txn_name_from_function(NRPRG(txn), wraprec->funcname,
                              wraprec->classname);
  }
```

agent/php_execute.c
```diff
@@ -1715,48 +1823,143 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
...
   }
   return;
```
bduranleau-nr, line 1935: Since this function is void and nothing happens after this point, this `return` can be removed
```suggestion
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +155,397 @@
...
+ *
+ * But we aren't done yet.
+ * current stacked_segment->metadata->this still doesn't equal the
+ * execute_data->prev_execute_data->This provided by
+ * nr_php_observer_fcall_begin(D). We pop the current segment *B and apply the
+ * exception. Because it has an exception, the segment is kept so we copy the
+ * contents of the stacked segment *B into a segment b we obtained from the
+ * slab allocator, and we make b a child of the stacked segment *A which
+ * becomes the current segment.  Now current stacked_segment->metadata->this
+ * still DOES equal the execute_data->prev_execute_data->This provided by
```
bduranleau-nr, line 334:
```suggestion
 * DOES equal the execute_data->prev_execute_data->This provided by
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +155,397 @@
...
+ *
+ *
+ * But we aren't done yet.
+ * current stacked_segment->metadata->this still doesn't equal the
+ * execute_data-> this provided by nr_php_observer_fcall_end(A). We pop the
+ * current segment *B and apply the exception. Because it has an exception, the
+ * segment is kept so we copy the contents of the stacked segment *B into a
+ * segment b we obtained from the slab allocator, and we make b a child of the
+ * stacked segment *A which becomes the current segment.  Now current
+ * stacked_segment->metadata->this still DOES equal the execute_data-> this
```
bduranleau-nr, line 388:
```suggestion
 * stacked_segment->metadata->this DOES equal the execute_data-> this
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +155,397 @@
...
+ *   *B                          *B <
+ *   |                           |
+ *   *C <                        c
+ *
+ * nr_php_observer_fcall_begin(C) starts *C gets started as child
+ * of *B. Function C throws an uncaught exception which B does not catch so
+ * netiher nr_php_observer_fcall_end(B) nor nr_php_observer_fcall_end(C) is
+ * called and *C remains the current segment. A does not catch the exception,
+ * but the txn has ended. Because we didn't get any nr_php_observer_fcall_end we
+ * know no segment caught the exception. We'll apply the acception and
```
bduranleau-nr, line 419:
```suggestion
 * know no segment caught the exception. We'll apply the exception and
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +155,397 @@
...
+ *                               |                           |
+ *                               *B                          *B <
+ *                               |                           |
+ *                               *C <                        c
+ *
+ * nr_php_observer_fcall_begin(C) starts *C gets started as child
+ * of *B. Function C throws an uncaught exception which B does not catch so
+ * netiher nr_php_observer_fcall_end(B) nor nr_php_observer_fcall_end(C) is
+ * called and *C remains the current segment. B catches the exception throws
+ * another exception. At this point we realize the current
```
bduranleau-nr, line 466: For the line 
```c
B catches the exception throws another exception.
```
Is the exception `B` catches here a separate exception to the one thrown by `C`, since `B` did not catch `C`?

agent/php_stacked_segment.h
```diff
@@ -155,6 +155,397 @@
...
+ * an exception, the segment is kept so we copy the contents of the stacked
+ * segment *B into a segment b we obtained from the slab allocator, and we make
+ * b a child of the stacked segment *A which becomes the current segment.  The
+ * this` value of the function that called newrelic_notice_error and see it is
+ * not the same as the current segment `this` so it proceeds and applies the
+ * notice error to the current segment *A.
+ * Note that this only works with segments on the default parent stack.
+ * Stacked segments cannot be used to model async segments.
+ */
+
```
bduranleau-nr, line 548: This whole writeup/comment block is really great! Thanks for putting all these detailed cases together and using ASCII diagrams to demonstrate what is happening. Everything is explained very clearly.

agent/php_stacked_segment.h
zsistla: Please see: [2cb814c](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/2cb814c660a71927b9287e3474bb7425bd2f45dc)

bduranleau-nr (approved):

agent/php_execute.c
zsistla: Please see: 
[fed7f9c](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/fed7f9c6c7a980c3d8bb1d87f43f9218363fc105)

and 

[4c64593](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/4c645933f57d20c7395153e36b409cc2358b0463)

agent/php_execute.c
zsistla: Please see: 
[fed7f9c](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/fed7f9c6c7a980c3d8bb1d87f43f9218363fc105)

and 

[4c64593](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/4c645933f57d20c7395153e36b409cc2358b0463)

agent/php_execute.c
zsistla: Please see: 
[fed7f9c](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/fed7f9c6c7a980c3d8bb1d87f43f9218363fc105)

and 

[4c64593](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/4c645933f57d20c7395153e36b409cc2358b0463)

agent/php_agent.h
```diff
@@ -863,8 +880,28 @@ extern void nr_php_txn_add_code_level_metrics(
...
+  function_name = execute_data->func->common.function_name;
+
+  if ((NULL == function_name)
+      && (ZEND_USER_FUNCTION == execute_data->func->type)) {
+    /*
+     * This is the case of a filename being called so there is no function name.
+     * It is broken out separately here in case we need to do something special
+     * with it in the future.
+     */
+    return NULL;
```
ZNeumann, line 900: For performance gains in the present, do we want to comment out this entire block?

ZNeumann (approved):

agent/php_observer.c
```diff
@@ -104,4 +113,23 @@ void nr_php_observer_minit() {
...
 
+void nr_throw_exception_hook(zend_object* exception) {
+  zval new_exception;
+  zval* exception_zval = NULL;
+
+  /*
+   * On PHP 7, EG(exception) is stored as a zend_object, and is only
+   * wrapped in a zval when it actually needs to be.
+   */
+  ZVAL_OBJ(&new_exception, exception);
```
mfulb, line 124: I'm not clear if the code here is handling PHP 7 differently (based on comment) than other PHP versions since there is no conditional code in this function.  Also all of this code is in a conditional block for PHP >= 8.0 so maybe the comment needs to be adjusted for this code block?

mfulb:
I think some documentation at the top of `php_api.c` would be good to explain when `    nr_php_api_ensure_current_segment()` need to be called for the benefit of future developers working on this file.

zsistla:
ok jenkins

zsistla:
ok jenkins

agent/php_observer.c
zsistla: Please see pr [e98d885](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/e98d8852f20913be01e41faec16fb1ee794ed576)

zsistla:
> I think some documentation at the top of `php_api.c` would be good to explain when `nr_php_api_ensure_current_segment()` need to be called for the benefit of future developers working on this file.

please see [c8a008b](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/c8a008b5bee3e6174f21513e18d04601bfb53b6d)

zsistla:
With OAPI exceptions, the registered function handler (nr_php_observer_fcall_end) doesn't get called when an uncaught exception occurs, and therefore doesn't decrement the stack_depth counter.  
1) All OAPI unhandled exception cleanup filters through: `nr_php_observer_segment_end` so we decrement `php_cur_stack_depth` there when we cleanup orphaned segments.  
2) Additionally, since `nr_php_observer_segment_end` is an exit path, also call `nr_php_show_oapi_metadata` 
3) New function `nr_php_show_oapi_metadata` called via the segment exception handling exit path (to correspond to nr_php_show_exec_return) to show the all the available function details when the special_flags.show_execute_* is toggled on.  This will help when debugging.
4) Added additional test cases to ensure proper php_cur_stack_depth counting


Above additions are in these commits:
@zsistla
[fix(agent): properly account for php_cur_stack_depth when closing dan…](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/2f2f0d05e1cc9364dfd720cb1bc33e5cae15af00) 
[2f2f0d0](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/2f2f0d05e1cc9364dfd720cb1bc33e5cae15af00)
@zsistla
[fix(agent): properly handle php_cur_stack_depth for OAPI exceptions.](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/e9c2418c277c9c8063cf4c51a57b3333c3a8ad55) 
[e9c2418](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/e9c2418c277c9c8063cf4c51a57b3333c3a8ad55)
@zsistla
[tests(unit): add additional tests to verify stack depth counter.](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/143851d33d5745a4555932de56f863359abd88a4)
[143851d](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/143851d33d5745a4555932de56f863359abd88a4)

agent/php_observer.c
mfulb: Thank you!

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
  * Stacked segments cannot be used to model async segments.
  */
 
+/*
```
misiekn, line 167: I suggest disabling formatting this comment in case clang-format ever decides to adjust indentation in the cartoons (which would result in them being not readable anymore). This is probably what happened to [this comment](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_internal_instrument.c#L386-L400) and [this comment](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_internal_instrument.c#L3285-L3290).
```suggestion
// clang-format off
/*
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
...
+ * `this`. We pop the current segment *B and apply the exception. Because it has
+ * an exception, the segment is kept so we copy the contents of the stacked
+ * segment *B into a segment b we obtained from the slab allocator, and we make
+ * b a child of the stacked segment *A which becomes the current segment.  The
+ * this` value of the function that called newrelic_notice_error and see it is
+ * not the same as the current segment `this` so it proceeds and applies the
+ * notice error to the current segment *A.
+ * Note that this only works with segments on the default parent stack.
+ * Stacked segments cannot be used to model async segments.
+ */
```
misiekn, line 545:
```suggestion
 */
 // clang-format on
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
...
+ *
+ * But we aren't done yet.
+ * current stacked_segment->metadata->this still doesn't equal the
+ * execute_data-> this provided by nr_php_observer_fcall_end(A). We pop the
+ * current segment *B and apply the exception. Because it has an exception, the
+ * segment is kept so we copy the contents of the stacked segment *B into a
+ * segment b we obtained from the slab allocator, and we make b a child of the
+ * stacked segment *A which becomes the current segment.  Now current
+ * stacked_segment->metadata->this DOES equal the execute_data-> this
+ * provided by nr_php_observer_fcall_end(A) so it proceeds, decides to keep the
```
misiekn, line 387: Does the decision to keep segment `A` is based on `A` having kept child segment `b` (it was kept because it threw exceptions)? Or having child segment `b` is irrelevant and `A` can get discarded? If `A` will get discarded, will `b` be correctly re-parented to root segment?

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
...
+ *                               |                           |
+ *                               *B                          *B <
+ *                               |                           |
+ *                               *C <                        c
+ *
+ * nr_php_observer_fcall_begin(C) starts *C gets started as child
+ * of *B. Function C throws an exception which B catches but 
+ * nr_php_observer_fcall_end(C) is not called so *C remains the 
+ * current segment. B catches the exception and throws another 
+ * exception which triggers the exception hook. At this point we realize the current
```
misiekn, line 464: Why isn't the exception hook triggered in case 3, which also deals with an exception that's not caught by anything up to the root? It is a little murky for me when exception hook is called by Zend.

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
...
+ *   |                           |
+ *   *B                          *B <
+ *   |                           |
+ *   *C <                        c
+ *
+ * nr_php_observer_fcall_begin(C) starts *C gets started as child
+ * of *B. Function C throws an uncaught exception which B does not catch so
+ * neither nr_php_observer_fcall_end(B) nor nr_php_observer_fcall_end(C) is
+ * called and *C remains the current segment. A catches the exception and calls
+ * newrelic_notice_error. We check the `this` value of the function that called
```
misiekn, line 512: The `this` value check happens at the beginning of the `newrelic_notice_error`, yes? If so then consider making it clear:
```suggestion
 * newrelic_notice_error. In newrelic_notice_error we check the `this` value of the function that called
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
...
+ *   |
+ *   *A <
+ *    |
+ *    b
+ *    |
+ *    c
+ *
+ *
+ * But we aren't done yet.
+ * The `this` value of the function that called
```
misiekn, line 534:
```suggestion
 * We check the `this` value of the function that called
```

agent/php_stacked_segment.h
```diff
@@ -155,6 +164,386 @@
...
+ * newrelic_notice_error and see it is not the same as the current segment
+ * `this`. We pop the current segment *B and apply the exception. Because it has
+ * an exception, the segment is kept so we copy the contents of the stacked
+ * segment *B into a segment b we obtained from the slab allocator, and we make
+ * b a child of the stacked segment *A which becomes the current segment.  The
+ * this` value of the function that called newrelic_notice_error and see it is
+ * not the same as the current segment `this` so it proceeds and applies the
```
misiekn, line 541:
```suggestion
 * b a child of the stacked segment *A which becomes the current segment.  We
 * check the this` value of the function that called newrelic_notice_error and see it is
 * the same as the current segment `this` so newrelic_notice_error proceeds and applies the
```

agent/tests/test_php_execute.c
```diff
@@ -101,20 +101,238 @@ static void test_php_cur_stack_depth(TSRMLS_D) {
...
+  expr = nr_php_call(NULL, "uncaught", arg);
+  tlib_pass_if_int_equal(
+      "PHP stack depth tracking should be 3 after function call", 3,
+      NRPRG(php_cur_stack_depth));
+  tlib_pass_if_null("Exception so expr should be null.", expr);
+
+  /*
+   * Trigger the unwind.
+   */
+  tlib_php_request_eval("newrelic_end_transaction(); ");
```
misiekn, line 201: What happens in a PHP application that doesn't call `newrelic_end_transaction();`? I guess `nr_php_txn_end` will get called eventually (at RSHUTDOWN?) which will call `nr_php_stacked_segment_unwind`, yes?

agent/tests/test_php_execute.c
```diff
@@ -101,20 +101,238 @@ static void test_php_cur_stack_depth(TSRMLS_D) {
...
+      "PHP stack depth tracking should be 0 after transaction ends", 0,
+      NRPRG(php_cur_stack_depth));
+  nr_php_zval_free(&expr);
+  nr_php_zval_free(&arg);
+  tlib_php_request_end();
+
+  /*
+   * call a function and trigger an exception that is caught but another
+   * uncaught exception is thrown and causes two segments to dangle stack_depth
+   * should be initially stuck at 2 after unwind, stack_depth should be zero
```
misiekn, line 277:
```suggestion
   * should be initially stuck at 2 but after unwind, stack_depth should be zero
```

agent/tests/test_php_execute.c
```diff
@@ -101,20 +101,238 @@ static void test_php_cur_stack_depth(TSRMLS_D) {
...
+      "PHP stack depth tracking should be 0 after transaction ends", 0,
+      NRPRG(php_cur_stack_depth));
+  nr_php_zval_free(&expr);
+  nr_php_zval_free(&arg);
+  tlib_php_request_end();
+
+  /*
+   * call a function and trigger an exception that is caught then rethrown
+   * stack_depth should be initially stuck at 2
+   * after unwind, stack_depth should be zero again
```
misiekn, line 308:
```suggestion
   * but after unwind, stack_depth should be zero again
```

misiekn (approved):
Very nice unit and integration tests! Thank you for those! 👍

agent/php_stacked_segment.h
zsistla: excellent point!

agent/tests/test_php_execute.c
zsistla: Yes, it gets called automatically here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_txn.c#L1053 if a txn is detected as active at the end of rshutdown: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_rshutdown.c#L121
Using newrelic_end_transaction was a way for me to force it to occur while still retaining request state that I could query and check for correct values.

agent/php_stacked_segment.h
zsistla: None of the APIs do this checking explicitly.  The API functions call `nr_php_api_ensure_current_segment` (before doing anything related to segments).  `nr_php_api_ensure_current_segment` passes the `This` pointer of the calling function to `nr_php_observer_handle_uncaught_exception` which calls [nr_php_observer_handle_uncaught_exception](https://github.com/newrelic/newrelic-php-agent/pull/580/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1864) which does the checking in php_execute.c.  `nr_php_observer_handle_uncaught_exception` is the function that all    code paths will call when they detect dangling segments due to exceptions.  Please see clarification here: [80f1acd](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/80f1acd0406ea78a52ac0bb0bf8ebc604bad9903)

agent/php_stacked_segment.h
zsistla: The thing to remember about the exception hook is it is called EVERY time an exception occurs but it has no idea if that exception is caught or not so we need to have the extra functionality to check `This` pointers before cleaning up segments.  This specifically calls out that the exception hook is triggered because we want to do any segment cleanup related to the old exception before we record the new exception that was triggered.  In this way we only have to keep track of one exception at a time.

zsistla: Tried to clarify case 3 here: [7be484d](https://github.com/newrelic/newrelic-php-agent/pull/580/commits/7be484d69eb7a10f65828220a8d6a2a27fa9b99b)

agent/php_stacked_segment.h
zsistla: `A` can possibly get discarded.  The decision to keep/discard segments is always determined in this function: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1265 

If `A` is discarded, `b` will be reparented to root.

zsistla: The "decides to keep" was to indicate for the the example that `A` passed the checks in [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1265) and was kept and that segment A was chosen; otherwise, it would never be moved to the heap.

PR #583, opened 2022-11-16
tests(agent): Add tests to exercise JIT.

zsistla:


ZNeumann (approved):

zsistla:
ok jenkins

zsistla:
ok jenkins

zsistla:
ok jenkins

tests/integration/jit/function/test_computations.php
```diff
@@ -0,0 +1,75 @@
...
+
+/*INI
+newrelic.distributed_tracing_enabled=1
+newrelic.transaction_tracer.threshold = 0
+error_reporting = E_ALL
+opcache.enable=1
+opcache.enable_cli=1
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=function
```
mfulb, line 27: The default appears to be 'tracing/on' instead of 'function'.  Is there a reason we change this from what might be the most common value for our users?

mfulb: Ok I think get it there is a 'function' and 'tracing' directory for testing each mode?

tests/integration/jit/tracing/test_computations.php
```diff
@@ -0,0 +1,73 @@
...
+
+/*INI
+newrelic.distributed_tracing_enabled=1
+newrelic.transaction_tracer.threshold = 0
+error_reporting = E_ALL
+opcache.enable=1
+opcache.enable_cli=1
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=function
```
mfulb, line 26: I think all of these in the `tracing` directory need to have `opcache.jit` set to `tracing` instead of `function`.

mfulb (approved):

mfulb:
ok jenkins

zsistla:
ok jenkins

tests/integration/jit/function/skipif.inc
```diff
@@ -0,0 +1,10 @@
+<?php
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+if (version_compare(PHP_VERSION, "8.0", "<")) {
+  die("skip: PHP < 8.0.0 not supported\n");
+}
```
lavarou, line 10:
```suggestion
}
if (!extension_loaded('Zend OPcache')) {
  die("skip: Zend OPcache extension required\n");
}
```
These tests require `Zend OPcache` extension to be available and loaded. Only then Zend engine is capable of detecting incompatibility and issuing this warning:
```
Warning: JIT is incompatible with third party extensions that override zend_execute_ex(). JIT disabled. in Unknown on line 0
```
if it detects that `newrelic.so` under test is not using Observer API to hook into PHP. Only this warning will cause the tests to fail. Otherwise all tests pass regardless if instrumentation is done via `zend_execute_ex()` hook override or via Observer API. I think these tests should prove that the agent is compatible with Zend's OPcache JIT and it's best if they fail when they're executed with an agent without Observer API or Zend detects other incompatibility.

tests/integration/jit/tracing/skipif.inc
```diff
@@ -0,0 +1,10 @@
+<?php
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+if (version_compare(PHP_VERSION, "8.0", "<")) {
+  die("skip: PHP < 8.0.0 not supported\n");
+}
```
lavarou, line 10:
```suggestion
}
if (!extension_loaded('Zend OPcache')) {
  die("skip: Zend OPcache extension required\n");
}
```

lavarou (changes requested):
I think these tests should prove that the agent is compatible with Zend's OPcache JIT and it's best if they fail when they're executed with an agent without Observer API or Zend OPcache detects other incompatibility and issues other warnings. This is only possible if the tests execute with an environment that has `Zend OPcache` extension loaded. If `Zend OPcache` extension is not loaded, these tests should either be skipped or fail.

tests/integration/jit/tracing/skipif.inc
mfulb: I made this change in commit [b2a47c](https://github.com/newrelic/newrelic-php-agent/pull/583/commits/b2a47cf5a9fec76c1557d24f184983559db66fef).

After ensuring ```opcache.so``` was loaded the following tests now fail (pass w/o opcache):

tests/integration/jit/tracing/test_span_class_function.php
tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php

tests/integration/jit/function/skipif.inc
mfulb: I made this change in commit [b2a47c](https://github.com/newrelic/newrelic-php-agent/pull/583/commits/b2a47cf5a9fec76c1557d24f184983559db66fef).

After ensuring ```opcache.so``` was loaded the following tests now fail (pass w/o opcache):

tests/integration/jit/tracing/test_span_class_function.php
tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php

lavarou (approved):

mfulb (changes requested):
The JIT integration test failures that turned up when opcache was properly used need to be investigated and addressed.

zsistla:
ok jenkins

bduranleau-nr (approved):

zsistla:
ok jenkins

zsistla:
Closing this forked branch in lieu of main repo branch: https://github.com/newrelic/newrelic-php-agent/pull/715

PR #513, opened 2022-08-25
feat(agent): Add Custom Event Configuration Options - For reference only see PR 584 for final work

hahuja2:
Work proceeds in PR #584 - this is for reference only.

- Add Custom Event config options to INI file
- Increase default Custom Events Harvest Limit from 10,000 to 30,000 events per minute
- Set NR_MAX_CUSTOM_EVENTS_MAX_SAMPLES_STORED macro to 100,000

PR #584, opened 2022-11-16
feat(agent): Add Custom Event Configuration Options

mfulb:
Based on work by @hahuja2 (PR #513).

- Add Custom Event config options to INI file
- Increase default Custom Events Harvest Limit from 10,000 to 30,000 events per minute
- Set NR_MAX_CUSTOM_EVENTS_MAX_SAMPLES_STORED macro to 100,000


agent/scripts/newrelic.ini.template
```diff
@@ -834,6 +834,15 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;
 ;newrelic.custom_insights_events.enabled = true
 
+; Setting: newrelic.custom_events.max_samples_stored
+; Type   : integer
+; Scope  : per-directory
+; Default: 30000
+; Info   : The default Custom Events reservoir limit in the agent is increased
+;          from 10,000 events per minute to 30,000 events per minute.
```
zsistla, line 842: What is the maximum?

mfulb: The maximum is 100000.

zsistla: Please note that in the ini template.

Makefile
```diff
@@ -315,13 +315,74 @@ src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go: protocol/infinite_
...
+.PHONY: integration-events-limits
+integration-events-limits: daemon
+	# create array with the collector response for each agent requested custom events max samples
+	# currently based on fast harvest cycle being 5 seconds so ratio is 12:1
+	declare -A custom_limits_tests; \
+	custom_limits_tests[240]=20; \
+	custom_limits_tests[7000]=583; \
+	custom_limits_tests[30000]=2500; \
+	custom_limits_tests[100000]=8333; \
+	for PHP in $${PHPS:-8.1 8.0 7.4 7.3 7.2 7.1 7.0 5.6 5.5}; do \
```
ZNeumann, line 345: Is this something we can extract out to a single place (function/variable)? It's added a lot here (and is already used extensively in the Makefile), but it would be nice to be able to configure the PHPS list in one place

src/newrelic/processor_test.go
```diff
@@ -1409,13 +1421,18 @@ func runMockedCollectorHarvestLimitTest(t *testing.T, eventType string, agentLim
...
+	// defaults that collector will send if no harvest limit requested in connect request
 	logHarvestLimit := 833
+	customHarvestLimit := 2500
 
 	switch eventType {
 	case "log_event_data":
 		logHarvestLimit = int(collectorLimit)
 		appInfo = NewAppInfoWithLogEventLimit(int(agentLimit))
-
+	case "custom_event_data":
```
ZNeumann, line 1433: Can we update the comment on line 1411 to reflect this addition?

src/newrelic/processor_test.go
```diff
@@ -1471,11 +1497,31 @@ func runMockedCollectorHarvestLimitTest(t *testing.T, eventType string, agentLim
...
+	runMockedCollectorHarvestLimitTest(t, "log_event_data", 100*12, 0, "agent log limit != 0, collector == 0")
+
+}
+
+func TestConnectNegotiateCustomEventLimits(t *testing.T) {
+
+	runMockedCollectorHarvestLimitTest(t, "custom_event_data", 110*12, 100, "agent custom limit larger than collector")
+	runMockedCollectorHarvestLimitTest(t, "custom_event_data", 100*12, 100, "agent custom limit equal to collector")
+
+	// these tests exist for the log events (TestConnectNegotiateLogEventsLimit()) because at some point
```
ZNeumann, line 1516:
```suggestion
	// these tests exist for the custom events (TestConnectNegotiateLogEventsLimit()) because at some point
```

ZNeumann (approved):

Makefile
mfulb: Good idea - I'd prefer not to do that refactoring in this particular PR but maybe we can add it to a list of build/CI/CD improvements we want to do?

src/newrelic/processor_test.go
mfulb: Updated in commit  [c1c0d8](https://github.com/newrelic/newrelic-php-agent/pull/584/commits/c1c0d8a61f54707563418906303dd952ecbfd592).

src/newrelic/processor_test.go
mfulb: It is a confusing comment but it is correct - it is attempting to explain why I had commented out the tests below this comment which for logging are still run.  I left them for reference but they could possibly be removed.

ZNeumann: Ah gotcha. Could we remove the commented out tests and simplify this comment to just reference the log tests above this?

mfulb: I made this change in commit  [12eb8f](https://github.com/newrelic/newrelic-php-agent/pull/584/commits/12eb8fb5b0f1a6a04dd5341fa11341a573c3a660).

mfulb:
ok jenkins

Makefile
```diff
@@ -315,13 +315,74 @@ src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go: protocol/infinite_
...
 	  env NRLAMP_PHP=$${PHP} bin/integration_runner $(INTEGRATION_ARGS) || exit 1; \
 	  echo "# PHP=$${PHP}"; \
 	done
 
+#
+# Agent event limits integration testing
+#
+# Because of how the integration_runner connects to the collector once before
+# running any tests we have to tell the runner the value the agent would
+# have passed to it for each test.  This means these tests are not testing
```
misiekn, line 330:
```suggestion
# Because of how the integration_runner connects to the collector (once before
# running any tests) we have to tell the runner the value the agent would
# have passed to it for each test.  This means these tests are not testing
```

src/newrelic/limits/limits.go
```diff
@@ -34,7 +34,7 @@ const (
 	DefaultReportPeriod          = 60 * time.Second
 	MaxMetrics                   = 2 * 1000
 	MaxTxnEvents                 = 10 * 1000
-	MaxCustomEvents              = 10 * 1000
+	MaxCustomMaxEvents           = 100 * 1000
```
misiekn, line 37: Not something to be addressed before merge but somehow `MaxCustomMaxEvents` with two `Max` does not clearly convey the purpose of this constant. I guess it has been following `MaxSpanMaxEvents` and `MaxLogMaxEvents` but there also is `MaxTxnEvents` which IMHO better matches the intention.

misiekn (approved):
Nice work! 👍

ZNeumann (approved):

mfulb:
ok jenkins

zsistla (approved):
Thanks for getting this over the finish line!

PR #585, opened 2022-11-22
feat(agent): Add code level metrics functionality.

zsistla:
* feat(agent): Add code level metrics functionality. 
1) Add INI value to disable/enable code level metrics. 
2) Create new function `nr_php_txn_add_code_level_metrics` that uses nr_php_zend_execute_data_* family of functions to extract CLM and add it as an `agent attribute`. 
3) Added new NR_TXN_ATTRIBUTES for `code.namespace`, `code.lineno`, `code.filepath`, and `code.function`. 
4) Added call to php_execute_enabled to call the new CLM function 
5) Added new integration tests to exercise the new functionality. 
6) updated `nr_php_execute_metadata_t` to hold additional information and moved the definition of the struct to php_execute.h. 
7) `nr_php_execute_enabled` uses the metadata at the beginning so now we can simply release at the end. 
8) In the case of a super short segment that would have been ignored, we do NOT add CLM (otherwise it wouldn't be ignored).  We wait until after we decide to ignore or not to add the CLM data. 
9) check clm string lengths so don't add attributes if the string will be truncated

Note:  CLM functionality is only compatible with PHP 7+.

ETA:
10) tailored CLM to current metadata utilizing zrefs instead of strdups; otherwise, it was adding too much overhead.


bduranleau-nr (approved):

zsistla:
ok jenkins

agent/php_execute.c
```diff
@@ -1126,6 +1114,17 @@ static inline void nr_php_execute_segment_end(
...
+      /*
+   * Check if code level metrics are enabled in the ini.
+   * If they aren't, exit and don't add any attributes.
+   */
+  if (NRINI(code_level_metrics_enabled)) {
+    if (NULL == s->attributes) {
+      s->attributes = nr_attributes_create(s->txn->attribute_config);
+    }
+      nr_php_txn_add_code_level_metrics(s->attributes, metadata);
+  }
```
ZNeumann, line 1127: minor formatting issues in this block

agent/php_txn.c
```diff
@@ -1109,3 +1109,93 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
 
   return NR_SUCCESS;
 }
+
+extern void nr_php_txn_add_code_level_metrics(
+    nr_attributes_t* attributes,
+    const nr_php_execute_metadata_t* metadata) {
+#if ZEND_MODULE_API_NO < ZEND_7_0_X_API_NO /* PHP7+ */
```
ZNeumann, line 1116: Version comment mismatch

agent/php_txn.c
```diff
@@ -1109,3 +1109,93 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
...
+#else
+  /*
+   * Check if code level metrics are enabled in the ini.
+   * If they aren't, exit and don't add any attributes.
+   */
+  if (!NRINI(code_level_metrics_enabled)) {
+    return;
+  }
+
+  /* Current CLM functionality only works with PHP 7+ */
```
ZNeumann, line 1130: This comment seems to be in the wrong place

ZNeumann (approved):

agent/php_execute.c
zsistla: Please see [fdbf4a2](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/fdbf4a2df8224ef2425a5e6a56291f374f33a645)

agent/php_txn.c
zsistla: Please see [fdbf4a2](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/fdbf4a2df8224ef2425a5e6a56291f374f33a645)

agent/php_txn.c
zsistla: Please see [fdbf4a2](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/fdbf4a2df8224ef2425a5e6a56291f374f33a645)

agent/php_txn.c
```diff
@@ -1109,3 +1109,92 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
...
+
+#define CHK_CLM_STRLEN(s)                         \
+  if (CLM_STRLEN_MAX < NRSAFELEN(nr_strlen(s))) { \
+    s = NULL;                                     \
+  }
+
+  CHK_CLM_STRLEN(filepath)
+  CHK_CLM_STRLEN(namespace)
+  CHK_CLM_STRLEN(function)
+
```
mfulb, line 1162: Since these strings are originated as Zend values do we not already have the length from the Zend engine?  Might save some time.

zsistla: Please see [8c2a251](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/8c2a251d0ac61bb3b9868ce1297f0095540aa0d3)

zsistla:
ok jenkins

agent/php_agent.h
```diff
@@ -791,8 +793,114 @@ extern bool nr_php_function_is_static_method(const zend_function* func);
...
+ * Returns : A pointer to string, ownership of does NOT pass to the caller and
+ *           string must be dupped if it needs to persist,
+ *           or NULL if the zend_execute_data is invalid.
+ *
+ */
+static inline const char* nr_php_zend_execute_data_filename(
+    const zend_execute_data* execute_data) {
+  zend_string* filename = NULL;
+
+  while (NR_ZEND_USER_FUNC_EXISTS(execute_data)) {
```
mfulb, line 867: Based on comment for `NR_ZEND_USER_FUNC_EXISTS()` above it seems this loop is really looping until a user function is found?  
It is just nomenclature but just wanting to understand if I follow the flow correctly.

agent/php_agent.h
```diff
@@ -791,8 +793,114 @@ extern bool nr_php_function_is_static_method(const zend_function* func);
...
+ * Returns : A pointer to string, ownership of does NOT pass to the caller and
+ *           string must be dupped if it needs to persist,
+ *           or NULL if the zend_execute_data is invalid.
+ *
+ */
+static inline const char* nr_php_zend_execute_data_scope_name(
+    const zend_execute_data* execute_data) {
+  zend_class_entry* ce = NULL;
+
+  while (execute_data) {
```
mfulb, line 890: How deep can this walk of execute_data get?  Just wondering if it would be possible to walk execute_data and get filename (like in `nr_php_zend_execute_data_filename()`) and scope from a single pass?

agent/php_agent.h
```diff
@@ -791,8 +793,114 @@ extern bool nr_php_function_is_static_method(const zend_function* func);
...
+ *       Current CLM functionality only works with PHP 7+
+ */
+extern void nr_php_txn_add_code_level_metrics(
+    nr_attributes_t* attributes,
+    const nr_php_execute_metadata_t* metadata);
+
 #if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP7+ */
 
+#define NR_ZEND_USER_FUNC_EXISTS(x) \
+  (x && (!x->func || !ZEND_USER_CODE(x->func->type)))
```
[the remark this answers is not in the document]
zsistla: renamed for clarity: [83dd96f](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/83dd96f1240f4d3225d834aa8586b717fc88c3e5)

agent/php_agent.h
zsistla: renamed macro for clarity [83dd96f](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/83dd96f1240f4d3225d834aa8586b717fc88c3e5)

agent/php_agent.h
mfulb: I was asking if both pieces could be extracted simultaneously from a single loop through the execute data linked list  but it sounds like there wouldn't be much performance benefit.

agent/php_agent.h
```diff
@@ -791,8 +793,113 @@ extern bool nr_php_function_is_static_method(const zend_function* func);
...
+ */
+static inline const char* nr_php_zend_execute_data_scope_name(
+    const zend_execute_data* execute_data) {
+  zend_class_entry* ce = NULL;
+
+  while (execute_data) {
+    if (execute_data->func
+        && (ZEND_USER_CODE(execute_data->func->type)
+            || execute_data->func->common.scope)) {
+      ce = execute_data->func->common.scope;
```
mfulb, line 894: Would `ZEND_USER_CODE(execute_data->func->type)` ever be True and `execute_data->func->common.scope` be NULL?  As written this code would set `ce = NULL` in this case.  Is that ok?

zsistla: Yes, when a user function doesn't have a class, zend_user_code would be true, but common.scope would be null.

agent/php_execute.c
```diff
@@ -1013,7 +977,7 @@ typedef struct {
  */
 static void nr_php_execute_metadata_init(nr_php_execute_metadata_t* metadata,
                                          zend_op_array* op_array) {
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP7+ */
   if (op_array->scope && op_array->scope->name && op_array->scope->name->len) {
     metadata->scope = op_array->scope->name;
```
mfulb, line 982: Could we use `zend_copy_string()` to simplify code?

agent/php_execute.c
```diff
@@ -1027,6 +991,20 @@ static void nr_php_execute_metadata_init(nr_php_execute_metadata_t* metadata,
   } else {
     metadata->function = NULL;
   }
+  if (!NRINI(code_level_metrics_enabled)
+      || ZEND_USER_FUNCTION != op_array->type) {
+    metadata->filepath = NULL;
+    return;
+  }
+  if (op_array->filename && op_array->filename->len) {
+    metadata->filepath = op_array->filename;
```
mfulb, line 1000: Could we use `zend_copy_string()` to simplify code?

agent/php_execute.c
```diff
@@ -1050,7 +1028,7 @@ static void nr_php_execute_metadata_metric(
   const char* function_name;
   const char* scope_name;
 
-#ifdef PHP7
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP7+ */
   scope_name = metadata->scope ? ZSTR_VAL(metadata->scope) : NULL;
   function_name = metadata->function ? ZSTR_VAL(metadata->function) : NULL;
 #else
```
mfulb, line 1034: Is it OK if scope_name or function_name are NULL for the `snprintf()` on line1039 or could that cause a exception depending on how `snprintf()` is implemented?

zsistla: It's okay, it handles the null.  That is legacy code.

agent/php_execute.c
zsistla: Thanks for keeping an eye out for possible simplifications.

It's possible we could call zend_string_copy instead of zend_string_addref but it wouldn't simplify the code since all other lines would need to remain the same.  Also, changing the current implementation that uses addref and that affects all user functions that come through php_execute_enabled and the testing effort involved seems beyond the scope of this PR.

```
static zend_always_inline zend_string *zend_string_copy(zend_string *s)
{
        if (!ZSTR_IS_INTERNED(s)) {
                GC_ADDREF(s);
        }
        return s;
}
```

vs 

```
static zend_always_inline uint32_t zend_string_addref(zend_string *s)
{
        if (!ZSTR_IS_INTERNED(s)) {
                return GC_ADDREF(s);
        }
        return 1;
}
```

agent/php_txn.c
mfulb: Looks good!

axiom/nr_txn.c
```diff
@@ -157,6 +173,32 @@ void nr_txn_set_long_attribute(nrtxn_t* txn,
...
+  if ('\0' == value[0]) {
+    return;
+  }
+  nr_attributes_agent_add_string(attributes, attribute->destinations,
+                                 attribute->name, value);
+}
+
+void nr_txn_attributes_set_long_attribute(nr_attributes_t* attributes,
+                                          const nr_txn_attribute_t* attribute,
+                                          long value) {
```
mfulb, line 194: Would it make sense for this to be `const long value`?

axiom/nr_txn.c
```diff
@@ -157,6 +173,32 @@ void nr_txn_set_long_attribute(nrtxn_t* txn,
                                attribute->name, value);
 }
 
+void nr_txn_attributes_set_string_attribute(nr_attributes_t* attributes,
```
mfulb, line 176: Could this be tested with a unit test to verify it handles bad inputs, etc?

axiom/nr_txn.c
```diff
@@ -157,6 +173,32 @@ void nr_txn_set_long_attribute(nrtxn_t* txn,
...
+    return;
+  }
+  if ('\0' == value[0]) {
+    return;
+  }
+  nr_attributes_agent_add_string(attributes, attribute->destinations,
+                                 attribute->name, value);
+}
+
+void nr_txn_attributes_set_long_attribute(nr_attributes_t* attributes,
```
mfulb, line 192: Does it make sense to have a unit test to exercise invalid values?

agent/php_execute.c
zsistla: Same comment: https://github.com/newrelic/newrelic-php-agent/pull/585#discussion_r1042762736

mfulb:
I believe we need to have `newrelic.template.ini` included in this PR to include the new CLM INI variable.

agent/php_execute.c
```diff
@@ -1126,6 +1111,17 @@ static inline void nr_php_execute_segment_end(
...
+    /*
+     * Check if code level metrics are enabled in the ini.
+     * If they aren't, exit and don't add any attributes.
+     */
+    if (NRINI(code_level_metrics_enabled)) {
+      if (NULL == s->attributes) {
+        s->attributes = nr_attributes_create(s->txn->attribute_config);
+      }
+      nr_php_txn_add_code_level_metrics(s->attributes, metadata);
+    }
```
misiekn, line 1124: What if:
1. `nr_php_txn_add_code_level_metrics` would be renamed to `nr_php_execute_segment_add_clm_attributes`. The latter name seems to match functions purpose - the agent is adding clm attributes from execute metadata to the segment. Also CLM attributes apply more to the segment/span rather than transaction as a whole, which the first name might suggest.
2. Have `nr_php_execute_segment_add_clm_attributes` take the `s` and `metadata` as arguments to be more in line with what it's actually doing - adding attributes to the segment rather than having addition of attributes split: `nr_attributes_create` is called outside but the attributes are updated inside. Let's couple what shouldn't be separated.
3. Define `nr_php_execute_segment_add_clm_attributes` only for PHP7+. This would simplify this block of code  [here](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/agent/php_txn.c#L1117-L1122).
5. Move the definition of `nr_php_execute_segment_add_clm_attributes` from php_txn.c to php_execute.c and make it static - no other code uses this function. This would make the code base more focused. Additionally this would make the extraction of `nr_php_execute_metadata_t` unnecessary.
After these changes the lines above could read like this:
```
  #if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP >= PHP7 */
    if (NRINI(code_level_metrics_enabled)) {
      nr_php_execute_segment_add_clm_attributes(s, metadata);
    }
#endif
```
You can have a look at a more complete suggestion [here](https://github.com/zsistla/newrelic-php-agent/pull/2).

agent/php_agent.h
```diff
@@ -791,6 +793,25 @@ extern bool nr_php_function_is_static_method(const zend_function* func);
  */
 extern zend_execute_data* nr_get_zend_execute_data(NR_EXECUTE_PROTO TSRMLS_DC);
 
+/*
+ * Purpose : If code level metrics are enabled, extract the data segment
+ * metadata. Add the CLM as agent attributes to the attributes data structure.
```
misiekn, line 798: I don't think this description is accurate. How about something like this:
```suggestion
 * Purpose : If code level metrics are enabled, use the execute metadata as CLM agent attributes in the segment.
```

agent/php_txn.c
```diff
@@ -1109,3 +1109,101 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
...
+extern void nr_php_txn_add_code_level_metrics(
+    nr_attributes_t* attributes,
+    const nr_php_execute_metadata_t* metadata) {
+/* Current CLM functionality only works with PHP 7+ */
+#if ZEND_MODULE_API_NO < ZEND_7_0_X_API_NO /* PHP < PHP7 */
+  (void)attributes;
+  (void)metadata;
+  return;
+}
+#else
```
misiekn, line 1122: Why not make whole function available only for PHP 7+? See [this comment](https://github.com/newrelic/newrelic-php-agent/pull/585/files?diff=split&w=1#r1043702320).

agent/php_execute.c
```diff
@@ -1126,6 +1111,17 @@ static inline void nr_php_execute_segment_end(
...
     nr_segment_t* s = nr_php_stacked_segment_move_to_heap(stacked TSRMLS_CC);
     nr_php_execute_segment_add_metric(s, metadata, create_metric);
+
+    /*
+     * Check if code level metrics are enabled in the ini.
+     * If they aren't, exit and don't add any attributes.
+     */
+    if (NRINI(code_level_metrics_enabled)) {
+      if (NULL == s->attributes) {
+        s->attributes = nr_attributes_create(s->txn->attribute_config);
```
misiekn, line 1121: How likely it is that attributes will have to be created here? I ask because they potentially get created (malloc'ed) here and maybe not even used if [this condition](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/agent/php_txn.c#L1181) or [this condition](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/agent/php_txn.c#L1187) are not met. Since this will get called for every user function execution, when CLM is enabled, maybe the malloc could be postponed until it's absolutely necessary?

zsistla: I'll see if I can move it into the nr_php_txn_add_code_level_metrics function to after we do our valid clm checks.  It's very likely to need to be created but it's only called for the segments we keep not for every user function.

axiom/nr_txn.c
zsistla: Please see: [921224e](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/921224e3c10ba59cc9249787ad25d1dcdf519d2a)

axiom/nr_txn.c
zsistla: Please see: [921224e](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/921224e3c10ba59cc9249787ad25d1dcdf519d2a)

axiom/nr_txn.c
zsistla: Please see: [921224e](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/921224e3c10ba59cc9249787ad25d1dcdf519d2a)

axiom/tests/test_attributes.c
```diff
@@ -655,6 +656,74 @@ static void test_add(void) {
...
+  uint32_t all = NR_ATTRIBUTE_DESTINATION_ALL;
+  nrobj_t* obj = NULL;
+
+  config = nr_attribute_config_create();
+  attributes = nr_attributes_create(config);
+
+  /*
+   * Invalid values are attribute=NULL, value = NULL, value = empty string.
+   */
+  // amber
```
mfulb, line 682:
```suggestion
```

axiom/tests/test_attributes.c
```diff
@@ -655,6 +656,74 @@ static void test_add(void) {
...
+  nr_txn_attributes_set_string_attribute(attributes, nr_txn_clm_code_function,
+                                         "");
+  obj = nr_attributes_agent_to_obj(attributes, all);
+  tlib_pass_if_null("Shouldn't have any attributes", obj);
+  nro_delete(obj);
+
+  /*
+   * Invalid values are attribute=NULL.
+   */
+  // amber
```
mfulb, line 701:
```suggestion
```

agent/php_execute.c
misiekn: Maybe consider changing `nr_php_txn_add_code_level_metrics` signature to take the segment as argument as I suggest [here](https://github.com/newrelic/newrelic-php-agent/pull/585#discussion_r1043702320). That should make it possible to move attributes allocation after all conditions to add them are met.

axiom/tests/test_attributes.c
```diff
@@ -655,6 +656,74 @@ static void test_add(void) {
...
+unit tests which do data validation. This test case will just be verify the
+checks the wrappers employ.
+ */
+static void test_nr_txn_attributes_set_attribute(void) {
+  nr_attributes_t* attributes;
+  nr_attribute_config_t* config;
+  //  uint32_t event = NR_ATTRIBUTE_DESTINATION_TXN_EVENT;
+  // uint32_t trace = NR_ATTRIBUTE_DESTINATION_TXN_TRACE;
+  // uint32_t error = NR_ATTRIBUTE_DESTINATION_ERROR;
+  // uint32_t browser = NR_ATTRIBUTE_DESTINATION_BROWSER;
```
mfulb, line 672:
```suggestion
```

agent/php_execute.c
zsistla: regarding point 1 and 2:
Originally, the parameters were attributes and metadata to allow it also to be called from here to provide CLM for files: https://github.com/newrelic/newrelic-php-agent/pull/506/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1226

Additionally, the function needed access to a variable or function that it could only access while living in php_txn.

This is also why the attributes were created outside of the function because we didn't want to double create when adding to a txn instead of just a segment.

however, it's been refactored significantly including us no longer adding it to a filename/txn, so I'll see if it can be moved back and simplified.

misiekn: Thanks for explaining the reason why the code looks like it is. It's hard to see this background by only looking at this PR.

axiom/tests/test_attributes.c
zsistla: oh dear!

axiom/nr_txn.c
```diff
@@ -157,6 +171,32 @@ void nr_txn_set_long_attribute(nrtxn_t* txn,
...
+                                            const char* value) {
+  if (NULL == attribute) {
+    return;
+  }
+  if (NULL == value) {
+    return;
+  }
+  if ('\0' == value[0]) {
+    return;
+  }
```
misiekn, line 185:
```suggestion
  if (nr_strempty(value)) {
    return;
  }
```

agent/php_txn.c
```diff
@@ -1109,3 +1109,101 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
...
+    nr_txn_attributes_set_string_attribute(attributes, nr_txn_clm_code_filepath,
+                                           filepath);
+  }
+  if (!nr_strempty(namespace)) {
+    nr_txn_attributes_set_string_attribute(
+        attributes, nr_txn_clm_code_namespace, namespace);
+  }
+
+  nr_txn_attributes_set_long_attribute(attributes, nr_txn_clm_code_lineno,
+                                       metadata->function_lineno);
```
misiekn, line 1207: This code could be simplified to this:
```
#define CLM_ATTRIBUTE_DESTINATION\
  (NR_ATTRIBUTE_DESTINATION_TXN_TRACE | NR_ATTRIBUTE_DESTINATION_ERROR \
   | NR_ATTRIBUTE_DESTINATION_TXN_EVENT | NR_ATTRIBUTE_DESTINATION_SPAN)

#define SEGMENT_ATTRIBUTES_ADD_CLM(clm_attribute_name, clm_attribute_value, type) do { \
  nr_attributes_agent_add_##type(attributes, \
      CLM_ATTRIBUTE_DESTINATION, \
      clm_attribute_name, \
      clm_attribute_value); \
} while (0);

  SEGMENT_ATTRIBUTES_ADD_CLM("code.function", function, string);
  SEGMENT_ATTRIBUTES_ADD_CLM("code.filepath", filepath, string);
  SEGMENT_ATTRIBUTES_ADD_CLM("code.namespace", namespace, string);
  SEGMENT_ATTRIBUTES_ADD_CLM("code.lineno", metadata->function_lineno, long);
```
This would simplify a lot of code and re-use existing and tested functions:
1. Keep all CLM related things (attribute names and attribute destinations) in a single place - the function that adds the attributes.
2. Re-use existing and already tested functions: `nr_attributes_agent_add_string` and `nr_attributes_agent_add_long`  (reduced test maintenance).
3. Remove [these lines](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/axiom/nr_txn.c#L108-L119), [these functions](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/axiom/nr_txn.c#L174-L198) and tests for them.

tests/integration/attributes/test_transaction_non_web_clm.php
```diff
@@ -0,0 +1,144 @@
...
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+In a non-web transaction that has no user defined functions, code level metrics (CLM)
+should return the filename as the function name (because we are instrumenting the file)
+and lineno 1.
+The agent should include CLM agent attributes in error traces, error
+events, analytic events and span events.
```
misiekn, line 12: Is this test doing what the description say? I don't see CLM attributes in expected sections...

tests/integration/attributes/test_transaction_web_clm.php
```diff
@@ -0,0 +1,215 @@
...
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+In a web transaction that has no user defined functions, code level metrics (CLM)
+should return the filename as the function name (because we are instrumenting the file)
+and lineno 1. The agent should include CLM agent attributes in error traces, error
+events, analytic events and span events.
```
misiekn, line 11: Is this test doing what the description say? I don't see CLM attributes in expected sections...

agent/php_execute.c
misiekn: Resolved by 3e7ec823504a22e68c094b947fd86af06ebd41d9. Thank you @zsistla!

agent/php_execute.c
misiekn: Addressed in 3e7ec823504a22e68c094b947fd86af06ebd41d9.

agent/php_agent.h
```diff
@@ -41,6 +41,7 @@
 #include "php_zval.h"
 #include "util_memory.h"
 #include "util_strings.h"
+#include "php_execute.h"
```
misiekn, line 44: No longer needed to include after 3e7ec823504a22e68c094b947fd86af06ebd41d9.
```suggestion
```

agent/php_txn.c
zsistla: Yes, now it's been decoupled from txn->attributes, we can simplify a bit more.

agent/php_execute.c
```diff
@@ -107,6 +107,44 @@ typedef void (*nr_framework_enable_fn_t)(TSRMLS_D);
  */
 typedef void (*nr_library_enable_fn_t)(TSRMLS_D);
 
+/*
```
misiekn, line 110: I wish 3e7ec823504a22e68c094b947fd86af06ebd41d9 moved to where it used to be (right before the doc string of `nr_php_execute_metadata_init`). This way diff would only show these two lines as changes done for CLM:
```
  zend_string* filepath;
  uint32_t function_lineno;
```

agent/php_execute.h
```diff
@@ -24,6 +24,8 @@
 #define OP_ARRAY_IS_METHOD(OP, FNAME) \
   (0 == nr_strcmp(nr_php_op_array_function_name(OP), (FNAME)))
 
+#define CLM_STRLEN_MAX (255)
+
```
misiekn, line 28: Maybe this could be moved closer to where it's needed? Maybe [here](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/agent/php_execute.c#L1098)? It's not needed anywhere else in the codebase.

misiekn (commented):
Really nice work! Especially on the integration tests! I was wondering if PHP function names and/or namespaces have characters that need to be escaped in JSON strings (namespaces do - `\`) and I saw that integration tests test that. So I started wondering where is the string escaped and found it in this call stack: nr_txndata_prepend_transaction->nr_txndata_prepend_span_events->nr_span_event_to_json_buffer->nro_to_json_buffer->recursive_obj_to_json->[nr_buffer_add_escape_json](https://github.com/zsistla/newrelic-php-agent/blob/clm_to_dev/axiom/util_object.c#L1261). I have a few code improvement suggestions here and there and two questions about integration tests:
1. [This one](https://github.com/newrelic/newrelic-php-agent/pull/585#pullrequestreview-1210786247) and [This one](https://github.com/newrelic/newrelic-php-agent/pull/585#pullrequestreview-1210787532).
2. Should there be a test for CLM in auto instrumented functions? (`Monolog\Logger::addRecord` from Monolog package is a PHP function that the agent auto instruments. Should CLM attributes be expected in `Monolog\Logger::addRecord` spans?) If so, then maybe a few of the logging integration tests should be updated to have CLM on, and have EXPECT_SPAN_EVENTS sections added. What do you think?

agent/php_txn.c
misiekn: Addressed in 3e7ec823504a22e68c094b947fd86af06ebd41d9.

agent/php_txn.c
zsistla: Please see: [b05f098](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/b05f0980b22385746503dc4272ca59d8228eccb8)

While we can localize destinations and attribute names and remove no longer required functions, in practice, the macro doesn't seem to save us much since it is a misdirection to the actual function call and we still have to check file path and namespace for strempty but we don't need to check function or lineno.

It would still need to look like this and in this case the macro seems to hide more than help because we are passing to a macro to just pass straight to a function:
```
SEGMENT_ATTRIBUTES_ADD_CLM("code.function", function, string);

/*
 * If the string is empty, CLM specs say don't add it.
 * nr_attributes_agent_add_string is okay with an empty string attribute.
 * Already checked function for strempty no need to check again.
 */
if (!nr_strempty(filepath)) {
  SEGMENT_ATTRIBUTES_ADD_CLM("code.filepath", filepath, string);
}

if (!nr_strempty(namespace)) {
  SEGMENT_ATTRIBUTES_ADD_CLM("code.namespace", namespace, string);
}

SEGMENT_ATTRIBUTES_ADD_CLM("code.lineno", metadata->function_lineno, long);
```

Seems in this case, better to just be direct: https://github.com/newrelic/newrelic-php-agent/pull/585/commits/b05f0980b22385746503dc4272ca59d8228eccb8#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1169

axiom/nr_txn.c
zsistla: Function removed.

agent/php_execute.c
zsistla: Please see: [59bb8ff](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/59bb8ff4132dfbb5b6efc25001a103e794752756)

agent/php_execute.h
zsistla: Please see: [6c23e7a](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/6c23e7a214ef06257731f1b3c1e3bb03d0be49cd)

tests/integration/attributes/test_transaction_web_clm.php
zsistla: This no longer shows CLM data because we are only putting it on segments not txns now.  I'll update the comment.

zsistla: Please see: [49a22d7](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/49a22d7a6abb50633602e0320f09f65ac8d0384a)

tests/integration/attributes/test_transaction_non_web_clm.php
zsistla: Please see: [49a22d7](https://github.com/newrelic/newrelic-php-agent/pull/585/commits/49a22d7a6abb50633602e0320f09f65ac8d0384a)

agent/php_txn.c
misiekn: Fair enough, looks much cleaner now! Thank you!

agent/php_execute.c
misiekn: Thank you! Now it's easy to see in this PR what CLM actually changed/added.

tests/integration/logging/monolog3/test_monolog_basic_clm.php
```diff
@@ -0,0 +1,205 @@
...
+    usleep(10000);
+    $logger->error("error");
+    usleep(10000);
+    $logger->critical("critical");
+    usleep(10000);
+    $logger->alert("alert");
+    usleep(10000);
+    $logger->emergency("emergency");
+}
+
```
misiekn, line 204: I wonder how would the EXPECT_SPAN_EVENTS look like if a custom tracer `newrelic_add_custom_tracer("test_logging");` was added here...
```suggestion
newrelic_add_custom_tracer("test_logging");
```

zsistla: custom tracing is already handled in 4 other CLM tests.  In this case even though we log it, it would be no different from another custom_tracer. 

Many of the tests will need updating once CLM is on by default.

zsistla:
> I believe we need to have `newrelic.template.ini` included in this PR to include the new CLM INI variable.

Please see: 
0e2369d

zsistla:
ok jenkins

zsistla:
ok jenkins

tests/integration/logging/monolog3/test_monolog_basic_clm.php
```diff
@@ -0,0 +1,159 @@
...
+        "name": "Custom\/Monolog\\Logger::addRecord",
+        "guid": "??",
+        "timestamp": "??",
+        "duration": "??",
+        "category": "generic",
+        "parentId": "??"
+      },
+      {},
+      {
+        "code.lineno": 314,
```
misiekn, line 97: Maybe use a wildcard for line number in case this changes after Monolog update?
```suggestion
        "code.lineno": "??",
```

tests/integration/logging/monolog3/test_monolog_basic_clm.php
misiekn: Thank you! This is exactly what I wanted to see tested: CLM attributes in spans for auto-instrumented functions.

misiekn (approved):
Awesome work! 👍

tests/integration/logging/monolog3/test_monolog_basic_clm.php
zsistla: Wildcard is a really good idea, but I'd rather leave a small subset of tests that actually checks the line number.
When CLM goes on by default and we update all the rest of the tests, that would be a good place to use the wild card.

PR #587, opened 2022-11-29
refactor(agent): user instrumentation lookup

misiekn:
Refactor the code that enables/installs and retrieves special instrumentation (wraprecs) for user functions.

1. Cleanup conditional compilation around user_instrument_get and user_instrument_set to make the code more readable and maintainable

2. Add two new wraprec lookup methods: util_hashmap (axiom implementation) and wraprec_hashmap (specialized implementation that takes advantage of the structure of data that needs to be looked up which makes matching routine to be highly optimized

3. Factor out each lookup method implementation into separate modules. Following implementations are available:
- 0 - **op_array** (legacy method). This method stores wraprecs in a vector and uses reserved array in zend_function's op_array to store the index into the vector that has wraprec associated with zend_function. This method no longer works when agent runs within php-fpm with opcache enabled (7.4+).
- 1 - **linked list** (legacy method for PHP 7.4+). This method stores wraprecs in a linked list and uses zend_function's metadata (filename, lineno, scope and function name) to find a match in linked list when looking for a wraprec associated with zend_function. This method was created to address the issue with op_array method but was found to have performance issues when list of wraprecs is long.
- 2 - **util_hashmap** using axiom's implementation (experimental). This method stores wraprecs in a hashmap (axiom's implementation) and uses zend_function's metadata (filename, lineno, scope and function name) to create a string key that axiom's implementation of a hashmap uses to find a match when looking for a wraprec associated with zend_function. Axiom's hashmap implementation uses linked list to store values that have the same hash. This method was created to address the performance issue with linked list method however it was found to be slower than linked list method because the operation of creating string key and then converting it to hash is slower than walking a short list and comparing metadata.
- 3 - **wraprec_hashmap** using specialized hashmap (new default for PHP 7.4+). This method stores wraprecs in a hashmap (wraprec specialized implementation) and uses zend_function's metadata (filename, lineno, scope and function name) to create a numeric hash that wraprec specialized implementation of a hashmap uses to find a bucket (a linked list) that is further scanned for exact match of metadata when looking for a wraprec associated with zend_function. wraprec specialized implementation uses linked list to store values that have the same hash. This method was created to address the performance issue with axiom's hashmap.  It uses optimized hash generation as well as optimized metadata matcher.

4. The lookup method implementation used by the agent is selected at build time through LOOKUP_METHOD env variable. It takes following values: 0 - legacy method, 1 - linked list, 2 - axiom's
implementation of hashmap, 3 - specialized wraprec hashmap.with the wraprec_hashmap method being default for PHP 7.4+ and op_array method being default for older PHPs.

TODO: add unit tests for wraprec_hashmap lookup method in `agent/tests/test_user_instrument.c` module.


agent/php_user_instrument_wraprec_hash.h
```diff
@@ -0,0 +1,249 @@
...
+extern nr_php_wraprec_hashmap_t* nr_php_wraprec_hashmap_create(void);
+#else
+nr_php_wraprec_hashmap_t* nr_php_wraprec_hashmap_create() {
+  nr_php_wraprec_hashmap_t* hashmap;
+
+  hashmap = (nr_php_wraprec_hashmap_t*)nr_malloc(sizeof(nr_php_wraprec_hashmap_t));
+    /*
+     * Encode the default value in one place: namely, here.
+     */
+  hashmap->hash_modulo = 257; //521 or 509 //53; // or 97 or other prime number
```
misiekn, line 34: The closer the number to the average the length of a PHP file, the better distribution of wraprecs between buckets and the shorter the list in the bucket. The number affects the memory used by buckets, i.e. `hash_modulo x sizeof( nr_wraprec_hashmap_bucket_t`).

agent/php_user_instrument_wraprec_hash.h
```diff
@@ -0,0 +1,249 @@
...
+
+  if (!wraprec_streq(&nr_id->function_name, &zf_id.function_name)) {
+    return false;
+  }
+
+  if (!wraprec_streq(&nr_id->filename, &zf_id.filename)) {
+    return false;
+  }
+
+  /* No need to check scope - function has the same name is at the same location in the same file */
```
misiekn, line 183: On a second thought it may be better to leave scope matching in. See [this comment](https://github.com/newrelic/newrelic-php-agent/pull/587/files#r1034950492).

agent/php_user_instrument_wraprec_hash.h
```diff
@@ -0,0 +1,249 @@
...
+  SET_META_STRING(id, scope, zf, scope_name, nr_strdup);
+  SET_META_STRING(id, function_name, zf, function_name, nr_strdup);
+}
+
+static inline void wraprec_metadata_set(nruserfn_metadata* id, const zend_function* zf) {
+ if (id->is_set) {
+    return;
+  }
+  zf_to_id(id, zf);
+  id->is_set = true;
```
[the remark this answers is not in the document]
mfulb: Could we somehow expire these in the RSHUTDOWN if it a wraprec hasn't been touched in a time period?

agent/php_agent.h
```diff
@@ -670,6 +672,13 @@ nr_php_op_array_file_name(const zend_op_array* op_array) {
 #endif
 }
 
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
```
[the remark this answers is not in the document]
mfulb: Granted it may not be inlined like this?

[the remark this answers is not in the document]
misiekn: The reason why `nr_php_op_array_*_length` were added was to make the `nr_php_op_array_*` API more complete. There already existed `nr_php_function_name` with nicely paired `nr_php_function_name_length` for zend_function's common field and there existed `nr_php_op_array_function_name` but didn't have `nr_php_op_array_function_name_length`. So I thought instead of mixing `nr_php_op_array_function_name` and raw access to zend_function's op_array.function_name->len (or through Zend's API) I decided to make `nr_php_op_array_*` API more complete. Because the API became complete it made it possible to simplify the `zf_to_id` code.

agent/php_user_instrument.c
```diff
@@ -644,89 +509,71 @@ void nr_php_user_function_add_declared_callback(const char* namestr,
...
-  return (nruserfn_t*)nr_vector_get(NRPRG(user_function_wrappers), index);
+  p = wraprec_hashmap_get(&n, NRPRG(user_function_wrappers), func);
+
+#else
+
+#error "Unknown wraprec lookup method"
+
+#endif
+
+  if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).debug_user_instrument_lookup)) {
```
mfulb, line 574: If this is staying in the PR perhaps the output should go to the agent log file?

misiekn: This probably shouldn't stay in the PR.

agent/php_rinit.c
```diff
@@ -35,8 +35,17 @@ PHP_RINIT_FUNCTION(newrelic) {
...
+#if LOOKUP_METHOD == LOOKUP_USE_OP_ARRAY
   NRPRG(pid) = getpid();
   NRPRG(user_function_wrappers) = nr_vector_create(64, NULL, NULL);
+#elif LOOKUP_METHOD == LOOKUP_USE_UTIL_HASHMAP
+  /* This hashmap will store pointers to wraprecs. Since wraprecs
+   * are persistent (they're not destroyed between requests),
+   * there's no need for hashmap value destructor. */
+  NRPRG(user_function_wrappers) = nr_hashmap_create(NULL);
+#elif LOOKUP_METHOD == LOOKUP_USE_WRAPREC_HASHMAP
+  NRPRG(user_function_wrappers) = nr_php_wraprec_hashmap_create();
```
mfulb, line 47: How is the memory allocated by nr_php_wraprec_hashmap_create() released or is it ever?  Does it persist between requests if being run under php-fpm?

agent/php_user_instrument.c
```diff
@@ -273,6 +261,8 @@ static void nr_php_user_wraprec_destroy(nruserfn_t** wraprec_ptr) {
     return;
   }
 
+  wraprec_id_dtor(&wraprec->id);
```
mfulb, line 264: Maybe a good time to change the '0's on line 256 and 260 to NULLs?

agent/php_user_instrument.c
```diff
@@ -308,121 +298,6 @@ static void nr_php_add_custom_tracer_common(nruserfn_t* wraprec) {
...
-    }
-    p = p->next;
-  }
-
-  return NULL;
-}
-#endif
-
 #define NR_PHP_UNKNOWN_FUNCTION_NAME "{unknown}"
 nruserfn_t* nr_php_add_custom_tracer_callable(zend_function* func TSRMLS_DC) {
```
mfulb, line 302: Should this be `const zend_function*`?

agent/php_user_instrument.c
```diff
@@ -262,6 +244,12 @@ static nruserfn_t* nr_php_user_wraprec_create_named(const char* full_name,
   return wraprec;
 }
 
+static void wraprec_id_dtor(nruserfn_metadata *id) {
+  nr_free(id->filename.value);
```
mfulb, line 248: This seems to assume the `value` union member `cp` was set but there doesn't appear to be a way from the `zf_metadata_t` to actually verify if the data represented was allocated or not.  In this case you probably know the only way the `value` member is populated is using `strdup()` but maybe it would be good to have this info from the struct?

agent/php_user_instrument.h
```diff
@@ -26,6 +27,22 @@ typedef struct nrspecialfn_return_t (*nrspecialfn_t)(
...
 typedef void (*nruserfn_declared_t)(TSRMLS_D);
 
+typedef struct {
+  bool is_set;
+  uint32_t len;
+  union {
+    char *cp; /* cp = char pointer */
+    const char *ccp; /* ccp = const char pointer */
+  } value;
+} zf_metadata_t;
```
mfulb, line 37: Mentioned else where but maybe having a way to tell if the `cp` or `ccp` union field is populated would be helpful for robustly releasing the memory for this structure without having to know how it was populated?

agent/php_user_instrument_llist.h
```diff
@@ -0,0 +1,155 @@
...
+static void wraprec_metadata_set(nruserfn_t* wraprec, zend_function* func) {
+    const char* filename = nr_php_function_filename(func);
+  /*
+   * Before setting a filename, ensure it is not NULL and it doesn't equal the
+   * "special" designation given to funcs without filenames. If the function is
+   * an evaluated expression or called directly from the CLI there is no
+   * filename, but the function says the filename is "-". Avoid setting in this
+   * case; otherwise, all the evaluated/cli calls would match.
+   */
+  if ((NULL == wraprec->filename) && (NULL != filename)
```
mfulb, line 26: Should this also be checking if `wraprec` is NULL?

misiekn: Maybe. However this code is a verbatim copy of https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_user_instrument.c#L134-L151. The intention of this PR is not to change existing wraprec lookup implementations but to factor them out into separate modules for easier comparison testing, maintenance and review.

agent/php_user_instrument.c
misiekn: You are correct - this code assumes that `cp` was set and at this point it should be set. However I will apply your suggestion to remove this assumption.

agent/php_user_instrument.c
misiekn: Although it's a good suggestion, I'd rather keep the PR focused and not have it change the code it doesn't have to. I'd say the `0` -> `NULL` style change in 256 and 260 is not related to this PR and IMHO deserves its own PR.

agent/php_user_instrument.c
misiekn: Not necessarily. When `LOOKUP_USE_OP_ARRAY` method is used (PHP < 7.4), `func->op_array.reserved` is modified by `nr_php_op_array_set_wraprec` which is called by `nr_php_user_instrument_set` which is called by `nr_php_wrap_zend_function` which is called later in `nr_php_add_custom_tracer_callable`.

agent/php_user_instrument_llist.h
```diff
@@ -0,0 +1,155 @@
...
+/*
+ * Purpose : Store zend_function's metadata (filename, lineno) in wraprec
+ *           for use in looking up the instrumentation.
+ * 
+ * Params  : 1. The wraprec to update with metadata
+ *           2. The zend function with meta data
+ *
+ * Returns : void
+ */
+static void wraprec_metadata_set(nruserfn_t* wraprec, zend_function* func) {
```
misiekn, line 17: This code is a verbatim copy of https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_user_instrument.c#L134-L151.

agent/php_user_instrument_llist.h
```diff
@@ -0,0 +1,155 @@
...
+/*
+ * Purpose : Determine if a func matches a wraprec.
+ *
+ * Params  : 1. The wraprec to match to a zend function
+ *           2. The zend function to match to a wraprec
+ *
+ * Returns : True if the class/function of a wraprec match the class function
+ *           of a zend function.
+ */
+static inline bool nr_php_wraprec_matches(nruserfn_t* p, zend_function* func) {
```
misiekn, line 47: This code is a verbatim copy of https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_user_instrument.c#L322-L403. It's been factored out into lookup method implementation specific module for easier testing.

agent/php_user_instrument_op_array.h
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#if LOOKUP_METHOD == LOOKUP_USE_OP_ARRAY
```
misiekn, line 6: This module/code is verbatim copy of https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_user_instrument.c#L647-L732. It's been factored out into lookup method implementation specific module for easier testing and is enabled at compile time for PHP < 7.4 as configured in `agent/php_user_instrument_lookup_config.h`

agent/php_user_instrument_util_hashmap.h
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#if LOOKUP_METHOD == LOOKUP_USE_UTIL_HASHMAP
```
misiekn, line 6: This module implements an experimental lookup method and will probably be removed completely in the final version of this PR. It's left in for now as a reference.

agent/php_user_instrument_wraprec_hash.h
```diff
@@ -0,0 +1,249 @@
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#if LOOKUP_METHOD == LOOKUP_USE_WRAPREC_HASHMAP
+
+/* This hashmap will store pointers to wraprecs stored in the linked list.
+ * Since wraprecs are persistent (they're not destroyed between requests),
```
misiekn, line 9: This comment probably needs an update to be more accurate - non-transient wraprecs are persistent. Transient wraprecs are destroyed in `nr_php_remove_transient_user_instrumentation` called from `nr_php_post_deactivate` at the end of request.

agent/php_user_instrument.c
misiekn: Is ef8ebf666e0a17b49c8cde41492782371f5dce95 ok?

agent/php_user_instrument.h
misiekn: Is ef8ebf666e0a17b49c8cde41492782371f5dce95 ok?

agent/php_user_instrument_wraprec_hash.h
misiekn: Updated in [237ca07](https://github.com/newrelic/newrelic-php-agent/pull/587/commits/237ca0707c2b5335f562a9c0ed0ce58f5b4ed0bc)

agent/php_user_instrument.c
mfulb: True but like some other similar chores (PHP7 defines, etc) we've been doing it opportunistically as we update nearby code. 
 Just a consideration.

agent/php_user_instrument_llist.h
mfulb: Understood but if for robustness the value needs to be validated we need to make sure this update occurs in some fashion.

misiekn: I'm thinking that maybe in the final version of this PR, the *linked list* and *util_hashmap* lookup methods (and their implementations in `agent/php_user_instrument_llist.h` and `agent/php_user_instrument_util_hashmap.h`) should be removed completely. I left them in for now for following reasons:
1. In case someone would like to experiment with this codebase
2. In case when reviewing the code someone would like to quickly see how it was done before
If these reasons are no longer valid, I'll go ahead and remove these two implementations from the PR.

misiekn:
Superseded by #588

PR #588, opened 2022-12-02
refactor(agent): user instrumentation lookup

misiekn:
A more performant data structure is used to lookup user instrumentation:
a hashmap with zend function's metadata (lineno, functioname, scope
and filename) as a hash key, and a pointer to a wraprec as a value.

In order to improve performance of the specialized hashmap key matching,
zend_function's metadata is stored in a key as a refcounted zend_string
rather than raw char*. This allows key matching to take advantage of
the following zend_string attributes: length (len) and hash (h) - values for
both are pre-calculated by Zend engine and available for use by the
agent.  Because the agent instruments only a small fraction of all user
functions in a PHP application, there will be more missed hashmap
lookups than hits. Therefore delaying string value comparison until
absolutely necessary (len and h must match first) is a significant
performance gain.

To save time on lookups, a single hashmap will be used to store pointers
to both types of wraprecs: transient and reusable. For efficiency, the
wraprecs that are re-usable between requests (e.g. for well known
library or framework user functions), are additionally stored in a
linked list.  These wraprecs are created once per interesting function
detection, and destroyed at module shutdown. Transient wraprecs, that
are not re-usable between requests, are not stored in linked list.
Transient wraprecs are created on the fly and destroyed at request
shutdown.  However user function instrumentation installation (aka
wrapping) is done the same way for both types of wraprecs and happens
once per each request, i.e. for each request the hashmap is created
anew, when user function is instrumented, its wraprec is added to the
hashmap, and at the end of the request the hashmap is destroyed together
with transient wraprecs. Re-usable wraprecs are not destroyed - they
remain in linekd list and are re-set (marked as not wrapping a zend
function). This happens because with new request/transaction php is
loading all new user code.

### DIFF
--- a/agent/php_rshutdown.c
+++ b/agent/php_rshutdown.c
@@ -109,7 +109,9 @@ int nr_php_post_deactivate(void) {
   nr_free(NRPRG(predis_ctx));
   nr_hashmap_destroy(&NRPRG(predis_commands));
 
+#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO
   nr_vector_destroy(&NRPRG(user_function_wrappers));
+#endif
 
   NRPRG(cufa_callback) = NULL;
 
--- a/agent/php_rshutdown.c
+++ b/agent/php_rshutdown.c
@@ -109,7 +109,9 @@ int nr_php_post_deactivate(void) {
   nr_free(NRPRG(predis_ctx));
   nr_hashmap_destroy(&NRPRG(predis_commands));
 
-#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+  nr_php_reset_user_instrumentation();
+#else
   nr_vector_destroy(&NRPRG(user_function_wrappers));
 #endif
 
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -439,6 +439,7 @@ void nr_php_reset_user_instrumentation(void) {
   }
 #if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
   if (NULL != user_function_wrappers) {
+    // send a metric with the number of transient wrappers
     nr_hashmap_destroy(&user_function_wrappers);
   }
   /* wraprecs are re-usable and stored in linked list - no destructor */
@@ -451,6 +452,7 @@ void nr_php_reset_user_instrumentation(void) {
  */
 void nr_php_remove_transient_user_instrumentation(void) {
 #if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+  // send a metric with the number of transient wrappers
   nr_hashmap_destroy(&transient_wrappers);
 #else
   nruserfn_t* p = nr_wrapped_user_functions;
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -416,7 +416,7 @@ void nr_php_reset_user_instrumentation(void) {
     nr_hashmap_destroy(&user_function_wrappers);
   }
   // send a metric with the number of transient wrappers
-  user_function_wrappers = nr_hashmap_create_buckets(10, reset_wraprec);
+  user_function_wrappers = nr_hashmap_create_buckets(1024, reset_wraprec);
 #else
   nruserfn_t* p = nr_wrapped_user_functions;
   while (0 != p) {
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -79,21 +79,22 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
 
 #if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
 /* Hashmap with pointers to wraprecs. Some, that are re-usable between requests,
- * are stored in linked list. These wraprecs are created once per interesting 
- * function detection, and destroyed at module shutdown. Some, that are transient
- * and not re-usable between requests, are not stored in linked list. Transient
- * wraprecs are created on the fly and destroyed at request shutdown. However
- * wrapping is done the same way for both types of wraprecs and happens once
- * per each request, i.e. for each request the hashmap is created anew, when
- * user function is instrumented, its wraprec is added to the hashmap, and at
- * the end of the request the hashmap is destroyed together with transient
+ * are stored in linked list. These wraprecs are created once per interesting
+ * function detection, and destroyed at module shutdown. Some, that are
+ * transient and not re-usable between requests, are not stored in linked list.
+ * Transient wraprecs are created on the fly and destroyed at request shutdown.
+ * However wrapping is done the same way for both types of wraprecs and happens
+ * once per each request, i.e. for each request the hashmap is created anew,
+ * when user function is instrumented, its wraprec is added to the hashmap, and
+ * at the end of the request the hashmap is destroyed together with transient
  * wraprecs. Re-usable wraprecs are not destroyed - they are re-set. */
 static nr_php_wraprec_hashmap_t* user_function_wrappers;
 
-static inline void nr_php_wraprec_lookup_set(nruserfn_t* wr, zend_function* zf) {
+static inline void nr_php_wraprec_lookup_set(nruserfn_t* wr,
+                                             zend_function* zf) {
   nr_php_wraprec_hashmap_update(user_function_wrappers, zf, wr);
 }
-static inline nruserfn_t* nr_php_wraprec_lookup_get(zend_function *zf) {
+static inline nruserfn_t* nr_php_wraprec_lookup_get(zend_function* zf) {
   nruserfn_t* wraprec = NULL;
 
   nr_php_wraprec_hashmap_get_into(user_function_wrappers, zf, &wraprec);
@@ -106,26 +107,29 @@ static inline nruserfn_t* nr_php_wraprec_lookup_get(zend_function *zf) {
  * This creates wraprec lookup hashmap and registers wraprec destructor
  * callback - reset_wraprec - which is called on request shutdown.
  */
-static void reset_wraprec(nruserfn_t *);
+static void reset_wraprec(nruserfn_t*);
 void nr_php_init_user_instrumentation(void) {
   if (NULL != user_function_wrappers) {
     /* Should not happen */
-    nrl_verbosedebug(NRL_INSTRUMENT, "user_function_wrappers lookup hashmap already initialized!");
+    nrl_verbosedebug(
+        NRL_INSTRUMENT,
+        "user_function_wrappers lookup hashmap already initialized!");
     return;
   }
-  user_function_wrappers = nr_php_wraprec_hashmap_create_buckets(1024, reset_wraprec);
+  user_function_wrappers
+      = nr_php_wraprec_hashmap_create_buckets(1024, reset_wraprec);
 }
 
 /*
  * This callback resets user instrumentation. It is called at request shutdown
  * when user instrumentation is reset - lookup hashmap is destroyed together
- * with transient wraprecs and non-transient wraprecs are reset (mark as not 
+ * with transient wraprecs and non-transient wraprecs are reset (mark as not
  * wrapped). This happens because with new request/transaction php is loading
  * all new user code.
  */
 static void nr_php_user_wraprec_destroy(nruserfn_t** wraprec_ptr);
 static void reset_wraprec(nruserfn_t* wraprec) {
-  nruserfn_t*p = wraprec;
+  nruserfn_t* p = wraprec;
   nr_php_wraprec_hashmap_key_release(&p->key);
   if (p->is_transient) {
     nr_php_user_wraprec_destroy((nruserfn_t**)&wraprec);
@@ -429,25 +433,28 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
   return wraprec; /* return the new wraprec */
 }
 
-
 /*
  * Reset the user instrumentation records because we're starting a new
  * transaction and so we'll be loading all new user code.
- * 
+ *
  * For PHP 7.4+ this function is called on request shutdown to release memory
- * allocated for lookup hashmap! Additionally hashmap's value destructor 
+ * allocated for lookup hashmap! Additionally hashmap's value destructor
  * callback will reset all non-transient wraprecs (mark them as not wrapped),
  * destroy all transient wraprecs.
- * 
+ *
  */
 void nr_php_reset_user_instrumentation(void) {
 #if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
-   // send a metric with the number of transient wrappers
+  // send a metric with the number of transient wrappers
   if (NULL != user_function_wrappers) {
-    nr_php_wraprec_hashmap_stats_t stats = nr_php_wraprec_hashmap_destroy(&user_function_wrappers);
+    nr_php_wraprec_hashmap_stats_t stats
+        = nr_php_wraprec_hashmap_destroy(&user_function_wrappers);
 
-    nrl_debug(NRL_INSTRUMENT, "# elements: %lu, # buckets used: %lu", stats.elements, stats.buckets_used);
-    nrl_debug(NRL_INSTRUMENT, "collisions - min: %lu, max: %lu, avg: %lu", stats.collisions_min, stats.collisions_max, stats.collisions_mean);
+    nrl_debug(NRL_INSTRUMENT, "# elements: %lu, # buckets used: %lu",
+              stats.elements, stats.buckets_used);
+    nrl_debug(NRL_INSTRUMENT, "collisions - min: %lu, max: %lu, avg: %lu",
+              stats.collisions_min, stats.collisions_max,
+              stats.collisions_mean);
   }
 #else
   nruserfn_t* p = nr_wrapped_user_functions;
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -416,7 +416,7 @@ void nr_php_reset_user_instrumentation(void) {
     nr_hashmap_destroy(&user_function_wrappers);
   }
   // send a metric with the number of transient wrappers
-  user_function_wrappers = nr_hashmap_create(reset_wraprec);
+  user_function_wrappers = nr_hashmap_create_buckets(10, reset_wraprec);
 #else
   nruserfn_t* p = nr_wrapped_user_functions;
   while (0 != p) {
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -77,17 +77,17 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
 }
 
 #if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
-/* Hashmap with pointers to wraprecs stored in llist that are reusable between requests.
- * These wraprecs are created once per interesting function detection and destroyed
- * at module shutdown. However wrapping is done per each request, i.e. for each request
- * the hashmap is created anew and is destroyed at the end of the request. */
+/* Hashmap with pointers to wraprecs. Some, that are re-usable between requests,
+ * are stored in linked list. These wraprecs are created once per interesting 
+ * function detection, and destroyed at module shutdown. Some, that are transient
+ * and not re-usable between requests, are not stored in linked list. Transient
+ * wraprecs are created on the fly and destroyed at request shutdown. However
+ * wrapping is done the same way for both types of wraprecs and happens once
+ * per each request, i.e. for each request the hashmap is created anew, when
+ * user function is instrumented, its wraprec is added to the hashmap, and at
+ * the end of the request the hashmap is destroyed together with transient
+ * wraprecs. Re-usable wraprecs are not destroyed - they are re-set. */
 static nr_hashmap_t* user_function_wrappers;
-/* hashmap with pointers to transient wraprecs that are not reusable between requests.
- * These wraprecs are created once per interesting function detection and destroyed
- * at request shutdown. Wrapping is done at the same time the wraprec is created, i.e.
- * once for each request. The hashmap is created when requests starts and is destroyed
- * together with all wraprecs in it at the end of the request. */
-static nr_hashmap_t* transient_wrappers;
 
 static inline void nr_php_wraprec_hashmap_set(nr_hashmap_t* h, zend_function* zf, nruserfn_t* wr) {
   nr_hashmap_update(h, (const char *)&zf, sizeof(zend_function*), wr);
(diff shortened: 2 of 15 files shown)


agent/php_user_instrument.c
```diff
@@ -507,6 +412,17 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
...
 }
 
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+static void destroy_transient_wrapper(void* wraprec) {
+  nr_php_user_wraprec_destroy((nruserfn_t**)&wraprec);
+}
+
+void nr_php_init_transient_user_instrumentation(void) {
+  /* wraprecs are not re-usable and destroyed together with hashmap */
+  transient_wrappers = nr_hashmap_create(destroy_transient_wrapper);
```
misiekn, line 422: What should be the initial size of this hashmap? 64 for user_function_wrappers and 1024 for transient_wrappers?

misiekn:
ok jenkins

misiekn:
ok jenkins

agent/php_user_instrument.c
```diff
@@ -515,15 +431,28 @@ void nr_php_reset_user_instrumentation(void) {
...
 
   while (0 != p) {
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+    p->zf = NULL;
+#endif
     p->is_wrapped = 0;
     p = p->next;
   }
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+  if (NULL != user_function_wrappers) {
```
misiekn, line 441: Create a supportability metric with number of wraprecs for named user functions. What should be the name of the metric?

agent/php_user_instrument.c
```diff
@@ -515,15 +431,28 @@ void nr_php_reset_user_instrumentation(void) {
...
+  /* wraprecs are re-usable and stored in linked list - no destructor */
+  user_function_wrappers = nr_hashmap_create(NULL);
+#endif
 }
 
 /*
  * Remove any transient wraprecs. This must only be called on request shutdown!
  */
 void nr_php_remove_transient_user_instrumentation(void) {
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
```
misiekn, line 453: Create a supportability metric with number of wraprecs for transient user functions. What should be the name of the metric?

zsistla:
ok jenkins

agent/php_user_instrument.c
mfulb: The only penalty for making it larger is the memory usage correct?  Each bucket is two pointers (for linked list) and a pointer to a wraprec?  If this is the case even 1024 seems completely reasonable in memory usage.

agent/php_user_instrument.c
```diff
@@ -644,6 +576,18 @@ void nr_php_user_function_add_declared_callback(const char* namestr,
...
+nruserfn_t* nr_php_get_wraprec(zend_function* zf) {
+  nruserfn_t* wraprec = NULL;
+  /* Look in user function wrappers first: */
+  wraprec = nr_php_wraprec_hashmap_get(user_function_wrappers, zf);
+  if (NULL == wraprec) {
+    /* Look in transient user function wrappers next: */
+    wraprec = nr_php_wraprec_hashmap_get(transient_wrappers, zf);
+  }
+  return wraprec;
+}
```
mfulb, line 589: Is there a way we could put all wraprecs in a single hash and remote the transients at the end of a request?  It seems for the case where the number of transient wraprecs is large (>100?) this is going to incur an additional hash lookup for each one and we know there can be a large number.

agent/php_user_instrument.c
zsistla: Does making it larger make the lookups slower especially if it is sparsely populated?

agent/php_user_instrument.c
zsistla: `Supportability/PHP/Instrumented/Named` ?

agent/php_user_instrument.c
```diff
@@ -699,7 +643,6 @@ void nr_php_op_array_set_wraprec(zend_op_array* op_array,
   op_array->reserved[NR_PHP_PROCESS_GLOBALS(zend_offset)] = (void*)index;
 }
 
-#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO
```
zsistla, line 702: Is there a reason not to exclude it?

misiekn:
ok jenkins

agent/php_user_instrument.c
misiekn: > Does making it larger make the lookups slower especially if it is sparsely populated?
I need to measure more closely how well the hashing function distributes the addresses of zend_functions but making the hashmap larger (more buckets) should reduce the lookup time because there will be less collisions to resolve (fewer keys, zend_function addresses, will be mapped to the same bucket).

agent/php_user_instrument.c
misiekn: `nr_php_op_array_get_wraprec` and `nr_php_op_array_set_wraprec` are not built for PHP 7.4+. There's an `#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO` on line [579](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/agent/php_user_instrument.c#L579) with an ending `#else` on line [590](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/agent/php_user_instrument.c#L590).

agent/php_user_instrument.c
misiekn: It is possible to put all wraprecs in a single hashmap. This would require a slight change to hashmap destruction at the end of the request.

agent/php_user_instrument.c
misiekn: How about `Supportability/PHP/Instrumented/UserFunctions/Auto` for wraprecs created automatically by an agent (by calls to `nr_php_wrap_user_function` in `{fw|lib}_*_enable)`, `Supportability/PHP/Instrumented/UserFunctions/Transient` for wraprecs created 'in-flight' by intercepting `call_user_function` (WordPress hooks, Drupal hooks), and `Supportability/PHP/Instrumented/UserFunctions/Custom` for wraprecs created from PHP with call to [newrelic_add_custom_tracer](https://docs.newrelic.com/docs/apm/agents/php-agent/php-agent-api/newrelic_add_custom_tracer)? This opens up `Supportability/PHP/Instrumented/InternalFunctions/*` metric namespace in case we needed to count those too.

zsistla: LGTM

agent/php_user_instrument.c
zsistla: Thanks for the links, I see it now.

agent/php_user_instrument.c
misiekn: Here's how it could be done: https://github.com/misiekn/newrelic-php-agent/commit/c82ba80c2030aac92de5881fdb55a24a6340065e.

misiekn:
ok jenkins

agent/php_user_instrument.c
misiekn: For now I made them 64 and 1024: 02406bda54989b60a5af873e30fe1d5740b676bb.

misiekn:
ok jenkins

agent/php_user_instrument.c
misiekn: This will be added in a separate PR.

agent/php_user_instrument.c
misiekn: This will be added in a separate PR.

agent/php_user_instrument.c
```diff
@@ -644,6 +552,11 @@ void nr_php_user_function_add_declared_callback(const char* namestr,
   }
 }
 
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
```
mfulb, line 555: Seems like this >= 7.4 block might flow better with the code if it was in the block up at line 79 above?

agent/php_user_instrument.c
mfulb: Your new addition looks good Michal.

agent/php_user_instrument.c
misiekn: It was actually 1e7ec63f5395d469264ad8c2ba5718f9546ba435 that made the hashmap size to be 1024. And hashing function distribution seems to work fairly ok. Some test results:
```
# elements: 1001, # buckets used: 649, collisions - min: 1, max: 29, avg: 2
```

agent/php_rshutdown.c
```diff
@@ -109,7 +109,9 @@ int nr_php_post_deactivate(void) {
   nr_free(NRPRG(predis_ctx));
   nr_hashmap_destroy(&NRPRG(predis_commands));
```
zsistla, line 111: add a hashmap destroy here for the user_function_wrappers?
nr_hashmap_destroy(&NRPRG(user_function_wrappers));

this only gets called on txn start:
https://github.com/newrelic/newrelic-php-agent/pull/588/files#diff-9408f6b45682dc2cebef6e4beccd6aae78d1f1c542c82fafdb333a7d8fc83262R413

agent/php_user_instrument.c
misiekn: I think 32a63ed07d08f5af1784a69b42aec22ea398adc3 addressed this remark.

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,315 @@
...
+    collisions = 0;
+
+    if (bucket) {
+      stats.buckets_used++;
+    }
+
+    while (bucket) {
+      nr_wraprecs_bucket_t* next = bucket->next;
+
+      collisions++;
```
ZNeumann, line 114: Is it intended to count a collision even if the bucket list is of length one?

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,315 @@
...
+      stats.collisions_max = collisions;
+    }
+    if (collisions > 1) {
+      stats.buckets_with_collisions++;
+      stats.collisions_mean += collisions;
+    }
+  }
+
+  if (0 != stats.buckets_with_collisions) {
+    stats.collisions_mean /= stats.buckets_with_collisions;
```
ZNeumann, line 133: Why do we not want to divide by buckets_used instead?

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,315 @@
...
+
+  if (zf->op_array.function_name && !nr_zf_is_unnamed_closure(zf)) {
+    if (!zstr_equal(key->function_name, zf->op_array.function_name)) {
+      return false;
+    }
+    /* compare scope if it is set */
+    if (zf->op_array.scope && !zstr_equal(key->scope_name, zf->op_array.scope->name)) {
+      return false;
+    }
+    return true;
```
ZNeumann, line 233: Some comments explaining this if-else structure would be helpful

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,315 @@
...
+
+void nr_php_wraprec_hashmap_update(nr_php_wraprec_hashmap_t* hashmap, zend_function* zf, nruserfn_t* wr) {
+  nr_wraprecs_bucket_t* bucket = NULL;
+  size_t bucketidx;
+
+  if ((NULL == hashmap) || (NULL == zf) || (0 == wr)) {
+    return;
+  }
+
+  nr_php_wraprec_hashmap_key_set(&wr->key, zf);
```
ZNeumann, line 282: Do we need to worry about a case where this function fails to set the key? Is it even possible for that to happen?

agent/php_user_instrument_hashmap.c
misiekn: Yes it is intentional. [This condition](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/agent/php_user_instrument_hashmap.c#L126) checks if the `collisions` counter is greater than 1 and only then updates collisions mean stats. I guess maybe the variable name is not the most fortunate. Do you think `bucket_items_cnt` is better variable name?

agent/php_user_instrument_hashmap.c
misiekn: `collisions_mean` is only updated when number of elements in bucket is greater than 1. So is `buckets_with_collisions`. `buckets_used` is total number of buckets, even the ones without collisions (i.e. with number of elements in bucket equal to 1). Therefore dividing by `buckets_used` would not give the mean of collisions for buckets with collisions. It would give the mean of collisions including buckets without collisions.

agent/php_user_instrument_hashmap.c
misiekn: @ZNeumann Is this what you hoped to see: d51b4ed34fce72f93bec321480d2dc2622eedbc5?

agent/php_user_instrument_hashmap.c
ZNeumann: Yeah, I think that would be more clear when reading this code.

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+  } else if (NULL != zf->op_array.filename) {
+    /* and as last resort try hash of filename if available */
+    hash = ZSTR_HASH(zf->op_array.filename);
+  }
+
+  /* normalize to stay in-bound */
+  return (hash & ((1 << log2_num_buckets) - 1));
+}
+
+static inline bool zstr_equal(zend_string* zs1, zend_string* zs2) {
```
ZNeumann, line 206: Is this a function that we can extract into a common util file (like agent/php_zval.c)?

ZNeumann (approved):

agent/tests/test_user_instrument_hashmap.c
```diff
@@ -272,81 +362,229 @@ static void test_zend_string_hash_after_set_before_get() {
...
+
+  /* A function with the same in the same file but different scope */
+  mock_user_function_with_scope(&zf2, FILE_1_NAME, LINENO_BASE, SCOPE_2_NAME,
+                                FUNC_1_NAME);
+  nr_php_wraprec_hashmap_update(h, &zf2, &wr2);
+
+  wraprec_found = NULL;
+  rc = nr_php_wraprec_hashmap_get_into(h, &zf1, &wraprec_found);
+  tlib_pass_if_int_equal(
+      "Two functions with the same in the same file but different scope are "
```
mfulb, line 519:
```suggestion
      "Two functions with the same name in the same file but different scope are "
```

agent/tests/test_user_instrument_hashmap.c
```diff
@@ -272,81 +362,229 @@ static void test_zend_string_hash_after_set_before_get() {
...
+      "stored separetely",
+      1, rc);
+  tlib_pass_if_ptr_equal(
+      "Two functions with the same in the same file but different scope are "
```
mfulb, line 523:
```suggestion
      "Two functions with the same name in the same file but different scope are "
```

agent/tests/test_user_instrument_hashmap.c
```diff
@@ -272,81 +362,229 @@ static void test_zend_string_hash_after_set_before_get() {
...
+      "stored separetely",
+      &wr1, wraprec_found);
+
+  wraprec_found = NULL;
+  rc = nr_php_wraprec_hashmap_get_into(h, &zf2, &wraprec_found);
+  tlib_pass_if_int_equal(
+      "Two functions with the same in the same file but different scope are "
```
mfulb, line 530:
```suggestion
      "Two functions with the same name in the same file but different scope are "
```

agent/tests/test_user_instrument_hashmap.c
```diff
@@ -272,81 +362,229 @@ static void test_zend_string_hash_after_set_before_get() {
...
+      "stored separetely",
+      1, rc);
+  tlib_pass_if_ptr_equal(
+      "Two functions with the same in the same file but different scope are "
```
mfulb, line 534:
```suggestion
      "Two functions with the same name in the same file but different scope are "
```

mfulb: More like this below I didn't mark them all...

agent/php_user_instrument_hashmap.c
misiekn: @ZNeumann  Please see a09d04db3ce424dcb6012891c85f331c7a31d76a. If this is acceptable, please mark this conversation as resolved.

agent/tests/test_user_instrument_hashmap.c
misiekn: @mfulb Please see 050d68d9531f03d123725e6e35dbd73e97c3c012. Is this approach acceptable?

agent/tests/test_user_instrument_hashmap.c
misiekn: @mfulb Please see 050d68d9531f03d123725e6e35dbd73e97c3c012. Is this approach acceptable?

agent/tests/test_user_instrument_hashmap.c
misiekn: @mfulb Please see 050d68d9531f03d123725e6e35dbd73e97c3c012. Is this approach acceptable?

agent/tests/test_user_instrument_hashmap.c
misiekn: @mfulb Please see 050d68d9531f03d123725e6e35dbd73e97c3c012. Is this approach acceptable?

mfulb (approved):
Thanks for pulling all this work together Michal!

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+typedef struct _nr_php_wraprec_hashmap {
+  nr_php_wraprec_hashmap_dtor_fn_t dtor_func;
+  size_t log2_num_buckets;
+  nr_wraprecs_bucket_t** buckets;
+  size_t elements;
+} nr_php_wraprec_hashmap_t;
+
+static inline size_t nr_count_buckets(const nr_php_wraprec_hashmap_t* hashmap) {
+  return (size_t)(1 << hashmap->log2_num_buckets);
+}
```
bduranleau-nr, line 27: This is purely a style nit, but is there a reason to have these defined here as opposed to the .h file?  I think it would be slightly cleaner if all the necessary struct definitions were extracted out to the header.

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+    /*
+     * Basic sanity check: it's extremely unlikely that we'll ever need a
+     * hashmap for the user function wraprecs that has more than 2^24 buckets.
+     */
+    log2_num_buckets = 24;
+  }
+
+  hashmap
+      = (nr_php_wraprec_hashmap_t*)nr_malloc(sizeof(nr_php_wraprec_hashmap_t));
+  hashmap->dtor_func = dtor_fn;
```
bduranleau-nr, line 49: Would `NULL` be a valid value for `dtor_func`? Is there any kind of sanity checking that should be done on this arg?

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+  if (0 == buckets) {
+    actual_buckets = 0;
+  } else {
+    /*
+     * Calculate ceil(log2(buckets)), since that's what
+     * nr_hashmap_create_internal requires.
+     */
+
+    actual_buckets = 1;
+    while (((size_t)(1 << actual_buckets)) < buckets) {
```
bduranleau-nr, line 72: is there a maximum value that `buckets` should be checked for before entering this loop?

agent/php_user_instrument_hashmap.c
misiekn: This is intentionally not exposed in .h file. This type of code structure is called [opaque data type](https://en.wikipedia.org/wiki/Opaque_data_type). The internals of hashmap implementation should stay hidden away from the interface in case they need change.

agent/php_user_instrument_hashmap.c
misiekn: `NULL` is a valid value for `dtor_func`. `dtor_func == NULL` means that whatever is stored in the hashmap does not have to be destroyed whenever hashmap bucket is destroyed. The sanity check is done [here](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/agent/php_user_instrument_hashmap.c#L85)

agent/php_user_instrument_hashmap.c
misiekn: The final sanity check for maximum buckets that can ever be in the hashmap is done [here](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/agent/php_user_instrument_hashmap.c#L39-L44). The code itself is a copy of [axiom's hashmap implementation](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/axiom/util_hashmap.c#L23-L42) which also doesn't have any checks for `buckets` argument.

agent/php_user_instrument_hashmap.c
misiekn: Possibly. But this is not part of this PR.

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+  /* default to lineno */
+  uint32_t hash = nr_php_zend_function_lineno(zf);
+
+  if (NULL != zf->op_array.function_name && !nr_zf_is_unnamed_closure(zf)) {
+    /* but use hash of function name when possible */
+    hash = ZSTR_HASH(zf->op_array.function_name);
+  } else if (NULL != zf->op_array.filename) {
+    /* and as last resort try hash of filename if available */
+    hash = ZSTR_HASH(zf->op_array.filename);
+  }
```
bduranleau-nr, line 200: my hash implementation knowledge is rusty, but couldn't we hypothetically increase hash variability and reduce collisions if we added the hash values rather than just setting them based on an individual value? 
For example, 
lineno hash + function_name hash should be more unique than just lineno or just function name?

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+  if (!zstr_equal(key->filename, zf->op_array.filename)) {
+    /* no match: filename is different */
+    return false;
+  }
+
+  /* match: line number and filename are the same */
+  return true;
+}
+
+static int nr_php_wraprec_hashmap_fetch_internal(
```
bduranleau-nr, line 265: this looks like a truthy function, is there a reason we don't want the return type to be `bool` rather than int?

agent/php_user_instrument_hashmap.c
misiekn: Not really. Actually since lineno has very little entropy combining it with function_name or filename hash may reduce entropy from function_name hash especially that Zend's engine hashing function seems to be pretty well designed for strings and function names tend to be fairly unique (IMHO more unique than line numbers). Additionally the key comparison in this hashmap compares either function_name or filename so combining these two hashes also doesn't seem to make sense. I guess we'll have to wait for collisions stats from the field.

agent/php_user_instrument_hashmap.c
misiekn: API copy/pasted from [axiom's hashmap implementation](https://github.com/misiekn/newrelic-php-agent/blob/refactor/wraprec-lookup/axiom/util_hashmap.c#L286-L305).

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    bucket->wraprec = wr;
+    return;
+  }
+
+  nr_php_wraprec_hashmap_add_internal(hashmap, bucketidx, wr);
+}
+
+int nr_php_wraprec_hashmap_get_into(nr_php_wraprec_hashmap_t* hashmap,
```
bduranleau-nr, line 326: `bool` seems more appropriate than `int` for this function

bduranleau-nr (approved):

agent/php_user_instrument_hashmap.c
```diff
@@ -0,0 +1,345 @@
...
+    stats.collisions_mean /= stats.buckets_with_collisions;
+  }
+
+  nr_free(hashmap->buckets);
+  nr_realfree((void**)hashmap_ptr);
+
+  return stats;
+}
+
+static bool nr_zf_is_unnamed_closure(const zend_function* zf) {
```
zsistla, line 147: Could consider making this inline.

zsistla (approved):
Great job and thanks for the thorough test cases.

agent/php_user_instrument_hashmap.c
misiekn: Good idea! Done in 3930b15d3c4e8552676dea1b87bdf3c5a3b6f33f.

PR #607, opened 2023-02-01
feat(build): add 8.2 to Makefile and condense the php version lists

ZNeumann:


make/release.mk
```diff
@@ -40,17 +42,15 @@ ifeq (osx,$(RELEASE_OS))
...
 
-# Set the default PHP versions we will build.
-PHPS?=8.1 8.0 7.4 7.3 7.2 7.1 7.0 5.6 5.5
 # Set the PHP versions we support on 64-bit OSs.
-SUPPORTED_64=8.1 8.0 7.4 7.3 7.2 7.1 7.0 5.6 5.5
+SUPPORTED_64=$(PHP_VERSION_LIST)
 # Set the PHP versions we support on 32-bit OSs.
 SUPPORTED_32=
 # Of the PHP versions we want to build, determine which are supported on the detected OS.
 ifeq (x64,$(ARCH))
```
mfulb, line 50: Does this work with ARM64?  Also can we remove the 32 bit parts?

ZNeumann: I think both of those questions deserve a separate PR.

mfulb (approved):

mfulb:
You should also fix `make/release.mk`  to have a line (down around line 141) like:

`$(eval $(call RELEASE_AGENT_TARGET,8.2,<8.2 API NUM>))`

make/release.mk
misiekn: This code is not used by new pipeline. New pipeline needs https://github.com/newrelic/newrelic-php-agent/commit/04e2a8afe96d09a4cf1dd2d241b22bae4f98e09c. I will open a PR for changes to build system needed by GHA. And I agree with Zach - removal of code around `SUPPORTED_*` and `release-agent` target (which is used by legacy pipeline) deserves a separate PR.

ZNeumann:
ok jenkins

PR #613, opened 2023-02-06
build: enhance pcre and protobuf-c dependencies

misiekn:
Give the user an option to use `protobuf-c` from location alternative to
`vendor` subdirectory via `PROTOBUF_C_PREFIX` environmental variable.
`$PROTOBUF_C_PREFIX/bin` needs to have `protoc-c` binary, 
`$PROTOBUF_C_PREFIX/include` needs to have `protobuf-c/protobuf-c.h`
header file and `$PROTOBUF_C_PREFIX/lib` needs to have `libprotobuf-c.a`. 
By default agent's build system will try to look for `protobuf-c` dependency
in location returned by `pkg-config libprotobuf-c --variable=prefix`, a
location where `protobuf-c` is installed in the system via package
managers. If that location does not have `protobuf-c`, agent's build
system will fallback to using `protobuf-c` from `vendor` directory. This
enhancement speeds up agent build because neither `vendor/protobuf` nor
`vendor/protobuf-c` need to be build if `protobuf-c` is installed in the
system by the user.

If user or system provided `protobuf-c` does not meet agent's build system
requirements, the build will abort with an error message.

Force the agent to use static version of pcre library. This avoids the
issue with runtime dependency on libpcre shared object that is hard to
satisfy universally because different Linux distributions use different
name of the shared object: libpcre.so or libpcre3.so. Additionally the
user has an option to point to pcre library installation via
`PCRE_PREFIX` environmental variable. By default agent's build system 
will try to look for pcre dependency using `pkg-config`.

If user or system provided `pcre` does not meet agent's build system
requirements, the build will abort with an error message.

To troubleshoot build issues caused by dependencies, agent will log which
dependencies it is using.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,17 @@ echo_config:
 	@echo GIT_COMMIT=$(GIT_COMMIT)
 
 #
+# Print information where do the build dependencies come from
+#
+.PHONY: show-vendors
+show-vendors:
+	@echo ""
+	@echo -----------------------------------------------------------------------
+	@echo "| Using pcre library from $(PCRE_PREFIX) (from $(origin PCRE_PREFIX))"
+	@echo "| Using protobuf-c library from $(VENDOR_PREFIX) (from $(origin VENDOR_PREFIX))"
+	@echo -----------------------------------------------------------------------
+	@echo ""
+#
 # Let's build an agent! Building an agent is a three step process: using phpize
 # to build a configure script, using configure to build a Makefile, and then
 # actually using that Makefile to build the agent extension.
@@ -116,7 +127,7 @@ PHPIZE := phpize
 PHP_CONFIG := php-config
 
 .PHONY: agent
-agent: agent/Makefile
+agent: show-vendors agent/Makefile
 	$(MAKE) -C agent
 
 agent/configure: agent/config.m4 agent/Makefile.frag
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,7 @@ agent/configure: agent/config.m4 agent/Makefile.frag
 	cd agent; $(PHPIZE) --clean && $(PHPIZE)
 
 agent/Makefile: agent/configure | axiom
-	cd agent; ./configure $(SILENT) --enable-newrelic --with-axiom=$(realpath axiom) --with-php-config=$(PHP_CONFIG) --with-protobuf-c=$(VENDOR_PREFIX) --with-pcre=$(PCRE_PREFIX)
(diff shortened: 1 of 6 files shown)


ZNeumann (approved):

zsistla (approved):
LGTM, thanks.
Have you tried it with bigchange to ensure it still works with the legacy release pipeline?

Makefile
```diff
@@ -305,7 +316,7 @@ daemon-protobuf: src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go
 src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go: protocol/infinite_tracing/v1.proto
 	$(MAKE) vendor # Only build vendor stuff if v1.proto has changed. Otherwise
 	               # this rule will be triggered every time the daemon is built.
-	$(VENDOR_BASE)/local/bin/protoc \
+	$(VENDOR_PREFIX)/bin/protoc \
```
mfulb, line 319: Should this be `PROTOBUF_C_PREFIX`?

misiekn: I was thinking about it but decided this will require a separate PR for the following reasons:
1. `protoc` is not `protoc-c`. `protoc` is protobuf compiler with support for many languages but C. `protoc` is provided by `protobuf`. protoc-c is built on top of `protoc` and adds support for C language. In this particular target `protoc` is needed, not `protoc-c` and this PR only makes updates to provider of `protoc-c`.
2. [src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go](https://github.com/newrelic/newrelic-php-agent/blob/dev/src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go) is stored in the repository and this target is not usually triggered in day to day builds. `src/newrelic/infinite_tracing/com_newrelic_trace_v1/v1.pb.go` has to be rebuild manually after `protocol/infinite_tracing/v1.proto` gets an update and the result of the manual build has to be checked into the repository. If this target needs to be rebuilt, not only `protoc` needs to be available, but also Go protocol buffer plugin needs to be manually installed - see [Makefile#L295-L298](https://github.com/newrelic/newrelic-php-agent/blob/dev/Makefile#L295-L298).

Given 1. and especially 2., which requires additional manual effort to properly configure build environment, I decided for this PR not to touch this area of the build system and have it use `protoc` from vendor subdirectory, which will be built by the command on line 317.

mfulb (approved):

bduranleau-nr (approved):

misiekn:
ok jenkins

PR #620, opened 2023-02-17
build: refactor release targets

lavarou:
Refactor release targets to give more flexibility in building them 
without triggering legacy release workflow.

Separate release-installer-script and release-installer-iutil targets
from release-installer target. This is needed because newrelic-install
is not a binary artifact, which does not depend on platform arch, and
newrelic-iutil is a binary artrifact, which depens on platform arch.
Separating these targets out simplifies building agent release when
using containers as build platform.

Separate release-version target from release target. This is needed
when VERSION and COMMIT artifacts are needed without the need to do
a full build (which build the daemon and the agent for all PHPs), e.g.
while testing release workflow.

Add a target for building the agent for a given PHP version. Works well
when building agent using containers. This is useful not only in GitHub
Actions workflows, but also in a day to day development, because it
allows to preserve agent between PHP version switches.

### DIFF
--- a/make/release.mk
+++ b/make/release.mk
@@ -54,7 +54,10 @@ else
 endif
 
 
-release: Makefile release-daemon release-installer release-agent release-docs release-scripts | releases/$(RELEASE_OS)/
+.PHONY: release
+release: Makefile release-version release-daemon release-installer release-agent release-docs release-scripts | releases/$(RELEASE_OS)/
+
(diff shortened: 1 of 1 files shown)


ZNeumann (approved):

zsistla (approved):
Makes sense to modularize, thanks.

bduranleau-nr (approved):

PR #621, opened 2023-02-22
test: update integration tests for PHP 8.2

bduranleau-nr:


bduranleau-nr:
ok jenkins

tests/integration/memcache/test_memcache_add.php82.php
```diff
@@ -0,0 +1,116 @@
...
+/*DESCRIPTION
+The agent should report Memcache metrics when keys are added using the Memcache
+procedural api.
+*/
+
+/*SKIPIF
+<?php require('skipif.inc');
+
+if (version_compare(PHP_VERSION, "8.2", "<")) {
+  die("skip: PHP 8.2 exclusive\n");
```
mfulb, line 16: More accurately isn't this "PHP 8.2+ exclusive"?  Maybe good to change the message or it could be misleading.

mfulb (approved):

tests/integration/memcache/test_memcache_add.php82.php
bduranleau-nr: Since the issue here is the memcache dependency itself, I think it may be better to keep it tied to 8.2, and if memcache releases the 8.2 fix that resolves this, this problem will be resolved for 8.3 and this can be removed entirely

ZNeumann (approved):

agent/php_call.c
```diff
@@ -48,8 +52,29 @@ zval* nr_php_call_user_func(zval* object_ptr,
...
+#if ZEND_MODULE_API_NO >= ZEND_8_2_X_API_NO /* PHP 8.2+ */
+  /*
+   * With PHP 8.2, functions that do not exist will cause a fatal error to
+   * be thrown. `zend_call_method_if_exists` will attempt to call a function and
+   * silently fail if it does not exist
+   */
+  if (NULL != object_ptr) {
+    object = Z_OBJ_P(object_ptr);
+  } else {
+    object = NULL;
```
zsistla, line 64: Since object is already assigned NULL [here](https://github.com/newrelic/newrelic-php-agent/pull/621/files#diff-92a37f49fed3190bcf77a05cd9b996d03bc1a904264d7131a86b64a4860b1ccaR19) is this else statement necessary?

zsistla (approved):

agent/php_call.c
bduranleau-nr: No, technically it isn't- I believe when we were going through this fix it was added for developer clarity, but isn't necessary.

bduranleau-nr: Fixed in d340b89bd3739c2c2b91ba7051ee1e2434f05ff5

mfulb: I think this change is dangerous - if someone sees there is an `object` variable to use above this code and assign it to something other than NULL then this existing code will not work properly.  I don't think it is good to assume someone will catch this side effect so being defensive and setting it NULL is just being safe.

bduranleau-nr: If someone sees `object` and assigns it to a non-null value, it will just be overwritten when it reaches this point, which is right before it gets passed to `zend_call_method_if_exists`, after which point it isn't used again. Given that I think it would take a degree of intentional effort to break this functionality.
That said I don't have an issue restoring the `else` statement to make the intention clear to the developer, even if it is strictly unnecessary from a code point of view.
@zsistla - are you comfortable with @mfulb's suggestion or do you prefer the `else` statement be removed?

zsistla: If someone assigns it to something other than NULL and we overwrite it with NULL (possibly not freeing it), wouldn't that also cause issues?  I'm okay with returning the else statement, it wasn't a blocker for me either way.

bduranleau-nr: Reverted in 6249464d66a8ef6a2acca661ac7187cf6f21dcf4.

bduranleau-nr:
ok jenkins

bduranleau-nr:
ok jenkins

PR #624, opened 2023-02-28
build: pin libc's symbol versions

bduranleau-nr:
To maintain New Relic's PHP Agent compatibility with older runtime platforms, use GNU's Assembler's [as](https://sourceware.org/binutils/docs/as/index.html) [.symver](https://sourceware.org/binutils/docs/as/Symver.html) directive to pin glibc's symbols to versions compatible with RHEL7, which uses glibc version 2.17.

### DIFF
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -25,6 +25,8 @@
 #if (((__GLIBC__) > 2) || (((__GLIBC__) == 2) && ((__GLIBC_MINOR__) > 17)))
 #if defined (__x86_64__)
   __asm__(".symver pow,pow@GLIBC_2.2.5");
+#elif defined (__aarch64__)
+  __asm__(".symver pow,pow@GLIBC_2.17");
 #endif
 #endif
 #endif
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -23,7 +23,11 @@
  */
 #if defined(__GLIBC__)
 #if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
-__asm__(".symver pow,pow@GLIBC_2.17");
+#if defined (__x86_64__)
+  __asm__(".symver pow,pow@GLIBC_2.2.5");
+#elif defined (__aarch64__)
+  __asm__(".symver pow,pow@GLIBC_2.17");
+#endif
 #endif
 #endif
 
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -19,7 +19,7 @@
  *
  * To pin the version of `pow` to a version compatible with all NR PHP Agent
  * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
- * select an older version- in this case, 2.17.
+ * select an older version of the symbol.
  */
 #if defined(__GLIBC__)
 #if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -22,9 +22,7 @@
  * select an older version- in this case, 2.17.
  */
 #if defined(__GLIBC__)
-#if ZEND_MODULE_API_NO == ZEND_8_2_API_NO
 __asm__(".symver pow,pow@GLIBC_2.17");
-#endif
 #endif
 
 void nr_app_harvest_init(nr_app_harvest_t* ah,
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -22,7 +22,7 @@
  * select an older version of the symbol.
  */
 #if defined(__GLIBC__)
-#if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
+#if (((__GLIBC__) > 2) || (((__GLIBC__) == 2) && ((__GLIBC_MINOR__) > 17)))
 #if defined (__x86_64__)
   __asm__(".symver pow,pow@GLIBC_2.2.5");
 #endif
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -22,7 +22,9 @@
  * select an older version- in this case, 2.17.
  */
 #if defined(__GLIBC__)
+#if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
 __asm__(".symver pow,pow@GLIBC_2.17");
+#endif
 #endif
 
 void nr_app_harvest_init(nr_app_harvest_t* ah,
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -25,8 +25,6 @@
 #if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
 #if defined (__x86_64__)
   __asm__(".symver pow,pow@GLIBC_2.2.5");
-#elif defined (__aarch64__)
-  __asm__(".symver pow,pow@GLIBC_2.17");
 #endif
 #endif
 #endif
--- a/axiom/nr_app_harvest.c
+++ b/axiom/nr_app_harvest.c
@@ -12,6 +12,21 @@
 #include "nr_app_harvest_private.h"
 #include "util_logging.h"
 
+/*
+ * The build pipeline for PHP 8.2 uses a Debian image that is built on a
+ * version of glibc that contains an newer glibc version symbol for `pow()`
+ * which is incompatible with older distributions, causing RPM installs to fail.
+ *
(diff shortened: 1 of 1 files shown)


axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,21 @@
...
+/*
+ * The build pipeline for PHP 8.2 uses a Debian image that is built on a
+ * version of glibc that contains an newer glibc version symbol for `pow()`
+ * which is incompatible with older distributions, causing RPM installs to fail.
+ *
+ * To pin the version of `pow` to a version compatible with all NR PHP Agent
+ * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
+ * select an older version- in this case, 2.17.
+ */
+#if defined(__GLIBC__)
```
mfulb, line 24:
```suggestion
#ifdef __GLIBC__
#if __GLIBC__ >= 2
#if __GLIBC_MINOR__ > 17
  __asm__(".symver pow,pow@GLIBC_2.17");
#endif
#endif
#endif
```

mfulb: You'll also need to add a 

#if defined(__GLIBC__)
#include <gnu/libc-version.h>
#endif

at the top of the file to get the include with the __GLIBC__ and __GLIBC_MINOR__ defines.

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,21 @@
...
+#if ZEND_MODULE_API_NO == ZEND_8_2_API_NO
+__asm__(".symver pow,pow@GLIBC_2.17");
+#endif
```
misiekn, line 27: axiom has no knowledge about PHP version, i.e. `#if ZEND_MODULE_API_NO == ZEND_8_2_API_NO` directive is meaningless. It needs to be removed:
```suggestion
__asm__(".symver pow,pow@GLIBC_2.17");
```

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,19 @@
...
+ * version of glibc that contains an newer glibc version symbol for `pow()`
+ * which is incompatible with older distributions, causing RPM installs to fail.
+ *
+ * To pin the version of `pow` to a version compatible with all NR PHP Agent
+ * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
+ * select an older version- in this case, 2.17.
+ */
+#if defined(__GLIBC__)
+__asm__(".symver pow,pow@GLIBC_2.17");
+#endif
```
misiekn, line 26: Extending range of @mfulb's [suggestion](https://github.com/newrelic/newrelic-php-agent/pull/624/files#r1120806448) so that it can be applied from GitHub's web UI:
```suggestion
#if defined(__GLIBC__)
#if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
__asm__(".symver pow,pow@GLIBC_2.17");
#endif
#endif
```

axiom/nr_app_harvest.c
misiekn: First suggestion added in 661beed2b6ba197adefdb386b3a59cbad149a5fe.

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,21 @@
...
+ * version of glibc that contains an newer glibc version symbol for `pow()`
+ * which is incompatible with older distributions, causing RPM installs to fail.
+ *
+ * To pin the version of `pow` to a version compatible with all NR PHP Agent
+ * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
+ * select an older version- in this case, 2.17.
+ */
+#if defined(__GLIBC__)
+#if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
+__asm__(".symver pow,pow@GLIBC_2.17");
```
misiekn, line 26: libm does not provide the same versions of a symbol across different architectures.
x86_64 has 2.2.5 and 2.29:
```
root@fa2762fd1382:/# objdump -T /lib/x86_64-linux-gnu/libm-2.31.so | grep -w pow
0000000000040040 g    DF .text	00000000000000b6  GLIBC_2.29  pow
0000000000010020 g    DF .text	000000000000014a (GLIBC_2.2.5) pow
root@fa2762fd1382:/# uname -m
x86_64
root@fa2762fd1382:/# grep PRETTY /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
```
aarch64 has 2.17 and 2.29:
```
root@6d553b8b57f8:/# objdump -T /lib/aarch64-linux-gnu/libm-2.31.so | grep -w pow
0000000000035ff0 g    DF .text	000000000000044c  GLIBC_2.29  pow
000000000000ebb0 g    DF .text	000000000000013c (GLIBC_2.17) pow
root@6d553b8b57f8:/# uname -m
aarch64
root@6d553b8b57f8:/# grep PRETTY /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
```

```suggestion
#if defined (__x86_64__)
  __asm__(".symver pow,pow@GLIBC_2.2.5");
#elif defined (__aarch64__)
  __asm__(".symver pow,pow@GLIBC_2.17");
#endif
```

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,25 @@
...
 #include "util_logging.h"
 
+/*
+ * The build pipeline for PHP 8.2 uses a Debian image that is built on a
+ * version of glibc that contains an newer glibc version symbol for `pow()`
+ * which is incompatible with older distributions, causing RPM installs to fail.
+ *
+ * To pin the version of `pow` to a version compatible with all NR PHP Agent
+ * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
+ * select an older version- in this case, 2.17.
```
misiekn, line 22: Symbol version is dependent on arch. Adjust the comment to be more generic and let the code do the specifics:
```suggestion
 * select an older version of the symbol.
```

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,25 @@
...
+ *
+ * To pin the version of `pow` to a version compatible with all NR PHP Agent
+ * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
+ * select an older version of the symbol.
+ */
+#if defined(__GLIBC__)
+#if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
+#if defined (__x86_64__)
+  __asm__(".symver pow,pow@GLIBC_2.2.5");
+#elif defined (__aarch64__)
```
mfulb, line 28:
```suggestion
```
I don't think we need this since we are not packaging on ARM64.

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,25 @@
...
+  __asm__(".symver pow,pow@GLIBC_2.17");
+#endif
```
mfulb, line 30:
```suggestion
```

mfulb: I think this is only needed if we are trying to package the ARM64 binaries.  Probably better not to symbol version when we don't have to.

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,25 @@
...
+ * The build pipeline for PHP 8.2 uses a Debian image that is built on a
+ * version of glibc that contains an newer glibc version symbol for `pow()`
+ * which is incompatible with older distributions, causing RPM installs to fail.
+ *
+ * To pin the version of `pow` to a version compatible with all NR PHP Agent
+ * supported OSes, we utilize `asm()` and `.symver` to instruct the linker to
+ * select an older version of the symbol.
+ */
+#if defined(__GLIBC__)
+#if ((__GLIBC__) >= 2) && ((__GLIBC_MINOR__) > 17)
```
zsistla, line 25: I think this logic wouldn't quite work once glibc goes to a version 3.

axiom/nr_app_harvest.c
```diff
@@ -12,6 +12,25 @@
...
+#if defined (__x86_64__)
+  __asm__(".symver pow,pow@GLIBC_2.2.5");
+#elif defined (__aarch64__)
+  __asm__(".symver pow,pow@GLIBC_2.17");
```
bduranleau-nr, line 29:
```suggestion
```

axiom/nr_app_harvest.c
bduranleau-nr: Good eye, I think this is probably what we would need

```suggestion
#if (((__GLIBC__) > 2) || (((__GLIBC__) == 2) && ((__GLIBC_MINOR__) > 17)))
```

mfulb: Looks good.

mfulb (approved):

axiom/nr_app_harvest.c
bduranleau-nr: Resolved in de1769ac677c613c9bc34cc229e668f207a86780.

PR #625, opened 2023-03-01
build: don't enforce linking to static libpcre

misiekn:
By default let the agent link to whatever libpcre is available on the system (most likely a shared object). However for release builds the agent needs to link with a static version of libpcre to ensure it runs on all Linux distros. This is needed because different Linux distributions use different names for the shared object and its impossible to link to a shared object that works universally on all Linux distributions. If PCRE_STATIC is set to yes, agent's build system will enforce linking to static version of libpcre and if it is not available, the build will abort.

Supplements #613


zsistla (approved):
Thanks for working to get this the flexibility it needs for the multiple environments/scenarios!

bduranleau-nr (approved):

PR #629, opened 2023-03-06
test(daemon): fix non-deterministic failures

ZNeumann:
Some tests were assuming that the time the metrics were created would be the same time that the metrics were sent. This non-deterministically fails if the brief period between those times in the test spanned a unix time step. ([code that does this seen here](https://github.com/newrelic/newrelic-php-agent/blob/main/src/newrelic/metrics.go#L282))

This PR fixes that assumption and properly grabs both the different times for creation of the "truth" metrics object.

### DIFF
--- a/src/newrelic/processor_test.go
+++ b/src/newrelic/processor_test.go
@@ -257,8 +257,9 @@ func TestProcessorHarvestDefaultData(t *testing.T) {
 			t.Fatal(string(append(cp.data, cp2.data...)))
 		}
 	}
-	time := strings.Split(string(cp3.data), ",")[1]
-	usageMetrics := `["one",` + time + `,` + time + `,` +
+	time1 := strings.Split(string(cp3.data), ",")[1]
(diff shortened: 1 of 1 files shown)


misiekn (approved):
It took me a moment to get it but the gist is to fill the timestamps in the expectation using the data from the results and the two timestamps can be different therefore both need to be extracted from the results.

zsistla (approved):

ZNeumann:
ok jenkins

mfulb:
Does that make this a self-fullfilling test?  Or are the times not important to what is being tested?

ZNeumann:
> Does that make this a self-fullfilling test? Or are the times not important to what is being tested?

The times are not what are being tested and were already self-fulfilling. This just makes it self-fulfilling in a way that doesn't occasionally fail.

mfulb (approved):

ZNeumann:
ok jenkins

PR #630, opened 2023-03-07
style(daemon): go test warnings

ZNeumann:
Fixes two minor daemon issues that newer go test versions (1.10+) dislike. ([further discussion here](https://github.com/golang/go/issues/23062))

1) format directives in Fprintln. There is an Fprintln call that should be an Fprintf because it contains formatting directives. Previously, the directive was ignored and the additional arguments were added to the end of the string.

2) passing an extra newline to Fprintln. Go dislikes Fprintln being passed a string that ends in a newline. To achieve the exact same previous functionality, it is now a Fprint call with an additional newline character.


zsistla (approved):
Thanks for fixing this.

misiekn (approved):

bduranleau-nr (approved):

PR #542, opened 2022-09-29
feat(agent): use Observer API to hook into Zend Engine for PHPs 8.0+

zsistla:
Use Zend's Observer API to hook into Zend Engine and instrument user functions for PHPs 8.0+. This method of hooking into Zend Engine is compatible with Zend Engine's JIT, which will no longer be disabled when the agent is installed.

PR #638, opened 2023-03-11
fix(agent):Update logic needed to call PHP functions from the agent.

zsistla:
The try/catch block was added for PHP 8+ because in the case of exceptions according to:  https://www.php.net/manual/en/function.call-user-func.php Callbacks registered with functions such as call_user_func() and call_user_func_array() will not be called if there is an uncaught exception thrown in a previous callback. So if we call something that causes an exception, it will block us from future calls that use call_user_func or call_user_func_array.

Valgrind showed the agent and/or the php system wasn't properly cleaning up after such cases.

Newer compilers had issues with this when compiling with any optimization and generated the following error: `might be clobbered by 'longjmp' or 'vfork' [-Werror=clobbered]`

This PR addresses the error by moving assignment of the offending variables into the try/catch block then @lavarou used a local variable to store retval until we are sure we will use it.

@lavarou read more about using the local variable here:
While I was looking for some information about -Wclobber and learning about zend_try/zend_catch I found these two interesting resources:
https://github.com/php/php-src/pull/5151 - this basically shows that PHP team is aware of issues caused by -Wclobber and basically turn these warnings off (I don’t know if it’s good or bad that PHP is ignoring compiler warnings though; maybe they know what they’re doing)
https://github.com/php/php-src/blob/master/main/streams/userspace.c#L335-L340 - this is an example in PHP source code how zend_call_method_if_exists is called. It is a nice little trick of using an automatic variable to store user function call result. Based on this I came up with this idea: https://github.com/newrelic/newrelic-php-agent/compare/compiler_error...lavarou/fix/php-call-try-catch?expand=1. It seems to resolve the issue - I haven’t tested it past agent’s unit tests (make agent-valgrind) for PHP 8.2. Using automatic/local variable has this advantage that memory allocation is delayed until it is certain retval is good for use - i.e. zend_call_method_if_exists succeeded. Clobbering (longjmp called from within zend_call_method_if_exists which will cause [this line](https://github.com/newrelic/newrelic-php-agent/blob/compiler_error/agent/php_call.c#L73) to jump [here](https://github.com/newrelic/newrelic-php-agent/blob/compiler_error/agent/php_call.c#L107)) may invalidate the value of retval pointer and at the time it is freed [here](https://github.com/newrelic/newrelic-php-agent/blob/compiler_error/agent/php_call.c#L92) it may point to invalid memory and cause segfault.

There are unit tests that test that were added with oapi that exercise this functionality(nr_php_call).  Adding the tests [here](https://github.com/newrelic/newrelic-php-agent/pull/542/files#diff-c175fdeafd6ad73c1b624822523084b3f5a3a98a303c9db389083a4bb2d50387R122) and [here](https://github.com/newrelic/newrelic-php-agent/blob/4209f55f749c5718bd86ed9adec53965cc57a86d/agent/tests/test_php_wrapper.c)(extensively used, with exception and non exception cases search for nr_php_call) and also [here](https://github.com/newrelic/newrelic-php-agent/pull/542/files#diff-c175fdeafd6ad73c1b624822523084b3f5a3a98a303c9db389083a4bb2d50387R146) are what initially caused the try/catch to be added because we weren't handling those cases properly and were causing multiple valgrind issues.

Additionally, combined the 8+ blocks for functionality and simplicity.

`zend_call_method_if_exists` that had been used in the 8.2+ block was suppressing a valid warning message.

When we ran it, because we forced Wordpress as a framework, we made a call in fw_wordpress.c which makes a call to get_theme_roots.
However, in the test case 
get_theme_roots (which normally exists in Wordpress) does not exist in our "mocked" Wordpress so we received a perfectly valid error msg.

Fatal error: Uncaught Error: Invalid callback get_theme_roots, function "get_theme_roots" not found or invalid function name in /usr/local/src/newrelic-php-agent/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php:66
Stack trace:
#0 /usr/local/src/newrelic-php-agent/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php(52): g('string2')
#1 /usr/local/src/newrelic-php-agent/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php(75): apply_filters('g', 'string2')
#2 /usr/local/src/newrelic-php-agent/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php(52): f('string1')
#3 /usr/local/src/newrelic-php-agent/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php(79): apply_filters('f', 'string1')
#4 {main}
  thrown in /usr/local/src/newrelic-php-agent/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php on line 66

If we use nr_php_call we don't want it silently suppressing exceptions and continuing on as if nothing had happened.  The exceptions are warnings for us to keep things updated.

The error message is good, the problem is that after this, we weren't detecting that something invalid happened and were copying/returning retval even though we shouldn't.
The fix is to 
a) check retval for undef and if so, return NULL.
b) add get_theme_roots to the test case.

2) initialize retval to prevent segfaults.
zval retval = {0};

ETA:
Unrolling the change to use `call_user_function` for all.  Using call_user_function for all was simpler and more maintainable and worked for tested PHP versions across multiple apps but could not handle the Wordpress/php8.2 combo.
zend_call_method_if_exists seems to be the more robust function that wordpress/php 8.2 needs.

### DIFF
--- a/agent/php_call.c
+++ b/agent/php_call.c
@@ -14,14 +14,17 @@ zval* nr_php_call_user_func(zval* object_ptr,
                             const char* function_name,
                             zend_uint param_count,
                             zval* params[] TSRMLS_DC) {
-#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ if clause 1*/
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP 7.0+ */
+#if ZEND_MODULE_API_NO >= ZEND_8_2_X_API_NO
+  zend_object* object = NULL;
+  zend_string* method_name = NULL;
+#endif
   int zend_result = FAILURE;
   zval* fname = NULL;
   HashTable* symbol_table = NULL;
   zval* param_values = NULL;
   zval retval = {0};
   zval* retval_copy = NULL;
-
 #ifndef PHP8
   int no_separation = 0;
 #endif /* PHP8 */
@@ -45,10 +48,8 @@ zval* nr_php_call_user_func(zval* object_ptr,
       param_values[i] = *params[i];
     }
   }
-
   fname = nr_php_zval_alloc();
   nr_php_zval_str(fname, function_name);
-
   /*
    * For PHP 8+, in the case of exceptions according to:
    * https://www.php.net/manual/en/function.call-user-func.php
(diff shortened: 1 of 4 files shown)


agent/php_call.c
```diff
@@ -48,53 +49,91 @@ zval* nr_php_call_user_func(zval* object_ptr,
...
+  nr_php_zval_free(&fname);
+
+  nr_free(param_values);
+  if (SUCCESS == zend_result) {
+    return retval;
   }
-  zend_catch { zend_result = FAILURE; }
+  nr_php_zval_free(&retval);
+  return NULL;
   zend_end_try();
```
lavarou, line 101: I don't know a whole lot about zend_try/zend_catch/zend_end_try but intuitively the order of these lines does not seem right. I.e. returning from function before `zend_end_try`.

zsistla: good point.  please see: [d818132](https://github.com/newrelic/newrelic-php-agent/pull/638/commits/d8181329c518cfe29489eb30e4141afe978bd745)

lavarou: Resolved with #640

lavarou (approved):
👍

mfulb (approved):

tests/integration/frameworks/wordpress/test_wordpress_do_action.php8.php
```diff
@@ -52,6 +52,10 @@ function do_action($tag, ...$args) {
     call_user_func_array($tag, $args);
 }
 
+//Simple mock of wordpress's get_theme_roots
+function get_theme_roots() {
+}
+
```
lavarou, line 58: How is this function used in the test?

lavarou (approved):
Great find about catching retval being undefined! Also thanks for cleaning `nr_php_call_user_func` - it looks 💯 better! 👍

tests/integration/frameworks/wordpress/test_wordpress_do_action.php8.php
zsistla: We are mocking wordpress here and forcing the framework detection to the wordpress.
In wordpress, we'll always try to call get_theme_roots, a wordpress provided function.
https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_wordpress.c#L217-L229

zsistla:
ok jenkins

ZNeumann (approved):

tests/integration/frameworks/wordpress/test_wordpress_do_action.php8.php
lavarou: Could we make it a more complete mock like this:
```suggestion
//Simple mock of wordpress's get_theme_roots
function get_theme_roots() {
    return '/themes';
}
```
(based on https://github.com/WordPress/wordpress-develop/blob/6.1/src/wp-includes/theme.php#L380). I know the agent will be able to work with `get_theme_roots` that doesn't return anything useful but maybe it's worthwhile to test using something closer to real life. Or even better have three tests: 
1. do not return anything useful
2. return '/themes' string
3. return array of strings

zsistla:
ok jenkins

tests/integration/frameworks/wordpress/test_wordpress_do_action.php8.php
zsistla: Yes, more mock could be added, but it's not part of this PR.
It was to have a function that the internal instrumentation could call without erroring.
The fw_wordpress logic handles if get_themes_roots doesn't return anything useful, so leaving one test like this and more that return other things helps exercise more code paths.

zsistla:
ok jenkins

zsistla:
Closed in lieu of PR #703 .

PR #641, opened 2023-03-16
feat(daemon): Add loglevel for daemon health

mfulb:
This PR adds a new log level for the daemon called "healthcheck".  It is between "debug" and "info" in verbosity. 
If enabled this log level will currently output a daemon log message of type "HealthCheck" which will output the current number of applications connected to the agent as well as the maximum number of allowed applications.
The format of the line is:
```log.Healthf("current number of apps is %d of a max of %d", numapps, limits.AppLimit)```
where ```numapps``` is the current number of applications and ```limits.AppLimit``` is the compile-time limit for the number of applications a daemon instance will allow.

### DIFF
--- a/agent/scripts/newrelic.ini.template
+++ b/agent/scripts/newrelic.ini.template
@@ -160,7 +160,7 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;          rarely need to change this from the default, and usually only under
 ;          the guidance of technical support.
 ;          Must be one of the following values:
-;            always, error, warning, healthcheck, info, debug
+;            always, error, warning, info, healthcheck, debug
 ;
 ;          The values verbose and verbosedebug are deprecated aliases for debug.
 ;
--- a/agent/scripts/newrelic.ini.template
+++ b/agent/scripts/newrelic.ini.template
@@ -160,7 +160,7 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;          rarely need to change this from the default, and usually only under
 ;          the guidance of technical support.
 ;          Must be one of the following values:
-;            always, error, warning, info, debug
+;            always, error, warning, healthcheck, info, debug
 ;
 ;          The values verbose and verbosedebug are deprecated aliases for debug.
 ;
--- a/src/newrelic/processor.go
+++ b/src/newrelic/processor.go
@@ -377,7 +377,7 @@ func (p *Processor) processAppInfo(m AppInfoMessage) {
 	app = NewApp(m.Info)
 	p.apps[key] = app
 	numapps = len(p.apps)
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: agent/scripts/newrelic.ini.template [hunk 1]
```diff
@@ -160,7 +160,7 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
 ;          rarely need to change this from the default, and usually only under
 ;          the guidance of technical support.
 ;          Must be one of the following values:
-;            always, error, warning, info, debug
+;            always, error, warning, info, healthcheck, debug
```
bduranleau-nr, line 163: If we're adding this to the existing daemon.loglevel `ini`, are we forcing anyone who wants this functionality to also have error+warning+info? What if they want the healthcheck, but don't want logs any higher than `warning`?

bduranleau-nr: I think it would be better suited in the long run to make this feature it's own INI value that can be set independently of other values. In my opinion it serves a specific purpose that doesn't fit neatly into the existing `newrelic.daemon.loglevel` INI, or any other currently existing INI's that I can think of.

mfulb: My understanding is the request was for a log level more verbose than 'info' that provided the health information.  There was no specification of picking and choosing the log level messages that would be reported.

src/newrelic/processor.go
```diff
@@ -818,10 +823,13 @@ func (p *Processor) doHarvest(ph ProcessorHarvest) {
 	if p.cfg.AppTimeout > 0 && app.Inactive(p.cfg.AppTimeout) {
 		log.Infof("removing %q with run id %q for lack of activity within %v",
 			app, id, p.cfg.AppTimeout)
-		if len(p.apps) == limits.AppLimitNotifyLow {
+		numapps := len(p.apps)
+		if numapps == limits.AppLimitNotifyLow {
 			log.Infof("current number of apps is %d",
 				limits.AppLimitNotifyLow)
 		}
+		log.Healthf("current number of apps is %d of a max of %d",
```
ZNeumann, line 831: This line will double log when numapps == limites.AppLimitNotifyLow. Do we want this, do we want to change the Infof log a little, or do we want the Healthf log in an else?

lavarou (approved):
LGTM

bduranleau-nr (approved):

zsistla (approved):
Thanks!

src/newrelic/processor.go
mfulb: The existing ```Infof ```output could potentially be in use by some customers so I was not comfortable removing it.  It was added to service a similar request from a customer but I don't know they would want to transition to the 'healthcheck' log level due to the considerably higher potential log output.

mfulb:
ok jenkins

PR #642, opened 2023-03-17
feat(integration_runner): Add 'warn' test result

mfulb:
This commit adds a 'warn' test result which is intended to be used to represent tests which could not be run due to something not being correct in the environment.  

The intention is 'skip' is for tests which cannot be run because of the PHP version or some other expected condition which would make the test not run in all environments.   However, currently 'skip' is used to also represent when a test which is expected to run could not run due to a misconfiguration of the environment.  This can cover up potential test run deficiencies.

The use of 'warn' means the test environment is mis-configured. 

A new integration_runner boolean option 'warnisfail' will cause any warnings to be treated as a failure and the integration_runner will exit with a non-zero exit code to signify the failure.

The 'warn' result can be returned by doing something like:

 ```die("warn: could not load important thing so test not run")```.

### DIFF
--- a/src/newrelic/integration/test.go
+++ b/src/newrelic/integration/test.go
@@ -63,6 +63,7 @@ type Test struct {
 
 	// Remaining fields are populated after the test is run.
 	Skipped bool
+    Warned bool
 
 	// If the test was skipped or the test could not be run due to an
 	// error, describes the reason.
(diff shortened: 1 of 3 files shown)


src/newrelic/integration/test.go
```diff
@@ -63,6 +63,7 @@ type Test struct {
 
 	// Remaining fields are populated after the test is run.
 	Skipped bool
+    Warned bool
```
ZNeumann, line 66:
```suggestion
        Warned bool
```

ZNeumann: I think my suggestion probably will not do what we want (tabs vs spaces etc), but this line is misaligned

ZNeumann (approved):

lavarou (approved):
Nice addition! It would be good to see it in action, i.e. have a few tests updated for demonstration purposes.

src/newrelic/integration/test.go
mfulb: I think [07f7c3ca](https://github.com/newrelic/newrelic-php-agent/pull/642/commits/07f7c3cab645a6d79e453fa2d478eac7fa2a4731) fixes the spacing?

mfulb:
ok jenkins

mfulb:
ok jenkins

mfulb:
ok jenkins

PR #649, opened 2023-03-27
fix(tests): Match logic of test to test intention.

zsistla:
Tests were incorrectly throwing an uncaught exception when intention of test was to catch the exception.

### DIFF
--- a/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php
+++ b/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php
@@ -5,8 +5,8 @@
  */
 
 /*DESCRIPTION
-Test that span events are correctly created from any eligible segment, even
-when an exception is handled by the exception handler.
+Test that span events are correctly created from any eligible segment when a caught exception occurs.
+Putting in a try/catch block means an exception is NOT handled by the exception handler.
 */
 
 /*SKIPIF
@@ -32,7 +32,7 @@ null
   "?? agent run id",
   {
     "reservoir_size": 10000,
-    "events_seen": 4
+    "events_seen": 3
   },
   [
     [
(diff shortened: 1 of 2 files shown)


mfulb (approved):
Thanks Amber.

bduranleau-nr (approved):

lavarou (changes requested):
I see there's a discrepancy between the file name and the `DESCRIPTION` in the file: filename says `caught exception` vs `DESCRIPTION` says `handled exception`. The `DESCRIPTION` in the filename matched the code in the test before this PR changed it. I.e. it tested that span events are correctly created from any eligible segment, even ***when an exception is handled by the exception handler***. This PR changes the original intent of this test by adding `try/catch` block to catch the exception. I'm not convinced the code in this test should change, especially that it's passing. I think we need a little more time to understand why it stopped passing when Observer API is used to hook into Zend engine. Maybe we need a different version of this test with different expectation, but IMHO we need the code in this test to run in an unchanged form, i.e. without the try/catch block. There already exist tests with `try/catch` block: [tests/integration/span_events/test_span_events_exception_caught_same_span.php](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/integration/span_events/test_span_events_exception_caught_same_span.php).

mfulb:
I agree with @lavarou that after some more reflection there seems to be more that needs to be understood about these tests and the interaction with the OAPI changes which are causing integration test failures which were not previously seen.

zsistla:
To recap discussion,
the original _caught_exception tests were actually testing an uncaught exception being handled by an exception handler.

Closing this PR.
Another PR will handle:

renaming the uncaught_exception test to uncaught_unhandled_exception
renaming the caught_exception to uncaught_handled_exception
creating new caught_exception to test the caught exception case.

PR #653, opened 2023-03-29
ci: add test-agent workflow to test pull requests

lavarou:
test-agent workflow replaces prototype agent-build workflow. test-agent workflow is capable of:
- building daemon and running daemon unit tests
- building agent (and axiom) for all supported platforms
- building agent's (and axiom's) unit tests for all supported platforms
- running agent's (and axiom's) unit tests for all supported platforms using valgrind as much as possible


.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+#
+
+name: test-pull-request
+
+#
+# Controls when the action will run. 
+#
+on:
+  #
+  # Triggers the workflow on push or pull request events but only for the main branch
```
bduranleau-nr, line 12: `dev` branch is also a trigger

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+      - main
+      - 'dev'
+  pull_request:
+    branches: 
+      - main
+      - 'dev'
+
+jobs:
+  daemon-unit-tests:
+    runs-on: ubuntu-latest
```
bduranleau-nr, line 25: I'm torn on whether it would technically be safer to specify the version (like `ubuntu-22.04`) rather than `latest` just in case there's an issue with compatibility in a future release

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+        arch: [amd64, arm64]
+    steps:
+      - name: Checkout newrelic-php-agent code
+        uses: actions/checkout@v3
+        with:
+          path: php-agent
+      - name: Enable arm64 emulation
+        if: ${{ matrix.arch == 'arm64' }}
+        run: |
+          docker run --privileged --rm tonistiigi/binfmt --install arm64
```
bduranleau-nr, line 42: since this is only one line, you should be able to omit the literal block scalar
```suggestion
        run: docker run --privileged --rm tonistiigi/binfmt --install arm64
```

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+          fi
+          if [[ ${{matrix.arch}} = 'arm64' ]]; then
+            echo "AGENT_CHECK_VARIANT=check" >> $GITHUB_OUTPUT
+          elif [[ ${{ matrix.platform }} = 'gnu' ]]; then
+            echo "AGENT_CHECK_VARIANT=valgrind" >> $GITHUB_OUTPUT
+          elif [[ ${{matrix.php}} = '5.5' || ${{matrix.php}} = '5.6' || ${{matrix.php}} = '7.0' || ${{matrix.php}} = '7.1' ]]; then
+            echo "AGENT_CHECK_VARIANT=check" >> $GITHUB_OUTPUT
+          else 
+            echo "AGENT_CHECK_VARIANT=valgrind" >> $GITHUB_OUTPUT
+          fi
```
bduranleau-nr, line 113: In order to cut down on the amount of code within the yml file itself, would you consider turning this into a script and invoking the script instead? 
There's an example of this kind of feature [here](https://docs.github.com/en/actions/learn-github-actions/essential-features-of-github-actions#adding-scripts-to-your-workflow)
Basically create a `./.github/scripts` directory for more complex actions to live.

.github/workflows/test-agent.yml
bduranleau-nr: Actually, looking at this again, I think I prefer the "cleaner" look of the command being on it's own line

.github/workflows/test-agent.yml
mfulb: This begs the question if we need the ability to specify the target for these tests so we can try them on different OSes.

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Build axiom
+        run: >
+          docker run --rm --platform linux/${{matrix.arch}}
+          -v "${GITHUB_WORKSPACE}/php-agent":"/usr/local/src/newrelic-php-agent" 
+          $IMAGE_NAME:$IMAGE_TAG-${{matrix.php}}-${{matrix.platform}}-$IMAGE_VERSION axiom
```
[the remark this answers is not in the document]
lavarou: It seems like a good suggestion. Let me think about it.

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_NAME: newrelic/nr-php-agent-builder
+      IMAGE_TAG: make-php
+      IMAGE_VERSION: v1
+    strategy:
+      matrix:
+        platform: [gnu, musl]
+        arch: [amd64, arm64]
+        php: ['5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
```
ZNeumann, line 68: I'm worried about all of the different places we have these php version lists (not just in this workflow/PR). When removing/adding a php version, it seems very easy for us to miss a spot. Is there any way to abstract this?

ZNeumann (approved):

.github/workflows/test-agent.yml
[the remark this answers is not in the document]
lavarou: Option 1. Using the script it would look something like this:
```
ARCH=${{matrix.arch}} LIBC=${{matrix.platform}} PHP=${{matrix.php}} make.sh axiom
```
`make.sh` would have to check if all input params are provided and then hide away `docker run ...` command.


Option 2. Using reusable workflow it would look something like this:
```
  uses: ./.github/workflows/make-for-php.yml
    with:
      php: ${{matrix.php}}
      arch: ${{matrix.arch}}
      platform: ${{matrix.platform}}
      make-target: axiom
```
`make-for-php.yml` would hide away `docker run ...` command.

Does it look any better? I think Option 2. is more in line with GitHub Action's design.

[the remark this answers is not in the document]
lavarou: On the other hand I think this refactor could be done in its own PR. The workflow in its current shape achieves what is intended - it builds and unit tests daemon and agent on all supported platforms. Unless there's something functionally wrong with it, I'd like it to be merged and used, and then let's iterate over it.

.github/workflows/test-agent.yml
lavarou: How about this:
```suggestion
  # Triggers the workflow on push or pull request events but only for listed branches
```

mfulb (approved):

.github/workflows/test-agent.yml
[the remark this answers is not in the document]
zsistla: Agree with seeing the change in it's own PR.
Also really like the option 2 reusable workflow to reuse the matrix.* params and to keep the docker call in GHA.

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+          - arch: arm64
+            php: '7.4'
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+        with:
+          path: php-agent
+      - name: Enable arm64 emulation
+        run: |
+          docker run --privileged --rm tonistiigi/binfmt --install arm64
```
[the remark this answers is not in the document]
lavarou: Thanks for this good suggestion. This is best to be done by using [docker-setup-qemu](https://github.com/marketplace/actions/docker-setup-qemu) action, and providing a pinned version of `tonistiigi/binfmt` image as `image` input. It will be done in its own PR.

zsistla (approved):

.github/workflows/test-agent.yml
```diff
@@ -0,0 +1,143 @@
...
+  #
+  push:
+    branches: 
+      - main
+      - 'dev'
+  pull_request:
+    branches: 
+      - main
+      - 'dev'
+
```
zsistla, line 22:
```suggestion
- 'oapi'
```

lavarou:
ok jenkins

PR #661, opened 2023-04-26
chore(install): Disable ZTS

lavarou:
Don't check if zts binaries are in agent's install bundle. 
Abort installation on system with PHP with ZTS.


zsistla (approved):

lavarou:
ok jenkins

lavarou:
ok jenkins

PR #663, opened 2023-05-01
feat: add support for installing on aarch64 systems

lavarou:
First round of changes to support installing an agent on aarch64 systems.


mfulb (approved):
LGTM thanks Michal!

zsistla (approved):
Looks good, thanks.

PR #666, opened 2023-05-08
fix: typo in install-newrelic script

bduranleau-nr:
Expand `pi_php8` variable correctly when checking aarch64 support.


zsistla (approved):
Thanks!

lavarou (approved):
Thanks for catching this!

PR #678, opened 2023-06-13
fix: rework generic callable instrumentation to properly handle transience

ZNeumann:
Adds a transient argument to `nr_php_add_custom_tracer_named`, so that when adding a tracer from name we can decide the transience at creation. Previously, all tracers from name were non-transient. To facilitate this change, an additional `nr_php_wrap_*` function has been created: `nr_php_wrap_user_function_transience`. This wrapping function takes transience as a parameter and properly plumbs it into `nr_php_add_custom_tracer_named`. Due to this change and to reduce code duplication, `nr_php_wrap_user_function` simply calls `nr_php_wrap_user_function_transience` with is_transient=false. This effectively works as a default parameter and keeps the top-level API consistent for all previous instrumentation.

Additional tests have been added to ensure `nr_php_wrap_generic_callable` functionality when 1) wrapping an already wrapped function and 2) wrapping an already "specially" wrapped function.


ZNeumann:
closing for PR based from non-fork repo, for CI/CD https://github.com/newrelic/newrelic-php-agent/pull/679

PR #679, opened 2023-06-13
fix: rework generic callable instrumentation to properly handle transience

ZNeumann:
Adds a transient argument to `nr_php_add_custom_tracer_named`, so that when adding a tracer from name we can decide the transience at creation. Previously, all tracers from name were non-transient. To facilitate this change, an additional `nr_php_wrap_*` function has been created: `nr_php_wrap_user_function_with_transience`. This wrapping function takes transience as a parameter and properly plumbs it into `nr_php_add_custom_tracer_named`. Due to this change and to reduce code duplication, `nr_php_wrap_user_function` simply calls `nr_php_wrap_user_function_with_transience` with `transience` argument value set to `NR_WRAPREC_NOT_TRANSIENT`. This effectively works as a default parameter and keeps the top-level API consistent for all previous instrumentation.

Additional tests have been added to ensure `nr_php_wrap_generic_callable` functionality when 1) wrapping an already wrapped function and 2) wrapping an already "specially" wrapped function.

### DIFF
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -173,14 +173,16 @@ static void reset_wraprec(nruserfn_t* wraprec) {
  * `newrelic.transaction_tracer.custom` 3)
  * nr_php_user_function_add_declared_callback (prior to PHP 7.3) 4) from
  * function `nr_php_wrap_user_function` called from php_wrapper sets the wraprec
- * with framework specific instrumentation. Optionally sets `is_transient`.
+ * with framework specific instrumentation. Optionally specifies transience.
    5) from function `nr_php_wrap_callable` (in `php_wrapper.c`) used only by
- * Wordpress and predis for custom instrumentation that adds `is_transient`.
+ * Wordpress and predis for custom instrumentation that sets
+ * `NR_WRAPREC_IS_TRANSIENT`.
+ *
  * Transient wrappers get disposed of at the end of each request at RSHUTDOWN lifecycle.
  *
  * When overwriting the zend_execute_ex function, every effort was made to
  * reduce performance overhead because until the agent returns control, we are
- * the bottleneck of PHP execution on a customer's machine.  Overwriting the
+ * the bottleneck of PHP execution on a customer's machine. Overwriting the
  * reserved field was seen as a quick way to check if a function is instrumented
  * or not.
  *
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -433,7 +433,7 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
     wraprec->transience = NR_WRAPREC_IS_TRANSIENT;
   } else {
     /* non-transient wraprecs are added to both the hashmap and linked list.
-     * At request shutfown, the hashmap will free transients, but leave
+     * At request shutdown, the hashmap will free transients, but leave
      * non-transients to be freed by the linked list */
     nr_php_add_custom_tracer_common(wraprec);
   }
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -131,7 +131,7 @@ static void nr_php_user_wraprec_destroy(nruserfn_t** wraprec_ptr);
 static void reset_wraprec(nruserfn_t* wraprec) {
   nruserfn_t* p = wraprec;
   nr_php_wraprec_hashmap_key_release(&p->key);
-  if (p->is_transient) {
+  if (p->transience == NR_WRAPREC_IS_TRANSIENT) {
     nr_php_user_wraprec_destroy((nruserfn_t**)&wraprec);
   } else {
     p->is_wrapped = 0;
@@ -381,7 +381,7 @@ nruserfn_t* nr_php_add_custom_tracer_callable(zend_function* func TSRMLS_DC) {
   }
 
   wraprec = nr_php_user_wraprec_create();
-  wraprec->is_transient = 1;
+  wraprec->transience = NR_WRAPREC_IS_TRANSIENT;
 
   nrl_verbosedebug(NRL_INSTRUMENT, "adding custom for callable '%s'", name);
   nr_free(name);
@@ -396,7 +396,7 @@ nruserfn_t* nr_php_add_custom_tracer_callable(zend_function* func TSRMLS_DC) {
 
 nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
                                            size_t namestrlen,
-                                           bool is_transient TSRMLS_DC) {
+                                           nr_transience_t transience TSRMLS_DC) {
   nruserfn_t* wraprec;
   nruserfn_t* p;
 
@@ -429,8 +429,8 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
       (0 == wraprec->classname) ? "" : "::", NRP_PHP(wraprec->funcname));
 
   nr_php_wrap_user_function_internal(wraprec TSRMLS_CC);
-  if (is_transient) {
-    wraprec->is_transient = true;
+  if (transience == NR_WRAPREC_IS_TRANSIENT) {
+    wraprec->transience = NR_WRAPREC_IS_TRANSIENT;
   } else {
     /* non-transient wraprecs are added to both the hashmap and linked list.
      * At request shutfown, the hashmap will free transients, but leave
@@ -482,7 +482,7 @@ void nr_php_remove_transient_user_instrumentation(void) {
   nruserfn_t* prev = NULL;
 
   while (NULL != p) {
-    if (p->is_transient) {
+    if (p->transience == NR_WRAPREC_IS_TRANSIENT) {
       nruserfn_t* trans = p;
 
       if (prev) {
@@ -519,7 +519,7 @@ void nr_php_add_transaction_naming_function(const char* namestr,
                                             int namestrlen TSRMLS_DC) {
   nruserfn_t* wraprec
       = nr_php_add_custom_tracer_named(namestr, namestrlen,
-                                       false /*is_transient*/ TSRMLS_CC);
+                                       NR_WRAPREC_NOT_TRANSIENT TSRMLS_CC);
 
   if (NULL != wraprec) {
     wraprec->is_names_wt_simple = 1;
@@ -529,7 +529,7 @@ void nr_php_add_transaction_naming_function(const char* namestr,
 void nr_php_add_custom_tracer(const char* namestr, int namestrlen TSRMLS_DC) {
   nruserfn_t* wraprec
       = nr_php_add_custom_tracer_named(namestr, namestrlen,
-                                       false /*is_transient*/ TSRMLS_CC);
+                                       NR_WRAPREC_NOT_TRANSIENT TSRMLS_CC);
 
   if (NULL != wraprec) {
     wraprec->create_metric = 1;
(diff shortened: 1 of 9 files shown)


agent/php_user_instrument.c
```diff
@@ -510,7 +518,8 @@ void nr_php_add_user_instrumentation(TSRMLS_D) {
 void nr_php_add_transaction_naming_function(const char* namestr,
                                             int namestrlen TSRMLS_DC) {
   nruserfn_t* wraprec
-      = nr_php_add_custom_tracer_named(namestr, namestrlen TSRMLS_CC);
+      = nr_php_add_custom_tracer_named(namestr, namestrlen,
+                                       false /*is_transient*/ TSRMLS_CC);
```
bduranleau-nr, line 522: Instead of adding `/*is_transient*/` as a comment in the function call, why not create a `bool is_transient = false` and pass the new `is_transient` variable instead?

ZNeumann: Mainly as a standard. Creating a local variable is not very extendable if, say, 5 variables are passed as constants. Additionally, this doesn't clutter the local variable space (less chance of accidentally passing the same variable to a different function with the wrong value), protects this call from a similar affect, and I find it easier to read.

lavarou: If only C had named function arguments...

agent/php_user_instrument.c
```diff
@@ -395,7 +395,8 @@ nruserfn_t* nr_php_add_custom_tracer_callable(zend_function* func TSRMLS_DC) {
 }
 
 nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
-                                           size_t namestrlen TSRMLS_DC) {
+                                           size_t namestrlen,
+                                           bool is_transient TSRMLS_DC) {
```
lavarou, line 399: To minimize code changes and keep backwards compatibility this code should be:
```
nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
                                           size_t namestrlen TSRMLS_DC) {

  nr_php_add_custom_tracer_named_with_transience(namestr, namestrlen, false, TSRMLS_CC);
}

nruserfn_t* nr_php_add_custom_tracer_named_with_transience(const char* namestr,
                                           size_t namestrlen,
                                           bool is_transient TSRMLS_DC) {
...
```
As in:
1. Leave behavior of `nr_php_add_custom_tracer_named` unchanged, which is never create transient wraprec
2. Add a new function `nr_php_add_custom_tracer_named_with_transience` that will add new behavior, i.e. it can create a wraprec that can be transient or not depending on the input (the value of new argument).
This way no place in the code will require change except for the place that requires transient wraprec for `custom_tracer_named` to be added.

agent/php_user_instrument.c
lavarou: I think [this suggestion](https://github.com/newrelic/newrelic-php-agent/pull/679/files#r1228571233) will make this change unnecessary.

lavarou (changes requested):
I suggest to use similar approach for `nr_php_add_custom_tracer_named` as for `nr_php_wrap_user_function` - see [this suggestion](https://github.com/newrelic/newrelic-php-agent/pull/679/files#r1228571233) or [this suggestion](https://github.com/newrelic/newrelic-php-agent/pull/679/files#r1228573781).

agent/php_user_instrument.c
ZNeumann: I had considered this approach but ultimately decided against it. This complicates the API by adding yet another option to call. Because the `add_custom_tracer` type functions are quite limited in where they used, I do not think it is justified to add confusion with this interface (unlike with the `wrap_user_function` family which is invoked in MANY locations, and 99% of these calls are with is_transient=false). Ultimately when reasonable, I would always rather have a function with more parameters than multiple functions used to emulate default parameters (thanks C for not having those).

The side-effects here is that in the future if we ever add a place calling nr_php_add_custom_tracer_named, we will be forced to consider whether each call should be transient- making us less likely to make the same mistake that caused this PR.

lavarou: [This approach](https://github.com/newrelic/newrelic-php-agent/pull/679#discussion_r1228571233) does not complicate things - it simplifies things and removes the `/* is_transient */` comment from the code which eventually will become confusing and dead.

mfulb: Something like this so when looking at code it is obvious what type is being created would be nice.  Does this get rid of may of the /* is_transient */ in the code?

ZNeumann: But what if in the future we want to add another option (similar to this one) to this interface? Even just 1 more option here would immediately balloon into 4 more function headers for the same underlying call, quickly becoming unwieldy and unreadable.

lavarou: > Something like this so when looking at code it is obvious what type is being created would be nice. Does this get rid of may of the /* is_transient */ in the code?

Yes. Both suggestions eliminate `/* is_transient */` comments in the code.

ZNeumann: > Having two APIs with `_transient` and `_persistent` suffixes makes it clear what the API will do and what caller intention was which makes it precisely the case in point - the author calling this API knows what they're doing.

If you are reading the API/header, yes. If one is just reading the code and see the call somewhere, then to get this info one would need to code dive.

agent/php_user_instrument.c
ZNeumann: refactored in [ea328b5](https://github.com/newrelic/newrelic-php-agent/pull/679/commits/ea328b5b408d4efa889f4e6bc9d2733cca2301e2)

agent/php_user_instrument.c
ZNeumann: refactored in [ea328b5](https://github.com/newrelic/newrelic-php-agent/pull/679/commits/ea328b5b408d4efa889f4e6bc9d2733cca2301e2)

agent/tests/test_user_instrument.c
```diff
@@ -97,7 +97,7 @@ static void test_hashmap_wraprec() {
 
   /* instrument user function */
   user_func1_wraprec = nr_php_add_custom_tracer_named(
-      user_func1_name, nr_strlen(user_func1_name));
+      user_func1_name, nr_strlen(user_func1_name), false /*is_transient*/ );
```
zsistla, line 100: Does this need updating with the enum?

ZNeumann: good catch- still passed the test because false was cast to the NOT_TRANSIENT enum. fixed in [59e20f4](https://github.com/newrelic/newrelic-php-agent/pull/679/commits/59e20f46755defa8283848df0ecc1dd36804df09)

mfulb (approved):

bduranleau-nr (approved):

agent/php_user_instrument.c
```diff
@@ -428,7 +429,14 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
       (0 == wraprec->classname) ? "" : "::", NRP_PHP(wraprec->funcname));
 
   nr_php_wrap_user_function_internal(wraprec TSRMLS_CC);
-  nr_php_add_custom_tracer_common(wraprec);
+  if (transience == NR_WRAPREC_IS_TRANSIENT) {
+    wraprec->transience = NR_WRAPREC_IS_TRANSIENT;
+  } else {
+    /* non-transient wraprecs are added to both the hashmap and linked list.
+     * At request shutfown, the hashmap will free transients, but leave
```
lavarou, line 436:
```suggestion
     * At request shutdown, the hashmap will free transients, but leave
```

agent/php_user_instrument.c
```diff
@@ -428,7 +429,14 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
...
+     * non-transients to be freed by the linked list */
```
lavarou, line 437:
```suggestion
     * non-transients to be freed when the linked list is disposed of which is at module shutdown */
```

agent/php_user_instrument.h
```diff
@@ -71,14 +75,21 @@ typedef struct _nruserfn_t {
...
-  int is_user_added;      /* True if user requested this instrumentation */
-  int create_metric; /* 1 if a metric should be made for calls of this function
-                      */
+  int is_exception_handler;   /* 1 if the current exception handler; 0 otherwise
+                               */
+  int is_names_wt_simple;     /* True if this function "names" its enclosing WT;
+                                 the first such function does the naming */
+  nr_transience_t transience; /* Wraprecs that are transient are destroyed
+                                 after each request. Wraprecs that are
+                                 non-transient are kept indefinitely.
```
lavarou, line 84:
```suggestion
                                 non-transient are kept until module shutdown.
```

agent/php_user_instrument.h
```diff
@@ -27,6 +27,10 @@ typedef struct nrspecialfn_return_t (*nrspecialfn_t)(
 
 typedef void (*nruserfn_declared_t)(TSRMLS_D);
 
+typedef enum {
+  NR_WRAPREC_NOT_TRANSIENT = 0,
+  NR_WRAPREC_IS_TRANSIENT = 1
```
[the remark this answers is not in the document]
ZNeumann: I could be convinced to change to the second option listed here if this is something you feel strongly about. The naming convention I used is based on what I would call a local variable carrying this information without the prefix. For example, if I needed a constant local `true` boolean, I would call it `is_transient`. Likewise, if I needed a constant local `false` boolean, I would call it `not_transient`. I was not aiming to standardize the english readability.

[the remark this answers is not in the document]
lavarou: A local variable `is_transient` can mean both things depending on the value it stores. Also in the context of a local variable of boolean type, `!is_transient` reads 'not is transient', not 'not transient' hence my first suggestion. So if consistency is not the deciding factor I would even prefer to use `NR_WRAPREC_TRANSIENT` and `NR_WRAPREC_PERSISTENT`. However, I do not feel strongly about this. I was just curious why this inconsistency.

agent/php_user_instrument.c
```diff
@@ -173,10 +173,10 @@ static void reset_wraprec(nruserfn_t* wraprec) {
  * `newrelic.transaction_tracer.custom` 3)
  * nr_php_user_function_add_declared_callback (prior to PHP 7.3) 4) from
  * function `nr_php_wrap_user_function` called from php_wrapper sets the wraprec
- * with framework specific instrumentation. 5) from function
- * `nr_php_wrap_callable` (in `php_wrapper.c`) used only by Wordpress and predis
- * for custom instrumentation that adds `is_transient` wrappers that get cleaned
- * up with each shutdown.
+ * with framework specific instrumentation. Optionally sets `is_transient`.
```
lavarou, line 176: 1. There is no `is_transient` field in `nruserfn_t` data type any more.
2. Is setting `transience` field optional? The new API requires explicit transience value to be used at the time the wraprec is created.

agent/php_user_instrument.c
```diff
@@ -173,10 +173,10 @@ static void reset_wraprec(nruserfn_t* wraprec) {
...
+   5) from function `nr_php_wrap_callable` (in `php_wrapper.c`) used only by
+ * Wordpress and predis for custom instrumentation that adds `is_transient`.
```
lavarou, line 178: There is no is_transient field in nruserfn_t data type any more.

ZNeumann: reworded in [a52a440](https://github.com/newrelic/newrelic-php-agent/pull/679/commits/a52a440e0fc1aae7e8c3a2a08d38037201596ede)

agent/php_user_instrument.c
ZNeumann: 1. reworded in [a52a440](https://github.com/newrelic/newrelic-php-agent/pull/679/commits/a52a440e0fc1aae7e8c3a2a08d38037201596ede)
2. it is optional. There is the regular call to nr_php_wrap_user_function (which sets not_transient) and then the call nr_php_wrap_user_function_with_transience which takes transience as a parameter.

agent/php_user_instrument.c
lavarou: Thank you for compromising.

lavarou (approved):

PR #680, opened 2023-06-13
ci: add on-demand run-integration-tests workflow

lavarou:
An on-demand workflow that will run integration tests for PHPs 8.0, 8.1 and 8.2 on gnu and musl platforms on x86_64 and aarch64 architectures. This workflow runs integration tests only because unit tests as well as custom events limits tests and LASP tests are already part of the test-pull-request workflow.

It is expected that running tests on aarch64 architecture will cause test failures because tests run slower on emulated platform and more spans are generated. aarch64 environment needs some adjustments - maybe experiment with `expensive_node_min`.

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -100,11 +100,17 @@ jobs:
   test:
     needs: [build-test-runner, build-agent]
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
+      fail-fast: true
       matrix:
         platform: [gnu, musl]
         arch: [amd64, arm64]
         php: ['8.0', '8.1', '8.2']
+        experimental: [false]
+        include:
+          - arch: arm64
+            experimental: true
     steps:
       - name: Checkout integration tests
         uses: actions/checkout@v3
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -1,0 +1,167 @@
+# Copyright 2020 New Relic Corporation. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+name: run-integration-tests
+run-name: Testing ${{ inputs.origin }}:${{ inputs.ref }} (by @${{ github.actor }})
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-integration-tests.yml
```diff
@@ -0,0 +1,173 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build integration_runner
+        run: >
+          docker run --rm --platform linux/${{matrix.arch}}
+          -v "${GITHUB_WORKSPACE}/php-agent":"/usr/local/src/newrelic-php-agent" 
+          -e ACCOUNT_supportability=${{secrets.ACCOUNT_SUPPORTABILITY}}
+          -e APP_supportability=${{secrets.APP_SUPPORTABILITY}}
+          $IMAGE_NAME:$IMAGE_TAG-${{ matrix.platform }}-$IMAGE_VERSION bin/integration_runner
```
mfulb, line 55: Below 'make' is part of the docker run command building axiom and the agent, and here it isn't - do these all either need 'make' or not?

zsistla (approved):
This'll be so useful, thanks!
The `experimental` toggle is a great addition.

ZNeumann (approved):

.github/workflows/run-integration-tests.yml
bduranleau-nr: Also looking for clarification on this

lavarou: `make-go` variant of `newrelic/nr-php-agent-builder` and `make-php` variant of the same image have different entry points. The `make-go` entry point is make. The `make-php` entry point is `docker-php-entrypoint`:
```
$ docker inspect -f '{{.Config.Entrypoint}}' newrelic/nr-php-agent-builder:make-go-gnu-v1
[make -r]
$ docker inspect -f '{{.Config.Entrypoint}}' newrelic/nr-php-agent-builder:make-php-8.2-gnu-v1
[docker-php-entrypoint]
```

Yes, this is a bit unfortunate.

bduranleau-nr (approved):

PR #681, opened 2023-06-15
Release 10.11 Amendment

ZNeumann:


lavarou (approved):
👍

zsistla:
Do we need to call it a patch?  Technically 10.11 didn't go out yet.

lavarou:
> Do we need to call it a patch? Technically 10.11 didn't go out yet.

Maybe the word 'patch' could be improved upon. However PR titles are by default used as merge commit messages and therefore it makes sense to differentiate this pull request from #677, which is already called 'Release 10.11'. This PR is an amendment to #677 so maybe 'amendment' instead of 'patch'? @ZNeumann @zsistla

ZNeumann:
ok jenkins

zsistla (approved):

PR #685, opened 2023-06-20
fix(agent): Update drupal msg logging level

zsistla:
Since this is called with every request, the message, when triggered, can fill up the logs. It's more appropriate for it to be a `verbosedebug` level similar to the majority of other drupal messages.  Generally, msgs that have a probability of logging with every request should not be info level.


ZNeumann (approved):

lavarou (approved):

mfulb (approved):

zsistla:
ok jenkins

PR #692, opened 2023-07-12
chore: make dev-shell compatible with Debian 12

bduranleau-nr:
Fix the Dockerfile used by `make dev-shell` in development.
Upstream PHP docker images have migrated to Debian 12 bookworm, which modifies the sources location as well as dependency names.

This fix only resolves running PHP versions 8.0+.

### DIFF
--- a/files/Dockerfile
+++ b/files/Dockerfile
@@ -13,18 +13,6 @@ FROM php:${PHP_VER:-8.2}
 
 RUN docker-php-source extract
 
-#
-# Uncomment deb-src lines for all enabled repos. First part of single-quoted
-# string (up the the !) is the pattern of the lines that will be ignored.
-# Needed for apt-get build-dep call later in script
-#
-
-# PHP 8.0
-#RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list
-
-# PHP 8.1+
-RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list.d/debian.sources
-
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update
 RUN apt-get install -y build-essential
--- a/files/Dockerfile
+++ b/files/Dockerfile
@@ -18,7 +18,12 @@ RUN docker-php-source extract
 # string (up the the !) is the pattern of the lines that will be ignored.
 # Needed for apt-get build-dep call later in script
 #
-RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list
+
+# PHP 8.0
+#RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list
+
+# PHP 8.1+
+RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list.d/debian.sources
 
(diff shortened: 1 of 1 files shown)


files/Dockerfile
```diff
@@ -18,7 +18,12 @@ RUN docker-php-source extract
 # string (up the the !) is the pattern of the lines that will be ignored.
 # Needed for apt-get build-dep call later in script
 #
-RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list
+
+# PHP 8.0
```
zsistla, line 22: Can we automatically toggle between the two depending on the value of PHP_VER?

bduranleau-nr: We could, but having investigated that before I decided it would make too much of a mess here and opted for the simpler solution

ZNeumann (approved):

hahuja2 (approved):

files/Dockerfile
```diff
@@ -18,7 +18,12 @@ RUN docker-php-source extract
...
+#RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list
+
+# PHP 8.1+
+RUN sed -Ei '/.*partner/! s/^# (deb-src .*)/\1/g' /etc/apt/sources.list.d/debian.sources
```
lavarou, line 26: Is this line even needed? The comment on lines 17-19 says it is needed to enable some? repo so that `apt-get build-dep` called later in the script would work. However I don't see `apt-get build-dep` called anywhere. Hence my question if this sed magic with debian sources file is needed at all.

files/Dockerfile
```diff
@@ -28,7 +33,7 @@ RUN apt-get install -y build-essential
 # PHP dependencies
 #
 RUN apt-get update \
- && apt-get -y install gcc git netcat \
+ && apt-get -y install gcc git netcat-openbsd \
```
lavarou, line 36: What is `netcat` needed for? It is not needed to build the agent. Maybe this `netcat/netcat-openbsd` dependency could be removed altogether.

files/Dockerfile
bduranleau-nr: The effort to determine that as well as the question surrounding netcat is outside of the scope of this PR and requires it's own investigation IMO

lavarou: If it is not needed, why try to fix it? Have you tried removing it and building the agent inside `dev-shell`? That should be enough to determine if it is needed or not.

bduranleau-nr: Tested this and confirmed that it does not appear to impact agent build or tests.
Removed in 9b172b52c8341e1cefa938d8378369d252850835

lavarou: @bduranleau-nr Thank you for checking this and removing dead code from the codebase!

files/Dockerfile
bduranleau-nr: I don't believe `netcat` is used to build the agent, but it's difficult to rule out that it isn't used in execution in some way or another (searching for `nc` against the repo returns a couple thousand results). It could have been used historically for testing agent functionality and may not be necessary anymore, but I don't think it hurts anything to keep it included for the time being.

lavarou: Agree. I did check that `netcat` package in Debian 11 is a virtual package that redirects to `netcat-openbsd` so this is not a breaking change and I'm ok with leaving it as is.

lavarou (approved):

mfulb (approved):

PR #696, opened 2023-07-21
tests(agent): add Drupal integration tests

lavarou:


bduranleau-nr:
Are these test scripts unchanged from the internal repo, with the only major change being the addition of the `drupal.php` file to download drupal source?

tests/integration/external/drupal6/skipif.inc
```diff
@@ -0,0 +1,15 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+if (version_compare(PHP_VERSION, "7.4", ">")) {
+  die("skip: PHP > 7.4.0 not supported\n");
```
mfulb, line 8: Is only 7.4.0 supported and not 7.4.x where x != 0?

bduranleau-nr (approved):

tests/integration/external/drupal6/skipif.inc
lavarou: This skips PHP versions 7.4.x and above. 7.3.12 will pass but 7.4.12 will not.

PR #700, opened 2023-07-25
fix(agent): set oapi callbacks the right way

ZNeumann:
When wrapping a user function do not add special instrumentation is ANY special instrumentation (before/after/cleanup) has already been set. This approach matches pre-oapi instrumentation.

### DIFF
--- a/agent/php_wrapper.c
+++ b/agent/php_wrapper.c
@@ -25,8 +25,8 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
 
   /* If any of the callbacks we are attempting to set are already set to
    * something else, we want to exit without setting new callbacks */
-  if (is_instrumentation_set(wraprec->special_instrumentation,
-                             after_callback)) {
+  if (is_instrumentation_set_and_not_equal(wraprec->special_instrumentation,
+                                           after_callback)) {
     nrl_verbosedebug(
         NRL_INSTRUMENT,
         "%s: attempting to set special_instrumentation for %.*s, but "
@@ -35,8 +35,8 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
     return wraprec;
   }
 
-  if (is_instrumentation_set(wraprec->special_instrumentation_before,
-                             before_callback)) {
+  if (is_instrumentation_set_and_not_equal(wraprec->special_instrumentation_before,
+                                           before_callback)) {
     nrl_verbosedebug(NRL_INSTRUMENT,
                      "%s: attempting to set special_instrumentation_before "
                      "for %.*s, but "
@@ -45,8 +45,8 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
     return wraprec;
   }
 
-  if (is_instrumentation_set(wraprec->special_instrumentation_clean,
-                             clean_callback)) {
+  if (is_instrumentation_set_and_not_equal(wraprec->special_instrumentation_clean,
+                                           clean_callback)) {
     nrl_verbosedebug(NRL_INSTRUMENT,
                      "%s: attempting to set special_instrumentation_clean "
                      "for %.*s, but "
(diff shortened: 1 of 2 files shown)


agent/php_wrapper.c
```diff
@@ -23,6 +23,8 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
     return wraprec;
   }
 
+  /* If any of the callbacks we are attempting to set are already set to
```
zsistla, line 26: The logic doesn't seem quite right yet.  It looks like if `after_callback` is null and the `wraprec->special_instrumentation after_callback` is already set but wraprec->before/clean are null, we don't get the return wraprec line so at the end, before/clean get overwritten when they shouldn't have because they might not match the after.

Plus when we add the 3 additional if checks again to see if the passed in callbacks are null it adds additional time to wraprec handling.

What if is_instrumentation_set just checked to see if any before/after/clean was set otherwise returns false.  Then we just do one is_instrumentation_set at beginning and return wraprec if true; otherwise, we can skip the individual checks and just assign all of them.
We could probably even streamline and save more time by skipping the null check and just assign the wraprec->callbacks to the passed in callbacks because if the wraprec callbacks aren't set they are NULL and if the passed in callbacks are NULL we'd just be reassigning them to NULL anyway.

ZNeumann: This was created with in mind the edge case of setting previously null callbacks on an already existing wrapper. If this is something that we won't want, or want a separate API for, then I agree with the proposal to streamline the checks. I am leaning towards not having this functionality, and then adding a separate API if ever needed.

zsistla: Ah, I thought the reasoning was that if any callback is already set and we subsequently try to add another callback, it should fail because they might not match.

ZNeumann: I think on retrospect I tend towards not allowing this edge case and streamlining the checks.

ZNeumann: Changed to not allow any instrumentation if any is set in [6bb143f](https://github.com/newrelic/newrelic-php-agent/pull/700/commits/6bb143f072580b241ef436ea582da0deaa6c7458)

lavarou: The name `is_instrumentation_set` is misleading. This function does two things, not one as the name suggests. This function:
1. checks if the instrumentation is set (not `NULL`)
2. additionally(!), if instrumentation is already set, checks if the old callback matches the new callback 

It took me some time to dig it out and understand what `is_instrumentation_set` is meant to do. I liked [the code](https://github.com/zsistla/newrelic-php-agent/blob/cbe76a49d459abe28b5b7f22c13485a62ea7a4c7/agent/php_wrapper.c) before [this refactor](https://github.com/zsistla/newrelic-php-agent/commit/136249c70c332aeac014bcccf65b3a003e5ed973) better - it was easier to read and understand.

lavarou (approved):

agent/php_wrapper.c
ZNeumann: I'm under the impression that it should be rare that instrumentation is already set. As such, I would rather have the separated logging structure than a slightly faster exit in that rare case.

I agree that is_instrumentation_set is a misleading name, but I believe that we wish to keep its current functionality. Is there a different name that you would prefer?

lavarou: Maybe `is_instrumentation_set_and_equal_to`? But to be frank I really liked [the code](https://github.com/zsistla/newrelic-php-agent/blob/cbe76a49d459abe28b5b7f22c13485a62ea7a4c7/agent/php_wrapper.c) before [this refactor](https://github.com/zsistla/newrelic-php-agent/commit/136249c70c332aeac014bcccf65b3a003e5ed973) better. I'm fine with leaving things as they are in this PR, i.e. just leave `is_instrumentation_set` as is. But if you are willing to re-factor things, I'd rather restore the code as it was before the refactor and have explicit comparisons vs hidden behind `is_instrumentation_set` or any other function. It is almost the same amount of text but explicit checks make the intention of the code very clear.

ZNeumann: renamed to `is_instrumentation_set_and_not_equal` in [fa05f1b](https://github.com/newrelic/newrelic-php-agent/pull/700/commits/fa05f1bdbff86d21f725960188ec2dc540e518b0)

zsistla (approved):

PR #701, opened 2023-07-25
feat(agent): instrument Drupal 9.4 hooks with oapi

ZNeumann:
In addition to OAPI Drupal 9.4 support, this PR does some refactoring:

1. The hook stacks have been renamed from `module_invoke_all` to
   `invoke_all` to be more aligned with the name of Drupal functions that
can now push to those stacks.

2. The generic wrapper function (originally added when Drupal 9.4
   instrumentation was introduced) has been updated to support OAPI.

3. `nr_php_wrap_callable_before_after_clean` has been introduced as a
   parallel for `nr_php_wrap_user_function_before_after_clean`. To
facilitate this, many of the internals have been extracted to a common
function.

4. Speaking of, lots of the maintenance of the Drupal's hook stacks has
   been extracted to common functions to reduce code reuse

### DIFF
--- a/agent/fw_drupal.c
+++ b/agent/fw_drupal.c
@@ -753,13 +753,13 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_before) {
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL);
 
   hook_copy = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
-  invoke_all_push_hook_stacks(hook_copy);
+  nr_drupal_invoke_all_hook_stacks_push(hook_copy);
 }
 NR_PHP_WRAPPER_END
 
 NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_after) {
   (void)wraprec;
-  invoke_all_clean_hook_stacks();
+  nr_drupal_invoke_all_hook_stacks_pop();
 }
 NR_PHP_WRAPPER_END
 
@@ -767,7 +767,7 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_clean) {
   NR_UNUSED_SPECIALFN;
   NR_UNUSED_FUNC_RETURN_VALUE;
   (void)wraprec;
-  invoke_all_clean_hook_stacks();
+  nr_drupal_invoke_all_hook_stacks_pop();
 }
 NR_PHP_WRAPPER_END
 
--- a/agent/fw_drupal.c
+++ b/agent/fw_drupal.c
@@ -470,7 +470,7 @@ static void nr_drupal_wrap_hook_within_module_invoke_all(
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
   zval* curr_hook
-      = (zval*)nr_stack_get_top(&NRPRG(drupal_module_invoke_all_hooks));
+      = (zval*)nr_stack_get_top(&NRPRG(drupal_invoke_all_hooks));
   if (!nr_php_is_zval_non_empty_string(curr_hook)) {
     nrl_verbosedebug(NRL_FRAMEWORK,
                      "%s: cannot extract hook name from global stack",
@@ -480,8 +480,8 @@ static void nr_drupal_wrap_hook_within_module_invoke_all(
   char* hook_name = Z_STRVAL_P(curr_hook);
   size_t hook_len = Z_STRLEN_P(curr_hook);
 #else
-  char* hook_name = NRPRG(drupal_module_invoke_all_hook);
-  size_t hook_len = NRPRG(drupal_module_invoke_all_hook_len);
+  char* hook_name = NRPRG(drupal_invoke_all_hook);
+  size_t hook_len = NRPRG(drupal_invoke_all_hook_len);
 #endif
   if (NULL == hook_name) {
     nrl_verbosedebug(NRL_FRAMEWORK,
@@ -753,25 +753,13 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_before) {
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL);
 
   hook_copy = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
-  if (nr_php_is_zval_non_empty_string(hook_copy)) {
-    nr_stack_push(&NRPRG(drupal_module_invoke_all_hooks), hook_copy);
-    nr_stack_push(&NRPRG(drupal_module_invoke_all_states), (void*)!NULL);
-  } else {
-    nr_stack_push(&NRPRG(drupal_module_invoke_all_states), NULL);
-  }
+  invoke_all_push_hook_stacks(hook_copy);
 }
 NR_PHP_WRAPPER_END
 
-static void module_invoke_all_clean_stacks() {
-  if ((bool)nr_stack_pop(&NRPRG(drupal_module_invoke_all_states))) {
-    zval* hook_copy = nr_stack_pop(&NRPRG(drupal_module_invoke_all_hooks));
-    nr_php_arg_release(&hook_copy);
-  }
-}
-
 NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_after) {
   (void)wraprec;
-  module_invoke_all_clean_stacks();
+  invoke_all_clean_hook_stacks();
 }
 NR_PHP_WRAPPER_END
 
@@ -779,7 +767,7 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_clean) {
   NR_UNUSED_SPECIALFN;
   NR_UNUSED_FUNC_RETURN_VALUE;
   (void)wraprec;
-  module_invoke_all_clean_stacks();
+  invoke_all_clean_hook_stacks();
 }
 NR_PHP_WRAPPER_END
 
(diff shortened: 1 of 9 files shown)


agent/fw_drupal8.c
```diff
@@ -71,6 +71,48 @@ static void nr_drupal8_add_method_callback(const zend_class_entry* ce,
   }
 }
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+static void nr_drupal8_add_method_callback_before_after_clean(
```
ZNeumann, line 76: This is essentially a copy of https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_drupal8.c#L32 that instead calls the underlying before_after_clean wrapping function

agent/php_wrapper.c
```diff
@@ -52,12 +48,53 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
...
+  /*
+   * For logging purposes, let's create a name if we're logging at verbosedebug.
+   */
+  if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+    name = nr_php_function_debug_name(callable);
+  }
+  nr_php_wraprec_add_before_after_clean_callbacks(name, nr_strlen(name), wraprec,
+                                                  before_callback,
+                                                  after_callback,
+                                                  clean_callback);
```
lavarou, line 97: According to [this comment](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_agent.h#L585-L586) the `name` should be freed if it got created:
```suggestion
                                                  clean_callback);
  if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT) && NULL != name) {
    nr_free(name);
  }
```

agent/fw_drupal_common.c
```diff
@@ -294,3 +294,22 @@ void nr_drupal_headers_add(zval* arg, bool is_drupal_7 TSRMLS_DC) {
 
   nr_php_zval_free(&newrelic_headers);
 }
+
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+void invoke_all_push_hook_stacks(zval* hook_copy) {
```
lavarou, line 300: I think the name `nr_drupal_{invoke_all_hooks_}stack{s}_push()` would be better here for the following reasons:
1. The `nr_` prefix follows the convention of scoping the function to public, internal function. This convention is described in point 3. of [our style guide](https://github.com/newrelic/newrelic-php-agent/blob/main/STYLEGUIDE.md)
2. The `drupal_` sub-prefix follows the undocumented, yet silently followed (although not always), convention of scoping the function to the module related to Drupal instrumentation. This convention, although undocumented, helps when searching for symbols related to particular functionality/features of the agent
3. The `{invoke_all_hooks_}stack{s}` part describe the object on which this function operates. The `invoke_all_hooks` part is optional (hence in curly braces), but helpful to more precisely describe the object on which this function operates - drupal invoke_all hooks stack. The `s` (plural form of `stack`) is also optional but also better describe that this function operates on two stacks - the hooks stack and the states stack. To sum up, it'll be fine to have this portion as just `stack` in its simplest form, but other variations are also acceptable, `invoke_all_stacks` being the most verbose variation.
4. The `(noun/object)_verb` convention (where the noun/object in this case is `{invoke_all_}stack{s}` follows the convention for util_stack.h APIs and this API for `invoke_all_stacks` is an extension to/concrete use of generic stack.

Similar remarks apply to `invoke_all_clean_hook_stacks`.

Using this name would make reading the code in fw_drupal.c module easier.

agent/fw_drupal.c
```diff
@@ -753,33 +753,21 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_before) {
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL);
 
   hook_copy = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
-  if (nr_php_is_zval_non_empty_string(hook_copy)) {
-    nr_stack_push(&NRPRG(drupal_module_invoke_all_hooks), hook_copy);
-    nr_stack_push(&NRPRG(drupal_module_invoke_all_states), (void*)!NULL);
-  } else {
-    nr_stack_push(&NRPRG(drupal_module_invoke_all_states), NULL);
-  }
+  invoke_all_push_hook_stacks(hook_copy);
```
lavarou, line 756: Consider renaming this according to [this comment](https://github.com/newrelic/newrelic-php-agent/pull/701/files#r1275366315). Then it would read as:
```suggestion
  nr_drupal_invoke_all_hook_stacks_push(hook_copy);
```

agent/fw_drupal.c
```diff
@@ -753,33 +753,21 @@ NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_before) {
...
-  if ((bool)nr_stack_pop(&NRPRG(drupal_module_invoke_all_states))) {
-    zval* hook_copy = nr_stack_pop(&NRPRG(drupal_module_invoke_all_hooks));
-    nr_php_arg_release(&hook_copy);
-  }
-}
-
 NR_PHP_WRAPPER(nr_drupal_wrap_module_invoke_all_after) {
   (void)wraprec;
-  module_invoke_all_clean_stacks();
+  invoke_all_clean_hook_stacks();
```
lavarou, line 762: Consider renaming this according to [this comment](https://github.com/newrelic/newrelic-php-agent/pull/701/files#r1275366315). Then it would read as:
```suggestion
  nr_drupal_invoke_all_hook_stacks_clean();
```

agent/fw_drupal8.c
```diff
@@ -406,9 +440,24 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_callback) {
     goto leave;
   }
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+  zval* curr_hook
+      = (zval*)nr_stack_get_top(&NRPRG(drupal_invoke_all_hooks));
```
lavarou, line 446: Should the `states` stack be queried first to see if `hook` name can be popped?

agent/fw_drupal8.c
```diff
@@ -508,7 +578,14 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
    * for oapi so the hook vars this is referencing do not exist.
    *
    */
```
lavarou, line 580: Shouldn't this comment be removed? It looks like the temporary exclusion has been addressed.

agent/fw_drupal8.c
```diff
@@ -436,40 +488,58 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with) {
...
   nr_php_arg_release(&hook);
+#endif // not OAPI
+}
+NR_PHP_WRAPPER_END
+
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_after) {
+  (void)wraprec;
+  invoke_all_clean_hook_stacks();
```
lavarou, line 532: Should the 'after' callback cleanup the whole stack or just pop the last pushed item? Pre-oapi seemed to do the later, i.e. pop the last item from the stack.

agent/fw_drupal8.c
```diff
@@ -436,40 +488,58 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with) {
...
+  nr_free(NRPRG(drupal_invoke_all_hook));
+  NRPRG(drupal_invoke_all_hook) = prev_hook;
+  NRPRG(drupal_invoke_all_hook_len) = prev_hook_len;
+#endif // not OAPI
+
+leave: ;
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
+    || defined OVERWRITE_ZEND_EXECUTE_DATA
+  /* for OAPI, the _after callback handles this free */
   nr_php_arg_release(&hook);
```
lavarou, line 523: I think `hook` needs to be released for OAPI too.

lavarou (commented):
I have a few questions and suggestions before I hit 'approve'.

agent/fw_drupal8.c
ZNeumann: It is rather confusing, and part of why I've abstracted management of the stacks into a separate API. The states stack keeps track whether of not the current invoke_all* execution successfully pushed a hook onto the hooks stack. If it failed to add the hook (`nr_php_zval_is_non_empty_string` returned `false`) then the states stack reflects this with a false value. If it succeeded, the state stack reflects it with a true value. In all cases, the states stack will match our depth of invoke_all* calls, while the hooks stack might not.

So no, we do not need to query the states stack to peek at the top of the hooks stack. Because if the top of the states stack were false, that is simply telling us that the top of the hook stack is from an upstream nesting execution of invoke_all*, or its NULL. Either way is fine. This functionality is inline with how it was handled pre-OAPI.

The real purpose of the states stack is to facilitate communication from the _before callback to the _after callback. At that time, we free the top of the hooks stack if and only if the top of the states stack is `true`. This saves us another copy of the hook argument, and another call to `nr_php_zval_is_non_empty_string`.

agent/fw_drupal8.c
ZNeumann: This call does just pop the last elements of the stack, explained more thoroughly here: https://github.com/newrelic/newrelic-php-agent/pull/701#discussion_r1276267265.

"clean" is perhaps the wrong word to use. Maybe "pop" instead?

agent/fw_drupal8.c
ZNeumann: In reality, I am unsure how often nested invoke_all calls really happen

agent/fw_drupal8.c
ZNeumann: It is unconditionally retrieved, and then ownership is transferred to the stack if it is a valid hook. In the case of an invalid hook, ownership is not transferred and it does need to be freed. Fixed in [fcedfbf](https://github.com/newrelic/newrelic-php-agent/pull/701/commits/fcedfbf632d4c692abb9238a94cd54f480c44a04)

agent/fw_drupal8.c
lavarou: Yes, 'pop' as a verb would make things more obvious. If you consider re-naming the 'stacks' API, please consider [this suggestion](https://github.com/newrelic/newrelic-php-agent/pull/701#discussion_r1275366315) for naming the functions.

agent/fw_drupal8.c
lavarou: Thanks for these additional details about ownership of the `hook`!

agent/fw_drupal8.c
ZNeumann: On retrospect, the states stack might be needlessly complicated. As stated, the main purpose is to avoid copying the hook argument again. However, the function [nr_php_get_user_func_arg](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_agent.c#L417) can accomplish this same goal.

ZNeumann: Further investigations: during the "_clean" functions, we do not have access to the function arguments and would be unable to call nr_php_get_user_func_arg. Therefore, the state_stack is absolutely needed to properly convey the necessary cleaning information.

agent/fw_drupal_common.c
ZNeumann: renamed in [2592e6f](https://github.com/newrelic/newrelic-php-agent/pull/701/commits/2592e6fd08814c13540b734d72eb1928d80e1078)

agent/fw_drupal_common.h
```diff
@@ -147,4 +147,23 @@ nr_status_t module_invoke_all_parse_module_and_hook_from_strings(
...
 void nr_drupal_headers_add(zval* arg, bool is_drupal_7 TSRMLS_DC);
 
+
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+/*
+ * Purpose: Before an invoke_all style call, adds the hook to that hook states stacks
+ *
+ * Params  : 1. A zval holding a copy of the hook invoked, to be managed by the hook
+ *              states stacks and freed by invoke_all_clean_hook_stacks() after the
```
lavarou, line 157:
```suggestion
 *              states stacks and freed by nr_drupal_invoke_all_hook_stacks_pop() after the
```

lavarou (approved):
LGTM 👍

zsistla (approved):
Thanks for getting this complicated update in place.

PR #703, opened 2023-07-26
fix(agent): wrap zend_try in a func to avoid clobbering

ZNeumann:
Calls to `call_user_function` (in PHP 8.0 and 8.1) and `zend_call_method_if_exists` (in PHP 8.2+) need to be wrapped by the `zend_try`/`zend_catch`/`zend_end_try` block, which use [`setjmp`](https://linux.die.net/man/3/setjmp) and [`longjmp`](https://linux.die.net/man/3/longjmp), because according to [`call_user_func()`](https://www.php.net/manual/en/function.call-user-func.php):
> Callbacks registered with functions such as `call_user_func()` and [`call_user_func_array()`](https://www.php.net/manual/en/function.call-user-func-array.php) will not be called if there is an uncaught exception thrown in a previous callback.

So if we call something that causes an exception, it will block us from future calls that use `call_user_func` or `call_user_func_array`. Valgrind showed the agent and/or the Zend engine wasn’t properly cleaning up after such cases and newer compilers had issues with this when compiling with any optimization and generated the following error:
```
error: variable ‘retval’ might be clobbered by ‘longjmp’ or ‘vfork’) [-Werror=clobbered]
```
PHP developers solve this problem by using an automatic variable to store user function call result - see [here](https://github.com/php/php-src/blob/master/main/streams/userspace.c#L335-L340) for an example in PHP source code how zend_call_method_if_exists is called.

This solution wraps `zend_try`/`zend_catch`/`zend_end_try` constructs in a function to localize the `retval` and avoid variable clobbering.

### DIFF
--- a/agent/php_call.c
+++ b/agent/php_call.c
@@ -10,6 +10,54 @@
 
 #include "Zend/zend_exceptions.h"
 
+/*
+ * zend_try family of macros entail the use of setjmp and longjmp, which can cause clobbering issues with
+ * non-primitive local variables. Abstracting these constructs into separate functions protects from this.
+ */
+#if ZEND_MODULE_API_NO >= ZEND_8_2_X_API_NO
+static int nr_php_call_try_catch(zend_object* object,
+                                 zend_string* method_name,
+                                 zval* retval,
+                                 zend_uint param_count,
+                                 zval* param_values) {
+  /*
+   * With PHP 8.2, functions that do not exist will cause a fatal error to
+   * be thrown. `zend_call_method_if_exists` will attempt to call a function and
+   * silently fail if it does not exist
+   */
+  int zend_result = FAILURE;
+  zend_try {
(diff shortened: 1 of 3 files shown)


ZNeumann:
duplicate, but alternative approach to https://github.com/newrelic/newrelic-php-agent/pull/638

zsistla:
703 seems a lot cleaner and less convoluted than 638.  Could you please grab some of the description from 703 to describe why it needs updating and what we tried before going with this way?

ZNeumann:
@lavarou , you added a bunch of additional context to the solution provided by https://github.com/newrelic/newrelic-php-agent/pull/638 in the PR description. I am unsure how much of that is relevant to this implementation, but would appreciate the addition of anything you wish to carry over.

lavarou (approved):
LGTM 👍

zsistla (approved):
Looks good, thanks.

PR #708, opened 2023-08-02
fix: add guards to oapi cufa opcode check

ZNeumann:
When instrumenting wordpress/drupal hooks, we want to instrument the hook function calls, which are dispatched by php's call_user_function_array. This php function (cufa for short) is flattened/inlined, so to determine if we are in a function called by a cufa call, we need to check the opcodes of the caller to the current function. If the caller is a php internal function, checking these opcodes was not safe and lead to invalid reads. As such, before reading the opcodes, we first check if the caller is a user function. We only want to instrument cufa calls from user functions anyway.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1969,8 +1969,11 @@ static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
    */
 
   /* 
-   * We cannot safely access the opline of internal functions, and we only
-   * want to instrument cufa calls from user functions anyway
+   * When Observer API is used, this code executes in the context of
+   * zend_execute and not in the context of VM (as was the case pre-OAPI),
+   * therefore we need to ensure we're dealing with a user function. We cannot
+   * safely access the opline of internal functions, and we only want to
+   * instrument cufa calls from user functions anyway.
    */
   if (UNEXPECTED(NULL == execute_data->prev_execute_data->func)) {
     nrl_verbosedebug(NRL_AGENT, "%s: cannot get previous function", __func__);
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1985,7 +1985,7 @@ static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
 
   /* 
    * Extra safety check. Previously, we instrumented by overwritting ZEND_DO_FCALL.
-   * Within OAPI, for consistencies sake, we will ensure consistency here
+   * Within OAPI, for consistency's sake, we will ensure the same
    */
   if (ZEND_DO_FCALL != prev_opline->opcode) {
     nrl_verbosedebug(NRL_AGENT, "%s: cannot get previous function name", __func__);
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1934,6 +1934,7 @@ static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
     nrl_verbosedebug(NRL_AGENT, "%s: cannot get previous execute data", __func__);
     return;
   }
+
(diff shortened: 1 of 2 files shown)


agent/php_execute.c
```diff
@@ -1966,20 +1967,40 @@ static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
...
+
+  /* 
+   * Extra safety check. Previously, we instrumented by overwritting ZEND_DO_FCALL.
+   * Within OAPI, for consistency's sake, we will ensure the same
+   */
+  if (ZEND_DO_FCALL != prev_opline->opcode) {
+    nrl_verbosedebug(NRL_AGENT, "%s: cannot get previous function name", __func__);
+    return;
+  }
+  prev_opline -= 1; // Checks previous opcode
```
lavarou, line 1994: Maybe add a comment explaining why it is safe to check previous opcode. https://github.com/newrelic/newrelic-php-agent/pull/639 has a good explanation:
> ZEND_DO_FCALL will never be the first opcode in an op array -- minimally, there is always at least a ZEND_INIT_FCALL before it -- so it is safe to iterate backwards in the opcode like execute_data->prev_execute_data->opline - 1


```suggestion
  /* ZEND_DO_FCALL will never be the first opcode in an op array -- minimally,
   * there is always at least a ZEND_INIT_FCALL before it -- so it is safe to 
   * check what previous opcode is. */
  prev_opline -= 1; // Checks previous opcode
```

ZNeumann: The large block comment above has this explanation https://github.com/newrelic/newrelic-php-agent/pull/708/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1944

agent/php_execute.c
```diff
@@ -1966,20 +1967,40 @@ static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
    * cause additional performance overhead, this should be considered a last
    * resort.
    */
-  const zend_op* prev_opline = execute_data->prev_execute_data->opline - 1;
+
+  /* 
+   * We cannot safely access the opline of internal functions, and we only
+   * want to instrument cufa calls from user functions anyway
```
lavarou, line 1973:
```suggestion
   * Because this code executes in the context of zend_execute and not in the
   * context of VM, we need to ensure we're dealing with a user function. We
   * cannot safely access the opline of internal functions, and we only want
   * to instrument cufa calls from user functions anyway.
```

lavarou (approved):
LGTM 👍

agent/php_execute.c
```diff
@@ -1966,20 +1967,43 @@ static void nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_PROTO) {
...
-    prev_opline = execute_data->prev_execute_data->opline - 2;
+    prev_opline -= 1; // Checks previous opcode
   }
   if (ZEND_SEND_ARRAY == prev_opline->opcode) {
+
     if (UNEXPECTED((NULL == execute_data->func))) {
       nrl_verbosedebug(NRL_AGENT, "%s: cannot get current function", __func__);
       return;
     }
-    if (UNEXPECTED(NULL == execute_data->prev_execute_data->func)) {
```
zsistla, line 1978: Why was this check removed?

zsistla: Nevermind, I see it's called earlier.

ZNeumann: It was moved to https://github.com/newrelic/newrelic-php-agent/pull/708/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1978 because we need to check the func type (for internal functions) before checking opcodes

zsistla (approved):
Great job getting the issue fixed.

PR #709, opened 2023-08-08
fix build for PHP 5.x

lavarou:
Build fails when PHP 5.x ifdefs are missing - `zend_string` type is not defined for PHP 5.x.


ZNeumann:
I might prefer wrapping the entire function in a 7.0+ ifdef instead. This function is never called in 5.x code.

lavarou:
> I might prefer wrapping the entire function in a 7.0+ ifdef instead. This function should never be called in 5.x code. (and if it is, we can likely ifdef that function in a 7.0+ block as well.)

I tried to increase code readability with cec08d8da175666c499275d2b4a6a40e40295764. Does it look any better?

PR #713, opened 2023-08-16
tests(agent): enhance guzzle tests

lavarou:
* [cleanup skipif.inc for guzzle5](https://github.com/newrelic/newrelic-php-agent/commit/a86c38657abbb318bdd2c310f34c8af4825cbea1) 
skipif.inc for guzzle5 ensures tests run on correct version of PHP and skip
test if PHP version does not fall into >=5.4,<7.4 interval. Additionally,
it warns (and skips) if guzzle5 installation is not found.

* [add skipif.inc for guzzle6](https://github.com/newrelic/newrelic-php-agent/commit/442efcf9382cd4757c2b218a4b3fc25e44f48de8) 
skipif.inc for guzzle6 ensures tests run on correct version of PHP and skip
test if PHP version does not fall into >=5.5,<8.0 interval. Additionally,
it warns (and skips) if guzzle6 installation is not found.

* [add tests for version 7 of guzzle](https://github.com/newrelic/newrelic-php-agent/commit/5b9d6a73a8aaa8e176ca76cbf2500e5ad2cc4d60) 
Version 6 of Guzzle supports PHPs up to but not including 8.0. In order to test
agent's instrumentation of guzzle on PHPs 8.0+, version 7 of Guzzle needs to
be used. There are no fundamental changes in version 7 of Guzzle therefore
agent's instrumentation of version 6 of guzzle works for version 7 too. However,
this results in guzzle7 being detected as guzzle6 and that's why metrics in
integration tests expectation include `Supportability/library/Guzzle 6/detected`
as well as `Supportability/library/Guzzle 4-5/detected`.

* [composer update guzzle 6 & 7 in guzzle.tar.bz2](https://github.com/newrelic/newrelic-php-agent/commit/bf000082a211ddc151cee9d3eea6b15e01bce8d5)
  - Upgrading guzzlehttp/guzzle (6.3.3 => 6.5.8) (with dependencies)
  - Upgrading guzzlehttp/guzzle (7.4.5 => 7.7.0) (with dependencies)


zsistla (approved):
Thanks for updating.

tests/integration/external/guzzle7/test_no_cat_no_dt.php
```diff
@@ -0,0 +1,78 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that Cross Application Tracing (CAT) works with guzzle 7.
```
mfulb, line 8: This comment seems incorrect based on INI setting for CAT?

mfulb (approved):
Thanks Michal!

tests/integration/external/guzzle7/test_no_cat_no_dt.php
lavarou: @mfulb Funny thing - guzzle5 and guzzle6 tests have exactly the same problem :). I can update this comment in all 3 tests. Will this be more accurate description:
```suggestion
Test that neither CAT nor DT headers are added when CAT and DT are disabled.
```

lavarou: Done in 20a51012c057bb39c4bc303642c4136609eeb715

PR #498, opened 2022-08-01
feat(agent): Add Guzzle 7 Library Support

hahuja2:
This PR implements the following changes:

- Combine Guzzle 6 instrumentation with Guzzle 7
- Add guzzle version check function, which returns the correct guzzle version
- Remove emitter function, which checks for emitter file in guzzle 4-5 package
- Add Guzzle 7 test cases
- Install Guzzle 7 dependencies and create new tar file with all guzzle packages included (Guzzle 4-7)
- Guzzle 7 requires PHP 7.2 or above

PR #714, opened 2023-08-17
refactor(agent): improve guzzle detection

lavarou:
Use files that more uniquely identify Guzzle 5, and Guzzle 6. Because 
the file used to identify Guzzle 5 is unique to Guzzle 5, Guzzle 4-5
supportability metrics are no longer generated when predis is detected.
Guzzle 4-5 supportability metrics is also not generated when Guzzle 6
or Guzzle 7 are detected.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -476,12 +476,8 @@ typedef struct _nr_library_table_t {
 static nr_library_table_t libraries[] = {
     {"Doctrine 2", "doctrine/orm/query.php", nr_doctrine2_enable},
     {"Guzzle 3", "guzzle/http/client.php", nr_guzzle3_enable},
-    /*
-     * NOTE:  This file ClientInterface.php also exists in Guzzle 3 and 6.
-     *        This file also exists in Predis.
-     */
-    {"Guzzle 4-5", "clientinterface.php", nr_guzzle4_enable},
-    {"Guzzle 6", "handlerstack.php", nr_guzzle6_enable},
+    {"Guzzle 4-5", "hasemitterinterface.php", nr_guzzle4_enable},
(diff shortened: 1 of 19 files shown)


agent/php_execute.c
```diff
@@ -476,12 +476,8 @@ typedef struct _nr_library_table_t {
...
     {"Doctrine 2", "doctrine/orm/query.php", nr_doctrine2_enable},
     {"Guzzle 3", "guzzle/http/client.php", nr_guzzle3_enable},
-    /*
-     * NOTE:  This file ClientInterface.php also exists in Guzzle 3 and 6.
-     *        This file also exists in Predis.
-     */
-    {"Guzzle 4-5", "clientinterface.php", nr_guzzle4_enable},
-    {"Guzzle 6", "handlerstack.php", nr_guzzle6_enable},
+    {"Guzzle 4-5", "hasemitterinterface.php", nr_guzzle4_enable},
+    {"Guzzle 6", "guzzle/src/functions_include.php", nr_guzzle6_enable},
```
zsistla, line 480: Should this be `{"Guzzle 6-7", "guzzle/src/functions_include.php", nr_guzzle6_enable},`
for now?

lavarou: I was debating on this and decided to leave the functionality of the agent unchanged. This string is used to create a metric name, and that can be used to query data. Because this PR does not add feature to differentiate between Guzzle 6 and Guzzle 7, `Guzzle 6-7` means the same thing as `Guzzle 6` and I don't want to break existing queries. I hope future work, based on #498 , will leave `Guzzle 6` metric but add `Guzzle 7` supportability metric to precisely measure usage of different versions of Guzzle.

zsistla (approved):
Great work!

PR #715, opened 2023-08-17
tests(agent): Add tests that specifically exercise JIT.

zsistla:
Convert forked PR branch to main repo branch so the ci/cd will run...

Please see original forked PR here: https://github.com/newrelic/newrelic-php-agent/pull/583


tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
```diff
@@ -0,0 +1,142 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that span events are correctly created from any eligible segment when a caught exception occurs.
+Putting in a try/catch block means an exception is NOT handled by the exception handler.
```
lavarou, line 9: This test is a clone of `tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php`. Previous PR #583 in 3771c158d6c4c98e8bfd6b5f44d68787604e2db7 (and therefore this PR) as well as #649 modify this test. However, as [expressed before](https://github.com/newrelic/newrelic-php-agent/pull/649#pullrequestreview-1363871912), I'm not sure this test should be modified.

lavarou (changes requested):

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
zsistla: Please see:
[2a47c6c](https://github.com/newrelic/newrelic-php-agent/pull/715/commits/2a47c6cb820f730a3ccf2398dcd75e908c6f45c6)

[62e699b](https://github.com/newrelic/newrelic-php-agent/pull/715/commits/62e699b387d4013910431130bda4038b1cec8f9e)

zsistla: To recap discussion,
the original _caught_exception tests were actually testing an uncaught exception being handled by an exception handler.
Another PR will handle: 
* renaming the uncaught_exception test to uncaught_unhandled_exception
* renaming the caught_exception to uncaught_handled_exception 
* creating new caught_exception to test the caught exception case.

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
```diff
@@ -29,20 +31,17 @@
...
 
+
 /*PHPMODULES
 zend_extension=opcache.so
 */
 
-/*EXPECT_ERROR_EVENTS
-null
-*/
-
```
lavarou: Why was this removed? The 'baseline' for this test has this: https://github.com/newrelic/newrelic-php-agent/blob/acc7942cb672999fd87537be9aae508915bcc916/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php#L26-L28

tests/integration/jit/tracing/test_span_events_are_created_upon_caught_exception.php
```diff
@@ -33,16 +33,12 @@
 zend_extension=opcache.so
 */
 
-/*EXPECT_ERROR_EVENTS
-null
-*/
-
```
lavarou: Why was this removed? The 'baseline' for this test has this: https://github.com/newrelic/newrelic-php-agent/blob/acc7942cb672999fd87537be9aae508915bcc916/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php#L26-L28

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
```diff
@@ -21,6 +21,8 @@
 newrelic.transaction_tracer.threshold = 0
 newrelic.span_events_enabled=1
 newrelic.cross_application_tracer.enabled = false
+display_errors=1
+log_errors=0
 error_reporting = E_ALL
```
lavarou, line 26: I don't think the error related ini settings are required. The 'baseline' for this test doesn't have them: https://github.com/newrelic/newrelic-php-agent/blob/acc7942cb672999fd87537be9aae508915bcc916/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php#L19-L24

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
zsistla: The original tests, which provided the prototype for this test, pass; however, these are not exact clones or they would be duplicate tests.

These tests have differences, namely:
https://github.com/newrelic/newrelic-php-agent/pull/715/files/acc7942cb672999fd87537be9aae508915bcc916#diff-61ae0ae768b1f2332109c0aad270b311e54933445589dc3249ce023b009b78eaR24-R26

So we are using PHP's native error handling in a slightly different way.
Because of the differences and how PHP handles errors, for PHP 8.0/8.1 it creates an error when it handles the uncaught exception and the agent create an error event with that error. However, PHP 8.2 handles an uncaught exception slightly differently and doesn't generate an error when it handles an uncaught exception.

lavarou: > The original tests, which provided the prototype for this test, pass; however, these are not exact clones or they would be duplicate tests.

Yes, they are not duplicates. The modify runtime execution of original tests, by:
1. loading opcache module
2. enabling opcache
3. enabling JIT

However, the agent's behavior should persist regardless of this change in runtime. Yet, agent's behavior changes, and this code block:
```
/*EXPECT_ERROR_EVENTS
null
*/
```
exposes those behavior differences. I.e. ERROR_EVENTS are generated in PHPs 8.0 and 8.1 but are not generated in PHP 8.2.

lavarou: If the behavior of the agent is supposed to be different depending on PHP version, then this test should be PHP version specific.

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
zsistla: The tests here are not meant to necessarily be duplicates.

The INI settings are to be consistent with the rest of the `tracing` and `function` tests and to ensure the log settings are consistent as sometimes the setting on a particular machine have different log settings which could then lead to different test behavior.

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
zsistla: This behavior differs between PHP 8.0/8.1 and PHP 8.2 because PHP used to accept less strongly typed error handlers without explicit parameters. Starting with PHP 8.2, error handlers need to have explicit parameters.
As such, exceptions handled by old style error handlers will be correctly propagated for PHP 8.0/8.1,
but PHP 8.2 and above requires strongly typed error handlers otherwise the object it returns isn't explicitly recognizable as a Throwable.

The tests are split to accommodate the difference.

Please see: [d139ac6](https://github.com/newrelic/newrelic-php-agent/pull/715/commits/d139ac65b1e4d608579bae7120f5fd19d9e56392) which does the following:
* Split tests since PHP used to accept less strongly typed error handlers without explicit parameters.
Starting with PHP 8.2, error handlers need to have explicit parameters.
As such, exceptions handled by old style error handlers will be correctly propagated for PHP 8.0/8.1,
but PHP 8.2 and above requires strongly typed error handlers otherwise the object it returns isn't explicitly recognizable as a Throwable.
To cover both cases, there is now one test for PHP 8.0/8.1 with old style error handler and one test for PHP 8.0+ with new style handler.
* Updated description to explicitly determine if error events are created since previous description only noted if span events were created.
* Updated names to more accurately reflect test behavior.

tests/integration/jit/tracing/test_span_events_are_created_upon_caught_exception.php
zsistla: This behavior differs between PHP 8.0/8.1 and PHP 8.2 because PHP used to accept less strongly typed error handlers without explicit parameters. Starting with PHP 8.2, error handlers need to have explicit parameters.
As such, exceptions handled by old style error handlers will be correctly propagated for PHP 8.0/8.1,
but PHP 8.2 and above requires strongly typed error handlers otherwise the object it returns isn't explicitly recognizable as a Throwable.

The tests are split to accommodate the difference.

Please see: [d139ac6](https://github.com/newrelic/newrelic-php-agent/pull/715/commits/d139ac65b1e4d608579bae7120f5fd19d9e56392) which does the following:
* Split tests since PHP used to accept less strongly typed error handlers without explicit parameters.
Starting with PHP 8.2, error handlers need to have explicit parameters.
As such, exceptions handled by old style error handlers will be correctly propagated for PHP 8.0/8.1,
but PHP 8.2 and above requires strongly typed error handlers otherwise the object it returns isn't explicitly recognizable as a Throwable.
To cover both cases, there is now one test for PHP 8.0/8.1 with old style error handler and one test for PHP 8.0+ with new style handler.
* Updated description to explicitly determine if error events are created since previous description only noted if span events were created.
* Updated names to more accurately reflect test behavior.

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
zsistla: Please see: [d139ac6](https://github.com/newrelic/newrelic-php-agent/pull/715/commits/d139ac65b1e4d608579bae7120f5fd19d9e56392) which does the following:
* Split tests since PHP used to accept less strongly typed error handlers without explicit parameters.
Starting with PHP 8.2, error handlers need to have explicit parameters.
As such, exceptions handled by old style error handlers will be correctly propagated for PHP 8.0/8.1,
but PHP 8.2 and above requires strongly typed error handlers otherwise the object it returns isn't explicitly recognizable as a Throwable.
To cover both cases, there is now one test for PHP 8.0/8.1 with old style error handler and one test for PHP 8.0+ with new style handler.
* Updated description to explicitly determine if error events are created since previous description only noted if span events were created.
* Updated names to more accurately reflect test behavior.

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
lavarou: These INI settings are not relevant to this test.

zsistla: The INI values are not required for the test to run.  However, we can improve our tests.

Since these are tests, if an unexpected error occurs (especially in a complex situation like JIT entails), we want to be notified of the diagnostics whether or not the error notification setting is toggled on the system or whether or not the default has changed.

```
display_errors string
This determines whether errors should be printed to the screen as part of the output or if they should be hidden from the user.
```

error_reporting default has changed in the past and could in the future.
```
error_reporting int
Set the error reporting level. The parameter is either an integer representing a bit field, or named constants. The error_reporting levels and constants are described in [Predefined Constants](https://www.php.net/manual/en/errorfunc.constants.php), and in php.ini. To set at runtime, use the [error_reporting()](https://www.php.net/manual/en/function.error-reporting.php) function. See also the [display_errors](https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors) directive.

The default value is E_ALL.

Prior to PHP 8.0.0, the default value was: E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED. This means diagnostics of level E_NOTICE, E_STRICT and E_DEPRECATED were not shown. 
```

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
lavarou: I don't think 38eb42d6cde9d20bdaa579d715e828f775406a98 and d139ac65b1e4d608579bae7120f5fd19d9e56392 is the right approach here. The crux of the issue is that `nr_php_error_zval_is_exception` in PHP 8.2 when opcache is enabled is not able to confirm that the zval it got is of Throwable type. According to https://www.php.net/manual/en/function.set-exception-handler.php#refsect1-function.set-exception-handler-parameters, the `$callback` should be a function that accepts exactly one parameter, which will be the [Throwable](https://www.php.net/manual/en/class.throwable.php) object. The handler in the test is a closure that doesn’t expect and/or use any parameters. The incorrect PHP code leads to undefined behavior of the agent, which is what this issue is all about. To address this issue, the test code needs to be fixed in the first place. When the test code uses valid PHP code, the OAPI agent behaves the same for all PHP versions regardless if opcache is enabled or disabled and therefore the split between PHP versions is not needed. This PR https://github.com/newrelic/newrelic-php-agent/pull/748 reverts 38eb42d6cde9d20bdaa579d715e828f775406a98 and d139ac65b1e4d608579bae7120f5fd19d9e56392 and fixes the PHP code in tests. 

However, https://github.com/newrelic/newrelic-php-agent/pull/748 will expose a different issue: when DT is enabled, OAPI agent generates error_events and pre-OAPI agent does not generate error_events (`tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php` fails for OAPI agent because it has [this expectation](https://github.com/newrelic/newrelic-php-agent/blob/2b9e3b4195b7f61cab4e9252a802957193ee2d27/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php#L26-L28)). This behavior difference is caused by this code difference:
* pre-oapi agent checks if `nr_txn_should_create_span_events(NRPRG(txn))` [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1361) and generates error_event only when current request will not cause spans to be created
* oapi agent does not check if current request will or will not cause spans to be created [here](https://github.com/newrelic/newrelic-php-agent/blob/2b9e3b4195b7f61cab4e9252a802957193ee2d27/agent/php_execute.c#L2061-L2085) and and generates error_event unconditionally.

To address this issue we need to answer this question: "Which behavior is correct: pre-OAPI agent or OAPI agent?"

tests/integration/jit/function/test_span_events_error_collector_disabled.php
```diff
@@ -0,0 +1,127 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that no span events error attributes are captured when error_collector is disabled
```
lavarou, line 8:
```suggestion
Test that no span events error attributes are captured when error_collector is disabled. No error events should be generated either.
```

tests/integration/jit/function/test_span_events_error_collector_disabled.php
```diff
@@ -0,0 +1,127 @@
...
+opcache.enable_cli=1
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=function
+*/
+
+/*PHPMODULES
+zend_extension=opcache.so
+*/
+
```
lavarou, line 37:
```suggestion
/*EXPECT_ERROR_EVENTS
null
*/

```

tests/integration/jit/tracing/test_span_events_error_collector_disabled.php
```diff
@@ -0,0 +1,127 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that no span events error attributes are captured when error_collector is disabled
```
lavarou, line 8:
```suggestion
Test that no span events error attributes are captured when error_collector is disabled.
No error events should be generated either.
```

tests/integration/jit/tracing/test_span_events_error_collector_disabled.php
```diff
@@ -0,0 +1,127 @@
...
+opcache.enable_cli=1
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=tracing
+*/
+
+/*PHPMODULES
+zend_extension=opcache.so
+*/
+
```
lavarou, line 37:
```suggestion
/*EXPECT_ERROR_EVENTS
null
*/

```

tests/integration/jit/function/test_span_events_are_created_upon_caught_error.php
```diff
@@ -0,0 +1,161 @@
...
+  ]
+]
+*/
+
+/*EXPECT_REGEX
+^\s*(PHP )?Fatal error:\s*foo in .*? on line [0-9]+\s*$
+*/
+
+set_error_handler(
+    function () {
```
lavarou, line 140: Technically, since this PR fixes the invalid code in tests that set user exception handler, I think it makes sense to fix error handler callback signature as well. According to https://www.php.net/manual/en/function.set-error-handler.php, the callback should accept at least two parameters: `int $errno` and `string $errstr` (the others - `string $errfile`, `int $errline`, `array $errcontext` - are indicated as optional). Therefore consider this change:
```suggestion
    function (int $errno, string $errstr) {
```

tests/integration/jit/function/test_span_events_are_created_upon_caught_error.php
```diff
@@ -0,0 +1,161 @@
...
+opcache.enable_cli=1
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=function
+*/
+
+/*PHPMODULES
+zend_extension=opcache.so
+*/
+
```
lavarou, line 37:
```suggestion
/*EXPECT_ERROR_EVENTS
[
  "?? agent run id",
  {
    "reservoir_size": 100,
    "events_seen": 1
  },
  [
    [
      {
        "type": "TransactionError",
        "timestamp": "??",
        "error.class": "E_USER_ERROR",
        "error.message": "foo",
        "transactionName": "OtherTransaction\/php__FILE__",
        "duration": "??",
        "databaseDuration": "??",
        "databaseCallCount": 1,
        "nr.transactionGuid": "??",
        "guid": "??",
        "sampled": true,
        "priority": "??",
        "traceId": "??",
        "spanId": "??"
      },
      {},
      {}
    ]
  ]
]
*/

```

tests/integration/jit/function/test_span_events_are_created_upon_caught_error.php
```diff
@@ -0,0 +1,161 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that span events are correctly created from any eligible segment, even
+when an error is generated and handled.
```
lavarou, line 9:
```suggestion
Test that span events are correctly created from any eligible segment, even
when an error is generated and handled by user error handler. The span that
generated the error should have error attributes. Additionally error events
should be created.
```

tests/integration/jit/tracing/test_span_events_are_created_upon_caught_error.php
```diff
@@ -0,0 +1,161 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that span events are correctly created from any eligible segment, even
+when an error is generated and handled.
```
lavarou, line 9:
```suggestion
Test that span events are correctly created from any eligible segment, even
when an error is generated and handled by user error handler. The span that
generated the error should have error attributes. Additionally error events
should be created.
```

tests/integration/jit/tracing/test_span_events_are_created_upon_caught_error.php
```diff
@@ -0,0 +1,161 @@
...
+opcache.enable_cli=1
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=tracing
+*/
+
+/*PHPMODULES
+zend_extension=opcache.so
+*/
+
```
lavarou, line 37:
```suggestion
/*EXPECT_ERROR_EVENTS
[
  "?? agent run id",
  {
    "reservoir_size": 100,
    "events_seen": 1
  },
  [
    [
      {
        "type": "TransactionError",
        "timestamp": "??",
        "error.class": "E_USER_ERROR",
        "error.message": "foo",
        "transactionName": "OtherTransaction\/php__FILE__",
        "duration": "??",
        "databaseDuration": "??",
        "databaseCallCount": 1,
        "nr.transactionGuid": "??",
        "guid": "??",
        "sampled": true,
        "priority": "??",
        "traceId": "??",
        "spanId": "??"
      },
      {},
      {}
    ]
  ]
]
*/
```

tests/integration/jit/tracing/test_span_events_are_created_upon_caught_error.php
```diff
@@ -0,0 +1,161 @@
...
+  ]
+]
+*/
+
+/*EXPECT_REGEX
+^\s*(PHP )?Fatal error:\s*foo in .*? on line [0-9]+\s*$
+*/
+
+set_error_handler(
+    function () {
```
lavarou, line 140: Technically, since this PR fixes the invalid code in tests that set user exception handler, I think it makes sense to fix error handler callback signature as well. According to https://www.php.net/manual/en/function.set-error-handler.php, the callback should accept at least two parameters: `int $errno` and `string $errstr` (the others - `string $errfile`, `int $errline`, `array $errcontext` - are indicated as optional). Therefore consider this change:
```suggestion
    function (int $errno, string $errstr) {
```

lavarou (commented):
I have a few 'cosmetic' updates which, hopefully, improve test descriptions as well as test expectations.

tests/integration/jit/function/test_span_events_are_created_upon_caught_exception.php
lavarou: After further discussion it was determined that:
* this PR changes tests in `tests/integratio/jit` subtree only, and therefore the signature of user exception handler in `tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php` is not going to be fixed by this PR. https://github.com/newrelic/newrelic-php-agent/pull/748 will be re-purposed to address fixing invalid test code in tests outside `tests/integratio/jit` subtree.
* it is ok to leave `test_span_events_are_created_upon_uncaught_exception.php` -> `test_span_events_are_created_upon_uncaught_unhandled_exception.php` and `test_span_events_are_created_upon_caught_exception.php` -> {`test_span_events_are_created_upon_uncaught_handled_exception.php` and `test_span_events_are_created_upon_uncaught_handled_exception_invalid_handler.php`} file renames in this PR
* the behavior in oapi agent of generating error events unconditionally is the correct behavior; either re-purposed https://github.com/newrelic/newrelic-php-agent/pull/748 or new PR will address this behavior change and update/split `test_span_events_are_created_upon_caught_exception.php` and/or other tests that have this issue depending on PHP version.

tests/integration/jit/function/test_span_events_are_created_upon_uncaught_old_handled_exception.php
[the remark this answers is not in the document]
zsistla: Good idea, done.

tests/integration/jit/tracing/test_span_events_are_created_upon_old_uncaught_handled_exception.php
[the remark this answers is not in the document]
zsistla: Also done.

tests/integration/jit/tracing/test_span_events_are_created_upon_old_uncaught_handled_exception.php
```diff
@@ -0,0 +1,198 @@
...
+
+/*SKIPIF
+<?php
+
+if (version_compare(PHP_VERSION, "8.2", ">=")) {
+  die("skip: PHP > 8.1 not supported\n");
+}
+//else {
+require('skipif.inc');
+//}
```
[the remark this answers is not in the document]
zsistla: Already handled in another change.

Order seemed to matter for the test runner on x64 8.2 instances.

Since it tried to run the test on gnu and musl , amd64, 8.2 which it should have skipped (and which oddly skipped okay the first time).

Going with this version:
```
if (version_compare(PHP_VERSION, "8.2", ">=")) {
  die("skip: PHP > 8.1 not supported\n");
}
require('skipif.inc');
```

tests/integration/jit/tracing/test_span_events_are_created_upon_old_uncaught_handled_exception.php
```diff
@@ -0,0 +1,198 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+ 
+/*DESCRIPTION
+Test that span events are correctly created from any eligible segment, even
+when an exception is handled by the exception handler.  In the case of PHP 8.0/8.1,
+PHP OAPI additionally where PHP OAPI additionally passes exception information in the
+zend_execute_data for the agent to use to create an error_event.
+Check that error events are created.
+*/
```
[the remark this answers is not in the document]
zsistla: File was "outdated" so wouldn't automatically commit.
Change folded in here: [47469df](https://github.com/newrelic/newrelic-php-agent/pull/715/commits/47469dfd368ca6cbb8a8b0540ddf931d998838da)

lavarou (approved):

ZNeumann (approved):
Great work and very thorough conversations!

PR #717, opened 2023-08-21
feat(testing): Implement EXPECT_SPAN_EVENTS_LIKE

mfulb:
Adds a new feature (`EXPECT_SPAN_EVENTS_LIKE`) for the `integration_runner` which allows testing for the existence of a set of spans in the actual spans created.  It differs from `EXPECT_SPAN_EVENTS` in that it is not necessary to list ALL expected spans.  For tests involving PHP frameworks the number of generated spans can be in the hundreds so it is not practical to use `EXPECT_SPAN_EVENTS` and thus `EXPECT_SPAN_EVENTS_LIKE` was added.

Here is an example:

```
/*EXPECT_SPAN_EVENTS_LIKE
[
  [
    {
        "category": "generic",
        "type": "Span",
        "guid": "??",
        "traceId": "??",
        "transactionId": "??",
        "name": "WebTransaction\/Action\/dispatch",
        "timestamp": "??",
        "duration": "??",
        "priority": "??",
        "sampled": true,
        "nr.entryPoint": true,
        "transaction.name": "WebTransaction\/Action\/dispatch"
    },
    {},
    {
        "response.headers.contentType": "text\/html",
        "http.statusCode": 200,
        "response.statusCode": 200,
        "httpResponseCode": "200",
        "request.uri": "\/dispatch",
        "request.method": "GET",
        "request.headers.host": "127.0.0.1"
    }
  ],
  [
    {
        "category": "generic",
        "type": "Span",
        "guid": "??",
        "traceId": "??",
        "transactionId": "??",
        "name": "Custom\/App\\Http\\Controllers\\JobOne::dispatch",
        "timestamp": "??",
        "duration": "??",
        "priority": "??",
        "sampled": true,
        "parentId": "??"
    },
    {},
    {
        "code.lineno": "??",
        "code.namespace": "App\\Http\\Controllers\\JobOne",
        "code.filepath": "??",
        "code.function": "dispatch"
    }
  ]
]
*/
```

### DIFF
--- a/src/newrelic/integration/test.go
+++ b/src/newrelic/integration/test.go
@@ -328,6 +328,7 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
 
 	if err := json.Unmarshal(t.spanEventsLike, &x2); nil != err {
 		t.Fatal(fmt.Errorf("unable to parse expected spans like json for fuzzy matching: %v", err))
+		return
 	}
 
 	// expected will be represented as an array of "interface{}"
@@ -345,6 +346,7 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
 	actualJSON, err := es.Data(id, time.Now())
 	if nil != err {
 		t.Fatal(fmt.Errorf("unable to access span event JSON data: %v", err))
+		return
 	}
 
 	// scrub actual spans
@@ -355,6 +357,7 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
 	// parse to internal format
 	if err := json.Unmarshal(scrubjson, &x1); nil != err {
 		t.Fatal(fmt.Errorf("unable to parse actual spans like json for fuzzy matching: %v", err))
+		return
 	}
 
 	// expect x1 to be of type "[]interface {}" which wraps the entire span event data
@@ -371,12 +374,14 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
 	case []interface{}:
 	default:
 		t.Fatal(errors.New("span event data json doesnt match expected format"))
+		return
 	}
 
 	// expect array of len 3
 	v2, _ := x1.([]interface{})
 	if 3 != len(v2) {
 		t.Fatal(errors.New("span event data json doesnt match expected format - expected 3 elements"))
+		return
 	}
 
 	// get array of actual spans from 3rd element
@@ -422,7 +427,7 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
 				Expect: string(expectedJSON),
 				Actual: actualPretty.String(),
 			})
-
+			return
 		}
 	}
 }
--- a/src/newrelic/integration/test.go
+++ b/src/newrelic/integration/test.go
@@ -403,8 +403,8 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
 				continue
 			}
 
-			match := isFuzzyMatchRecursive(expected[j], actual[i])
-			if nil == match {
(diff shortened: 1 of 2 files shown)


ZNeumann (approved):

src/newrelic/integration/test.go
```diff
@@ -315,6 +316,117 @@ func (t *Test) compareResponseHeaders() {
...
+func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
+	// convert array of expected spans JSON to interface representation
+	var x2 interface{}
+
+	if nil == t.spanEventsLike {
+		return
+	}
+
+	if err := json.Unmarshal(t.spanEventsLike, &x2); nil != err {
+		t.Fatal(fmt.Errorf("unable to parse expected spans like json for fuzzy matching: %v", err))
```
lavarou, line 330: Isn't `return` needed here?

mfulb: Yes!  Thanks for catching this.
I misunderstood how the test failure mechanism worked but reviewing more of the existing code I see a return is required.

Fixed in commit [56511a1](https://github.com/newrelic/newrelic-php-agent/pull/717/commits/56511a1c7bb3aaa9a5fdb83077590f5dbe4721cf).

src/newrelic/integration/test.go
```diff
@@ -315,6 +316,122 @@ func (t *Test) compareResponseHeaders() {
...
+	}
+
+	// expect x1 to be of type "[]interface {}" which wraps the entire span event data
+	// within this generic array there will be:
+	// - a string of the form "?? agent run id"
+	// - a map container "events_seen" and "reservoir_size"
+	// - an array of arrays containing:
+	//    - map containing main span data
+	//    - map for attributes(?)
+	//    - map of CLM data
```
lavarou, line 370: Could you add an example of `/*EXPECT_SPAN_EVENTS_LIKE` to the description of the PR?

mfulb: But this comment is related to an actual span events as captured in the harvest by the daemon/integration_runner.  We only keep the final "an array of arrays containing" part as that is the array of spans, the other parts are metadata we don't care about for comparing span events.

src/newrelic/integration/test.go
```diff
@@ -315,6 +316,122 @@ func (t *Test) compareResponseHeaders() {
...
+	for i := 0; i < len(actual); i++ {
+		// check each expected span (interface representation) against current actual span
+		// only iterate over unmatched expected spans
+		for j := 0; j < len(expected); j++ {
+			if haveMatched[j] {
+				continue
+			}
+
+			match := isFuzzyMatchRecursive(expected[j], actual[i])
+			if nil == match {
```
lavarou, line 407: This is a bit confusing - `numMatched` is updated when `match` is `nil` 🤔 `isFuzzyMatchRecursive` returns an error, not the match result. Therefore I think this code would read better and match `isFuzzyMatchRecursive` signature and intention:
```suggestion
			err := isFuzzyMatchRecursive(expected[j], actual[i])
			if nil == err {
```
I.e. if no error when fuzzy matching data, update `numMatched`.

src/newrelic/integration/test.go
```diff
@@ -315,6 +316,122 @@ func (t *Test) compareResponseHeaders() {
...
+				haveMatched[j] = true
+				numMatched++
+				break
+			}
+		}
+
+		if len(expected) == numMatched {
```
lavarou, line 414: This is short-circuit optimization, right? No need to go through all spans in actual harvest if all expected have already been found.

mfulb: Correct.

src/newrelic/integration/test.go
mfulb: That makes sense - thanks for the suggestion!

PR #719, opened 2023-08-24
feat: add functionality to group errors

bduranleau-nr:
Implement Error Fingerprinting functionality in the PHP Agent.

Error Fingerprinting allows a user to call an Agent API `newrelic_set_error_group_callback` and provide it a PHP callback function as an argument. 

If the API call is successful, the Agent will register this PHP callback on a per-request basis and invoke it if the user application encounters a PHP error. The contents of the callback are completely at the user's discretion, but it must adhere to the following paradigm:
- The callback must accept *two* arguments that will be PHP arrays provided by the Agent: The first contains transaction data, the second contains error data.
- The callback must return a string value that is the error fingerprint name the user wishes to group the error under in their Errors Inbox.

If successful, an agent attribute `error.group.name` will be sent to the backend with the return value from the callback.


agent/php_api.c
```diff
@@ -1578,3 +1578,73 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
...
+                "callback parameters: %d",
+                fcc.function_handler->common.num_args);
+    RETURN_FALSE;
+  }
+
+  // Log debug message if the user is overwriting an existing callback
+  if (NULL != NRPRG(error_group_user_callback)) {
+    nrl_debug(
+        NRL_API,
+        "newrelic_set_error_group_callback: overwriting previous callback");
```
lavarou, line 1634: Should the `error_group_user_callback` be overwritten or left as it was?

bduranleau-nr: According to the Agent Spec, any subsequent calls to the API to register a callback will clobber any previously registered callbacks.

lavarou: Then I think this code needs to be updated as it currently doesn't overwrite `error_group_user_callback`.

lavarou: Also would be good to have a test for this case, i.e. that the callback gets overwritten.

lavarou: Ah! I think I know what is happening - this code only allocates memory if it was not allocated. Then the comment is misleading a bit. And I found the test for 'clobbering'.

bduranleau-nr: Yes, this is just a message for debugging purposes in case someone has multiple callbacks and can't understand why it isn't behaving how they might expect. The memory is already allocated, so we can just overwrite  `fci` and `fcc` directly without having to reallocate. I have included an integration test that tests this functionality.

agent/php_error.c
```diff
@@ -580,6 +696,13 @@ nr_status_t nr_php_error_record_exception(nrtxn_t* txn,
     error_message = nr_formatf("%s'%s'", prefix, klass);
   }
 
+  /*
+   * Error Fingerprinting Callback
+   */
+  if (NULL != NRPRG(error_group_user_callback)) {
+    nr_php_error_fingerprint_callback(txn, klass, message, file, stack_json);
```
lavarou, line 703: Somehow I feel a verb is missing from this function name. This function is not a callback, it is supposed to call user's callback, right? If that's the case then maybe `nr_php_error_call_error_group_callback` or `nr_php_error_set_group` would be better?

bduranleau-nr: It's a balance between being descriptive without being overly verbose- How about `nr_php_error_call_fingerprint_callback` or `nr_php_call_error_fingerprint_callback`? 
I think we have to maintain some reference to the fact that this function is calling a callback and not just us setting a value.

lavarou: It might be better (for searching purposes) to consistently name things pertaining to the same functionality. Some places use `error_group_user_callback`, others use `fingerprint_callback` and others use `error_group_callback`. Could we make it consistent throughout the codebase and use `error_group_callback` (after the API)? And if so, the function would be named `nr_php_error_call_error_group_callback`.

bduranleau-nr: Yeah, I can agree with that- changed to `nr_php_error_call_error_group_callback ` in 8d06bbc.

agent/php_newrelic.h
```diff
@@ -503,6 +510,9 @@ nrtxn_t* txn; /* The all-important transaction pointer */
 char* predis_ctx; /* The current Predis pipeline context name, if any */
 nr_hashmap_t* predis_commands;
 
+nrcallbackfn_t* error_group_user_callback; /* The user defined callback for
+                                              error group naming */
+
```
lavarou, line 515: Why does this have to be a pointer to allocated memory and not a plain variable without dynamic memory management overhead?

agent/php_newrelic.h
```diff
@@ -304,8 +309,9 @@ nr_php_ini_attribute_config_t
                                     */
 
 nrinibool_t custom_events_enabled; /* newrelic.custom_insights_events.enabled */
-nriniuint_t custom_events_max_samples_stored; /* newrelic.custom_events.max_samples_stored */
-nrinibool_t synthetics_enabled;    /* newrelic.synthetics.enabled */
+nriniuint_t custom_events_max_samples_stored; /* newrelic.custom_events.max_samples_stored
+                                               */
+nrinibool_t synthetics_enabled;               /* newrelic.synthetics.enabled */
```
lavarou, line 314: This change is unrelated to error grouping. To improve code review experience and avoid distractions caused by unrelated formatting changes, I strongly suggest using `git clang-format` command, which formats only the code that is changed in a commit, not the whole file. You can read more about it [here](https://ortogonal.github.io/cpp/git-clang-format/#:~:text=git%2Dclang%2Dformat%20is%20a,on%20the%20changes%20you%20made).

agent/php_newrelic.h
bduranleau-nr: It's easier to check if it's been set or not simply by checking `NULL != error_group_user_callback`. Since it is more likely to not be set, I prefer the need to handle memory management over the need for a more involved check to see if the struct is set or not.

lavarou: Why not add a `is_set` flag to `nrcallbackfn_t` type? This would achieve the same without the memory management overhead.

bduranleau-nr: > When error group user callback is used, isn't it set for every request?
No. This will only ever be allocated if the API is called, and only deallocated if set (which can only happen if the API is called). Users who do not call the API will not be exposed to any of this functionality.

lavarou: So for an application that uses the error grouping callback, and sets it up early in the request processing, the memory will be allocated and freed for every request. Maybe it is a rare use case, but I'm just pointing out possible issues if dynamic memory is used for storing error grouping callback - performance of calling malloc/free and heap fragmentation in a long running process.

tests/integration/api/error_group_callback/test_error_group_callback_empty_callback.php
```diff
@@ -0,0 +1,109 @@
...
+  ]
+]
+*/
+
+function alpha()
+{
+  newrelic_notice_error(new Exception('Sample Exception'));
+}
+
+newrelic_set_error_group_callback();
```
lavarou, line 107: Should the return value for `newrelic_set_error_group_callback` be tested? This test expects `false` to be returned.

lavarou: [`tap_assert`](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/include/tap.php#L12) and [`tap_refute`](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/include/tap.php#L21) with `/*EXPECT*` could be used to test return value of `newrelic_set_error_group_callback()`. 

For an example use of `tap_assert` with corresponding `/*EXPECT` see:

[`/*EXPECT ok - delete list with ltrim */`](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/integration/redis/test_list.php#L47) 

[`tap_assert($redis->ltrim($key, 1, 0), 'delete list with ltrim');`](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/integration/redis/test_list.php#L195)

Here it could be something like this:
```
/*EXPECT
ok - should reject zero args
*/

@$result = newrelic_set_error_group_callback();
tap_refute($result, 'should reject zero args')
```

bduranleau-nr: Good catch- I'll add a tap-ok/not ok EXPECT to these tests when calling the API

lavarou: I think this approach should help with eliminating the need for separate treatment of error handling by different PHPs in a few tests (like test_error_group_callback_bad_params.php), and use of very verbose `/*EXPECT_REGEX`.

agent/php_newrelic.h
bduranleau-nr: I think the point about heap fragmentation is valid- I made the changes to implement this in 16ce1ffc2772b5a7d25c57ff7c1a246fc1e13be1.

tests/integration/api/error_group_callback/test_error_group_callback_empty_callback.php
bduranleau-nr: > I think this approach should help with eliminating the need for separate treatment of error handling by different PHPs in a few tests (like test_error_group_callback_bad_params.php), and use of very verbose /*EXPECT_REGEX.

I know the EXPECT_REGEX is pretty intense, but I really do believe we need it- without it, we don't really have the ability to test that we are passing the correct data (or any data at all) to the customer callback. WIthout it we would only be testing that the API was called successfully and *something* happened that generated an `error.group.name` Agent attribute.

bduranleau-nr: (unless you're only referring to removing the EXPECT_REGEX for the error output in `bad_callback` and `bad_callback.php7` - those can go. The others are testing the values we provide, so they should really remain IMO.)

lavarou: @bduranleau-nr You have far more context on this so it's best for you to decide if and which EXPECT_REGEX could be simplified. I'll go with your recommendation on this.

bduranleau-nr: in practice it looks like merging the bad_callback tests is not possible- for  php 8.0 it treats this as a fatal error and will not continue execution, including printing out the tap test result. This is due to `zend_parse_parameters` in the API code expecting an argument of a function type, which bad_callback intentionally does not provide. This differs from PHP 7, where it is treated as an exception and execution continues as normal.
I can remove the EXPECT_ERROR_EVENTS and EXPECT_METRICS from both of these files, but it looks like EXPECT_REGEX has to stay.

However, I did manage to add tap tests to all the other tests to verify the correct return value from the API call.

bduranleau-nr: Integration test improvements adding tap output made in af742dccca975e2841f3d47d61e30d2361f53134

lavarou (approved):
LGTM 👍 I really like those tests!!!

tests/integration/api/error_group_callback/test_error_group_callback_bad_callback.php7.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Tests newrelic_set_error_group_callback() API handling when passed a non-function object.
```
hahuja2, line 8:
```suggestion
Tests newrelic_set_error_group_callback() API handling when passed a non-function object for PHP versions less than 8.0.
```

hahuja2 (approved):
Great job Brian!

agent/php_api.c
```diff
@@ -1578,3 +1578,70 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
...
+
+  // Log debug message if the user is overwriting an existing callback
+  if (is_error_callback_set()) {
+    nrl_debug(
+        NRL_API,
+        "newrelic_set_error_group_callback: overwriting previous callback");
+  }
+
+  // Set global values
+  NRPRG(error_group_user_callback).fci = fci;
```
mfulb, line 1638: I feel like since there is an accessor function for `is_set` here that perhaps there ought to be a setter function for these global values since there was a partial effort to hide the implementation.

bduranleau-nr: The `is_set` function wasn't meant to hide the implementation, just to make it easier to check if the values were set. Since the `is_set` function is currently called 4 times (and could conceivably need to be checked again at some point/effort in the future), adding logic to make it easier to check these values made sense to me. 
The actual setting of these values is only ever done once in this function and I don't think that's liable to change, so to me a setter function for this individual case makes less sense.

agent/php_api.c
```diff
@@ -1578,3 +1578,70 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
...
+ *              of returning a string for the 'error.group.name' Agent
+ *              Attribute.
+ *
+ */
+#ifdef TAGS
+void zif_newrelic_set_error_group_callback(void); /* ctags landing pad only */
+void newrelic_set_error_group_callback(void);     /* ctags landing pad only */
+#endif
+PHP_FUNCTION(newrelic_set_error_group_callback) {
+  zend_fcall_info fci;
```
zsistla, line 1595: consider setting both values to `= {0}` for safety like:
```
zend_fcall_info fci = {0};
zend_fcall_info_cache fci_cache = {0};
```

here's an example of php-src using it: https://github.com/php/php-src/blob/master/sapi/embed/README.md

agent/php_api.c
```diff
@@ -1578,3 +1578,70 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
...
+  // Populate function data
+  if (FAILURE == zend_parse_parameters(ZEND_NUM_ARGS(), "f", &fci, &fcc)) {
+    nrl_warning(
+        NRL_API,
+        "newrelic_set_error_group_callback failure: invalid argument passed");
+    RETURN_FALSE;
+  }
+
+  // Verify the user callback accepts 2 arguments
+  if (2 != fcc.function_handler->common.num_args) {
```
zsistla, line 1622: Is function_handler guaranteed to be non-null?

agent/php_error.c
```diff
@@ -29,6 +31,98 @@ typedef struct _nr_php_exception_filter_t {
...
+  if (!nr_php_is_zval_non_empty_string(group_name_zv)) {
+    nrl_debug(NRL_MISC,
+              "Error Group callback: Invalid return value (Non-Empty "
+              "String Required).");
+    FREE_MEM;
+    return;
+  }
+
+  group_name_str
+      = nr_strndup(Z_STRVAL_P(group_name_zv), NR_ATTRIBUTE_VALUE_LENGTH_LIMIT);
```
zsistla, line 115: Why NR_ATTRIBUTE_VALUE_LENGTH_LIMIT vs Z_STRLEN_P(group_name_zv) ?
The concern here is that if group_name_zv is less than NR_ATTRIBUTE_VALUE_LENGTH_LIMIT, we could be copying junk over.

agent/php_error.c
```diff
@@ -497,6 +593,23 @@ void nr_php_error_cb(int type,
...
+    if (is_error_callback_set()) {
+#if ZEND_MODULE_API_NO < ZEND_8_1_X_API_NO
+      file = nr_strdup(error_filename);
+#else
+      file = nr_strdup(ZSTR_VAL(error_filename));
+#endif
+      klass = nr_strdup(errclass);
+
+      nr_php_error_call_error_group_callback(NRPRG(txn), klass, msg, file,
+                                             stack_json);
```
zsistla, line 608: Why not pass `errclass` in directly instead of duping it?

agent/php_error.c
```diff
@@ -497,6 +593,23 @@ void nr_php_error_cb(int type,
...
                         errclass, stack_json);
 
+    /*
+     * Error Fingerprinting Callback
+     */
+    if (is_error_callback_set()) {
+#if ZEND_MODULE_API_NO < ZEND_8_1_X_API_NO
+      file = nr_strdup(error_filename);
+#else
+      file = nr_strdup(ZSTR_VAL(error_filename));
```
zsistla, line 603: I know the strdup of zvals is not handled consistently, but it's probably safer to use 
`file = nr_strdup(ZSTR_VAL(error_filename), ZSTR_LEN(error_filename));`

agent/php_error.c
```diff
@@ -29,6 +31,98 @@ typedef struct _nr_php_exception_filter_t {
...
+                                                   char* klass,
+                                                   char* message,
+                                                   char* file,
+                                                   char* stack_json) {
+  zval* txn_arr = NULL;
+  zval* error_arr = NULL;
+  zval* group_name_zv = NULL;
+  zend_fcall_info fci;
+  zend_fcall_info_cache fcc;
+  nrobj_t* agent_attributes;
```
zsistla, line 64:
```suggestion
  nrobj_t* agent_attributes = NULL;
```

agent/php_error.c
```diff
@@ -29,6 +31,98 @@ typedef struct _nr_php_exception_filter_t {
...
+
+  txn_arr = nr_php_zval_alloc();
+  error_arr = nr_php_zval_alloc();
+
+  array_init(txn_arr);
+  array_init(error_arr);
+
+  nr_php_add_assoc_string(txn_arr, "request_uri", request_uri);
+  nr_php_add_assoc_string(txn_arr, "path", path);
+  nr_php_add_assoc_string(txn_arr, "method", method);
```
zsistla, line 93: Are these null safe functions?

agent/php_rinit.c
```diff
@@ -36,6 +36,7 @@ PHP_RINIT_FUNCTION(newrelic) {
   NRPRG(php_cur_stack_depth) = 0;
   NRPRG(deprecated_capture_request_parameters) = NRINI(capture_params);
   NRPRG(sapi_headers) = NULL;
+  NRPRG(error_group_user_callback).is_set = false;
```
zsistla, line 39: what about just set `NRPRG(error_group_user_callback) = {0};`

zsistla: does this ever need to be cleaned up in rshutdown? or it's pretty much there for the life of the agent?

ZNeumann (approved):

agent/php_api.c
bduranleau-nr: Yes, otherwise the [zpp call above] (https://github.com/newrelic/newrelic-php-agent/pull/719/files#diff-aaa6dfffdf27a5da3f0c89176d8aa9b62226676fae89c5c2c8e0e1a35db06b66R1614)would have failed and the function returned

agent/php_error.c
bduranleau-nr: Looking at the implementation of `nr_strndup`, it closely follows the c library implementation of `strndup`, where *at most* n bytes are copied. So there shouldn't be any junk data at risk of being copied over- it searches for the null terminator to the string and only copied over the number of characters up to that null terminator.

bduranleau-nr: The rationale for using `NR_ATTRIBUTE_VALUE_LENGTH_LIMIT` is because any string longer than 255 characters must be truncated.

agent/php_api.c
bduranleau-nr: Added in 1d3ea0d3da9ac486dc728a1ebd6dc9a6518f9d6f

agent/php_api.c
bduranleau-nr: Added in 1d3ea0d3da9ac486dc728a1ebd6dc9a6518f9d6f

agent/php_error.c
zsistla: zval string handling is special which is why they have the length attribute as well. They aren't guaranteed to be null terminated.
https://www.oreilly.com/library/view/programming-php/1565926102/ch14s06.html
```
Strings do not need to be null-terminated, but since most third-party libraries expect null-terminated strings it is a good idea to always null-terminate any string you create.
```
Could copy over the lesser of NR_ATTRIBUTE_VALUE_LENGTH_LIMIT or Z_STRLEN_P(group_name_zv).
something like 
```
nr_strndup(Z_STRVAL_P(group_name_zv), NR_ATTRIBUTE_VALUE_LENGTH_LIMIT < Z_STRLEN_P(group_name_zv) ? NR_ATTRIBUTE_VALUE_LENGTH_LIMIT : Z_STRLEN_P(group_name_zv));
```

agent/php_error.c
bduranleau-nr: Initially I wanted to do this, but `errclass` is defined as a `const char*` in this function, which causes the compiler to error out due to discarding of the `const` qualifier for `add_assoc_string`.

See this failed GHA build for the details: https://github.com/newrelic/newrelic-php-agent/actions/runs/5969170113/job/16194501104#step:7:262

agent/php_error.c
bduranleau-nr: Ah the joys of PHP - added in 01c0f7f8fb2c9b0bf85fc156677c714fd3ea82d9

agent/php_error.c
bduranleau-nr: Added in 1fe745c6faabb68875e513e71221b0eea8e49bc5

agent/php_error.c
bduranleau-nr: Since they are handled by `nr_strdup`, the value will never be `NULL` for these variables- if the value is null, `nr_strdup` returns an empty string, which is what would be used here.

agent/php_rinit.c
bduranleau-nr: See this conversation for context on how/why this was implemented this way: https://github.com/newrelic/newrelic-php-agent/pull/719#discussion_r1307556367

bduranleau-nr: In general since this isn't a pointer there shouldn't be any cleanup required in rshutdown- resetting the `is_set` to false here in rinit allows us to set and unset it on a per-request basis without the need for allocating or deallocating memory.

zsistla: by cleanup, I meant resetting to unset so we won't be calling one application's callbacks on another application's errors.

bduranleau-nr: is that a risk if we set it to false at the beginning of every rinit?

zsistla: Nope, should be good.

zsistla (approved):
Great job, thanks!

PR #723, opened 2023-08-30
feat: add newrelic_set_user_id API

bduranleau-nr:
Implement `newrelic_set_user_id` API functionality in the PHP Agent.

Allows an application to track a user via calling a new `newrelic_set_user_id` API, providing a UUID string as an argument.

If the API call is successful, an agent attribute `enduser.id` will be generated containing the provided UUID string.


agent/php_api.c
```diff
@@ -1578,3 +1578,60 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
     nr_free(span_id);
   }
 }
+
+/*
+ * Purpose      (New Relic API) (User Tracking)
+ *              Allows a caller to uniquely identify a user via a unique ID
+ *              string Agent Attribute
```
lavarou, line 1585:
```suggestion
 * Purpose      Allows a caller to add a user id string to agent attributes in 
                transaction event, transaction trace, error and span.
```

tests/integration/api/set_user_id/test_set_user_id.php
```diff
@@ -0,0 +1,72 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Tests newrelic_set_user_id() API
```
lavarou, line 8:
```suggestion
Tests newrelic_set_user_id() API:
 * Supportability/api/set_user_id metric is created
 * enduser.id agent attribute is present in span event
```

tests/integration/api/set_user_id/test_set_user_id_empty_string.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+newrelic_set_user_id() should fail when provided an empty uuid string
```
lavarou, line 8:
```suggestion
When provided an empty uuid string, newrelic_set_user_id() should return false and enduser.id agent attribute should not be present in span event's agent attributes.
```

lavarou (changes requested):
Looks good, however https://github.com/newrelic/newrelic-php-agent/pull/723/files#r1317385066 must be addressed. Also shouldn't there be tests testing if enduser.id is present in error events and error traces?

tests/integration/api/set_user_id/test_set_user_id.php
```diff
@@ -0,0 +1,72 @@
...
+    [{"name": "OtherTransactionTotalTime"},                                 [1, "??", "??", "??", "??", "??"]],
+    [{"name": "OtherTransactionTotalTime/php__FILE__"},                     [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Supportability/Logging/Forwarding/PHP/enabled"},             [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Supportability/Logging/Metrics/PHP/enabled"},                [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Supportability/api/set_user_id"},                            [1, 0, 0, 0, 0, 0]]
+  ]
+]
+*/
+
+/*EXPECT_SPAN_EVENTS
```
lavarou, line 34: Shouldn't EXPECT_TXN_TRACES and EXPECT_ANALYTICS_EVENTS be tested for presence of enduser.id attribute as well?

lavarou (commented):
Shouldn't there be tests testing if enduser.id is present in error events and error traces?

bduranleau-nr:
Error event integration tests added in 2c57cc30441c3874c7cf69574b2c00fa17fb18e0.

lavarou (approved):
👍

agent/php_api.c
```diff
@@ -1578,3 +1578,64 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
...
+void newrelic_set_user_id(void);     /* ctags landing pad only */
+#endif
+PHP_FUNCTION(newrelic_set_user_id) {
+  zval* uuid_zv = NULL;
+  char* uuid_str = NULL;
+
+  NR_UNUSED_RETURN_VALUE_PTR;
+  NR_UNUSED_THIS_PTR;
+  NR_UNUSED_RETURN_VALUE_USED;
+
```
zsistla, line 1598: We should probably bail out early if we aren't recording, like this: https://github.com/newrelic/newrelic-php-agent/blob/dev/agent/php_api.c#L650-L652

agent/php_api.c
```diff
@@ -1578,3 +1578,64 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
...
+    RETURN_FALSE;
+  }
+
+  if (NR_ATTRIBUTE_VALUE_LENGTH_LIMIT < Z_STRLEN_P(uuid_zv)) {
+    nrl_warning(NRL_API,
+                "newrelic_set_user_id_failure: invalid string length.");
+    RETURN_FALSE;
+  }
+
+  uuid_str = nr_strdup(Z_STRVAL_P(uuid_zv));
```
zsistla, line 1631: It's probably safer to call nr_strdup with the length explicitly passed in `str = nr_strndup(Z_STRVAL_P(uuid_zv), Z_STRLEN_P(uuid_zv));` like was done [here](https://github.com/newrelic/newrelic-php-agent/blob/dev/agent/php_api.c#L582)
since we can't depend on zval strings to be guaranteed to be null terminated.

references:
https://www.phpinternalsbook.com/php7/internal_types/strings/zend_strings.html#the-zend-string-api

ZNeumann (approved):

agent/php_api.c
bduranleau-nr: Wouldn't that case already be handled by checking that the transaction is not NULL on line [1601](https://github.com/newrelic/newrelic-php-agent/pull/723/files#diff-aaa6dfffdf27a5da3f0c89176d8aa9b62226676fae89c5c2c8e0e1a35db06b66R1601)?

bduranleau-nr: Or is it possible to create a transaction object but not record anything?

agent/php_api.c
bduranleau-nr: Added in 18dd7acfa2344abb5b9d538a432acd2d33948513

agent/php_api.c
bduranleau-nr: Ah I suppose someone could hypothetically place a call to `nr_txn_ignore()` as well as a call to this API (although I'm not sure why somebody would do that) so checking to see if the transaction is recording is technically necessary for cases like that.

bduranleau-nr: Added in e8f8ac25d51924c2173e529db729f387ed5353e5

hahuja2 (approved):

PR #724, opened 2023-09-01
fix(agent): Fixes 8T distributed traces

mfulb:
When using 8T the 'parentId' intrinsics attribute was not being set for
the root span of a service to a value based on the inbound header.  This caused
fragmented traces.  This commit added the required 'parentId' attribute.

Normal DT (not 8T) was working correctly because the 'parentId' was
being added to a span returned from nr_segement_to_span_event()
in a different code path dedicated to JSON creation starting with
nr_segment_tree_finalise().

A test in test_segment.c was modified to check that the 'parentId'
is correctly set based on the inbound DT header.


bduranleau-nr (approved):

lavarou (approved):

zsistla (approved):
Looks good, thanks for finding and fixing.

PR #725, opened 2023-09-01
refactor(agent): improve phpunit instrumentation

lavarou:
* [improve phpunit detection](https://github.com/newrelic/newrelic-php-agent/commit/d2b65d4bb66b958c7fe1681dc9205e1fb6210243) 
Use file that is loaded also on PHP 8.2 with include/require optimizations that
don't execute files without executable statements.

* [improve skipped tests handling](https://github.com/newrelic/newrelic-php-agent/commit/f6928af15178c4a1eacd8fe4f088cafccd25aa72) 
PHPUnit passes skipped tests to `addError` as well as `endTest` regardless of
the version. The difference is that in PHPUnit 5.x it is not possible to obtain
test outcome for skipped test - `$test.getStatus()` returns 'null'. Because of
that, `endTest` instrumentation bails out and does not generate test event and
therefore `addError` needs to be instrumented. It is possible to obtain test
outcome for skipped tests in PHPUnit 9.x. However, the `$time` for skipped test
is not received in `endTest` as a valid double and therefore it needs to be
'fixed'. Moreover testcase.getStatusMessage returns 'null' for tests skipped due
to dependency failure and therefore it is not possible to expect it in the test
event.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -491,8 +491,8 @@ static nr_library_table_t libraries[] = {
      * /usr/local/bin. While BaseTestRunner isn't the very first file to load,
      * it contains the test status constants and loads before tests can run.
      */
-    {"PHPUnit", "phpunit/src/runner/basetestrunner.php", nr_phpunit_enable},
-    {"PHPUnit", "phpunit/runner/basetestrunner.php", nr_phpunit_enable},
+    {"PHPUnit", "phpunit/src/framework/test.php", nr_phpunit_enable},
+    {"PHPUnit", "phpunit/framework/test.php", nr_phpunit_enable},
 
     {"Predis", "predis/src/client.php", nr_predis_enable},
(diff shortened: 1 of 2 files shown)


ZNeumann (approved):

agent/php_execute.c
```diff
@@ -491,8 +491,8 @@ static nr_library_table_t libraries[] = {
      * /usr/local/bin. While BaseTestRunner isn't the very first file to load,
```
zsistla, line 491: Should the comment be updated since we don't use BaseTestRunner anymore?

zsistla (approved):
Thanks for updating this.

agent/php_execute.c
lavarou: Good catch! Comment updated in bd240c915ee0bce31a3c0802eaa4af287cd10fff.

PR #727, opened 2023-09-08
fix(agent): Removes deprecation warnings for mysqli::real_connect()

mfulb:
For PHP >= 8.1 deprecation warnings were being generated by the agent
when parsing arguments to the mysqli::real_connect() function wrapper.
This is a combination of the argument type specification string
enabling checking for null values and PHP 8.1 and later outputing a
deprecation warning if a null is provided in this scenario.
The fix is to change the format specification string to the same as what
is used by mysqli::real_connect() for PHP >= 8.1.

### DIFF
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -855,7 +855,7 @@ NR_INNER_WRAPPER(mysqli_real_connect) {
    * uses to avoid creating this deprecation warning.
    * For older PHPs continue to use the same specification string as previously
    * to minimize any chances of introducing new problems.
-   */ 
+   */
 #if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
   if (FAILURE
       == zend_parse_parameters_ex(
@@ -895,7 +895,7 @@ NR_INNER_WRAPPER(mysqli_real_connect) {
     }
   }
 #endif /* PHP >= 8.1 */
- 
+
   zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
                                      INTERNAL_FUNCTION_PARAM_PASSTHRU);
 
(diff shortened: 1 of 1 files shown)


agent/php_internal_instrument.c
```diff
@@ -849,6 +849,33 @@ NR_INNER_WRAPPER(mysqli_real_connect) {
...
+   * a non-null argument value is expected.  For these PHP versions we use the
+   * same argument type specification as the mysqli::real_connect() extension
+   * uses to avoid creating this deprecation warning.
+   * For older PHPs continue to use the same specification string as previously
+   * to minimize any chances of introducing new problems.
+   */
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
+  if (FAILURE
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "O|s!s!s!s!l!s!l",
```
bduranleau-nr, line 862:
```suggestion
          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "o|s!s!s!s!l!s!l",
```
I think this is the cause of the test failures for 8.1 and 8.2.
According to the zpp api [doc](https://github.com/php/php-src/blob/master/docs/parameter-parsing-api.md),
- `O` is for an object of specific type given by class entry
- `o` is for a generic object (typical `zval *`)

bduranleau-nr: It's also the primary difference between this and the previous logic, which used `o`. I tested this locally and it fixed the issue.

mfulb:
Closing - was fixed in PR https://github.com/newrelic/newrelic-php-agent/pull/750

PR #730, opened 2023-09-11
feat: optional instrumentated function metric

ZNeumann:
There still exists a default wrapper, `nr_php_wrap_user_function`, which (for OAPI) creates an 'after' wraprec that is non-transient and creates an `InstrumentedFunction` metric. Before this PR, there was the option to create 'before', 'after', and 'clean' wrappers and define the transience.

This PR extends the options available when creating a wraprec to no longer create the InstrumentedFunction metric. It wraps this option with the other bool option (transience) into an options struct.

Additionally, if the top-most default, `nr_php_wrap_user_function`, is not used, every single option must be defined ('before', 'after', 'clean', transience, and instrumented metric). This takes the concept that if a single one of these fields is non-default, it is much more likely that other fields will also be non-default. The code writer will be forced to explicitly consider each option available. 

This does not add the option to create the InstrumentedFunction metric when calling `nr_wrap_callable`.

### DIFF
--- a/agent/php_user_instrument.h
+++ b/agent/php_user_instrument.h
@@ -189,7 +189,7 @@ extern nruserfn_t* nr_php_add_custom_tracer_callable(
     zend_function* func TSRMLS_DC);
 extern nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
                                                   size_t namestrlen,
-                                                  nr_wrap_user_function_options_t options
+                                                  const nr_wrap_user_function_options_t* options
                                                   TSRMLS_DC);
 extern void nr_php_reset_user_instrumentation(void);
 extern void nr_php_remove_transient_user_instrumentation(void);
--- a/agent/php_user_instrument.h
+++ b/agent/php_user_instrument.h
@@ -27,10 +27,22 @@ typedef struct nrspecialfn_return_t (*nrspecialfn_t)(
 
 typedef void (*nruserfn_declared_t)(TSRMLS_D);
 
+/* Options for wrapping a user function */
 typedef enum {
   NR_WRAPREC_NOT_TRANSIENT = 0,
   NR_WRAPREC_IS_TRANSIENT = 1
 } nr_transience_t;
+
+typedef enum {
+  NR_WRAPREC_CREATE_INSTRUMENTED_FUNCTION_METRIC = 0,
+  NR_WRAPREC_NO_INSTRUMENTED_FUNCTION_METRIC = 1
+} nr_instrumented_function_metric_t;
+
+typedef struct _nr_wrap_user_function_options_t {
+  nr_transience_t                   transience;
+  nr_instrumented_function_metric_t instrumented_function_metric;
+} nr_wrap_user_function_options_t;
+
 /*
  * An equivalent data structure for user functions.
  *
@@ -177,7 +189,8 @@ extern nruserfn_t* nr_php_add_custom_tracer_callable(
     zend_function* func TSRMLS_DC);
 extern nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
                                                   size_t namestrlen,
-                                                  nr_transience_t transience TSRMLS_DC);
+                                                  nr_wrap_user_function_options_t options
+                                                  TSRMLS_DC);
 extern void nr_php_reset_user_instrumentation(void);
 extern void nr_php_remove_transient_user_instrumentation(void);
 extern void nr_php_add_user_instrumentation(TSRMLS_D);
--- a/agent/php_wrapper.c
+++ b/agent/php_wrapper.c
@@ -56,13 +56,28 @@ static void nr_php_wraprec_add_before_after_clean_callbacks(
   wraprec->special_instrumentation_clean = clean_callback;
 }
 
+nruserfn_t* nr_php_wrap_user_function_before_after_clean(
+    const char* name,
+    size_t namelen,
+    nrspecialfn_t before_callback,
+    nrspecialfn_t after_callback,
+    nrspecialfn_t clean_callback) {
+  nr_wrap_user_function_options_t options = {
+      NR_WRAPREC_NOT_TRANSIENT,
+      NR_WRAPREC_CREATE_INSTRUMENTED_FUNCTION_METRIC
+  };
+  return nr_php_wrap_user_function_before_after_clean_with_options(
+          name, namelen, before_callback, after_callback,
+          clean_callback, &options);
+}
+
 nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_options(
     const char* name,
     size_t namelen,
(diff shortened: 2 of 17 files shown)


agent/php_wrapper.c
```diff
@@ -182,6 +186,11 @@ nruserfn_t* nr_php_wrap_generic_callable(zval* callable,
   zend_fcall_info_cache fcc;
   zend_fcall_info fci;
 
+  nr_wrap_user_function_options_t options = {
+    NR_WRAPREC_IS_TRANSIENT,
+    NR_WRAPREC_NO_INSTRUMENTED_FUNCTION_METRIC
+  };
```
lavarou, line 192: Could a note about these options be added to the doc string of this function? I.e.:
```
callable is wrapped with a transient wraprec and no instrumented function metric is being created for callables
```

agent/php_user_instrument.h
```diff
@@ -177,7 +189,8 @@ extern nruserfn_t* nr_php_add_custom_tracer_callable(
     zend_function* func TSRMLS_DC);
 extern nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
                                                   size_t namestrlen,
-                                                  nr_transience_t transience TSRMLS_DC);
+                                                  nr_wrap_user_function_options_t options
```
lavarou, line 192: Should it be passed by value or by reference? Also nr_php_add_custom_tracer_named does not modify options. If that's the case shouldn't this be indicated with a `const`?

agent/php_wrapper.h
```diff
@@ -154,11 +154,11 @@ extern nruserfn_t* nr_php_wrap_user_function(const char* name,
                                              size_t namelen,
                                              nrspecialfn_t callback TSRMLS_DC);
 
-extern nruserfn_t* nr_php_wrap_user_function_with_transience(
+extern nruserfn_t* nr_php_wrap_user_function_with_options(
     const char* name,
     size_t namelen,
     nrspecialfn_t callback,
-    nr_transience_t TSRMLS_DC);
+    nr_wrap_user_function_options_t TSRMLS_DC);
```
lavarou, line 161: Should `options` be passed by value or by reference? Also `nr_php_wrap_user_function_with_options` does not modify options. If that's the case shouldn't this be indicated with a const?

agent/php_wrapper.c
```diff
@@ -56,16 +56,16 @@ static void nr_php_wraprec_add_before_after_clean_callbacks(
   wraprec->special_instrumentation_clean = clean_callback;
 }
 
-nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
+nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_options(
```
lavarou, line 59: There exist:
* `nr_php_wrap_user_function` that creates wraprec with options valid for library and framework wraprecs: `NR_WRAPREC_NOT_TRANSIENT`, `NR_WRAPREC_CREATE_INSTRUMENTED_FUNCTION_METRIC` (the options are embedded in the `nr_php_wrap_user_function` implementation)
* `nr_php_wrap_generic_callable` that creates wraprec with options valid for callables: `NR_WRAPREC_IS_TRANSIENT`, `NR_WRAPREC_NO_INSTRUMENTED_FUNCTION_METRIC` (the options are embedded in the `nr_php_wrap_generic_callable`)
* `nr_php_wrap_user_function_with_options` that creates wraprec with arbitrary options which is only used by `nr_php_wrap_generic_callable` and `nr_php_wrap_user_function` which both have options set properly for their use cases.
* `nr_php_wrap_user_function_before_after_clean_with_options` that creates wraprec with arbitrary options - used for both library and framework wraprecs and callables and in each case the options had to be set 'by hand'.

However, all library and framework instrumentation needs wraprecs that are not transient and create supportability metrics (that's why `nr_php_wrap_user_function` has the options 'hardcoded'). Moreover only callables wraprecs need to be transient and not create supportability metrics (that's why `nr_php_wrap_generic_callable` has the options 'hardcoded'). Therefore, wouldn't it makes sense to have `nr_php_wrap_user_function_before_after_clean` that corresponds to `nr_php_wrap_user_function`, and has the options for wraprecs for library and framework instrumentation 'hardcoded'? It will indicate and highlight that wraprecs for library and framework instrumentation need to be non-transient and create supportability metrics, regardless if zend_execute hook or oapi is used to instrument user functions. Such API will help to avoid mistakes of using incorrect options for wraprecs when instrumenting library and framework user functions.

agent/php_wrapper.c
ZNeumann: Added in [449afaa](https://github.com/newrelic/newrelic-php-agent/pull/730/commits/449afaacda9a9d734919944b3742d90d348eb3d9)

agent/php_user_instrument.h
ZNeumann: fixed in [449afaa](https://github.com/newrelic/newrelic-php-agent/pull/730/commits/449afaacda9a9d734919944b3742d90d348eb3d9)

agent/php_wrapper.h
```diff
@@ -136,13 +136,13 @@
...
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
-extern nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_transience(
+extern nruserfn_t* nr_php_wrap_user_function_before_after_clean_with_options(
     const char* name,
     size_t namelen,
     nrspecialfn_t before_callback,
     nrspecialfn_t after_callback,
     nrspecialfn_t clean_callback,
-    nr_transience_t transience);
+    nr_wrap_user_function_options_t options);
```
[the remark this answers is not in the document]
ZNeumann: fixed in [449afaa](https://github.com/newrelic/newrelic-php-agent/pull/730/commits/449afaacda9a9d734919944b3742d90d348eb3d9)

agent/php_wrapper.h
ZNeumann: fixed in [449afaa](https://github.com/newrelic/newrelic-php-agent/pull/730/commits/449afaacda9a9d734919944b3742d90d348eb3d9)

agent/php_wrapper.c
ZNeumann: I prefer explicit coding where options/parameters are specified and clear at every call, even when redundant. However, I recognize that the existence of `nr_php_wrap_user_function` begs the existence of `nr_php_wrap_user_function_before_after_clean`. added in [449afaa](https://github.com/newrelic/newrelic-php-agent/pull/730/commits/449afaacda9a9d734919944b3742d90d348eb3d9)

lavarou: Thank you!

lavarou (approved):
Looks nice and clean!

zsistla (approved):
looks good, thanks.

bduranleau-nr (approved):

PR #731, opened 2023-09-11
feat: redo wordpress hooks instrumentation for OAPI

ZNeumann:
Bypass the need to check for cufa calls during do_action/filter by instrumenting add_action/filter and wrapping the desired calls at hook time, rather than call time. Substantially improves performance and stability.


tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php8.php
```diff
@@ -36,6 +39,7 @@
     [{"name": "DurationByCaller/Unknown/Unknown/Unknown/Unknown/allOther"}, [1, "??", "??", "??", "??", "??"]],
     [{"name": "Framework/WordPress/Hook/f"},                          [1, "??", "??", "??", "??", "??"]],
     [{"name": "Framework/WordPress/Hook/g"},                          [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Framework/WordPress/Hook/h"},                          [1, "??", "??", "??", "??", "??"]],
```
ZNeumann, line 42: Previously, Hook/h metric wasn't generated because we were using an 'after' wrapper and h threw an exception, preventing our wrapper from being called. Now, we wrap with a 'before' wrapper.

lavarou: Thanks for this detail! This helps a lot!

agent/fw_wordpress.c
```diff
@@ -735,6 +745,11 @@ void nr_wordpress_enable(TSRMLS_D) {
       nr_wordpress_handle_tag_stack_after, nr_wordpress_handle_tag_stack_clean,
       options);
 
+  nr_php_wrap_user_function(NR_PSTR("add_action"),
+                            nr_wordpress_add_action_filter);
+  nr_php_wrap_user_function(NR_PSTR("add_filter"),
+                            nr_wordpress_add_action_filter);
```
lavarou, line 751: `add_action` is a wrapper around `add_filter` (quote from WordPress code base):
```
function add_action($tag, $function_to_add, $priority = 10, $accepted_args = 1) {
	add_filter($tag, $function_to_add, $priority, $accepted_args);
}
```
Wouldn't instrumenting `add_filter` be enough?

lavarou: Are `add_action` and `add_filter` the only functions that register user callbacks for various WordPress hooks?

lavarou (commented):
Knowing that in WordPress, `add_action` is actually a wrapper around `add_filter` (this is true for all WordPress versions), the WordPress instrumentation could be limited to only instrumenting add_filter. This change will require reworking WordPress mock used for integration testing.

agent/fw_wordpress.c
ZNeumann: Good point! I have done a manual audit of the WP source code (and a quick internet search) and do believe that there is no other callback to use besides `add_filter`. I will remove `add_action` instrumentation and change the tests.

ZNeumann: fixed in [28480e6](https://github.com/newrelic/newrelic-php-agent/pull/731/commits/28480e616ba88f418f171de4e5a1de8a9116f915)

tests/integration/frameworks/wordpress/test_wordpress_do_action.php
```diff
@@ -73,4 +71,21 @@ function f() {
...
+add_action("hook", "g");
+add_action("hook", "h");
+/* 
+ * pre-OAPI: Initiates a non-flattened call stack of internal->user_code
+ * to ensure that cufa instrumentation properly handles skipping
+ * opline lookups of internal functions
+ *
+ * OAPI: Initiates a call to an added action outside
+ * the context of do_action, to ensure we only instrument
+ * with an active hook
```
lavarou, line 86: This is great documentation!

zsistla (approved):
Nicely done, thanks.

lavarou (approved):
Nice and clean!

PR #732, opened 2023-09-13
build(daemon): transition to go modules [WIP]

mfulb:
Prototype of builds using go sources using Go modules.
Currently vendor directories are probably not correctly handled.
Had to update flatbufferdata/data_1.20_test.go because expected output changed - needs investigation to see why/if this is different than builds of same sources using GOPATH.
Removed data_1.7_test.go and data_1.8_test.go as very unlikely to build against those golang versions and conditional compilation can't handle 3 different possibilities.
Need to compare test results (daemon, etc) to GOPATH build as well.


lavarou:
Obsoleted by #824.

PR #735, opened 2023-09-27
fix(agent): fix wrapping transient user functions

ZNeumann:
Prevent adding transient wraprecs for already wrapped transient functions by checking if wraprec already exists in the hashmap. Additionally prevent adding transient wraprecs for if the transient function is already wrapped as non-transient function by checking if wraprec already exists in the linked list.

For non-transient wrappers (standard instrumentation), the wrapper
is stored in both the hashmap and linked-list. For transient wrappers,
the wrapper is only stored in the hashmap. HOWEVER! non-transient
wrappers MAY only be in the linked-list. This normally happens if it
is wrapping a function that hasn't been loaded by PHP yet. Once the
function is loaded, there are many hooks to ensure the wrapper is also
added to the hashmap.

The implications of the above are: For non-transient wrappers, we only
need to check the linked-list to ensure that we are not duplicating a
wrapper. If a wrapper is only in the hashmap, it is transient and will
be overwritten by the non-transient wrapper.

For transient wrappers, however, we must check both the linked
list and the hashmap. This is because it is possible to attempt to
create a transient wrapper around an already non-transiently wrapped
function. This will return the non-transient wrapper and will attempt
to set the transient callbacks if that wrapper has those callbacks free.
This means that it is possible for callbacks intended to be transient
are attached onto a wrapper that isn't.

### DIFF
--- a/agent/php_user_instrument.c
+++ b/agent/php_user_instrument.c
@@ -501,7 +501,11 @@ nruserfn_t* nr_php_add_custom_tracer_named(const char* namestr,
     }
 
     if (NULL != orig_func) {
+#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO
+      p = nr_php_op_array_get_wraprec(&orig_func->op_array TSRMLS_CC);
+#else
       p = nr_php_wraprec_lookup_get(orig_func);
+#endif
 
       if (p) {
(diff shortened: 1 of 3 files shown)


lavarou (approved):
Great job figuring this one out!

zsistla (approved):
Thanks for finding/fixing this tricky case.
Nicely handled and excellent comments!

PR #729, opened 2023-09-11
feat: optimize oapi cufa detection

ZNeumann:

PR #746, opened 2023-10-16
fix(agent): Correctly handle laravel queues

zsistla:
When run via artisan queue:listen or artisan queue:work, Laravel queue jobs are generated as a separate background transaction for each job that is processed. These transactions are named with the class name of the job (or IlluminateQueueClosure if a closure was queued) and the connection type which is stored in the application's config/queue.php configuration file that includes connection configurations for each of the queue drivers that are included with the framework(sync--for dev env to execute immediately, database, [Amazon SQS](https://aws.amazon.com/sqs/), [Redis](https://redis.io/), and [Beanstalkd](https://beanstalkd.github.io/) drivers, and null --discards queued jobs).

For example, current implementation shows a job like:
`OtherTransaction\/Custom\/App\\Http\\Controllers\\JobOne (database)`
`OtherTransaction` shows it is a background job
`App\\Http\\Controllers\\JobOne` is the class name of the job
`database` is the connection type

This PR:
* Removed WorkCommand::handle wrapper 
* removed worker::process wrapper
* removed all non-supported Laravel 4.0 instrumentation.
* updated link
* removed references to "fire"
* removed unneeded job parsing code since the laravel Job class already provides the info
* Added handling to detect when a txn is ended within a wrapped function (that could result in segfaults when transactions were restarted (either directly through newrelic_start_transaction() or indirectly through newrelic_set_appname(), or with laravel queue instrumentation).

1)With laravel updates, there are better functions to instrument queue jobs that also result in fewer extraneous segments.
For OAPI/PHP 8+ and moving forward, we can use:
*   Illuminate\\Queue\\Worker::raiseBeforeJobEvent(string $connectionName, Job $job):void
*   Illuminate\\Queue\\SyncQueue::raiseBeforeJobEvent(Job $job):void
*   Illuminate\\Queue\\Worker::raiseAfterJobEvent(string $connectionName, Job $job):void
*   Illuminate\\Queue\\SyncQueue::raiseAfterJobEvent(Job $job):void

2)Using updated laravel functionality, we can additionally simplify the txn name decoding quite a bit and add additional information.
Current naming is: "job_name (connection_type)"

New naming is: "job_name (connection_type:job_queue)"

3) git diff got confused. 
For the purpose of reviews, despite git saying anything else, in fw_laravel_queue.c all the code above line 159 and the 
Block is the same code/functionality as currently exists.

Multiverse PR that tests both sync and async (database) queue workers is forthcoming.


agent/fw_laravel_queue.c
```diff
@@ -624,8 +579,52 @@ NR_PHP_WRAPPER(nr_laravel_queue_workcommand_fire) {
...
-   * the transaction after WorkCommand::fire() has finished executing, at which
-   * point no more jobs can be run.
+   * the transaction after WorkCommand::handle() has finished executing, at
+   * which point no more jobs can be run.
+   */
+
+  /*
+   * Start listening for jobs.
+   */
+  nr_php_wrap_user_function_before_after_clean(
```
ZNeumann, line 589: Is there a reason why we are setting this wrapper here, instead of in `nr_laravel_queue_enable`? This creates a non-transient wrapper, so by next request, any context gained by not setting it in `nr_laravel_queue_enable` is almost certainly lost.

If the goal is to not instrument `Illuminate\Queue\Worker::process` unless within a call to `Illuminate\Queue\Console\WorkCommand::handle`, perhaps a global boolean that holds that context would be better? (similar to the check_cufa added here https://github.com/newrelic/newrelic-php-agent/pull/729/files)

agent/fw_laravel_queue.c
```diff
@@ -624,8 +579,52 @@ NR_PHP_WRAPPER(nr_laravel_queue_workcommand_fire) {
...
+
+NR_PHP_WRAPPER(nr_laravel_queue_workcommand_handle_after) {
+  NR_UNUSED_SPECIALFN;
+  (void)wraprec;
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_LARAVEL);
+  /*
+   * Stop recording the transaction and throw it away.
+   */
+  nr_php_txn_end(1, 0 TSRMLS_CC);
```
ZNeumann, line 606: Do we not need to then start a new transaction here?

agent/fw_laravel_queue.c
zsistla: No need to rewrap the non-transient wrapper.
We can wrap both at once, since they are both used together; it doesn't make sense to have one without the other.`WorkCommand::handle` is encountered, `Worker::Process` will be in the calling chain for async queues.

Please see: [25c1652](https://github.com/newrelic/newrelic-php-agent/pull/746/commits/25c16523602c113c894d568f45902bd028337d94)

agent/fw_laravel_queue.c
zsistla: No, we want to record individual transactions for each job that is executed, but don't want to record a transaction for the actual queue:work command, since it spends most of its time sleeping.  Calling `nr_php_txn_end` here ensures that we ignore the transaction after WorkCommand::handle() has finished executing, at which point no more jobs can be run.  All of the job and txn naming is handled in the `Worker::process` wrapper which has already completed by the time it reaches this point.

zsistla: WorkCommand::handle was removed entirely for PHP 8+. (but tests before removing showed that it did indeed need to restart the txn)

agent/fw_laravel_queue.c
zsistla: Both `WorkCommand::handle` and `Worker::Process` are no longer used for PHP8+.

agent/fw_laravel_queue.c
```diff
@@ -40,7 +39,324 @@ static int nr_laravel_queue_is_sync_job(zval* job TSRMLS_DC) {
...
+    nr_header_set_synthetics_txn(NRPRG(txn), headers.synthetics);
+  }
+
+  if (headers.dt_payload || headers.traceparent) {
+    char* dt_payload = nr_strdup(headers.dt_payload);
+    char* tracestate = nr_strdup(headers.tracestate);
+    char* traceparent = nr_strdup(headers.traceparent);
+
+    nr_hashmap_t* header_map = nr_header_create_distributed_trace_map(
+        dt_payload, traceparent, tracestate);
```
ZNeumann, line 144: nr_header_create_distributed_trace_map already calls a [strdup](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_header.c#L42), so we shouldn't need to do it here.

ZNeumann (approved):

lavarou (approved):

agent/fw_laravel_queue.c
zsistla: Good call.
The reason to stdrup would be if we are storing a zval value that might go away, but that doesn't seem to be the case here.  I followed the preceeding and following calls and it does look like we strdup it right away, so we can claw back a little bit of performance in the dupping and the freeing.
Also updated the struct init at the top of the function to null ALL the values....

Please see:
[75e61fa](https://github.com/newrelic/newrelic-php-agent/pull/746/commits/75e61fafdf3bd689d3b87b21b3b06632f5311e27)

PR #748, opened 2023-10-18
tests: fix user exception and error handler implementation

lavarou:
* [fix user exception handler implementation](https://github.com/newrelic/newrelic-php-agent/pull/748/commits/571e8ac855cda1e1b657d0e0fcc6ab360cf7acbd)
User exception handler, set with [`set_exception_handler`](https://www.php.net/manual/en/function.set-exception-handler) is a callable that
needs to accept one parameter, which will be the Throwable object that was
thrown. If that callable is not defined according to these rules, PHP's
behavior, and therefore agent's behavior, is undefined. E.g. in PHPs 8.0, 8.1
and 8.2, when opcache is disabled Zend Engine gives the agent a zval which is
an object of Throwable type. The same thing happens in PHPs 8.0 and 8.1 when
opcache is enabled. However, when opcache is enabled in PHP 8.2, Zend Engine
no longer gives the agent a zval which is an object of Throwable type.

* [fix user error handler implementation](https://github.com/newrelic/newrelic-php-agent/pull/748/commits/e35ecff370fbb9dafa68c8c70b4d20c1a3981b0b)
User error handler, set with `set_error_handler` is a callable that needs to
accept at least two parameters: int errno and string errstr set to error level
and error message respectively. If that callable is not defined according to
these rules, PHP's behavior, and therefore agent's behavior, is undefined.

* [split integration test](https://github.com/newrelic/newrelic-php-agent/pull/748/commits/1b5e93378d51229ed6e66235f527d8356136ae25) 
An agent hooking into Zend Engine using zend_execute (pre-oapi) hook behaves
differently than the agent hooking into Zend Engine using observer API (oapi).
Namely pre-oapi agent does not generate error events when an exception is
handled with user exception handler wheras oapi agent generates error events
in such case. Therefore the test testing agent's behavior when an exception is
handled by user exception handler must be split depending on what method is used
to hook into Zend Engine: zend_execute hook is used for PHPs < 8.0 and observer
API is used for PHPs >= 8.0.


zsistla:
When the tests/integration/span test gets updated/split, please also update [this](https://github.com/newrelic/newrelic-php-agent/blob/oapi/tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php5.php) test which  says it's for php5, but there's no skipif and it is triggering on PHP 8 as well.

tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php5.php
```diff
@@ -9,6 +9,13 @@
 when an exception is handled by the exception handler.
```
zsistla, line 9: Actually, do we need this file anymore, can we delete it? The name is misleading because it says php5, but, as written, will run on PHP 7 as well.
It also looks like a duplicate of [tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php7.php](https://github.com/newrelic/newrelic-php-agent/pull/748/files#diff-6346f826d98c0bba03ab77c273fc53aaa637f342e7c865818a19afda58a1fefc)

lavarou: You're right! Support for PHPs 5.5 and 5.6 ended with [10.11.0.3](https://github.com/newrelic/newrelic-php-agent/releases/tag/v10.11.0.3) release. I'll remove it.
P.S. It is not duplicate of the `php7` version of this test, because `php7` version checks for code level metrics attributes.

lavarou: `php5` version of the test removed in 91aa2e29a7699960f9e36df096e152e31c14f3e1.

tests/integration/span_events/test_span_events_are_created_upon_caught_exception.php
```diff
@@ -6,13 +6,15 @@
 
 /*DESCRIPTION
```
zsistla, line 7: Will the test name be updated to `tests/integration/span_events/test_span_events_are_created_upon_uncaught_handled_exception.php`
in this or another PR?

lavarou: Initially we discussed that refactoring renaming the tests will be its own PR. However, since then, test renames in `jit` subtree became part of #715. Therefore, I think it makes sense for this PR to complement #715, and have test renames in this PR as well.

lavarou: Files renamed in 13fd3ffd7dff3075a00958387c08d6340c2361a1.

ZNeumann (approved):

PR #749, opened 2023-10-18
feat(agent): Adds log attribute capture [DO NOT REVIEW]

mfulb:
- Adds a log destination for attributes.  Log attributes are created on log events for context data passed to logging library.
- Adds code to capture the context data for the Monolog library and convert it to attributes on the log event.
- Adds a unit test to exercise conversion of context data to attributes as well as the log destination for attributes.
- Note this commit does not contain code to convert log attributes to JSON for sending to the daemon/collector.

### DIFF
--- a/agent/lib_monolog.c
+++ b/agent/lib_monolog.c
@@ -30,6 +30,11 @@
 #define LOG_DECORATE_PROC_FUNC_NAME \
   "newrelic_phpagent_monolog_decorating_processor"
 
+/*
+ * Prefix for log context attributes 
+ */
+#define NR_LOG_CONTEXT_DATA_ATTRIBUTE_PREFIX "context."
+
 // clang-format off
 /*
  * This macro affects how instrumentation $context argument of
@@ -453,7 +458,7 @@ nr_attributes_t* nr_monolog_convert_context_data_to_attributes(
     nrobj_t* obj = nr_monolog_context_data_zval_to_attribute_obj(val);
 
     if (NULL != obj) {
-      char* buf = nr_formatf(NR_TXN_LOG_CONTEXT_DATA_ATTRIBUTE_PREFIX "%s",
+      char* buf = nr_formatf(NR_LOG_CONTEXT_DATA_ATTRIBUTE_PREFIX "%s",
                              ZSTR_VAL(key));
       st = nr_attributes_user_add(attributes, NR_ATTRIBUTE_DESTINATION_LOG, buf,
                                   obj);
--- a/agent/lib_monolog.c
+++ b/agent/lib_monolog.c
@@ -441,15 +441,13 @@ nr_attributes_t* nr_monolog_convert_context_data_to_attributes(
   nr_attributes_t* attributes = NULL;
   nr_status_t st;
 
-  if (NULL == context_data) {
+  if (NULL == context_data || !nr_php_is_zval_valid_array(context_data)) {
     return NULL;
   }
 
   attributes = nr_attributes_create(NRPRG(txn)->attribute_config);
 
   ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARR_P(context_data), key, val) {
-    (void)key;
-    (void)val;
 
     if (NULL == key) {
       continue;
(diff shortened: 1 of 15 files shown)


agent/lib_monolog.c
```diff
@@ -306,6 +314,165 @@ static char* nr_monolog_get_context(const size_t argc,
...
+      break;
+
+    case IS_FALSE:
+      retobj = nro_new_boolean(false);
+      break;
+#else
+    case IS_BOOL:
+      retobj = nro_new_boolean(Z_BVAL_P(z));
+      break;
+#endif /* PHP 7.0+ */
```
bduranleau-nr, line 364: Since we don't support any versions lower than 7.0, I think we can remove the need for this CC macro and just keep the `IS_TRUE`/`IS_FALSE` cases

axiom/nr_attributes.c
```diff
@@ -97,6 +97,15 @@ void nr_attribute_config_disable_destinations(nr_attribute_config_t* config,
   config->disabled_destinations |= disabled_destinations;
 }
 
+void nr_attribute_config_enable_destinations(nr_attribute_config_t* config,
+                                             uint32_t disabled_destinations) {
```
ZNeumann, line 101: This argument name is confusing to me. Shouldn't it be called enabled_destinations?

ZNeumann (approved):

agent/lib_monolog.c
```diff
@@ -27,6 +30,11 @@
 #define LOG_DECORATE_PROC_FUNC_NAME \
   "newrelic_phpagent_monolog_decorating_processor"
 
+/*
+ * Prefix for log context attributes 
+ */
+#define NR_LOG_CONTEXT_DATA_ATTRIBUTE_PREFIX "context."
```
bduranleau-nr, line 36: It looks like this macro is only used once [here](https://github.com/newrelic/newrelic-php-agent/pull/749/files#diff-d48bf43b4158fad631ce65b0464841310be10cff57ac0b937061fb9d6c33bdc3R459). Do you think it will be reused again in the future? If not, I would consider declaring it as a `const char *` instead directly in that function so the value of the variable is visible without having to scroll to the top of the file.

agent/tests/test_monolog.c
```diff
@@ -0,0 +1,333 @@
...
+                    attributes);
+  nr_php_zval_free(&z);
+
+  tlib_php_request_end();
+}
+
+void test_main(void* p NRUNUSED) {
+#if defined(ZTS) && !defined(PHP7)
+  void*** tsrm_ls = NULL;
+#endif /* ZTS && !PHP7 */
```
bduranleau-nr, line 324: Should be able to remove this since ZTS & PHP<7 are not supported

axiom/nr_attributes.c
mfulb: That would make it clearer - thanks for the suggestion.

agent/lib_monolog.c
mfulb: I would prefer keeping it as a define and not hard code it into a local variable.  I could not rule out its use in other parts of the code at this point.

axiom/nr_log_event.c
```diff
@@ -53,7 +54,8 @@ static bool add_log_field_to_buf(nrbuf_t* buf,
                                  const char* field_name,
                                  const char* field_value,
                                  const bool first,
-                                 const bool required) {
+                                 const bool required,
+                                 const bool quoted) {
```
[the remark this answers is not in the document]
mfulb: Fixed in [045fad6](https://github.com/newrelic/newrelic-php-agent/commit/045fad6b03dda513fec0fcc4883a53458308fa00).

agent/lib_monolog.c
mfulb: Fixed in commit [48db58c](https://github.com/newrelic/newrelic-php-agent/commit/48db58c9193277c088c67e6a07b8b698560637a8).

agent/tests/test_monolog.c
mfulb: Fixed in commit [35975ba](https://github.com/newrelic/newrelic-php-agent/commit/35975baa425ae10fd6fec82a429d76c5c8f24685).

mfulb:
Closing due to replace [PR 753](https://github.com/newrelic/newrelic-php-agent/pull/753).

PR #750, opened 2023-10-18
fix(agent): Resolve mysqli deprecation warnings on PHP 8.1+

bduranleau-nr:
Certain wrapped mysqli functions allow null values to be passed for optional values on PHP 8.1+. 
This PR fixes the issue of the agent not properly expecting these null values in zpp resulting in deprecation warnings being thrown.

### DIFF
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -645,19 +645,20 @@ NR_INNER_WRAPPER(mysqli_construct) {
 
 #if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
   bool port_is_null = 1;
-  
+  const char *type_spec = "|s!s!s!s!l!s!";
   if (FAILURE
       == zend_parse_parameters_ex(
-          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "|s!s!s!s!l!s!", &host,
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, type_spec, &host,
           &host_len, &username, &username_len, &password, &password_len,
           &database, &database_len, &port, &port_is_null, &socket, &socket_len)) {
     nr_wrapper->oldhandler(INTERNAL_FUNCTION_PARAM_PASSTHRU);
     return;
   }
 #else
+  const char *type_spec =  "|ssssls";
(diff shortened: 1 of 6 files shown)


agent/php_internal_instrument.c
```diff
@@ -849,6 +879,33 @@ NR_INNER_WRAPPER(mysqli_real_connect) {
...
+   * a non-null argument value is expected.  For these PHP versions we use the
+   * same argument type specification as the mysqli::real_connect() extension
+   * uses to avoid creating this deprecation warning.
+   * For older PHPs continue to use the same specification string as previously
+   * to minimize any chances of introducing new problems.
+   */
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
+  if (FAILURE
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "o|s!s!s!s!l!s!l",
```
ZNeumann, line 892: I think, in all cases where we are changing the specification string, I would prefer a smaller ifdef- perhaps around a local variable. As it is, I expect some other differences in these blocks besides the sting. Something like this?
```
#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
char string_spec[] = "o|s!s!s!s!l!s!l";
#else
char string_spec[] = "|s!s!s!s!l!s!l";
#endif
```

ZNeumann:
Do we have a test that was previously showing these warnings? If not, can we add a test?

agent/php_internal_instrument.c
bduranleau-nr: I could do this for certain cases, but for cases where we have to include `port_is_null` conditionally for 8.1+, (like this one) I'll have to duplicate the zpp calls anyway to optionally include that arg.

bduranleau-nr: See the changes in 87ae53c5252fdd79b0eec65646ab893678e16063 that implement this

bduranleau-nr:
> Do we have a test that was previously showing these warnings? If not, can we add a test?

@ZNeumann I added some integration tests exercising this specific scenario in d66f7b2bc61b93a0e6462621aceeecde43db15e2

ZNeumann (approved):

mfulb (approved):
Nice Brian thanks for finishing this work.

PR #752, opened 2023-10-22
feat(tests): Enable opcache by default for the integration runner.

zsistla:
* Enable the opcache.so extension by default in the integration_runner
* Ensure the opcache INI settings in the integration_runner environment are enabled (these can be overwritten in the test case INI section if needing to test without opcache)
* Added few tests to ensure compatibility with opcache disabled and to demonstrate how to overwrite the value.

### DIFF
--- a/src/integration_runner/main.go
+++ b/src/integration_runner/main.go
@@ -347,14 +347,14 @@ func main() {
 		ctx.Settings["newrelic.loglevel"] = *flagLoglevel
 	}
 
-    if false == *flagOpcacheOff {
-        // PHP Modules common to all tests
-        ctx.Settings["zend_extension"] = "opcache.so"
+	if false == *flagOpcacheOff {
+		// PHP Modules common to all tests
+		ctx.Settings["zend_extension"] = "opcache.so"
 
-        // PHP INI values common to all tests
-        // These settings can be overwritten by adding new values to the INI block
-        ctx.Settings["opcache.enable"] = "1"
-        ctx.Settings["opcache.enable_cli"] = "1"
+		// PHP INI values common to all tests
+		// These settings can be overwritten by adding new values to the INI block
+		ctx.Settings["opcache.enable"] = "1"
+		ctx.Settings["opcache.enable_cli"] = "1"
 	}
 
 	// If the user provided a custom agent extension, use it.
--- a/src/integration_runner/main.go
+++ b/src/integration_runner/main.go
@@ -50,6 +50,7 @@ var (
 	flagTime            = flag.Bool("time", false, "time each test")
 	flagMaxCustomEvents = flag.Int("max_custom_events", 30000, "value for newrelic.custom_events.max_samples_stored")
 	flagWarnIsFail      = flag.Bool("warnisfail", false, "warn result is treated as a fail")
+	flagOpcacheOff      = flag.Bool("opcacheoff", false, "run without opcache. Some tests are intended to fail when run this way")
 
 	// externalPort is the port on which we start a server to handle
 	// external calls.
@@ -346,6 +347,7 @@ func main() {
 		ctx.Settings["newrelic.loglevel"] = *flagLoglevel
 	}
 
+    if false == *flagOpcacheOff {
         // PHP Modules common to all tests
         ctx.Settings["zend_extension"] = "opcache.so"
 
@@ -353,6 +355,7 @@ func main() {
         // These settings can be overwritten by adding new values to the INI block
         ctx.Settings["opcache.enable"] = "1"
         ctx.Settings["opcache.enable_cli"] = "1"
+	}
 
 	// If the user provided a custom agent extension, use it.
 	if len(*flagAgent) > 0 {
--- a/src/integration_runner/main.go
+++ b/src/integration_runner/main.go
@@ -346,6 +346,14 @@ func main() {
 		ctx.Settings["newrelic.loglevel"] = *flagLoglevel
 	}
 
+        // PHP Modules common to all tests
+        ctx.Settings["zend_extension"] = "opcache.so"
+
+        // PHP INI values common to all tests
+        // These settings can be overwritten by adding new values to the INI block
+        ctx.Settings["opcache.enable"] = "1"
+        ctx.Settings["opcache.enable_cli"] = "1"
+
 	// If the user provided a custom agent extension, use it.
 	if len(*flagAgent) > 0 {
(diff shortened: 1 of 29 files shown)


tests/integration/opcache/disabled/test_computations.php
```diff
@@ -0,0 +1,78 @@
...
+error_reporting = E_ALL
+opcache.enable=0
+opcache.enable_cli=0
+opcache.file_update_protection=0
+opcache.jit_buffer_size=32M
+opcache.jit=function
+*/
+
+/*PHPMODULES
+zend_extension=opcache.so
```
mfulb, line 30: Does this still need to be in tests if the default is for opcache to be enabled?

zsistla: Good point, please see: [5061df7](https://github.com/newrelic/newrelic-php-agent/pull/752/commits/5061df70afadd56e54d08258142317996ae606af)

src/integration_runner/main.go
```diff
@@ -346,6 +346,14 @@ func main() {
 		ctx.Settings["newrelic.loglevel"] = *flagLoglevel
 	}
 
+        // PHP Modules common to all tests
+        ctx.Settings["zend_extension"] = "opcache.so"
```
lavarou, line 350: What happens if `php` used by integration runner is not able to load the extension?

tests/integration/opcache/disabled/skipif.inc
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+/* Opcache only supported with PHP 7+ */
+if (version_compare(PHP_VERSION, "7.1", "<")) {
+  die("skip: PHP < 7.1 not supported\n");
+}
+
+/* Test should not run with opache */
+if (extension_loaded('Zend OPcache')) {
+  die("warn: Zend OPcache extension required\n");
```
ZNeumann, line 14:
```suggestion
  die("warn: Zend OPcache extension required to be not loaded\n");
```

src/integration_runner/main.go
zsistla: Good point, currently the tests just run even if it's not loaded unless a test specifically required it (like [here](https://github.com/newrelic/newrelic-php-agent/blob/ads/oapi-jit-tests/tests/integration/jit/function/skipif.inc#L12)).

We could do a few options:
1) detect and warn in log with something along the lines of "detected the default module X is missing. Tests will run, but results not guaranteed" but run tests anyway.
2) terminate with warning message "detected the default module X is missing. No tests will run." and do not run any tests.
3) A flag to toggle between option 1 and option 2

lavarou: My vote goes for 3. with the default being behavior described in 2.

tests/integration/opcache/disabled/skipif.inc
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+/* Opcache only supported with PHP 7+ */
+if (version_compare(PHP_VERSION, "7.1", "<")) {
+  die("skip: PHP < 7.1 not supported\n");
+}
+
+/* Test should not run with opache */
+if (extension_loaded('Zend OPcache')) {
+  die("warn: Zend OPcache extension required to be not loaded\n");
```
lavarou, line 14: So if integration_runner always loads opcache, won't extension_loaded('Zend OPcache') always return true and therefore none of the tests with this check will run?

lavarou: I don't think this check is needed at all - tests should run regardless if opcache is loaded or not. If it's loaded, it will be disabled with:
```
/*INI
opcache.enable=0
opcache.enable_cli=0
```
If it is not loaded, then opcache will be disabled as well.

src/integration_runner/main.go
ZNeumann: As far as I can tell, doing 2 would be quite involved. My understanding is that the integration runner is unable to know whether or not the module has been loaded until evaluating within the PHP `skipif`s. To implement this would then require an injection of the module check into the skips (which could be handled automatically), but then NOT injecting that check for tests that do not have opcache (like those added in this PR).

For simplicity's sake, I would rather avoid that altogether and just have a basic test verify that opcache was loaded. Then that singular test will fail if the extension fails to load. Then, if someone really wanted all the tests to pass without opcache, they could manually do something about that singular failure. I do not see a need to do anything more complicated.

tests/integration/opcache/disabled/skipif.inc
ZNeumann: I think that I would rather this check be a failure case within each test here. The tests expect opcache to be disabled and set the INI values as such. Therefore, failing this check should fail the test, not skip it. @lavarou do you agree?

src/integration_runner/main.go
ZNeumann: Added in [f7687f5](https://github.com/newrelic/newrelic-php-agent/pull/752/commits/f7687f561e5c57c30ef8cd5f23d8bc856665301a)

tests/integration/opcache/disabled/skipif.inc
ZNeumann: changed in [367b613](https://github.com/newrelic/newrelic-php-agent/pull/752/commits/367b6137b52604294ec32ff8df8f3fc06818d2f6)

src/integration_runner/main.go
lavarou: go fmt?

tests/integration/opcache/disabled/opcache_test.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* Test should not run with opache */
+if (extension_loaded('Zend OPcache')) {
+  if (false != opcache_get_status()) {
+    die("fail: Zend OPcache extension required to be not loaded\n");
```
lavarou, line 10:
```suggestion
    die("fail: OPcache enabled but needs to be disabled\n");
```

src/integration_runner/main.go
ZNeumann: fixed in [4f3f31e](https://github.com/newrelic/newrelic-php-agent/pull/752/commits/4f3f31e0ab3385aa3352329e40cf17c93f03a126)

lavarou (approved):
Looks good to me!

Issue #741, opened 2023-10-04
How to get monolog `$context`?


jamesmehorter:
## Summary

When we log a message with context data, that context data is not present on the log entry in New Relic.

We use Laravel, e.g. `Log::info('a log message', [ 'some' => 'helpful debugging data' ]);`. Under-the-hood, this is passed to monolog as `addRecord(6, 'a log message', [ 'some' => 'helpful debugging data' ]);`

See
* https://github.com/Seldaek/monolog/blob/main/src/Monolog/Logger.php#L320-L409
* https://laravel.com/docs/10.x/logging#contextual-information

However, if we send both the message and context as a JSON string in the message, the context is parsed by New Relic and log attributes are created. E.g. `Log::info(json_encode(['message' => 'a log message', 'some' => 'helpful debugging data' ]));`

## Desired Behavior
Context data supplied with a log entry is available as attributes in New Relic without the need of rewriting all `Log::` entries in our codebase.

jamesmehorter:
I found `HAVE_CONTET_IN_MESSAGE` here https://github.com/newrelic/newrelic-php-agent/blob/main/agent/lib_monolog.c#L158 and am curious if this is related? 

It's curious because, it seems that bool is hardcoded to `0` (disabled?). https://github.com/newrelic/newrelic-php-agent/blob/main/agent/lib_monolog.c#L31-L41

Huluti:
Should be fixed with https://github.com/newrelic/newrelic-php-agent/pull/753?

Huluti:
On my side with Symfony with automatic log forwadding I have just unuseful logs

![image](https://github.com/newrelic/newrelic-php-agent/assets/5782198/d55edbed-7c0c-4850-821a-6d5f9edfffba)

Huluti:
Thanks for your answer! But why prefer using an external tool if New Relic has its own log forwarder? Placeholder is a native monolog feature, so it would make sense to support it no?

jamesmehorter:
@Huluti I wonder if your comment is more appropriate on https://github.com/newrelic/newrelic-php-agent/issues/742? I.e. Does monolog us a processor to perform the find/replace of placeholders?

tsantos84:
> @Huluti I wonder if your comment is more appropriate on #742? I.e. Does monolog use a processor to perform the find/replace of placeholders?

[...]

PR #751, opened 2023-10-19
feat(agent): Adds JSON for log attributes

mfulb:
Adds json conversion for log attributes associated with log events.

Unit tests added to test log event JSON conversion function and conversion of log events to JSON into the flatbuffer used to communicate with the daemon.

Integration tests added for Monolog2 and Monolog3 to exercise:
 - simple addition of context data
 - filtering rules (include/exclude) under various scenarios
 - precedence of global and specific enable INI settings
 - attributes are not sent if HSM is enabled

PR #753, opened 2023-10-23
feat(agent): Adds log context attributes capture

mfulb:
Note: Replaces PR [749](https://github.com/newrelic/newrelic-php-agent/pull/749) and PR [751](https://github.com/newrelic/newrelic-php-agent/pull/751).

This work adds the capture of Monolog context data as log event attributes.  

Summary of work:
- Adds a log destination for attributes. Log attributes are created on log events for context data passed to logging library.
- Adds code to capture the context data for the Monolog library and convert it to attributes on the log event.
- Adds json conversion for log attributes associated with log events.
- Existing attribute filtering mechanism extended to implement the behavior of "include" rules for the log attribute destination, which differs from existing attribute destinations in that attribute names not included in "include" rules are dropped for the log destination.

Unit Tests Added:
- Tests to exercise Monolog context data capture and conversion to attributes
- Tests added to test log event JSON conversion function and conversion of log events to JSON into the flatbuffer used to communicate with the daemon.
- Tests added to exercise conversion of context data to attributes as well as the log destination for attributes.
- Log attribute destination added to filtering tests

Integration tests added for Monolog2 and Monolog3 to exercise:
 - simple addition of context data
 - filtering rules (include/exclude) under various scenarios
 - precedence of global and specific enable INI settings
 - attributes are not sent if HSM is enabled
 
Integration tests added for attributes in general:
 - exercise a variety of scenario to verify no regression from previous behavior


agent/lib_monolog.c
```diff
@@ -306,6 +309,150 @@ static char* nr_monolog_get_context(const size_t argc,
...
+      retobj = NULL;
+      break;
+  }
+
+  return retobj;
+}
+
+/*
+ * Purpose : Extract $context argument of Monolog\Logger::addRecord to
+ * attributes
```
lavarou, line 374: The function returns a pointer to `zval` with the value of `$context` argument. It does not deal with attributes.
```suggestion
 * Purpose : Get $context argument of Monolog\Logger::addRecord as `zval *`.
```

agent/lib_monolog.c
```diff
@@ -306,6 +309,150 @@ static char* nr_monolog_get_context(const size_t argc,
...
+ * Params  : # of Monolog\Logger::addRecord arguments, and
+ * Monolog\Logger::addRecord argument list
+ *
+ * Returns : zval* for context array on success (must be freed by caller)
+ *           NULL otherwise
+ *
+ */
+static zval* nr_monolog_extract_context_data(const size_t argc,
+                                             NR_EXECUTE_PROTO TSRMLS_DC) {
+  zval* context_arg;
```
lavarou, line 385:
```suggestion
  zval* context_arg = NULL;
```

agent/lib_monolog.c
```diff
@@ -306,6 +309,150 @@ static char* nr_monolog_get_context(const size_t argc,
...
+      if (!nr_php_is_zval_valid_string(z)) {
+        retobj = NULL;
+      } else {
+        retobj = nro_new_string(Z_STRVAL_P(z));
+      }
+      break;
+
+    default:
+      /* any other type conversion to attribute not supported */
+      retobj = NULL;
```
lavarou, line 365: Should there be a verbosedebug here to display what type was used in `$context` array?

agent/lib_monolog.c
```diff
@@ -306,6 +309,150 @@ static char* nr_monolog_get_context(const size_t argc,
...
+
+  nr_attributes_t* attributes = NULL;
+
+  if (NULL == context_data || !nr_php_is_zval_valid_array(context_data)) {
+    return NULL;
+  }
+
+  attributes = nr_attributes_create(NRPRG(txn)->attribute_config);
+
+  ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARR_P(context_data), key, val) {
```
lavarou, line 438: Will this work if `$context` is just an array and not a hashmap and the keys are numeric? Can `$context` be just an array with numeric keys? I'd suggest looking at and potentially re-using some pieces of `nr_monolog_fmt_context` (which uses more generic `nr_php_zend_hash_zval_apply`), `nr_monolog_fmt_context_item`, `nr_monolog_fmt_context_key`, `nr_monolog_fmt_context_value`.

agent/lib_monolog.c
```diff
@@ -306,6 +309,150 @@ static char* nr_monolog_get_context(const size_t argc,
...
+  zend_string* key;
+  zval* val;
+
+  nr_attributes_t* attributes = NULL;
+
+  if (NULL == context_data || !nr_php_is_zval_valid_array(context_data)) {
+    return NULL;
+  }
+
+  attributes = nr_attributes_create(NRPRG(txn)->attribute_config);
```
lavarou, line 436: Add a check if `attributes` is NULL? It is checked down the call stack but this would short-circuit things and avoid looping over $context.

agent/lib_monolog_private.h
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef LIB_MONOLOG_PRIVATE_HDR
+#define LIB_MONOLOG_PRIVATE_HDR
+
+/*
+ * Purpose : ONLY for testing to verify that the appropriate behavior of
+ *           the conversion of zvals to attribute via nro.
```
lavarou, line 11: Thank you for this comment 🙏!

agent/lib_monolog.c
mfulb: Monolog sends a hash currently so I don't know if this is a concern for code that is focused on the Monolog library.

mfulb: The current agent attribute implementation only accepts strings as attribute keys so any elements using an integer key are ignored.

agent/lib_monolog.c
```diff
@@ -397,21 +544,30 @@ NR_PHP_WRAPPER(nr_monolog_logger_addrecord) {
...
+    }
     api = nr_monolog_version(this_var TSRMLS_CC);
     timestamp
         = nr_monolog_get_timestamp(api, argc, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
   }
 
   /* Record the log event */
   nr_txn_record_log_event(NRPRG(txn), level_name, message, timestamp,
-                          NRPRG(app));
+                          context_attributes, NRPRG(app));
```
lavarou, line 570: Is the ownership of `context_attributes` transferred to `NRPRG(txn)` and `context_attributes` will get released when `NRPRG(txn)` is destroyed?

lavarou: I see it happening in `nr_log_event_destroy`.

agent/lib_monolog.c
lavarou: Fair enough. However, I just wanted to point out that this is valid code:
```
$logger->alert("context is array", ["foobar"]);
```
which yields following output when using `LineFormatter("%level_name%: %message% %context%\n");`:
```
ALERT: context is array ["foobar"]
```

tests/integration/logging/monolog3/test_monolog_context_filter_rule7.php
```diff
@@ -0,0 +1,112 @@
...
+
+/*INI
+newrelic.application_logging.enabled = true
+newrelic.application_logging.forwarding.enabled = true
+newrelic.application_logging.metrics.enabled = true
+newrelic.application_logging.forwarding.max_samples_stored = 10
+newrelic.application_logging.forwarding.log_level = DEBUG
+newrelic.application_logging.forwarding.context_data.enabled = 1
+newrelic.application_logging.forwarding.context_data.include = "A*"
+newrelic.application_logging.forwarding.context_data.exclude = ""
```
lavarou, line 31: Is there a test where `newrelic.application_logging.forwarding.context_data.exclude` is not defined in INI?

agent/lib_monolog.c
lavarou: I see there is a test for that - test_monolog_context_simple.php

lavarou (approved):
Great work! Awesome tests!

agent/lib_monolog.c
mfulb: That might get very chatty if they are using a lot of context data.

tests/integration/logging/monolog3/test_monolog_context_filter_rule7.php
mfulb: I modified `test_monolog_context_simple.php` to only enabled context attributes but does not specify any include or exclude rules.

Commit [fc11c62](https://github.com/newrelic/newrelic-php-agent/commit/fc11c620ee981dde9ac312c02e163f96db025eb3).

agent/lib_monolog.c
mfulb: Do you think any sort of report should be made if the creation failed (due to memory allocation failure most likely)?

agent/lib_monolog.c
```diff
@@ -397,21 +544,30 @@ NR_PHP_WRAPPER(nr_monolog_logger_addrecord) {
...
     argc = nr_php_get_user_func_arg_count(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
     message = nr_monolog_build_message(argc, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+
+    if (nr_txn_log_forwarding_context_data_enabled(NRPRG(txn))) {
+      zval* context_data = nr_monolog_extract_context_data(
+          argc, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+      context_attributes
+          = nr_monolog_convert_context_data_to_attributes(context_data);
+      nr_php_arg_release(&context_data);
+    }
```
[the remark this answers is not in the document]
mfulb: Removed in commit [8a57e36](https://github.com/newrelic/newrelic-php-agent/commit/8a57e3669fc10808026b2f1e7d4065b8cb497527).

agent/lib_monolog.c
lavarou: I meant verbose debug level to troubleshoot why attributes are not being sent.

agent/lib_monolog.c
lavarou: I don't this type of error reporting is done anywhere else in the agent's codebase.

agent/lib_monolog.c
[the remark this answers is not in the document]
lavarou: There are a couple more places: https://github.com/newrelic/newrelic-php-agent/blob/8a57e3669fc10808026b2f1e7d4065b8cb497527/agent/lib_monolog.c#L303-L306 and https://github.com/newrelic/newrelic-php-agent/blob/8a57e3669fc10808026b2f1e7d4065b8cb497527/agent/lib_monolog.c#L313-L319

agent/lib_monolog.c
```diff
@@ -298,44 +243,65 @@ static char* nr_monolog_get_context(const size_t argc,
...
+  }
+  ZEND_HASH_FOREACH_END();
+
+  return attributes;
 }
-#endif
 
 /*
  * Purpose : Combine $message and $context arguments of
```
lavarou, line 292:
```suggestion
 * Purpose : Convert $message argument of
```

agent/lib_monolog.c
```diff
@@ -298,44 +243,65 @@ static char* nr_monolog_get_context(const size_t argc,
...
 
-#if HAVE_CONTEXT_IN_MESSAGE
-  char* context = nr_monolog_get_context(argc, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
-  if (!nr_strempty(context)) {
-    message_and_context = nr_str_append(message_and_context, context, " ");
-  }
-  nr_free(context);
-#endif
-
   return message_and_context;
```
lavarou, line 305: I think this could be simplified even further. I.e. this function can be removed completely and a call to it on line 373 can be replaced with call to `nr_monolog_get_message`.

agent/lib_monolog.c
```diff
@@ -397,21 +363,30 @@ NR_PHP_WRAPPER(nr_monolog_logger_addrecord) {
...
   char* message = NULL;
+  nr_attributes_t* context_attributes = NULL;
   nrtime_t timestamp = nr_get_time();
 
   /* Values of $message and $timestamp arguments are needed only if log
    * forwarding is enabled so agent will get them conditionally */
   if (nr_txn_log_forwarding_enabled(NRPRG(txn))) {
     argc = nr_php_get_user_func_arg_count(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
-    message = nr_monolog_build_message(argc, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+    message = nr_monolog_build_message(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
lavarou, line 373: Since `$message` and `$context` are no longer need to be combined, I'd suggest saving two call frames on the stack by replacing call to `nr_monolog_build_message` with call to `nr_monolog_get_message`:
```suggestion
    message = nr_monolog_get_message(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
Moreover, `nr_monolog_build_message` function can be removed altogether.

Huluti:
Hello! Does this pr will fix https://github.com/newrelic/newrelic-php-agent/issues/741#issuecomment-1786057758 ?
Because monolog support placeholders for vars in context:

image

Source: https://github.com/Seldaek/monolog/blob/main/doc/message-structure.md

agent/lib_monolog.c
mfulb: Fixed in commit https://github.com/newrelic/newrelic-php-agent/pull/753/commits/486ae66fa911b8a1a440d178129d9bd397e9ca00.

agent/lib_monolog.c
mfulb: Added in commit https://github.com/newrelic/newrelic-php-agent/pull/753/commits/dbae61f9e4a8c3f2682eee3ca167ece8e0f29214.

agent/lib_monolog.c
mfulb: Nice catch - I made this suggestion in commit https://github.com/newrelic/newrelic-php-agent/pull/753/commits/aaf9d78eb42c9007ca8cd82769c53c1a02d2260c.

agent/lib_monolog.c
mfulb: I removed this function so I believe this suggestion is no longer applicable.

mfulb:
> Hello! Does this pr will fix [#741 (comment)](https://github.com/newrelic/newrelic-php-agent/issues/741#issuecomment-1786057758) ?

This effort will create attributes on New Relic log events for supported data types in the Monolog `$context`.  It does not affect the formatting of the log messages so I do not believe it will address the issue you are asking about.

axiom/nr_attributes.c
```diff
@@ -162,10 +189,133 @@ static int nr_attribute_destination_modifier_compare(
...
-void nr_attribute_config_modify_destinations(nr_attribute_config_t* config,
-                                             const char* match,
-                                             uint32_t include_destinations,
-                                             uint32_t exclude_destinations) {
+/*
+ * Purpose : Inspects current modifier list and adds finalize rules as needed.
+ *
+ * Params  : 1. Attribute configuration
+ *
+ * Notes   : Certain attriubtes like log context attributes expect the "include"
```
bduranleau-nr, line 197:
```suggestion
 * Notes   : Certain attributes like log context attributes expect the "include"
```

axiom/nr_attributes.c
```diff
@@ -162,10 +189,133 @@ static int nr_attribute_destination_modifier_compare(
...
+  for (cur = config->modifier_list; cur; cur = cur->next) {
+    if (cur->include_destinations & NR_ATTRIBUTE_DESTINATION_LOG) {
+      if (cur->has_wildcard_suffix) {
+        if (0 == cur->match_len) {
+          /* there is an include rule of "*" so no finalize is needed
+           * since all attributes are being explicitely included */
+          return;
+        }
+      }
+      add_finalize_rule = true;
```
bduranleau-nr, line 280: It looks like you could potentially remove a nesting level here by combining these two if statements:
```suggestion
      if (cur->has_wildcard_suffix && 0 == cur->match_len) {
          /* there is an include rule of "*" so no finalize is needed
           * since all attributes are being explicitely included */
          return;
        }
      add_finalize_rule = true;
```

axiom/nr_attributes.c
```diff
@@ -686,7 +874,17 @@ nrobj_t* nr_attributes_agent_to_obj(const nr_attributes_t* attributes,
...
-  return nr_attributes_to_obj_internal(attributes->agent_attribute_list,
+  return nr_attributes_to_obj_internal(attributes->agent_attribute_list, NULL,
+                                       destination);
+}
+
+nrobj_t* nr_attributes_logcontext_to_obj(const nr_attributes_t* attributes,
+                                         uint32_t destination) {
+  if (0 == attributes) {
+    return 0;
+  }
```
bduranleau-nr, line 885:
```suggestion
  if (NULL == attributes) {
    return NULL;
  }
```

axiom/nr_txn.c
```diff
@@ -3276,6 +3276,18 @@ bool nr_txn_log_forwarding_enabled(nrtxn_t* txn) {
...
+bool nr_txn_log_forwarding_context_data_enabled(nrtxn_t* txn) {
+  if (!nr_txn_log_forwarding_enabled(txn)) {
+    return false;
+  }
+
+  if (!txn->options.log_forwarding_context_data_enabled) {
+    return false;
+  }
+
+  return true;
```
bduranleau-nr, line 3288: Could simplify to this:
```suggestion
  return txn->options.log_forwarding_context_data_enabled
```

axiom/nr_attributes.c
```diff
@@ -162,10 +189,133 @@ static int nr_attribute_destination_modifier_compare(
...
+  nr_attribute_destination_modifier_t* next = NULL;
+  bool add_finalize_rule = false;
+
+  if (NULL == config || NULL == config->modifier_list) {
+    /* Since there is no configuration, no work to do. */
+    return;
+  }
+
+  /* remove any existing rules with is_finalize_rule = true */
+  for (cur = config->modifier_list; cur;) {
```
bduranleau-nr, line 240: Typically linked list traversal is done via a while loop rather than a for loop. I would suggest the following instead for clarity of purpose and keeping with implicit techniques, but the for loop method will work the same regardless:
```suggestion
  cur = config->modifier_list;
  while (NULL != cur) {
```

axiom/nr_attributes.c
```diff
@@ -162,10 +189,133 @@ static int nr_attribute_destination_modifier_compare(
...
+   * through the include rules and therefore are excluded implicitely
+   */
+  if (add_finalize_rule) {
+    nr_attribute_config_modify_destinations_internal(
+        config, "*", 0, NR_ATTRIBUTE_DESTINATION_LOG, true);
+  }
+}
+
+/*
+ * Purpose : Inserts a modifier rule into an attribute configuraiton.
```
hahuja2, line 295:
```suggestion
 * Purpose : Inserts a modifier rule into an attribute configuration.
```

axiom/nr_attributes.h
```diff
@@ -99,6 +105,22 @@ extern void nr_attribute_config_disable_destinations(
...
+/*
+ * Purpose : Enable attribute destinations.
+ *
+ * Params  : 1. The configuration to modify.
+ *           2. The set of destinations to enable.
+ *
+ * Note    : Destinations are enabled by default when a config
+ *           is created.  This function can be used to enable
+ *           a previously disabled destination.  Current
+ *           use is to enable destinationss disabled by default
```
hahuja2, line 117:
```suggestion
 *           use is to enable destinations disabled by default
```

hahuja2 (approved):
Awesome work!

axiom/nr_txn.c
mfulb: I actually prefer the current form where a sequence of filters must pass before returning `true`.  The compiler will generate the code you have but for reading and modifying I prefer the expanded form.

bduranleau-nr (approved):

Maxwell2022:
Where can we see these attributes in the UI? I'm using monolog, and I can only see the `message` even in the log details I cannot see any of my custom attributes, just the basic NR ones.

```php
$statusCode = $response->getStatusCode();
$method = $request->method();
$logLevel = $this->getLogLevel($response);
$path = '/' . strtolower($request->path());
$details = [
    'id' => $request->attributes->get('request_id'),
    'method' => $method,
    'url' => $request->url(),
    'request' => $this->obfuscateArray($request->toArray()),
    'response_status' => $statusCode,
    'response_body' => $response->getContent(),
];

logger()->log($logLevel, "$statusCode $method $path", $details);
```

I would expect to see all the `$details` attributes, but I cannot see anything.

image

PR #754, opened 2023-10-26
feat(agent, axiom): Add php package information to transaction

hahuja2:
Summary of this PR:

1. Creates a new file called "nr_php_packages.c" which adds new php package management functions
2. Adds new unit tests to ensure php package management functions are working correctly
3. Add new function called "nr_txn_add_php_package" which adds php package information to transaction
4. Create new unit test to ensure php package information is added to transaction
5. Add "nr_txn_add_php_package" in fw_laravel.c to show how once we find the version, we will call this function which will add the php package information to the transaction


axiom/nr_php_packages.h
```diff
@@ -0,0 +1,119 @@
...
+extern void nr_php_packages_add_package(nr_php_packages_t** h, nr_php_package_t* p);
+
+/*
+ * Purpose : Destroy/free the hashmap
+ *
+ * Params  : 1. A pointer to the pointer of a hashmap
+ * 
+ * Returns : Nothing, it is void
+ */
+extern void nr_php_packages_destroy_package(nr_php_packages_t** h);
```
lavarou, line 60: It destroys the packages hashmap, not a single package, right? I so then I suggest function name change:
```suggestion
extern void nr_php_packages_destroy(nr_php_packages_t** h);
```

axiom/tests/test_txn.c
```diff
@@ -8537,6 +8537,34 @@ static void test_txn_log_configuration(void) {
...
+  nr_txn_add_php_package(txn, package_name2, package_version2);
+  json = nr_php_packages_to_json(txn->php_packages);
+
+  tlib_pass_if_str_equal("correct json",
+                         "[{\"name\":\"Laravel\",\"version\":\"8.83.27\"},{"
+                         "\"name\":\"Slim\",\"version\":\"4.12.0\"}]",
+                         json);
+
+  nr_free(json);
+  nr_php_packages_destroy_package(&txn->php_packages);
```
lavarou, line 8564: I suggest calling `nr_php_packages_destroy` in `nr_txn_destroy`.

axiom/nr_txn.c
lavarou, line 1: I suggest destroying `nrtxn_t.php_packages` when transaction object is destroyed by adding call to `nr_php_packages_destroy` in `nr_txn_destroy_fields`.

axiom/tests/test_packages.c
```diff
@@ -0,0 +1,150 @@
...
+
+  tlib_pass_if_not_null("create package", package);
+  tlib_pass_if_str_equal("test package name", "Laravel", package->package_name);
+  tlib_pass_if_str_equal("test package version", "8.83.27",
+                         package->package_version);
+
+  nr_php_package_destroy(package);
+  nr_php_package_destroy(null_package);
+
+  // Test: passing NULL pointer should not cause crash
```
lavarou, line 25:
```suggestion

  // Test: passing NULL pointer should not cause crash
  nr_php_package_destroy(null_package);
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,119 @@
...
+extern nr_php_package_t* nr_php_package_create(char* name, char* version);
+
+/*
+ * Purpose : Destroy/free php package
+ *
+ * Params  : The php package to free
+ * 
+ * Returns : Nothing, it is void
+ */
+extern void nr_php_package_destroy(nr_php_package_t* p);
```
lavarou, line 39: I suggest `nr_php_package_destroy` to set `p` to `NULL`:
```suggestion
extern void nr_php_package_destroy(nr_php_package_t** p);
```

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,142 @@
...
+  p = (nr_php_package_t*)nr_malloc(sizeof(nr_php_package_t));
+
+  p->package_name = nr_strdup(name);
+  p->package_version = nr_strdup(version);
+
+  return p;
+}
+
+void nr_php_package_destroy(nr_php_package_t* p) {
+  if (p != NULL) {
```
lavarou, line 36: To avoid a situation of accessing an invalid memory by referencing a pointer to memory that has been freed, I suggest `nr_php_package_destroy` to set `p` to `NULL`:
```suggestion
void nr_php_package_destroy(nr_php_package_t** ptr) {
  nr_php_package_t* p;

  if ((NULL == ptr) || (NULL == *ptr)) {
    return;
  }

  p = *ptr;

  if (p != NULL) {
```

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,142 @@
...
+    nr_free(p->package_name);
+    nr_free(p->package_version);
+    nr_free(p);
+  }
+}
```
lavarou, line 41: To avoid a situation of accessing an invalid memory by referencing a pointer to memory that has been freed, I suggest `nr_php_package_destroy` to set `p` to `NULL` (like [here](https://github.com/newrelic/newrelic-php-agent/blob/4713dc219062b00e6771bbd5dfba92f5f7a0a60c/axiom/nr_log_event.c#L21-L38)):
```suggestion
void nr_php_package_destroy(nr_php_package_t** ptr) {
  nr_php_package_t* p;

  if ((NULL == ptr) || (NULL == *ptr)) {
    return;
  }

  p = *ptr;
  nr_free(p->package_name);
  nr_free(p->package_version);

  nr_realfree((void**)ptr);
}
```

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,142 @@
...
+#include "util_strings.h"
+
+nr_php_package_t* nr_php_package_create(char* name, char* version) {
+  nr_php_package_t* p = NULL;
+
+  if (name == NULL || version == NULL) {
+    return NULL;
+  }
+
+  p = (nr_php_package_t*)nr_malloc(sizeof(nr_php_package_t));
```
lavarou, line 27: Check if `p` is not `NULL`.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,142 @@
...
+  if (p != NULL) {
+    nr_free(p->package_name);
+    nr_free(p->package_version);
+    nr_free(p);
+  }
+}
+
+void nr_php_packages_add_package(nr_php_packages_t** h, nr_php_package_t* p) {
+  if (h == NULL || *h == NULL) {
+    *h = nr_hashmap_create((nr_hashmap_dtor_func_t)nr_php_package_destroy);
```
lavarou, line 45: Something is fishy here: if `h` is `NULL`, won't dereferencing with `*h` cause a seg fault?

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,142 @@
...
+
+char* nr_php_package_to_json(nr_php_package_t* package) {
+  char* json;
+  nrbuf_t* buf = nr_buffer_create(0, 0);
+
+  if (package == NULL || package->package_name == NULL
+      || package->package_version == NULL) {
+    return NULL;
+  }
+
```
lavarou, line 81: Defer creating the buffer only after ensuring it needs to be created. Otherwise line 79 causes mem leak:
```suggestion
  nrbuf_t* buf = NULL;

  if (package == NULL || package->package_name == NULL
      || package->package_version == NULL) {
    return NULL;
  }
  buf = nr_buffer_create(0, 0);
```

axiom/tests/test_packages.c
```diff
@@ -0,0 +1,150 @@
...
+  nr_php_package_t* package2;
+  nr_php_package_t* package3;
+  nr_php_packages_t* hm = NULL;
+  int count;
+
+  // Test: create multiple new packages and add to hashmap
+  package1 = nr_php_package_create("Package One", "10.1.0");
+  package2 = nr_php_package_create("Package Two", "11.2.0");
+  package3 = nr_php_package_create("Package Three", "12.3.0");
+
```
lavarou, line 40: Add a test when first parameter is `NULL`:
```suggestion
  /* Should not crash: */
  nr_php_packages_add_package(NULL, package1);
```

axiom/tests/test_packages.c
```diff
@@ -0,0 +1,150 @@
...
+  nr_php_packages_add_package(&hm, package1);
+  nr_php_packages_add_package(&hm, package2);
+  nr_php_packages_add_package(&hm, package3);
+
+  count = nr_php_packages_count(hm);
+
+  tlib_pass_if_int_equal("package count", 3, count);
+
+  nr_php_packages_destroy_package(&hm);
```
lavarou, line 49:
```suggestion
  nr_php_packages_destroy_package(&hm);
  tlib_pass_if_null("PHP packages hashmap destroyed", hm);
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,119 @@
...
+
+/*
+ * Purpose : Add new php package to hashmap. If the hashmap does not exist and this is the first package
+ *           being added, then a new hashmap will be created by calling nr_hashmap_create(). This function
+ *           also frees the package being added, so the user does not need to call nr_php_packages_destroy_package()
```
lavarou, line 44: Could you better describe what happens if the packages hashmap already contains a package with the same name? Could you add a test for that behavior?

axiom/nr_php_packages.c
mfulb: I think C will short circuit if `h == NULL` into the `if` statement body since it is an `||` logical operation.

axiom/nr_php_packages.c
hahuja2: After discussion, we will not make this change because it directly impacts the `nr_php_packages_add_package()` function since we are passing the `nr_php_package_destroy()` function as a destructor when `nr_hashmap_create()` is called and this change would cause the signatures to not match. It is also important to note that `nr_php_package_destroy` is only used as a dtor_func callback in the hashmap and not used anywhere else.

axiom/nr_php_packages.h
hahuja2: After discussion, this will stay the same. Explanation here: https://github.com/newrelic/newrelic-php-agent/pull/754#discussion_r1376736594

axiom/nr_php_packages.h
hahuja2: Yeah good idea, fixed: [ca1b1d7](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/ca1b1d7f5cfb31c2a07c983497c1871fefccc613)

axiom/tests/test_txn.c
hahuja2: Fixed: [5d95e78](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/5d95e7812f98fa74859f9edf83e110fa4d30b126)

axiom/nr_txn.c
hahuja2: Fixed: [5d95e78](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/5d95e7812f98fa74859f9edf83e110fa4d30b126)

axiom/nr_php_packages.c
hahuja2: Added: [89dce80](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/89dce80cf7bed926fcf047446043fa37c1713e30)

axiom/nr_php_packages.c
hahuja2: Yes, good catch! I have separated this NULL check into two different lines and check if `h == NULL` before checking `*h == NULL` [24a27c2](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/24a27c2df8aae92a5ab4e6f8693d814d4dc72ff4)

axiom/tests/test_packages.c
hahuja2: I have added this test and also added a test when the second parameter is NULL [f0c3107](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/f0c31076077ee319dd972f252fdd0425127ab191)

axiom/tests/test_packages.c
hahuja2: Added: [d4906b6](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/d4906b6e602bf49aedd1bd657a8fc20367ebd3ba)

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,149 @@
...
+void nr_php_packages_add_package(nr_php_packages_t** h, nr_php_package_t* p) {
+  if (h == NULL) {
+    return;
+  }
+
+  if (*h == NULL) {
+    *h = nr_hashmap_create((nr_hashmap_dtor_func_t)nr_php_package_destroy);
+  }
+
+  if (p == NULL || p->package_name == NULL || p->package_version == NULL) {
```
mfulb, line 55: Might consider moving this above the check for `*h == NULL` since it would keep us from creating a hashmap if the parameters are invalid.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,149 @@
...
+void nr_php_package_destroy(nr_php_package_t* p) {
+  if (p != NULL) {
+    nr_free(p->package_name);
+    nr_free(p->package_version);
+    nr_free(p);
+  }
+}
+
+void nr_php_packages_add_package(nr_php_packages_t** h, nr_php_package_t* p) {
+  if (h == NULL) {
```
mfulb, line 47: All of the conditional statements in your PR need to follow our [style guide](https://github.com/newrelic/newrelic-php-agent/blob/dev/STYLEGUIDE.md) and use 'yoda conditionals'.

mfulb, line 47:
```suggestion
  if (NULL == h) {
```

axiom/tests/test_txn.c
```diff
@@ -8537,6 +8537,33 @@ static void test_txn_log_configuration(void) {
...
+  char* package_name1 = "Laravel";
+  char* package_version1 = "8.83.27";
+  char* package_name2 = "Slim";
+  char* package_version2 = "4.12.0";
+  nrtxn_t* txn = new_txn(0);
+
+  /*
+   * NULL parameters: ensure it does not crash
+   */
+  nr_txn_add_php_package(NULL, NULL, NULL);
```
mfulb, line 8551: I would test for all combinations with a single `NULL` and otherwise valid parameters to make sure the first parameter being `NULL` doesn't shield other values being invalid having issues since the code probably checks for the first parameter being invalid and exits.

axiom/nr_php_packages.h
hahuja2: I have just pushed a commit that will handle the case where a package is added with the same name. If a new package is added with the same name but a different value, then the newer value will be kept and the older value will be freed. If a package is added with the same name and value, then no action will happen. [f5f2d0b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/f5f2d0b1802cd6ab3c1f71a62c28150008a223fa)

axiom/tests/test_packages.c
```diff
@@ -0,0 +1,150 @@
...
+  int count;
+
+  // Test: create multiple new packages and add to hashmap
+  package1 = nr_php_package_create("Package One", "10.1.0");
+  package2 = nr_php_package_create("Package Two", "11.2.0");
+  package3 = nr_php_package_create("Package Three", "12.3.0");
+
+  nr_php_packages_add_package(&hm, package1);
+  nr_php_packages_add_package(&hm, package2);
+  nr_php_packages_add_package(&hm, package3);
```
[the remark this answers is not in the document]
hahuja2: If a new package is added with the same key but different value, then the newer value will be kept and the older value will be freed. I have added this test case in this commit: [f5f2d0b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/f5f2d0b1802cd6ab3c1f71a62c28150008a223fa)

axiom/nr_php_packages.c
hahuja2: Good idea, fixed: [0d315e3](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/0d315e34cfaec65ddbc71a2e46c8901aef048550)

axiom/nr_php_packages.c
hahuja2: Fixed: [596aa6b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/596aa6b81613b00b6d142913cb7c73c8b935d3d3)

hahuja2: Fixed: [596aa6b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/596aa6b81613b00b6d142913cb7c73c8b935d3d3)

axiom/tests/test_txn.c
hahuja2: good idea, added more tests: [b8d3820](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/b8d3820c9de657c0f0be37ca566caf56db3bace6)

axiom/nr_php_packages.c
mfulb: I think you missed some in the rest of the file(s).

hahuja2: I noticed that I missed one in this same file. Fixed here: [64c3366](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/64c336626f1d08e02032ef42bd2ae8be51f22c78). I double checked right now and I don't think the other files have this issue unless I missed it.

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,131 @@
...
+#include "util_random.h"
+#include "util_vector.h"
+#include "util_hashmap.h"
+
+typedef struct _nr_php_package_t {
+  char* package_name;
+  char* package_version;
+} nr_php_package_t;
+
+typedef nr_hashmap_t nr_php_packages_t;
```
bduranleau-nr, line 19: I would suggest a different name for this hashmap- `nr_php_packages_t` does not indicate that it's actually a hashmap under the hood, and is very similar in name to the above struct `nr_php_package_t`. This can become confusing when trying to read and write code. 
Since we aren't changing anything about the underlying hashmap `nr_hashmap_t` implementation, do we need this typedef at all? Couldn't we just use the baseline `nr_hashmap_t` instead?

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright 2022 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef nr_php_packages_HDR
+#define nr_php_packages_HDR
```
bduranleau-nr, line 7: The paradigm for header `#ifdef`'s are all caps:
```suggestion
#ifndef NR_PHP_PACKAGES_HDR
#define NR_PHP_PACKAGES_HDR
```

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,166 @@
...
+
+void nr_php_package_destroy(nr_php_package_t* p) {
+  if (NULL != p) {
+    nr_free(p->package_name);
+    nr_free(p->package_version);
+    nr_free(p);
+  }
+}
+
+void nr_php_packages_add_package(nr_php_packages_t** h, nr_php_package_t* p) {
```
bduranleau-nr, line 46: Why use a pointer to a pointer here? 
All the subsequent operations on `h` in this function require accessing this variable via `*h`

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+    nr_php_package_destroy(p);
+    return;
+  }
+
+  nr_hashmap_set(*h, p->package_name, nr_strlen(p->package_name), p);
+}
+
+void nr_php_packages_destroy(nr_php_packages_t** h) {
```
bduranleau-nr, line 79: Pointers to pointers should be avoided to prevent unnecessary abstraction that can lead to confusion. Rather than using a pointer to a pointer, just use the pointer value. When you need to call `nr_hashmap_destroy`, just pass it `&h` instead.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,166 @@
...
+  nr_hashmap_destroy(h);
+}
+
+size_t nr_php_packages_count(nr_php_packages_t* h) {
+  return nr_hashmap_count(h);
+}
```
bduranleau-nr, line 85: Since these functions are just wrapping existing functions and not performing any operations on the data, I would suggest doing away with them and just calling the underlying functions directly instead.

axiom/nr_php_packages.h
mfulb: The idea was if we did change the implementation then we are not going around changing the data type.  This is a common form of hiding the implementation details so one can make changes in the future w/o having to change any consumers of the library.

mfulb: We do this all over the place - `nr_log_event_t` and `nr_log_events_t`, etc.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,166 @@
...
+
+size_t nr_hashmap_count_buckets(const nr_php_packages_t* hashmap) {
+  return (size_t)(1 << hashmap->log2_num_buckets);
+}
+
+int nr_php_packages_has_package(nr_php_packages_t* h,
+                                char* package_name,
+                                size_t package_len) {
+  return nr_hashmap_has(h, package_name, package_len);
+}
```
bduranleau-nr, line 95: Same as above- if we aren't making implementation specific modifications to the variables here, this wrapper isn't necessary and we can just use the `nr_hashmap_has` function directly instead.

axiom/nr_php_packages.h
bduranleau-nr: `nr_log_events_t` has a unique structure implementation though: https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_log_events.c#L17 
It isn't just an alias for `nr_hashmap_t`

bduranleau-nr: I can't find any occurrence in the code where we use an alias for the `nr_hashmap_t` struct alone

axiom/nr_php_packages.c
hahuja2: We need a double pointer here because we are allocating space for the hashmap if it is null

axiom/nr_php_packages.c
hahuja2: I think double pointer is necessary to ensure there are no dangling pointers. However, since this function is just wrapping an existing function, I will remove it and directly call nr_hashmap_destroy instead

axiom/nr_php_packages.c
lavarou: [New Relic PHP Agent's style guide](https://github.com/newrelic/newrelic-php-agent/blob/main/STYLEGUIDE.md) requires use of [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions).

axiom/nr_php_packages.c
lavarou: I think leaving the wrapper function makes sense as it completes the nr_php_packages API.

axiom/nr_php_packages.c
lavarou: I think they provide a readable API around nr_php_packages datatype (yes, it's a hashmap under the hood, but it just reads good and is easier to look for code related to php_packages). Maybe for efficiency make them inline.

axiom/nr_php_packages.h
lavarou: > The idea was if we did change the implementation then we are not going around changing the data type. This is a common form of hiding the implementation details so one can make changes in the future w/o having to change any consumers of the library.

I agree with the above. `nr_php_packages_t` with set of `nr_php_packages_*` functions hides the implementation nicely.

axiom/nr_php_packages.c
```diff
@@ -56,6 +56,23 @@ void nr_php_packages_add_package(nr_php_packages_t** h, nr_php_package_t* p) {
     return;
   }
 
+  // If package with the same key already exists, we will check if the value is
+  // different. If it is different, then we will update the value of the package
+  if (0
+      != nr_php_packages_has_package(*h, p->package_name,
+                                     nr_strlen(p->package_name))) {
+    nr_php_package_t* package = (nr_php_package_t*)nr_hashmap_get(
+        *h, p->package_name, nr_strlen(p->package_name));
```
lavarou, line 65: Wouldn't it be more efficient to just call `nr_hashmap_get`? This way the `nr_hashmap_hash_key` gets calculated once and `nr_hashmap_fetch` is also called once.

axiom/nr_txn.c
```diff
@@ -3457,3 +3459,24 @@ void nr_txn_record_log_event(nrtxn_t* txn,
 
   nr_txn_add_logging_metrics(txn, log_level_name);
 }
+
+extern void nr_txn_add_php_package(nrtxn_t* txn,
```
lavarou, line 3463: No need for `extern` here.

lavarou (approved):
Nice work! LGTM! 👍

axiom/nr_php_packages.h
lavarou: > > We do this all over the place - `nr_log_event_t` and `nr_log_events_t`, etc.

I believe this was in response to 

> `nr_php_packages_t` [...] is very similar in name to the above struct `nr_php_package_t`. This can become confusing when trying to read and write code.

and has nothing to do with `nr_php_packages_t` being an alias of `nr_hashmap_t`.

axiom/nr_php_packages.c
hahuja2: Great catch! I have removed the unnecessary function call [a486c77](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/a486c774a35994aa1441b77d5bcf199659e0c8c4)

axiom/nr_txn.c
hahuja2: Fixed: [9325d71](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/9325d719fcc1c0fae33a1f73ca9bc6eae6d74689)

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,131 @@
...
+ * Purpose : Destroy/free php package
+ *
+ * Params  : The php package to free
+ *
+ * Returns : Nothing, it is void
+ */
+extern void nr_php_package_destroy(nr_php_package_t* p);
+
+/*
+ * Purpose : Add new php package to hashmap. If the hashmap does not exist and
```
mfulb, line 43: General comment on this file - I think the references to 'hashmap' should be removed as the caller of these functions will (and shouldn't) have any idea how the package info is stored.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,163 @@
...
+char* nr_php_package_to_json(nr_php_package_t* package) {
+  char* json;
+  nrbuf_t* buf = NULL;
+
+  if (NULL == package || NULL == package->package_name
+      || NULL == package->package_version) {
+    return NULL;
+  }
+  buf = nr_buffer_create(0, 0);
+  nr_buffer_add(buf, NR_PSTR("{\"name\":\""));
```
mfulb, line 103:
```suggestion
  nr_buffer_add(buf, NR_PSTR("[\""));
  nr_buffer_add(buf, package->package_name, nr_strlen(package->package_name));
  nr_buffer_add(buf, NR_PSTR("\",\""));
  nr_buffer_add(buf, package->package_version,
                nr_strlen(package->package_version));
  nr_buffer_add(buf, NR_PSTR("\""));

  /* unclear if this empty dictionary should be included or not */
  nr_buffer_add(buf, NR_PSTR(",{}]"));
  nr_buffer_add(buf, NR_PSTR("\0"));
```

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,163 @@
...
+  nr_buffer_add(buf, package->package_name, nr_strlen(package->package_name));
+  nr_buffer_add(buf, NR_PSTR("\",\"version\":\""));
+  nr_buffer_add(buf, package->package_version,
+                nr_strlen(package->package_version));
+  nr_buffer_add(buf, NR_PSTR("\"}"));
+  nr_buffer_add(buf, NR_PSTR("\0"));
```
mfulb, line 109:
```suggestion
  nr_buffer_add(buf, NR_PSTR("[\""));
  nr_buffer_add(buf, package->package_name, nr_strlen(package->package_name));
  nr_buffer_add(buf, NR_PSTR("\",\""));
  nr_buffer_add(buf, package->package_version,
                nr_strlen(package->package_version));
  nr_buffer_add(buf, NR_PSTR("\""));

  /* unclear if this empty dictionary should be included or not */
  nr_buffer_add(buf, NR_PSTR(",{}]"));
  nr_buffer_add(buf, NR_PSTR("\0"));
```

mfulb: This change will generate JSON compatible with was is expected.

mfulb: You will need to fix unit tests which should be inspecting the generating JSON output.

axiom/nr_php_packages.c
hahuja2: I have changed them to inline for efficiency: [38061d5](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/38061d53dec5c8db86cb57d2321c40553806d39a)

axiom/nr_php_packages.c
hahuja2: changed it to inline: [38061d5](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/38061d53dec5c8db86cb57d2321c40553806d39a)

axiom/nr_php_packages.c
hahuja2: Thanks for fixing this! I have added these changes and updated the unit tests accordingly: [e997574](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/e9975744e9d6ad61800b6952faafe32c74ec5dd9)

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,152 @@
...
+#include "util_minmax_heap.h"
+#include "util_vector.h"
+#include "util_hashmap.h"
+#include "util_hashmap_private.h"
+#include "util_strings.h"
+
+nr_php_package_t* nr_php_package_create(char* name, char* version) {
+  nr_php_package_t* p = NULL;
+
+  if (NULL == name || NULL == version) {
```
zsistla, line 23: Is there a spec regarding this?  If we have the package name but can't determine the version, should not report anything or should we still report the package?

axiom/nr_php_packages.h
hahuja2: Good idea, fixed here: [a002793](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/a002793afbc17aa5753821ac3876802d3ac35db5)

axiom/nr_php_packages.c
hahuja2: Good point, the spec says that we should still pass an empty string with a space. I have added this case and added unit tests to ensure that when version is null, we will set the version to an empty string with a space. [2bb60f5](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/2bb60f58ff030adf798b02e29cde14e3ce51e7f0)

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,158 @@
...
+      package->package_version = nr_strdup(p->package_version);
+    }
+    nr_php_package_destroy(p);
+    return;
+  }
+
+  nr_hashmap_set(*h, p->package_name, nr_strlen(p->package_name), p);
+}
+
+size_t nr_hashmap_count_buckets(const nr_php_packages_t* hashmap) {
```
mfulb, line 82: Why is this function here?  It is an internal function from `util_hashmap.c` and should not be duplicated this way!

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,138 @@
...
+extern char* nr_php_packages_to_json(nr_php_packages_t* h);
+
+/*
+ * Purpose : Returns how many buckets are in the collection
+ *
+ * Params  : 1. A pointer to nr_php_packages_t
+ *
+ * Returns : The number of buckets in the collection
+ */
+extern size_t nr_hashmap_count_buckets(const nr_php_packages_t* hashmap);
```
mfulb, line 136: This should not be here it is an internal function from a separate source file.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,158 @@
...
+  size_t i;
+  nr_hashmap_bucket_t* bucket;
+  bool package_added = false;
+  char* package_json;
+
+  if (NULL == h || NULL == buf) {
+    return false;
+  }
+
+  num_buckets = nr_hashmap_count_buckets(h);
```
mfulb, line 123:
```suggestion
  num_buckets = nr_hashmap_count(h);
```

The consumer of the `nr_hashmap_t` data type does not need to concern itself with the internal structure of the data structure, in this case, the number of buckets in use to implement the hash map.  The correct call is `nr_hashmap_count()` which returns the number of items in the hashmap.

mfulb, line 123: This function needs to be rewritten to not iterate over the internal data structure of the `nr_hashmap_t` and instead use the function API for this data type.

axiom/tests/test_txn.c
```diff
@@ -8537,6 +8537,36 @@ static void test_txn_log_configuration(void) {
...
+  nr_txn_add_php_package(txn, package_name1, NULL);
+
+  // Test: add php packages to transaction
+  nr_txn_add_php_package(txn, package_name1, package_version1);
+  nr_txn_add_php_package(txn, package_name2, package_version2);
+  json = nr_php_packages_to_json(txn->php_packages);
+
+  tlib_pass_if_str_equal("correct json",
+                         "[[\"Laravel\",\"8.83.27\",{}],["
+                         "\"Slim\",\"4.12.0\",{}]]",
```
mfulb, line 8563:
```suggestion
                         "[[\"Laravel\",\"8.83.27\",{}],"
                         "[\"Slim\",\"4.12.0\",{}]]",
```

mfulb: That is the correct format for the JSON.

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,158 @@
...
+  nr_buffer_add(buf, NR_PSTR("[\""));
+  nr_buffer_add(buf, package->package_name, nr_strlen(package->package_name));
+  nr_buffer_add(buf, NR_PSTR("\",\""));
+  nr_buffer_add(buf, package->package_version,
+                nr_strlen(package->package_version));
+  nr_buffer_add(buf, NR_PSTR("\""));
+
+  /* unclear if this empty dictionary should be included or not */
+  nr_buffer_add(buf, NR_PSTR(",{}]"));
+  nr_buffer_add(buf, NR_PSTR("\0"));
```
mfulb, line 104:
```suggestion
   nr_buffer_add(buf, NR_PSTR("[\""));
  nr_buffer_add(buf, package->package_name, nr_strlen(package->package_name));
  nr_buffer_add(buf, NR_PSTR("\",\""));
  nr_buffer_add(buf, package->package_version,
                nr_strlen(package->package_version));
  nr_buffer_add(buf, NR_PSTR("\",{}]"));
  nr_buffer_add(buf, NR_PSTR("\0"));
```

mfulb: This should generate the proper JSON.

axiom/tests/test_packages.c
mfulb, line 1: I think this should be called `test_php_packages.c`.

axiom/nr_php_packages.h
mfulb: I think this is a better way to create the data structure for packages:
```
struct _nr_php_packages_t {
  nr_hashmap_t *data;
};

typedef nr_php_packages_t ) _nr_php_packages_t;
```

It is odd to see `nr_packages_t *` passed where the hashmap API expects `nr_hashmap_t *`.  This change would keep the isolation of how the php packages data type is implemented from consumers of the API and make the code more clear.

mfulb: Likewise `nr_php_package_t` probably could use a similar `struct` wrapper.

axiom/nr_php_packages.c
hahuja2: I have removed this function: [b4e182e](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/b4e182ea97bab700def5aa6c9e1943d8cc04fdf1)

axiom/nr_php_packages.h
hahuja2: I have removed this function: [b4e182e](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/b4e182ea97bab700def5aa6c9e1943d8cc04fdf1)

axiom/nr_php_packages.c
hahuja2: Thanks for pointing this out. I have simplified this function and instead use `nr_hashmap_apply()` now: [65c25ff](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/65c25ffcdd808620537c7e2f35b6a43010ec5125)

axiom/tests/test_packages.c
hahuja2: Fixed: [7cb8c10](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/7cb8c10e5c56b42fc10aec56c25e91e17c2a3cb6)

axiom/nr_php_packages.c
```diff
@@ -0,0 +1,173 @@
...
+  // If package with the same key already exists, we will check if the value is
+  // different. If it is different, then we will update the value of the package
+  package = (nr_php_package_t*)nr_hashmap_get(h->data, p->package_name,
+                                              nr_strlen(p->package_name));
+  if (NULL != package) {
+    if (0 != nr_strcmp(package->package_version, p->package_version)) {
+      nr_free(package->package_version);
+      package->package_version = nr_strdup(p->package_version);
+    }
+    nr_php_package_destroy(p);
```
mfulb, line 92: I don't think this function should destroy the `nr_php_package_t` that was passed in.  The default behavior is to not destroy it so it will cause bugs if sometimes it does and other times it doesn't.  The caller has no way to now what will happen.  On a name collision just overwrite the existing version but leave the passed in argument alone.

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+  char* package_name;
+  char* package_version;
+} nr_php_package_t;
+
+typedef struct _nr_php_packages_t {
+  nr_hashmap_t* data;
+} nr_php_packages_t;
+
+/*
+ * Purpose : Create a new php package
```
mfulb, line 24: I would add docs here to explain how to create a package with an unknown version.  

You could document that passing `version = NULL`  will do this.

An alternate way which I think makes the code easier to understand would be to have a define like `PHP_PACKAGE_VERSION_UNKNOWN` which is defined as:
```
#define PHP_PACKAGE_VERSION_UNKNOWN "PHP_PACKAGE_VERSION_UNKNOWN"
```

Then to create a package with an unknown version you would do:
```
pkg = nr_php_package_create("MyFramework", PHP_PACKAGE_VERSION_UNKNOWN);
```

Either way it is important to document this use case so one doesn't have to read the function to understand how to handle it.

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+ *           by nr_php_package_destroy()
+ */
+extern nr_php_package_t* nr_php_package_create(char* name, char* version);
+
+/*
+ * Purpose : Destroy/free php package
+ *
+ * Params  : The php package to free
+ *
+ * Returns : Nothing, it is void
```
mfulb, line 40:
```suggestion
 * Returns : Nothing
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+ */
+extern nr_php_packages_t* nr_php_packages_create(void);
+
+/*
+ * Purpose : Add new php package to collection. If a package with the same key
+ *           but different value is added, then the old value will be freed and
+ *           the newer value will be kept. If a duplicate package with the same
+ *           key and value is added, no action will be taken. This function also
+ *           frees the package being added, so the user does not need to call
+ *           nr_php_package_destroy()
```
mfulb, line 57:
```suggestion
 *           key and value is added, no action will be taken.
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Params  : 1. A pointer to nr_php_packages_t where packages
+ *              will be added
+ *           2. A pointer to the php package that needs to be added to the
+ *              collection
+ *
+ * Returns : Nothing, it is void
```
mfulb, line 64:
```suggestion
 * Returns : Nothing
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+/*
+ * Purpose : Destroy/free the collection
+ *
+ * Params  : 1. A pointer to the pointer of nr_php_packages_t
+ *
+ * Returns : Nothing, it is void
+ */
+static inline void nr_php_packages_destroy(nr_php_packages_t** h) {
+  if (nrlikely(h)) {
+    nr_hashmap_destroy(&(*h)->data);
```
mfulb, line 78:
```suggestion
    if (NULL != (*h)->data)) {
      nr_hashmap_destroy(&(*h)->data);
    }
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+ * Params  : 1. A pointer to nr_php_packages_t
+ *           2. The package to check
+ *           3. Length of package to check
+ *
+ * Returns : Returns non-zero if the package exists
+ */
+static inline int nr_php_packages_has_package(nr_php_packages_t* h,
+                                              char* package_name,
+                                              size_t package_len) {
+  if (nrlikely(h)) {
```
mfulb, line 110:
```suggestion
  if (nrlikely(NULL != h && NULL != h->data)) {
```
Need to check hashmap is defined.

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+ */
+extern bool nr_php_packages_to_json_buffer(nr_php_packages_t* hashmap,
+                                           nrbuf_t* buf);
+
+/*
+ * Purpose : Returns all of the packages in the collection as a JSON
+ *
+ * Params  : 1. A pointer to nr_php_packages_t
+ *
+ * Returns : A JSON
```
mfulb, line 142:
```suggestion
 * Returns : An allocated string containing the JSON representation of the packages collection.  Caller takes ownership of this string.
```

axiom/nr_php_packages.h
```diff
@@ -0,0 +1,146 @@
...
+  }
+  return 0;
+}
+
+/*
+ * Purpose : Converts a package to a json
+ *
+ * Params  : 1. A pointer to the package
+ *
+ * Returns : Returns the package in json format
```
mfulb, line 121:
```suggestion
 * Returns : An allocated string containing the JSON representation of the package.  Caller takes ownership of this string.
```

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+  nr_php_package_t* package3;
+  nr_php_package_t* package4;
+  int count;
+
+  package1 = nr_php_package_create("Package One", "1.0.0");
+  // Add package with same key, but different value. Newer value will be kept
+  package2 = nr_php_package_create("Package One", "11.0");
+  package3 = nr_php_package_create("Package Two", "2.0.0");
+  // Add package with same key and same value. No action will happen
+  package4 = nr_php_package_create("Package Two", "2.0.0");
```
mfulb, line 82: I'd add a package w/o a version also to verify the json creation.

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+  nr_php_package_t* package2;
+  nr_php_package_t* package3;
+  nr_php_packages_t* hm = nr_php_packages_create();
+  int count;
+
+  // Test: create multiple new packages and add to hashmap
+  package1 = nr_php_package_create("Package One", "10.1.0");
+  package2 = nr_php_package_create("Package Two", "11.2.0");
+  package3 = nr_php_package_create("Package Three", "12.3.0");
+  /* Should not crash: */
```
mfulb, line 40:
```suggestion
  /* Should not crash: */
  
```

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+  nr_php_packages_t* h = nr_php_packages_create();
+  nr_php_package_t* package1;
+  nr_php_package_t* package2;
+
+  // Test: passing NULL does not crash
+  tlib_pass_if_null("NULL package", nr_php_packages_to_json(NULL));
+
+  // Test: convert all packages in hashmap to json
+  package1 = nr_php_package_create("Package One", "10.1.0");
+  package2 = nr_php_package_create("Package Two", "11.2.0");
```
mfulb, line 118: Add a package w/o a version to verify the JSON creation.

axiom/nr_php_packages.h
mfulb, line 57: I would rewrite this to indicate the caller is not responsible for destroying the package whether or not there is a name collision.

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+
+  count = nr_php_packages_count(hm);
+
+  tlib_pass_if_int_equal("package count", 3, count);
+
+  nr_php_packages_destroy(&hm);
+  tlib_pass_if_null("PHP packages hashmap destroyed", hm);
+}
+
+static void test_php_package_to_json(void) {
```
mfulb, line 55: Test passing in a NULL pointer to `nr_php_package_to_json()`.

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+  // Test: convert package to json
+  package1 = nr_php_package_create("TestPackage", "7.2.0");
+  json = nr_php_package_to_json(package1);
+  tlib_pass_if_str_equal("valid package", "[\"TestPackage\",\"7.2.0\",{}]",
+                         json);
+  nr_free(json);
+  nr_php_package_destroy(package1);
+}
+
+static void test_php_packages_to_json_buffer(void) {
```
mfulb, line 68: Test passing in a NULL pointer to `nr_php_packages_to_json_buffer()`.

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+  nr_buffer_add(buf, NR_PSTR("\0"));
+  tlib_pass_if_str_equal(
+      "filled collection",
+      "[[\"Package One\",\"11.0\",{}],[\"Package Two\",\"2.0.0\",{}]]",
+      nr_buffer_cptr(buf));
+  nr_php_packages_destroy(&collection);
+  nr_buffer_destroy(&buf);
+}
+
+static void test_php_packages_to_json(void) {
```
mfulb, line 107: Test passing in a NULL pointer to `nr_php_packages_to_json()`.

axiom/tests/test_php_packages.c
hahuja2: added: [2d67c1b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/2d67c1bc407b743ecdbc9e6828bf42b03cdcecbc)

axiom/tests/test_php_packages.c
hahuja2: added: [2d67c1b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/2d67c1bc407b743ecdbc9e6828bf42b03cdcecbc)

axiom/tests/test_txn.c
```diff
@@ -8537,6 +8537,36 @@ static void test_txn_log_configuration(void) {
   // clang-format on
 }
 
+static void test_nr_txn_add_php_package(void) {
+  char* json;
+  char* package_name1 = "Laravel";
+  char* package_version1 = "8.83.27";
+  char* package_name2 = "Slim";
+  char* package_version2 = "4.12.0";
+  nrtxn_t* txn = new_txn(0);
```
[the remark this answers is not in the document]
hahuja2: good catch, added: [2d67c1b](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/2d67c1bc407b743ecdbc9e6828bf42b03cdcecbc)

axiom/tests/test_php_packages.c
hahuja2: This one is already added here: https://github.com/newrelic/newrelic-php-agent/blob/2d67c1bc407b743ecdbc9e6828bf42b03cdcecbc/axiom/tests/test_php_packages.c#L121

axiom/tests/test_php_packages.c
hahuja2: added: [2dc23ea](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/2dc23ea5ce9a0c5c0915502caf0396fce8002818)

axiom/tests/test_php_packages.c
hahuja2: added: [2dc23ea](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/2dc23ea5ce9a0c5c0915502caf0396fce8002818)

axiom/nr_php_packages.h
hahuja2: I have updated the description and decided to do it the first way for now: [90ece38](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/90ece38106c8b0fa9deb82bde4606dd92cd413bf)

axiom/nr_php_packages.h
hahuja2: updated: [90ece38](https://github.com/newrelic/newrelic-php-agent/pull/754/commits/90ece38106c8b0fa9deb82bde4606dd92cd413bf)

axiom/tests/test_php_packages.c
```diff
@@ -0,0 +1,189 @@
...
+}
+
+static void test_php_package_exists_in_hashmap(void) {
+  nr_php_package_t* package1;
+  nr_php_package_t* package2;
+  nr_php_packages_t* hm = nr_php_packages_create();
+  int exists;
+
+  // Test: check if package exists in hashmap
+  package1 = nr_php_package_create("Package One", "10.1.0");
```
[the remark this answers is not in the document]
hahuja2: After discussion, we will leave this out for now

mfulb (approved):
Great work Hitesh!

PR #756, opened 2023-10-31
feat: Send Docker cgroup V2 ID

bduranleau-nr:
Implements the following:
- Parse the Docker cgroup V2 ID from the Agent container contained in the file `/proc/self/mountinfo`
- Send the Docker ID from agent to Daemon
- Send Agent Docker ID from Daemon to Collector in the connect message


agent/php_environment.c
```diff
@@ -511,6 +511,94 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+
+  // open file
+  fd = fopen(cgroup_fname, "r");
+  if (NULL == fd) {
+    nrl_warning(NRL_AGENT, "Failed to open %s", cgroup_fname);
+    return;
+  }
+
+  // compile regex to verify hexadecimal chars
+  regex = nr_regex_create("^[a-fA-F0-9]+$", 0, 0);
```
mfulb, line 548: Can this call fail?

agent/php_environment.c
```diff
@@ -511,6 +511,94 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+   *  5. Verify the extracted ID is both:
+   *    a. 64 bytes long (not including null terminator)
+   *    b. comprised of only hexadecimal characters
+   *  6. Assign the extracted & verified ID to the global
+   *    a. Example ID: ec807d5258c06c355c07e2acb700f9029d820afe5836d6a7e19764773dc790f5
+   *  7. Set found = true and exit the loops
+   */
+  // clang-format on
+
+  nrl_verbosedebug(NRL_AGENT, "Reading %s:", cgroup_fname);
```
mfulb, line 575: Is this necessary for the production release?  Seems we are always reading the same filename.

agent/php_environment.c
```diff
@@ -511,6 +511,94 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+
+  while (FAILURE != getline(&line_ptr, &len, fd) && !found) {
```
mfulb, line 577: So this code will loop infinitely if for some reason the source continues to provide data - say "/dev/zero".  Understandably the file we're looking at shouldn't do that.  But would it be worth putting in some kind of limiter so if something goes wrong the entire PHP process doesn't lock up?  Like a maximum number of passes through the loop?

agent/php_environment.c
```diff
@@ -511,6 +511,94 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+    token = strtok(line_ptr, "/");
```
mfulb, line 577: I don't think `strtok()` is thread-safe.  It might become a problem at some point.
This function is not used elsewhere in the agent.
It appears `nr_strsplit()` is what is used instead.  I would suggest using this existing function to simplify the code and leverage a tested function.
At a minimum I would ask you mark this section of code as not thread safe so if the agent ever needs to be multi-threaded it will be easy to identify this code would need to be re-written.

agent/php_globals.h
```diff
@@ -75,6 +75,7 @@ typedef struct _nrphpglobals_t {
   int apache_threaded; /* 1 if a threaded MPM is in use, 0 otherwise */
   int preload_framework_library_detection; /* Enables preloading framework and
                                               library detection */
+  char* docker_id;
```
mfulb, line 78: Please add a description for this field.

agent/php_environment.c
bduranleau-nr: It technically can if the pcre expression fails to compile (for example), in which case it returns `NULL`. The subsequent calls that depend on `regex` (`nr_regex_match` and `nr_regex_destroy`) are both capable of handling `NULL` values. That being said, it probably makes sense to bail out here if the regex fails for any reason.

src/newrelic/app.go
```diff
@@ -244,6 +245,9 @@ func (info *AppInfo) ConnectPayloadInternal(pid int, util *utilization.Data) *Ra
 		utilCopy := *util
 		data.Util = &utilCopy
 		data.Util.Hostname = data.Host
+		if info.DockerId != "" {
```
mfulb, line 248: Does this handle if the value is nil?

agent/php_environment.c
bduranleau-nr: Removed in 1e0ca9167b57752f376bf2237291f61568eaa144

agent/php_environment.c
bduranleau-nr: Addressed in 1e0ca9167b57752f376bf2237291f61568eaa144

mfulb:
Not sure how to add this comment on a file so doing it here:

`test_cmd_appinfo.c`

- I believe you need to add a check in `test_create_empty_query()` to validate the `docker_id` field.
- Likewise check the value in `test_create_query()`

mfulb:
I believe there are tests on the daemon side as well - for example, in app_test.go there is a test `TestConnectPayloadInternalHostname()` which problably now needs a similar one for docker id.

src/newrelic/app.go
bduranleau-nr: The value at this point should never be `nil` since `DockerId` is defined as a string type and string types in Go can't be `nil`.

agent/php_environment.c
bduranleau-nr: I can add code to set an arbitrary loop count to bail out after X number of loops. How about 100? I don't think most mountinfo files will have more lines than that.

agent/php_environment.c
bduranleau-nr: Good point... What if I used `strtok_r` instead? That function is thread safe and doesn't have the overhead of having to convert a ton of nr objects. An early implementation of this function used `strtok_r` and I swapped it for `strtok` for brevity and not having to maintain additional position pointers, so it should be relatively simple and unintrusive to transition back to that model as oppsed to `nr_strsplit`.

agent/php_environment.c
bduranleau-nr: Added in d7039ee8dc39600064ce9631f28aa86d0d3ef600

agent/php_globals.h
bduranleau-nr: Added in fc0f631a710c509cb2935ce0c5a508ccea61694a

agent/php_environment.c
mfulb: Maybe higher - on my linux machine at home has 73 lines in it.  With more software coming in 'snap' format I believe this can cause a large number of mounts.  I think 1000 would be fine - it should never be reached unless something is probably very wrong with `/proc/self/mountinfo`.   Adding a log message if it bails out for this reason would be helpful in diagnosing the (rare hopefully) situation this limit is reached.

agent/php_environment.c
mfulb: That seems like a reasonable approach.

agent/php_environment.c
bduranleau-nr: Upped this value in b878efa816b73ccf9c186a00b4b9086450582bb8
Also took the opportunity to add a boundary to the inner loop with a max of 10 token parsing loops. The actual field we want shouldn't take more than 2 on a standard system, but just in case this file format changes in the future or a user has an unusual setup we'll have a reasonable buffer.

agent/php_environment.c
bduranleau-nr: fixed in a4e3dc49ac33500ca575d03dafa206397b067148

agent/php_environment.c
mfulb: Good idea!

src/newrelic/app_test.go
```diff
@@ -184,6 +184,54 @@ func TestConnectPayloadInternalHostname(t *testing.T) {
...
+func TestConnectPayloadInternalDocker(t *testing.T) {
+	util := &utilization.Data{}
+	info := &AppInfo{}
+
+	// No docker ID, nil utilization data
+	info.DockerId = ""
+
+	b := info.ConnectPayloadInternal(1, nil)
+
+	result, _ := utilization.GetDockerId(b.Util)
```
mfulb, line 196: Might be good to test the second (usually ignored) parameter matches expectations.

src/newrelic/app_test.go
```diff
@@ -184,6 +184,54 @@ func TestConnectPayloadInternalHostname(t *testing.T) {
...
+
+	if nil != b.Util {
+		t.Errorf("expected: %v\nactual: %v", nil, b.Util)
+	}
+
+	// No Docker ID, utilization data
```
mfulb, line 202: How is this test different that the one above?  Can't you check both test expectations after the first test runs?

src/newrelic/app_test.go
```diff
@@ -184,6 +184,54 @@ func TestConnectPayloadInternalHostname(t *testing.T) {
...
+	// Docker ID, nil utilization data
+	info.DockerId = "1056761e1f44969c959364a8e26e9345b37ccb91aef09a8173c90cf1d1d99156"
+
+	b = info.ConnectPayloadInternal(1, nil)
+
+	if nil != b.Util {
+		t.Errorf("expected: %v\nactual: %v", nil, b.Util)
+	}
+
+	// Docker ID, utilization data
```
mfulb, line 222: Is this test supposed to have non-nil utilization data?

src/newrelic/utilization/utilization_hash.go
```diff
@@ -204,6 +199,38 @@ func GatherDockerID(util *Data) error {
...
+		util.Vendors = nil
+	}
+}
+
+func GetDockerId(util *Data) (string, error) {
+    id := ""
+    if nil == util {
+        return id, fmt.Errorf("Util is nil")
+    }
+    if util.Vendors.isEmpty() {
```
mfulb, line 222: Do you need to check that `util.Vendors` is not `nil`?  Above when you create these structs you allocation `util.Vendors` separately from the `util` struct.

src/newrelic/utilization/utilization_hash.go
```diff
@@ -204,6 +199,38 @@ func GatherDockerID(util *Data) error {
...
+func OverrideDockerId(util *Data, id string) {
+	if nil == util.Vendors {
+		util.Vendors = &vendors{}
+		util.Vendors.Docker = &docker{}
+	}
+	util.Vendors.Docker = &docker{ID: id}		
+}
+
+func OverrideVendors(util *Data) {
+	if util.Vendors.isEmpty() {
```
mfulb, line 211: Should this have checks for util != nil and util.Vendors != nil?

src/newrelic/app_test.go
bduranleau-nr: This test passes a non-nil value for the utilization data object. It's testing the expected path for sending the Docker ID in the connect message.

src/newrelic/app_test.go
bduranleau-nr: The above test passes `nil` for the `util` object, this test passes a non-`nil` `util` object.

src/newrelic/app_test.go
bduranleau-nr: I debated checking that value- `GetDockerId` is just a utility/getter function that I defined specifically just to make getting the value easier by passing a `util` object and returning a `string` and optionally an `error`. The `error` in this case is being ignored, since it's just an error message I came up with for each scenario. I can include checks for it, but it would be testing the `GetDockerId` function moreso than the `ConnectPayloadInternal` function's handling of DockerId, so I left those checks out.

src/newrelic/utilization/utilization_hash.go
bduranleau-nr: I believe the logic to check if `Vendors == nil` is handled implicitly by the call to the isEmpty function here: https://github.com/newrelic/newrelic-php-agent/blob/main/src/newrelic/utilization/utilization_hash.go#L70

src/newrelic/utilization/utilization_hash.go
bduranleau-nr: I believe the logic to check if `Vendors == nil` is handled implicitly by the call to the isEmpty function here: https://github.com/newrelic/newrelic-php-agent/blob/main/src/newrelic/utilization/utilization_hash.go#L70

src/newrelic/app_test.go
mfulb: Is it reasonable to expect the error is predictable in this test if it succeeds.  If so I think it can't hurt to check it.  If the test fails it would indicate something went wrong with the test or the function itself.  Something I think we'd like to know about.

src/newrelic/utilization/utilization_hash.go
mfulb: If 'util' or `util.Vendors` is nil how can one call `util.Vendors.isEmpty()` w/o an error?

mfulb: Ok it makes sense that in golang you can run the functions for a data type on a data type value of `nil` so no changes are needed to check for `utils.Vendors`.  
However, if `util` is `nil` I think it could cause a problem since accessing `util.Vendors` will cause an error.

src/newrelic/utilization/utilization_hash.go
mfulb: OK this should work with how golang handles data types and functions for data types.

src/newrelic/utilization/utilization_hash.go
mfulb: See how you do it below in `GetDockerId()` where you return early if `util` is `nil`.

src/newrelic/utilization/utilization_hash.go
```diff
@@ -204,6 +199,38 @@ func GatherDockerID(util *Data) error {
 	return nil
 }
 
+func OverrideDockerId(util *Data, id string) {
+	if nil == util.Vendors {
```
mfulb, line 203: I think you need a `nil` check for `util` here as well.

src/newrelic/app_test.go
mfulb: Where is `util` set to non-nil? It is not clear in the test case.

agent/php_environment.c
```diff
@@ -511,6 +511,124 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+  // check if file exists
+  if (SUCCESS != access(cgroup_fname, F_OK)) {
+    nrl_verbosedebug(NRL_AGENT, "File not found: %s", cgroup_fname);
+    return NULL;
+  }
+
+  // open file
+  fd = fopen(cgroup_fname, "r");
+  if (NULL == fd) {
+    nrl_warning(NRL_AGENT, "Failed to open %s", cgroup_fname);
```
zsistla, line 547: Regarding all the debug statements:
It might make more sense to either add a little more detail to the msg or tag it with the function so it's immediately clear the context of what we were opening and why it matters that it failed.
Here's an example of tagging with the function name:
`nrl_verbosedebug(NRL_FRAMEWORK, "%s: cannot obtain 'this'", __func__);`

src/newrelic/app_test.go
mfulb: I'm not seeing how `util` is non-nil in the second test.

mfulb: I see how this code works now - I was overlooking the `nil` arg.

src/newrelic/app_test.go
mfulb: I see how this code works now - I was overlooking the `nil` arg.

agent/php_environment.c
```diff
@@ -511,6 +511,124 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+   *    b. comprised of only hexadecimal characters
+   *  6. Assign the extracted & verified ID to the retval
+   *    a. Example ID: ec807d5258c06c355c07e2acb700f9029d820afe5836d6a7e19764773dc790f5
+   *  7. Set found = true and exit the loops
+   */
+  // clang-format on
+
+  while (FAILURE != getline(&line_ptr, &len, fd) && !found
+         && line_count++ < MAX_LINE_COUNT) {
+    token = strtok_r(line_ptr, "/", &saveptr);
```
zsistla, line 586: Is it more efficient to tokenize the line by `/` vs making the regex something like:
`regex = nr_regex_create("^/docker/containers/([a-fA-F0-9]{DOCKER_ID_V2_STRLEN})/", 0, 0);`

Then using [nr_regex_match_capture and nr_regex_substrings_get](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/tests/test_regex.c#L145) to get the ID?

It could save the tokenizing step and the strlen comparison as well.

bduranleau-nr: `nr_regex_match_capture` requires a `const char* str` as it's second argument- what would that be if we don't know the actual string value of the id in advance? Looking at the examples, it seems they all require an explicit string containing the value to be passed, is this not the case?

zsistla: You can pass the modified regex and line_ptr in and match_capture will pull a valid docker id out.

[Wordpress](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_wordpress.c#L436) uses this as well.

agent/php_environment.c
bduranleau-nr: Added function name to debug statements in ff5cb1995e0b7f9815741f24eff316185c2de64e

agent/php_environment.c
bduranleau-nr: Ok I think I understand- my previous understanding of this function family was incorrect. I'll investigate this implementation and see if I can coerce it into to working for me

agent/php_environment.c
zsistla: Thanks!

agent/php_environment.c
```diff
@@ -511,6 +511,124 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+    nrl_warning(NRL_AGENT, "Failed to open %s", cgroup_fname);
+    return NULL;
+  }
+
+  // compile regex to verify hexadecimal chars
+  regex = nr_regex_create("^[a-fA-F0-9]+$", 0, 0);
+
+  if (NULL == regex) {
+    nrl_error(NRL_AGENT, "Error: regex creation failed");
+    return NULL;
```
zsistla, line 556: need an `fclose(fd);` before the return.

agent/php_environment.c
bduranleau-nr: Valgrind really despises trying to pass off the `line_ptr` as a `const char*` and is complaining about "Conditional jump depends on uninitialized values" in `nr_regex_match_capture`

zsistla: How are you declaring/passing in line_ptr?

The wordpress example just declares a `char *` [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_wordpress.c#L422) then passes it in [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_wordpress.c#L437)

Maybe it would prefer you to [strdup](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_wordpress.c#L437) it first?

src/newrelic/app_test.go
```diff
@@ -59,10 +60,14 @@ func TestConnectPayloadInternal(t *testing.T) {
 	}
 
 	expected.Util.Hostname = info.Hostname
+	utilization.OverrideDockerId(expected.Util, info.DockerId)
```
mfulb, line 63: I don't think this is necessary or a good idea to have the expected depend on a function.

src/newrelic/app_test.go
```diff
@@ -59,10 +60,14 @@ func TestConnectPayloadInternal(t *testing.T) {
...
 
 	pid := 123
 	b := info.ConnectPayloadInternal(pid, util)
 
+	expectedDockerId, _ := utilization.GetDockerId(expected.Util)
```
mfulb, line 68:
```suggestion
	expectedDockerId, _ := "1056761e1f44969c959364a8e26e9345b37ccb91aef09a8173c90cf1d1d99156"
```

agent/php_environment.c
bduranleau-nr: Tried passing `line_ptr` normally, tried casting it to `const char*`, tried `strdup`/`strndup`ing it, no luck. 
I think the critical difference between this case and the one in wordpress is that the `char *` is still based on the `const` zval `tag` argument.

src/newrelic/app_test.go
bduranleau-nr: It is necessary- I can't assign the `vendor` struct to `expected.Util` from outside the utilization package. If I can't call this function to set the docker id in the `vendor` struct, I have to remove this test entirely.

src/newrelic/app_test.go
bduranleau-nr: I think if we go this route we aren't really testing anything different from the `TestConnectPayloadInternalDocker` test

agent/php_environment.c
bduranleau-nr: Commenting for posterity: The issue was with trying to use the `len` value set by the `getline` function. Using an explicit `nr_strlen` on the `line_ptr` instead solved the valgrind issue.

agent/php_environment.c
bduranleau-nr: Fixed in 65bd871c117f5dddde7813c8e3f1aec9c04558e7

agent/php_environment.c
bduranleau-nr: Fixed in 65bd871c117f5dddde7813c8e3f1aec9c04558e7

agent/php_environment.c
```diff
@@ -511,6 +511,110 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+   *  1. scan file line-by-line
+   *  2. regex search each line for '/docker/containers/' string followed by a 64 byte hex value
+   *  3. extract the 64 byte field following "/docker/containers/"
+   *  4. Verify the extracted ID is both:
+   *    a. 64 bytes long (not including null terminator)
+   *    b. comprised of only hexadecimal characters
+   *  5. Assign the extracted & verified ID to the retval
+   *    a. Example ID: ec807d5258c06c355c07e2acb700f9029d820afe5836d6a7e19764773dc790f5
+   *  6. Set found = true and exit the loops
+   */
```
zsistla, line 578: This needs to be updated, 4a and 4b and now part of what happens in #2.

bduranleau-nr: Updated in 437bd73dadb85c0c665ed58ab20e89b7176db976

agent/php_environment.c
```diff
@@ -511,6 +511,110 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+    ss = nr_regex_match_capture(line_regex, line_ptr, nr_strlen(line_ptr));
+    if (NULL == ss) {
+      continue;
+    }
+    retval = nr_regex_substrings_get(ss, 1);
+    found = true;
+  }
+
+  nr_regex_destroy(&line_regex);
+  nr_regex_substrings_destroy(&ss);
```
zsistla, line 591: This looks SOOO much cleaner, great job!

Technically we can destroy it here because once we've found it we exist (since we set found to true) so regex_match_capture would only have been called once, but it might be better just to clean it after the retval is set since every regex_match_capture call creates an object that needs to be destroyed.

bduranleau-nr: Like this? be0cd8a4094d5e40ac5dc093f117a4af2dd2b239

src/newrelic/app_test.go
mfulb: Maybe this is redundant then as I don't think setting the expected value by calling `OverrideDockerId` is what we want.  
My understanding is this test is verifying `ConnectPayloadInternal()` will get the proper Docker Id when a value is present in the `info` struct.  Is that what the `TestConnectPayloadInternalDocker` test does?

bduranleau-nr: > My understanding is this test is verifying ConnectPayloadInternal() will get the proper Docker Id when a value is present in the info struct. Is that what the TestConnectPayloadInternalDocker test does?
Yes that's pretty much what the other test does. We can keep this though with your suggestions since this test has a more substantial info struct

src/newrelic/app_test.go
bduranleau-nr: See c28cb6775c1128750e7bb9b523fd7b1164bfdd52

src/newrelic/app_test.go
bduranleau-nr: See c28cb6775c1128750e7bb9b523fd7b1164bfdd52

src/newrelic/app_test.go
bduranleau-nr: added in 30f4a34c7813a4015cf8417b15edfe7524e7dc01

mfulb: Nice!

mfulb (approved):
Nice work on adding this needed feature!

agent/php_environment.c
```diff
@@ -511,6 +511,109 @@ static void nr_php_get_environment_variables(TSRMLS_D) {
...
+   *  3. extract the 64 byte substring following "/docker/containers/"
+   *  4. Assign the extracted & verified ID to the retval
+   *    a. Example ID: ec807d5258c06c355c07e2acb700f9029d820afe5836d6a7e19764773dc790f5
+   *  5. Set found = true and exit the loops
+   */
+  // clang-format on
+
+  while (FAILURE != getline(&line_ptr, &len, fd) && !found
+         && line_count++ < MAX_LINE_COUNT) {
+    ss = nr_regex_match_capture(line_regex, line_ptr, nr_strlen(line_ptr));
```
lavarou, line 582: Is the call to `nr_strlen(line_ptr)` required? Can't the value of `len` be used here?

lavarou (approved):
LGTM

agent/php_environment.c
lavarou: Missed that thread. Thanks!

zsistla (approved):

PR #767, opened 2023-11-14
feat(oapi): rework exception handling

ZNeumann:
Removing the need to manual handling of dangling segments, because Zend calls all of the hooks we need.

- adds a boolean argument to nr_php_error_record_exception to control whether we add the error to the current segment. This is needed because the OAPI context of when the above is called is no longer during a segment with an uncaught exception and was incorrectly adding the error to the root segment.
- removes all language of "dangling segments". These no longer exist. Zend calls all of the necessary `fcall_end`'s, even when an exception is thrown. When this happens,` func_return_value` is a C `NULL` pointer which is distinct from a `NULL` (but valid) zval when there is no return value from a function. We use this `NULL` value to determine the presence of an uncaught exception.
- no longer overwrites the exception hook; no longer stores a copy of exceptions locally
- replace storing metadata in the segment, which was used to pair segments from func_begin with func_end, with logic that always creates segment in func_begin

Mostly undoes the work of https://github.com/newrelic/newrelic-php-agent/pull/580


agent/php_execute.c
```diff
@@ -2071,13 +1876,7 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
     nr_php_observer_attempt_call_cufa_handler(NR_EXECUTE_ORIG_ARGS);
   }
   wraprec = nr_php_get_wraprec(execute_data->func);
-  /*
```
zsistla, line 2074: Why is this removed?

agent/php_execute.c
```diff
@@ -2071,13 +1876,7 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
...
-   * If there is custom instrumentation or tt detail is more than 0, start the
-   * segment.
-   */
-  if ((NULL == wraprec) && !(NRINI(tt_detail) && NR_OP_ARRAY->function_name)) {
-    return;
-  }
+
   /*
    * Check if it's a custom error handler.  Even with some custom error
    * handlers, fcall might not get called. But we don't need to wait for
```
zsistla, line 1882: This isn't true anymore,right?

agent/php_execute.c
ZNeumann: Because the metadata is no longer stored in the segment, we need to always start a segment. Previously, when we got to fcall_end, we were able to compare the metadata to the execute_data and see if a segment was started for the given execute_data.

Without the metadata and without always creating a segment, we would have no way of determining whether the active segment was actually for the given fcall_end

agent/php_execute.c
```diff
@@ -2290,19 +2017,34 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
   if (NULL != wraprec) {
     /*
      * This is the case for specifically requested custom instrumentation.
      */
     create_metric = wraprec->create_metric;
 
-    zcaught = nr_zend_call_orig_execute_special(wraprec, segment,
+    /*
+     * A NULL return value means that there was an uncaught exception, and
+     * therefore we want to call the 'clean' function type
```
zsistla, line 2029:
```suggestion
     * A NULL return value ptr means that there was an uncaught exception
     * and therefore we want to call the 'clean' function type
```

zsistla (commented):
Great job, looks like a clean rework.

agent/php_execute.c
ZNeumann: This needs to be investigated, because without these checks, the sheer amount of segment initializations and deinits it too computationally costly.

agent/php_error.h
```diff
@@ -137,6 +137,7 @@ extern void nr_php_error_install_exception_handler(TSRMLS_D);
 extern nr_status_t nr_php_error_record_exception(nrtxn_t* txn,
                                                  zval* exception,
                                                  int priority,
+                                                 bool add_to_segment,
```
lavarou, line 140: Please update doc string with explanation of the new parameter.

axiom/nr_txn.h
```diff
@@ -468,6 +468,7 @@ extern nr_status_t nr_txn_record_error_worthy(const nrtxn_t* txn, int priority);
 
 extern void nr_txn_record_error(nrtxn_t* txn,
                                 int priority,
+                                bool add_to_segment,
```
lavarou, line 471: Please update doc string with explanation of the new parameter.

agent/fw_laravel.c
```diff
@@ -619,7 +619,7 @@ NR_PHP_WRAPPER(nr_laravel5_exception_report) {
       nr_status_t st;
 
       st = nr_php_error_record_exception(NRPRG(txn), exception, priority,
-                                         NULL /* use default prefix */,
+                                         true, NULL /* use default prefix */,
```
lavarou, line 622: With the new argument added, this comment is now confusing. Is this comment about 'true' or 'NULL'?

agent/php_error.c
```diff
@@ -645,6 +645,7 @@ void nr_php_error_cb(int type,
 nr_status_t nr_php_error_record_exception(nrtxn_t* txn,
                                           zval* exception,
                                           int priority,
+                                          bool add_to_segment,
                                           const char* prefix,
                                           zend_llist* filters TSRMLS_DC) {
```
bduranleau-nr, line 650: Is there a "most common" use case in terms of whether this value should be true or false? I did a quick scan and it looks like only one invocation uses a `false` value. You may be able to avoid a lot of the argument-specific changes by turning `nr_php_error_record_exception` into a wrapper for an internal function where all the logic lives and the value passed is always true, and a sibling wrapper that where the value is always false. For cases where we want the `true` value, nothing would have to change. For cases where we want the false value, the deviant behavior could be captured in the function name itself i.e. `nr_php_error_record_exception_no_segment`.
Both of these functions would call some kind of `nr_php_error_record_exception_ex` with the `true`/`false` logic handled in the wrapper.
I think this would have the following effects:
1. No need to change the function call where the desired value is `true`.
2. The ability to give a detailed explanation of when to call each function in the function doc description that will be the focus of the wrappers rather than buried in the argument field.
3. Ease-of-reading for anyone looking at this code later- a `true`/`false` argument is hard to understand at a glance while a function that describes what it does in its name is eas(ier) (in theory).

agent/tests/test_php_execute.c
```diff
@@ -287,7 +269,7 @@ static void test_stack_depth_after_exception() {
   arg = tlib_php_request_eval_expr("0");
   expr = nr_php_call(NULL, "followup_uncaught", arg);
   tlib_pass_if_int_equal(
-      "PHP stack depth tracking should be 2 after function call", 2,
+      "PHP stack depth tracking should be 2 after function call", 0,
```
bduranleau-nr, line 272: Should this unit test description string be updated to `0` as well?
```suggestion
      "PHP stack depth tracking should be 0 after function call", 0,
```

agent/tests/test_php_execute.c
```diff
@@ -317,7 +297,7 @@ static void test_stack_depth_after_exception() {
   arg = tlib_php_request_eval_expr("0");
   expr = nr_php_call(NULL, "rethrow", arg);
   tlib_pass_if_int_equal(
-      "PHP stack depth tracking should be 2 after function call", 1,
+      "PHP stack depth tracking should be 2 after function call", 0,
```
bduranleau-nr, line 300:
```suggestion
      "PHP stack depth tracking should be 0 after function call", 0,
```
Does it even make sense to test for stack depth anymore considering your changes in this PR?

axiom/nr_txn.c
```diff
@@ -1549,6 +1549,7 @@ nr_status_t nr_txn_record_error_worthy(const nrtxn_t* txn, int priority) {
 
 void nr_txn_record_error(nrtxn_t* txn,
                          int priority,
+                         bool add_to_segment,
                          const char* errmsg,
                          const char* errclass,
                          const char* stacktrace_json) {
```
bduranleau-nr, line 1555: Similar feedback to [this comment](https://github.com/newrelic/newrelic-php-agent/pull/767/files#r1437898715) regarding the potential to use specialized wrappers instead of adding bools to each function call.

agent/php_execute.c
```diff
@@ -2244,39 +1982,28 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
-  /*
-   * If the metadata info doesn't match, an uncaught exception occurred which
-   * doesn't call fcall_end.
-   */
-  metadata = segment->metadata;
-  if ((metadata->execute_data_this != &execute_data->This)) {
+
+  wraprec = segment->wraprec;
+
+  if (wraprec && wraprec->is_exception_handler) {
```
bduranleau-nr, line 1991: This `if` statement doesn't seem to contain anything anymore and should be removed

agent/php_execute.c
```diff
@@ -2290,19 +2017,34 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
      * This is the case for specifically requested custom instrumentation.
      */
     create_metric = wraprec->create_metric;
 
-    zcaught = nr_zend_call_orig_execute_special(wraprec, segment,
+    /*
+     * A NULL return value means that there was an uncaught exception, and
+     * therefore we want to call the 'clean' function type
+     */
+    if (NULL != nr_php_get_return_value(NR_EXECUTE_ORIG_ARGS TSRMLS_CC)) {
```
bduranleau-nr, line 2031:
```suggestion
    if (NULL != nr_php_get_return_value(NR_EXECUTE_ORIG_ARGS)) {
```
`TSRM` macros aren't necessary anymore

agent/php_execute.c
```diff
@@ -2244,39 +1982,28 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
-    }
-    metadata = segment->metadata;
-    if (nrunlikely(metadata->execute_data_this != &execute_data->This)) {
-      /*
-       * Sanity check.
-       * If the pointers still aren't equal, let's exit.
-       *
-       */
-      return;
+  } else if (NULL == nr_php_get_return_value(NR_EXECUTE_ORIG_ARGS TSRMLS_CC)) {
```
bduranleau-nr, line 1997:
```suggestion
  } else if (NULL == nr_php_get_return_value(NR_EXECUTE_ORIG_ARGS)) {
```

agent/php_execute.c
```diff
@@ -2244,39 +1982,28 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
+    zval exception;
+    ZVAL_OBJ(&exception, EG(exception));
+    nr_status_t status = nr_php_error_record_exception_segment(
+      NRPRG(txn), &exception,
+      &NRPRG(exception_filters) TSRMLS_CC);
```
bduranleau-nr, line 2002:
```suggestion
      &NRPRG(exception_filters));
```

agent/php_execute.c
ZNeumann: My original reply still holds true and the performance cost has been addressed in my other PRs. (https://github.com/newrelic/newrelic-php-agent/pull/776 and https://github.com/newrelic/newrelic-php-agent/pull/801)

agent/php_execute.c
ZNeumann: Fixed the language in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3)

agent/php_error.h
ZNeumann: updated in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3)

agent/tests/test_php_execute.c
```diff
@@ -181,17 +173,15 @@ static void test_stack_depth_after_exception() {
...
+   * call a function and trigger an exception
    */
   tlib_pass_if_int_equal(
       "PHP stack depth tracking should be 0 before function call", 0,
       NRPRG(php_cur_stack_depth));
   arg = tlib_php_request_eval_expr("0");
   expr = nr_php_call(NULL, "uncaught", arg);
   tlib_pass_if_int_equal(
-      "PHP stack depth tracking should be 3 after function call", 3,
+      "PHP stack depth tracking should be 3 after function call", 0,
```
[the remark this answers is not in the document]
ZNeumann: fixed in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3)

axiom/nr_txn.h
ZNeumann: updated in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3)

axiom/nr_txn.c
```diff
@@ -1594,7 +1595,7 @@ void nr_txn_record_error(nrtxn_t* txn,
 
     current_segment = nr_txn_get_current_segment(txn, NULL);
 
-    if (current_segment) {
+    if (current_segment && add_to_segment) {
```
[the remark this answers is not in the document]
ZNeumann: addressed in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3)

agent/tests/test_php_execute.c
ZNeumann: changed in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3)

I believe that we still want to test and track stack depth.

agent/php_execute.c
ZNeumann: I added clarifying comments in [bce3e14](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/bce3e1462db26bb5f841542c199dae92ba96a1b3). I had this split off as an empty if statement (as opposed to combining this check with the `else if` check) to explicitly call out each case and have an explaining comment with each case. Do you still want the if's combined?

bduranleau-nr: Having a hollow conditional doesn't quite sit right with me. We don't have to combine the `if`'s necessarily if that's something you want to avoid, we could invert the logic and have the else if/else statements be nested within, right? Something like: 
```
if (!wraprec || !wraprec->is_exception_handler) {...}
```
Then just nest the `else if` statement as an `if` within it?

agent/php_error.c
ZNeumann: It has been decided to change this. Adding additional signatures for the same function clutters the API and makes knowing what is/isn't possible when reading or writing code increasingly difficult. Further, we do not hold any other argument type to this same standard (nor should we). For example, passing `NULL` in for the `prefix` argument is just as difficult to understand as `false` to the `add_to_segment` argument, yet we are not asking for a `nr_php_error_record_exception_no_prefix `signature.

lavarou (commented):
Here's my initial feedback - just a few comments so far.

agent/php_execute.c
```diff
@@ -2227,12 +1966,11 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
     nr_php_execute_file(NR_OP_ARRAY, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
-    php_observer_clear_uncaught_exception_globals();
     return;
   }
 
   /*
-   * Get the current segment and return if null.  The segment would only have
+   * Get the current segment and return if null. The segment would only have
    * been created if we are recording and if wraprec is set or if tt is greater
    * than 0.
```
lavarou, line 1975: This comment is no longer true, right? When Observer API is used to hook into Zend Engine, the segment is always created in `nr_php_instrument_func_begin` [here](https://github.com/newrelic/newrelic-php-agent/blob/cbeb5ee3eea483d9ccdf7bb6dc0d7cd62f5d0f3d/agent/php_execute.c#L1903).

agent/php_execute.c
ZNeumann: see https://github.com/newrelic/newrelic-php-agent/pull/825

axiom/styXoU99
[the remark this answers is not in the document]
ZNeumann: fixed in [d2c2d8a](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe)

axiom/tests/test_txn.c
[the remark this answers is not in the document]
ZNeumann: added in [d2c2d8a](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe)

agent/php_error.h
```diff
@@ -137,6 +138,7 @@ extern void nr_php_error_install_exception_handler(TSRMLS_D);
...
```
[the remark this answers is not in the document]
ZNeumann: changed in [d2c2d8a](https://github.com/newrelic/newrelic-php-agent/pull/767/commits/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe)

bduranleau-nr (approved):

axiom/tests/test_txn.c
```diff
@@ -8013,7 +8013,17 @@ static void test_segment_record_error(void) {
   tlib_pass_if_null("No segment error created", segment->error);
   txn->options.err_enabled = 1;
 
-  /* Normal operation */
+  /* Do not add to current segment */
+  nr_txn_record_error(txn, 0.5, false /* do not add to current segment*/,
+                      "low priority message", "low priority class", "[\"A\",\"B\"]");
+  tlib_pass_if_not_null("Txn error event created", txn->error);
+  tlib_pass_if_null("Segment error NOT created", segment->error);
```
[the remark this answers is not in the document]
ZNeumann: I think it is 100% certain. The segment is the only segment started on this transaction, and is assumed to be the current segment throughout this test. Later on, we are testing that this very same segment DOES get an error attached, without any added segment manipulation code.

lavarou (approved):
Really nice piece of work!!! 👏

PR #768, opened 2023-11-14
fix(agent): port the functionality to bypass wordpress cufa calls by wrapping add_filter

zsistla:
Port the work previously done here: reference: https://github.com/newrelic/newrelic-php-agent/pull/731

Bypass the need to check for cufa calls during do_action/filter by instrumenting add_action/filter and wrapping the desired calls at hook time, rather than call time. Substantially improves performance and stability.

### DIFF
--- a/agent/fw_wordpress.c
+++ b/agent/fw_wordpress.c
@@ -578,6 +578,63 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
 }
 NR_PHP_WRAPPER_END
 
+/*
+ * Wrap the wordpress function add_filter
+ * 
+ * function add_filter( $hook_name, $callback, $priority = 10, $accepted_args = 1 )
+ *
+ * @param string   $hook_name     The name of the filter to add the callback to.
+* @param callable $callback      The callback to be run when the filter is applied.
+* @param int      $priority      Optional. Used to specify the order in which the functions
+*                                associated with a particular filter are executed.
+*                                Lower numbers correspond with earlier execution,
+*                                and functions with the same priority are executed
+*                                in the order in which they were added to the filter. Default 10.
+* @param int      $accepted_args Optional. The number of arguments the function accepts. Default 1.
+* @return true Always returns true.
+ */
+
+NR_PHP_WRAPPER(nr_wordpress_add_filter) {
+  /* Wordpress's add_action() is just a wrapper around add_filter(),
+   * so we only need to instrument this function */
+  NR_UNUSED_SPECIALFN;
+  (void)wraprec;
+  const char* filename = NULL;
+  bool wrap_hook = true;
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);
+
+  /*
+   * We only want to hook the function being called if this is a WordPress
+   * function, we're instrumenting hooks, and WordPress is currently executing
+   * hooks (denoted by the wordpress_tag being set).
+   */
+  if ((NR_FW_WORDPRESS != NRPRG(current_framework))
+      || (0 == NRINI(wordpress_hooks))) {
+    wrap_hook = false;
+  }
+
+  if (NRINI(wordpress_hooks_skip_filename)
+      && 0 != nr_strlen(NRINI(wordpress_hooks_skip_filename))) {
+    filename = nr_php_op_array_file_name(NR_OP_ARRAY);
+
+    if (nr_strstr(filename, NRINI(wordpress_hooks_skip_filename))) {
+      nrl_verbosedebug(NRL_FRAMEWORK, "skipping hooks for function from %s",
+                       filename);
+      wrap_hook = false;
+    }
+  }
+
+  if (true == wrap_hook) {
+    zval* callback = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+    /* the callback here can be any PHP callable. nr_php_wrap_generic_callable
+   * checks that a valid callable is passed */
+    nr_php_wrap_generic_callable(callback, nr_wordpress_wrap_hook);
+    nr_php_arg_release(&callback);
+  }
+}
+NR_PHP_WRAPPER_END
+
 void nr_wordpress_enable(TSRMLS_D) {
   nr_php_wrap_user_function(NR_PSTR("apply_filters"),
                             nr_wordpress_apply_filters TSRMLS_CC);
@@ -590,7 +647,7 @@ void nr_wordpress_enable(TSRMLS_D) {
 
   nr_php_wrap_user_function(NR_PSTR("do_action_ref_array"),
                             nr_wordpress_exec_handle_tag TSRMLS_CC);
+  nr_php_wrap_user_function(NR_PSTR("add_filter"),
+                            nr_wordpress_add_filter);
 
-  nr_php_add_call_user_func_array_pre_callback(
-      nr_wordpress_call_user_func_array TSRMLS_CC);
 }
--- a/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
+++ b/tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
@@ -1,0 +1,88 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should properly instrument Wordpress apply_filters hooks.
+*/
+
+/*SKIPIF
+<?php
+if (version_compare(PHP_VERSION, "5.6", "<")) {
+  die("skip: PHP < 5.6 argument unpacking not supported\n");
+}
+*/
+
+/*INI
+newrelic.framework = wordpress
+*/
+
+/*EXPECT
+add filter
+add filter
+add filter
+f: string1
+add filter
+h: string3
+g: string2
+*/
+
+/*EXPECT_METRICS
+[
+  "?? agent run id",
+  "?? start time",
+  "?? stop time",
+  [
+    [{"name": "DurationByCaller/Unknown/Unknown/Unknown/Unknown/all"},      [1, "??", "??", "??", "??", "??"]],
+    [{"name": "DurationByCaller/Unknown/Unknown/Unknown/Unknown/allOther"}, [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Framework/WordPress/Hook/f"},                          [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Framework/WordPress/Hook/g"},                          [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Framework/WordPress/Hook/h"},                          [1, "??", "??", "??", "??", "??"]],
+    [{"name": "OtherTransaction/all"},                                [1, "??", "??", "??", "??", "??"]],
+    [{"name": "OtherTransaction/php__FILE__"},                        [1, "??", "??", "??", "??", "??"]],
+    [{"name": "OtherTransactionTotalTime"},                           [1, "??", "??", "??", "??", "??"]],
+    [{"name": "OtherTransactionTotalTime/php__FILE__"},               [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Supportability/Logging/Forwarding/PHP/enabled"},       [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Supportability/Logging/Metrics/PHP/enabled"},          [1, "??", "??", "??", "??", "??"]],
+    [{"name":"Supportability/Logging/LocalDecorating/PHP/disabled"},  [1, "??", "??", "??", "??", "??"]],
+    [{"name": "Supportability/framework/WordPress/forced"},           [1, "??", "??", "??", "??", "??"]]
+  ]
+]
+*/
+
+require_once __DIR__.'/mock_hooks.php';
+
+function h($str) {
+    echo "h: ";
+    echo $str;
+    echo "\n";
+    throw new Exception("Test Exception");
+}
+
+function g($str) {
+    echo "g: ";
+    echo $str;
+    echo "\n";
+}
+
+function f($str) {
+    echo "f: ";
+    echo $str;
+    echo "\n";
+    // For OAPI: attempt to overwrite the currently executing transient wrapper
+    add_filter("hook", "f");
+    try {
+        apply_filters("h", "string3");
+    } catch (Exception $e) {
+        apply_filters("g", "string2");
+    }
+}
+
+// Due to the mock simplification described above, the hook
+// is not used in this test, and the callback is treated as the hook
+add_filter("hook", "f");
+add_filter("hook", "g");
+add_filter("hook", "h");
+apply_filters("f", "string1");
--- a/tests/integration/frameworks/wordpress/test_wordpress_do_action.php
+++ b/tests/integration/frameworks/wordpress/test_wordpress_do_action.php
@@ -1,0 +1,92 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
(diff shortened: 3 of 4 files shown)


agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+
+    if (nr_strstr(filename, NRINI(wordpress_hooks_skip_filename))) {
+      nrl_verbosedebug(NRL_FRAMEWORK, "skipping hooks for function from %s",
+                       filename);
+      wrap_hook = false;
+    }
+  }
+
+  if (true == wrap_hook) {
+    zval* callback = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
bduranleau-nr, line 632:
```suggestion
    zval* callback = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS);
```

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
```diff
@@ -0,0 +1,88 @@
...
+
+/*DESCRIPTION
+The agent should properly instrument Wordpress apply_filters hooks.
+*/
+
+/*SKIPIF
+<?php
+if (version_compare(PHP_VERSION, "5.6", "<")) {
+  die("skip: PHP < 5.6 argument unpacking not supported\n");
+}
```
bduranleau-nr, line 15: Since we don't build/support PHP < 7.0 we should be able to remove this SKIPIF entirely

tests/integration/frameworks/wordpress/test_wordpress_do_action.php
```diff
@@ -0,0 +1,92 @@
...
+
+/*DESCRIPTION
+The agent should properly instrument Wordpress do_action hooks.
+*/
+
+/*SKIPIF
+<?php
+if (version_compare(PHP_VERSION, "5.6", "<")) {
+  die("skip: PHP < 5.6 argument unpacking not supported\n");
+}
```
bduranleau-nr, line 15: Since we don't build/support PHP < 7.0 we should be able to remove this SKIPIF entirely

agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+ *
+ * @param string   $hook_name     The name of the filter to add the callback to.
+* @param callable $callback      The callback to be run when the filter is applied.
+* @param int      $priority      Optional. Used to specify the order in which the functions
+*                                associated with a particular filter are executed.
+*                                Lower numbers correspond with earlier execution,
+*                                and functions with the same priority are executed
+*                                in the order in which they were added to the filter. Default 10.
+* @param int      $accepted_args Optional. The number of arguments the function accepts. Default 1.
+* @return true Always returns true.
```
lavarou, line 597: Fix off-by-one indent:
```suggestion
 * @param callable $callback      The callback to be run when the filter is applied.
 * @param int      $priority      Optional. Used to specify the order in which the functions
 *                                associated with a particular filter are executed.
 *                                Lower numbers correspond with earlier execution,
 *                                and functions with the same priority are executed
 *                                in the order in which they were added to the filter. Default 10.
 * @param int      $accepted_args Optional. The number of arguments the function accepts. Default 1.
 * @return true Always returns true.
```

agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+ */
+
+NR_PHP_WRAPPER(nr_wordpress_add_filter) {
+  /* Wordpress's add_action() is just a wrapper around add_filter(),
+   * so we only need to instrument this function */
```
lavarou, line 602: Could this comment be added to the PR description?

agent/fw_wordpress.c
```diff
@@ -376,8 +376,9 @@ NR_PHP_WRAPPER(nr_wordpress_wrap_hook) {
 NR_PHP_WRAPPER_END
 
 /*
- * A call_user_func_array() callback to ensure that we wrap each hook function.
+ * PHP 7.3 and below uses old-style wraprec, and will use old style cufa calls.
```
lavarou, line 379: Could you elaborate why `add_filter` cannot be wrapped for PHPs < 7.4?

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
```diff
@@ -0,0 +1,88 @@
...
+add filter
+add filter
+add filter
+f: string1
+add filter
+h: string3
+g: string2
+*/
+
+/*EXPECT_METRICS
```
lavarou, line 32: Would this be a good place to use `EXPECT_METRICS_EXIST` added in #771 to make the test more focused. This test only cares about following metrics being created:
`Framework/WordPress/Hook/f`, `Framework/WordPress/Hook/g`, `Framework/WordPress/Hook/h`.

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
```diff
@@ -0,0 +1,88 @@
...
+    try {
+        apply_filters("h", "string3");
+    } catch (Exception $e) {
+        apply_filters("g", "string2");
+    }
+}
+
+// Due to the mock simplification described above, the hook
+// is not used in this test, and the callback is treated as the hook
+add_filter("hook", "f");
```
lavarou, line 85: I know how the simplification works (and probably having a lot of context around WordPress instrumentation now helps), however it still confuses me (and in the future without additional context it will be even more confusing). For this code:
```
add_filter("hook", "g");
```
WordPress's instrumentation is expected to generate the following metric:
`Framework/WordPress/Hook/hook`, which accumulates totals for all callbacks registered under this hook. However, in this test the following metric will be generated `Framework/WordPress/Hook/f`! This expectation may surprise our future selves.

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should properly instrument Wordpress apply_filters hooks.
```
lavarou, line 8:
```suggestion
The agent should properly instrument Wordpress apply_filters hooks and generate
`Framework/WordPress/Hook/<hook_name>` metrics for all detected hooks.
```

agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+  bool wrap_hook = true;
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);
+
+  /*
+   * We only want to hook the function being called if this is a WordPress
+   * function, we're instrumenting hooks, and WordPress is currently executing
+   * hooks (denoted by the wordpress_tag being set).
+   */
+  if ((NR_FW_WORDPRESS != NRPRG(current_framework))
```
lavarou, line 615: This is already ensured by `NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);`

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
lavarou: However even the updated comment doesn't even give clear picture. WordPress instrumentation not only generates `Framework/WordPress/Hook/<hook_name>` metrics but also `Framework/WordPress/Plugin/<plugin_or_theme_name_that_hooks_onto_hook_name>` if the agent is able to convert `$callback` into `plugin_or_theme_name`. This metric will not be generated by this test because the agent is not able to convert `$callback` into `plugin_or_theme_name`.

agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+      || (0 == NRINI(wordpress_hooks))) {
```
lavarou, line 616: I think this check needs to be added in `oapi` branch too.

agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
 }
 NR_PHP_WRAPPER_END
 
+#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+/*
+ * Wrap the wordpress function add_filter
```
lavarou, line 585:
```suggestion
 * Wrap the wordpress function add_filter in order to wrap the `$callback`.
```

agent/fw_wordpress.c
```diff
@@ -578,6 +580,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+  if (NRINI(wordpress_hooks_skip_filename)
+      && 0 != nr_strlen(NRINI(wordpress_hooks_skip_filename))) {
+    filename = nr_php_op_array_file_name(NR_OP_ARRAY);
+
+    if (nr_strstr(filename, NRINI(wordpress_hooks_skip_filename))) {
+      nrl_verbosedebug(NRL_FRAMEWORK, "skipping hooks for function from %s",
+                       filename);
+      wrap_hook = false;
+    }
+  }
```
lavarou, line 629: This whole block of code is exactly the same as [this block](https://github.com/newrelic/newrelic-php-agent/blob/41bb3cd04d28d4472130717207cee71033bef064/agent/fw_wordpress.c#L386-L405) in `nr_wordpress_call_user_func_array`. The only difference is in how filename is retrieved. I think it makes sense to extract this into a helper function, say `static inline bool nr_wordpress_should_wrap_hook_callback(const char* filename)` to avoid this code duplication.

Then the function would start:
```
   char* filename = nr_php_op_array_file_name(<op_array_source>);

   if (nr_wordpress_should_wrap_hooks_callback(filename)) {
       <func-specific-wrapping>
   }
```

Note that the call to `nr_php_op_array_file_name` is 'cheap' because it is just pointer assignment, not `strcpy` or `strdup`, and therefore does not need to be wrapped with `if (is-wordpress-ini-set)`.

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
lavarou: https://github.com/newrelic/newrelic-php-agent/pull/786/commits/407e6703538dd5fcb8af1013d75943304a212c8d demonstrates a different way wordpress hook system can be mocked where callback is installed for a given hook and then called at hook time.

lavarou (approved):
None of my remarks are blocking and the updates can be done in a follow up PR.

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
lavarou: This will not be done to limit number of conflicts when syncing oapi branch.

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
lavarou: This will not be done to limit number of conflicts when syncing oapi branch.

agent/fw_wordpress.c
lavarou: This will not be done to limit number of conflicts when syncing oapi branch.

lavarou: For posterity: add_filter can be wrapped for PHPs < 7.4 but is not required as it wraprec lookup for PHPs < 7.4 is not causing increased agent's overhead.

agent/fw_wordpress.c
lavarou: This will not be done to limit number of conflicts when syncing oapi branch.

mfulb:
The code for PHP 7.4+ still seems to use the `nr_wordpress_apply_filters` function wrapper which appears to rely on cufa.  However for PHP 7.4+ my impression was cufa was not being used.  It is unclear to me if this is correct.

lavarou:
> The code for PHP 7.4+ still seems to use the `nr_wordpress_apply_filters` function wrapper which appears to rely on cufa.

@mfulb Are you talking about these places: [here](https://github.com/newrelic/newrelic-php-agent/blob/5a1ef2bcd7ed1e5ffde87c9175cce8e41abb8268/agent/fw_wordpress.c#L533) and [here](https://github.com/newrelic/newrelic-php-agent/blob/5a1ef2bcd7ed1e5ffde87c9175cce8e41abb8268/agent/fw_wordpress.c#L540). These have no effect on agent's operation when `add_filter` is instrumented. I agree that it's probably best to wrap them in `#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO`

agent/fw_wordpress.c
```diff
@@ -551,6 +553,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+ *                                Lower numbers correspond with earlier execution,
+ *                                and functions with the same priority are executed
+ *                                in the order in which they were added to the filter. Default 10.
+ * @param int      $accepted_args Optional. The number of arguments the function accepts. Default 1.
+ * @return true Always returns true.
+ */
+
+NR_PHP_WRAPPER(nr_wordpress_add_filter) {
+  /* Wordpress's add_action() is just a wrapper around add_filter(),
+   * so we only need to instrument this function */
```
mfulb, line 575: Do we have any way to test (integration/multiverse, not run-time) this assumption that add_action() is just a wrapper around add_filter()?   I'm thinking as a guard against if WP changes internally so this is not true.

agent/fw_wordpress.c
```diff
@@ -551,6 +553,65 @@ NR_PHP_WRAPPER(nr_wordpress_apply_filters) {
...
+   * function, we're instrumenting hooks, and WordPress is currently executing
+   * hooks (denoted by the wordpress_tag being set).
+   */
+  if ((NR_FW_WORDPRESS != NRPRG(current_framework))
+      || (0 == NRINI(wordpress_hooks))) {
+    wrap_hook = false;
+  }
+
+  if (NRINI(wordpress_hooks_skip_filename)
+      && 0 != nr_strlen(NRINI(wordpress_hooks_skip_filename))) {
```
mfulb, line 594: Would it be faster to just check if the string `NRINI(wordpress_hooks_skip_filename)` doesn't start with a byte equal to `0`?

lavarou: @mfulb I think it would even be better to replace these two checks (`NULL` and `nr_strlen`) with a call to `nr_strempty` which does just what is needed here. I would however prefer to do it in a separate PR as this code needs cleanup anyway because it is repeated twice: [here](https://github.com/newrelic/newrelic-php-agent/blob/5a1ef2bcd7ed1e5ffde87c9175cce8e41abb8268/agent/fw_wordpress.c#L344-L353) and [here](https://github.com/newrelic/newrelic-php-agent/blob/5a1ef2bcd7ed1e5ffde87c9175cce8e41abb8268/agent/fw_wordpress.c#L593-L602).

agent/fw_wordpress.c
lavarou: @mfulb This is a good suggestion and there already exists tests in multiverse to catch this.

agent/fw_wordpress.c
lavarou: This will be handled when `oapi` branch is synchronized with these changes after they land in `dev`.

tests/integration/frameworks/wordpress/test_wordpress_apply_filters.php
lavarou: Addressed with b9fb84d9c223bd4548c065e62a1cba4d8bdbe235.

tests/integration/frameworks/wordpress/test_wordpress_do_action.php
lavarou: Addressed with b9fb84d9c223bd4548c065e62a1cba4d8bdbe235.

mfulb (approved):

PR #770, opened 2023-11-15
fix(agent): fix wrapping transient user functions

zsistla:
Original commentary from https://github.com/newrelic/newrelic-php-agent/pull/735
Prevent adding transient wraprecs for already wrapped transient functions by checking if wraprec already exists in the hashmap. Additionally prevent adding transient wraprecs for if the transient function is already wrapped as non-transient function by checking if wraprec already exists in the linked list.

For non-transient wrappers (standard instrumentation), the wrapper is stored in both the hashmap and linked-list. For transient wrappers, the wrapper is only stored in the hashmap. HOWEVER! non-transient wrappers MAY only be in the linked-list. This normally happens if it is wrapping a function that hasn't been loaded by PHP yet. Once the function is loaded, there are many hooks to ensure the wrapper is also added to the hashmap.

The implications of the above are: For non-transient wrappers, we only need to check the linked-list to ensure that we are not duplicating a wrapper. If a wrapper is only in the hashmap, it is transient and will be overwritten by the non-transient wrapper.

For transient wrappers, however, we must check both the linked list and the hashmap. This is because it is possible to attempt to create a transient wrapper around an already non-transiently wrapped function. This will return the non-transient wrapper and will attempt to set the transient callbacks if that wrapper has those callbacks free. This means that it is possible for callbacks intended to be transient are attached onto a wrapper that isn't.

Fixes issue introduced in #768.


out.txt
bduranleau-nr, line 1: Did you mean to commit this file?

tests/integration/frameworks/drupal/skipif.inc
```diff
@@ -0,0 +1,9 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+if (version_compare(PHP_VERSION, "8.2", ">=")) {
```
bduranleau-nr, line 7: What breaks on 8.2?

mfulb (approved):

PR #771, opened 2023-11-15
tests: add `EXPECT_METRICS_EXIST` to integration tests

lavarou:
Integration tests sometimes don't care about the exact set of metrics that are being sent (particularly around things like datastore metrics); what is really interesting is that the framework is detected, the transaction is named correctly, and that there aren't errors. This adds a new `EXPECT_METRICS_EXIST` section which is simply a newline separated list of metric names that will be searched for: if any don't exist, the test will fail.

For example:

```php
<?php

/*EXPECT_METRICS_EXIST
Supportability/framework/WordPress/detected
*/
?>
```

This asserts that the test will generate
`Supportability/framework/WordPress/detected` metric, but doesn't assert anything else about the metric table.

---------

### DIFF
--- a/src/newrelic/integration/parse.go
+++ b/src/newrelic/integration/parse.go
@@ -32,6 +32,7 @@ var (
 		"EXPECT_SPAN_EVENTS_LIKE": parseSpanEventsLike,
 		"EXPECT_LOG_EVENTS":       parseLogEvents,
 		"EXPECT_METRICS":          parseMetrics,
+		"EXPECT_METRICS_EXIST":    parseMetricsExist,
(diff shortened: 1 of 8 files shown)


zsistla (commented):
Great addition.  Can you add one integration test case that uses it and can serve as an example?

mfulb (approved):
Nice addition Michal!

lavarou:
> Great addition. Can you add one integration test case that uses it and can serve as an example?

@zsistla Will this do: 0f98fcfd74b822a6ba0b7565692afde156745cbc?

zsistla (approved):
Nice addition, thanks!

PR #772, opened 2023-11-15
tests: fix guzzle integration tests

lavarou:
guzzle integration tests are flaky - they expect span for a call to `new Client();`
however when `transaction_tracer.detail` is set to 0, the agent sometimes drops it.
On the other hand when `transaction_tracer.detail` is set to 1, the agent
generates many more spans. Take advantage of `EXPECT_SPAN_EVENTS_LIKE` and only look
for limited number of spans: root and external.


bduranleau-nr (approved):

zsistla (approved):
Thank you!!!!

PR #774, opened 2023-11-15
Update test_module_invoke_all.php8.php

zsistla:
h is called twice and should have a 2
a total of five modules are called.

this can be seen in the expected output: https://github.com/newrelic/newrelic-php-agent/pull/773/files#diff-3379b8e322b13de0ff0eede236923dfb7f28c5cd4df20357fd6a6bbc0d9146bdR26-R30


bduranleau-nr (approved):
👍

PR #776, opened 2023-11-16
feat(oapi): use segment pool instead of stacked segments

ZNeumann:
Because OAPI doesn't create a nice callstack that replicates the PHP callstack, stacked segments make little/no sense in its context. Instead, we want to utilize the segment pool available in txn's.

force_current_segment was how stacked segments were able to maintain their context in a txn while not being on the segment stack. For OAPI, there is no reason to use this field, as the current segment will just be the top of the txn's segment stack.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1247,6 +1247,8 @@ static inline void nr_php_execute_segment_add_metric(
  *           custom metric?
  *
  * Params  : 1. The stacked segment to end.
+ *              OAPI does not use stacked segments, and should pass
+ *              a heap allocated segment instead
  *           2. The function naming metadata.
  *           3. Whether to create a metric.
  */
@@ -1272,7 +1274,13 @@ static inline void nr_php_execute_segment_end(
       || nr_vector_size(stacked->metrics) || stacked->id || stacked->attributes
       || stacked->error) {
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+    // There are no stacked segments for OAPI.
+    nr_segment_t* s = stacked;
+#else
     nr_segment_t* s = nr_php_stacked_segment_move_to_heap(stacked TSRMLS_CC);
+#endif // OAPI
     nr_php_execute_segment_add_metric(s, metadata, create_metric);
 
     /*
@@ -1287,7 +1295,13 @@ static inline void nr_php_execute_segment_end(
 
     nr_segment_end(&s);
   } else {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+    // There are no stacked segments for OAPI.
+    nr_segment_discard(&stacked);
+#else
     nr_php_stacked_segment_deinit(stacked TSRMLS_CC);
+#endif // OAPI
   }
 }
 
@@ -1298,6 +1312,8 @@ static inline void nr_php_execute_segment_end(
  * If the flag is NULL, then we've only added a couple of CPU instructions to
  * the call path and thus the overhead is (hopefully) very low.
  */
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
+    || defined OVERWRITE_ZEND_EXECUTE_DATA /* not OAPI */
 static void nr_php_execute_enabled(NR_EXECUTE_PROTO TSRMLS_DC) {
   int zcaught = 0;
   nrtime_t txn_start_time;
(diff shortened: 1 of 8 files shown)


agent/php_txn.c
```diff
@@ -1067,7 +1070,16 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
    * by calling newrelic_end_transaction inside nested function scopes)
    * the stack of stacked segments has to be cleaned up.
    */
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+  for (nr_segment_t* segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
+       segment != NULL && segment != NRTXN(segment_root);
+       segment = nr_txn_get_current_segment(NRPRG(txn), NULL)) {
+    nr_segment_end(&segment);
+  }
```
bduranleau-nr, line 1079: A `while` loop seems like a more appropriate loop here
```suggestion
  nr_segment_t* segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
  while(segment != NULL && segment != NRTXN(segment_root) {
    nr_segment_end(&segment);
    segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
  }
```

bduranleau-nr (approved):

agent/php_txn.c
ZNeumann: I don't see a compelling reason for either loop over the other. Is there a big reason to use `while `here?

bduranleau-nr: In my opinion a `while` loop in this situation makes the loop conditions easier to read and is typically more appropriate when you aren't sure how many elements you're looping over.

agent/php_execute.c
```diff
@@ -1974,7 +1983,7 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
    * been created if we are recording and if wraprec is set or if tt is greater
    * than 0.
```
lavarou, line 1984: See [this comment](https://github.com/newrelic/newrelic-php-agent/pull/767/files#r1471534153).

ZNeumann: addressed in said comment in previous PR

lavarou (commented):
I think [this comment](https://github.com/newrelic/newrelic-php-agent/blob/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe/agent/php_stacked_segment.h#L169-L256) needs some revisions. It still talks about Observer API using stacked segments, which it no longer does, right? I didn’t go through the whole comment just yet, but for example, these lines don’t seem accurate any more:
https://github.com/newrelic/newrelic-php-agent/blob/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe/agent/php_stacked_segment.h#L184
https://github.com/newrelic/newrelic-php-agent/blob/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe/agent/php_stacked_segment.h#L196-L198

agent/php_stacked_segment.h
```diff
@@ -256,6 +256,8 @@
 // clang-format on
 
 #include "php_agent.h"
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
+    || defined OVERWRITE_ZEND_EXECUTE_DATA /* not OAPI */
```
lavarou, line 260: This conditional compilation ensures that `nr_php_stacked_segment_*` API is only used by the agent that uses `zend_execute` hook and is not used by the agent using Observer API to hook into Zend Engine.

agent/php_execute.c
```diff
@@ -1900,7 +1908,8 @@ static void nr_php_instrument_func_begin(NR_EXECUTE_PROTO) {
         "Uncaught exception ", &NRPRG(exception_filters) TSRMLS_CC);
   }
 
-  segment = nr_php_stacked_segment_init(segment);
+  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+
   if (nrunlikely(NULL == segment)) {
     nrl_verbosedebug(NRL_AGENT, "Error initializing stacked segment.");
```
lavarou, line 1914:
```suggestion
    nrl_verbosedebug(NRL_AGENT, "Error starting segment.");
```

ZNeumann:
> I think [this comment](https://github.com/newrelic/newrelic-php-agent/blob/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe/agent/php_stacked_segment.h#L169-L256) needs some revisions. It still talks about Observer API using stacked segments, which it no longer does, right? I didn’t go through the whole comment just yet, but for example, these lines don’t seem accurate any more:
> 
> https://github.com/newrelic/newrelic-php-agent/blob/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe/agent/php_stacked_segment.h#L184
> 
> 
> https://github.com/newrelic/newrelic-php-agent/blob/d2c2d8a46b129b9d0ce622ec01a3d34bddde65fe/agent/php_stacked_segment.h#L196-L198

revised in [dcfab0f](https://github.com/newrelic/newrelic-php-agent/pull/776/commits/dcfab0fb492fa2315cb93895f3d2d7abdb759c0d)

lavarou (approved):

PR #777, opened 2023-11-16
refactor: compile wordpress hook regex at minit

bduranleau-nr:
Improve agent's performance by compiling `wordpress_hook_regex` once per process at MINIT rather than on every request at RINIT, (and be clean up at MSHUTDOWN rather than RSHUTDOWN).


lavarou (approved):
LGTM

PR #779, opened 2023-11-16
fix: correctly instrument WebdisConnection executeCommand in OAPI

ZNeumann:


agent/lib_predis.c
```diff
@@ -747,7 +747,44 @@ NR_PHP_WRAPPER(nr_predis_pipeline_executePipeline_clean) {
...
+    },
+  };
+
+  command_obj = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  conn = nr_php_scope_get(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+
+  operation = nr_predis_get_operation_name_from_object(command_obj TSRMLS_CC);
+  params.operation = operation;
+
+  params.instance = nr_predis_retrieve_datastore_instance(conn TSRMLS_CC);
```
bduranleau-nr, line 777:
```suggestion
  command_obj = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
  conn = nr_php_scope_get(NR_EXECUTE_ORIG_ARGS);

  operation = nr_predis_get_operation_name_from_object(command_obj);
  params.operation = operation;

  params.instance = nr_predis_retrieve_datastore_instance(conn);
```

agent/lib_predis.c
```diff
@@ -747,7 +747,44 @@ NR_PHP_WRAPPER(nr_predis_pipeline_executePipeline_clean) {
...
+  zval* command_obj = NULL;
+  zval* conn = NULL;
+  nr_segment_datastore_params_t params = {
+    .datastore = {
+      .type = NR_DATASTORE_REDIS,
+    },
+  };
+
+  command_obj = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  conn = nr_php_scope_get(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
bduranleau-nr, line 772: Are these guaranteed to be non-`NULL`, or are we not concerned if they are `NULL`?

ZNeumann: In pre-OAPI world, this call would not be alright. This is because (in pre-OAPI) after we call the underlying PHP function, zend frees the underlying resources associated with that function. We have callouts to that fact in php_wrapper.h.

In OAPI, this is fine, as zend keeps the resources available for any after callbacks. In this particular case, the function always has 1 argument (so command_obj is safe) and the function is a member of an object (so nr_php_scope_get is safe). These assumptions will be true even in the case of an exception, as an exception has no bearing on the argument or calling object. So we are safe to reuse this function as the "clean" callback. In the case of an exception, only the return value (not used here) would be NULL.

I have added some clarifications to the wrapping guide in php_wrapper.h in [00e3274](https://github.com/newrelic/newrelic-php-agent/pull/779/commits/00e32749ef833a77a291902f585f83fcb93c6d97) and [ffd6f4b](https://github.com/newrelic/newrelic-php-agent/pull/779/commits/ffd6f4b4af714ad4a2f6c1f70781bee91aa85381)

bduranleau-nr: > but that case is handled in `nr_predis_get_operation_name_from_object`.

Briefly glancing at `nr_predis_get_operation_name_from_object` it looks like the function will (eventually) just return `NULL` to be assigned to the `operation` variable. Is there a risk in `operation` being `NULL`? Do we want to continue the function execution if these values are set to `NULL`, or would it make sense to exit early?

ZNeumann: No matter what, we need to be calling `nr_segment_datastore_end`. It takes the params object that gets populated with this value (and is potentially null), and handles it accordingly (with a null check).

lavarou (approved):
LGTM

PR #780, opened 2023-11-17
fix(oapi): delay adding a segment to the parent's children for perfor…

ZNeumann:
…mance gains.

This seems to create some volatility in the test tests/integration/external/curl_multi_exec/test_span_sets_method.php, with the correct spans being created, but their order being non-deterministic. The integration runner is unable to properly match the spans due to the differences between them being in the metadata, so the runner has already matched the spans before seeing the differences in metadata

This is an alternative approach to https://github.com/newrelic/newrelic-php-agent/pull/801 and only 1 of the 2 should be merged.

### DIFF
--- a/axiom/nr_segment.c
+++ b/axiom/nr_segment.c
@@ -660,6 +660,17 @@ bool nr_segment_set_name(nr_segment_t* segment, const char* name) {
 }
 
 bool nr_segment_set_parent(nr_segment_t* segment, nr_segment_t* parent) {
+  if (nr_segment_set_parent_delayed_child(segment, parent)) {
+    // parent can be NULL here if the segment's parent was already NULL
+    if (NULL != parent) {
+      nr_segment_children_add(&parent->children, segment);
+    }
+    return true;
+  }
+  return false;
+}
+
+bool nr_segment_set_parent_delayed_child(nr_segment_t* segment, nr_segment_t* parent) {
   nr_segment_t* ancestor = NULL;
 
   if (NULL == segment) {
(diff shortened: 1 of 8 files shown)


axiom/nr_segment.c
```diff
@@ -852,6 +882,17 @@ void nr_segment_destroy_tree(nr_segment_t* root) {
...
 }
 
+void nr_segment_end_tree(nr_segment_t* curr, nr_segment_t* root) {
+  nr_segment_t* next;
+  if (NULL == curr || root == NULL) {return;}
+  next = curr->parent;
+  while (curr != root) {
+    nr_segment_end(&curr);
+    curr = next;
+    next = curr->parent;
```
bduranleau-nr, line 892: is `next` aka `curr->parent` always guaranteed to be non-`null`?

bduranleau-nr: (and `curr` for that matter as well)

ZNeumann: They should be guaranteed non-null. At the very least, if they are null, we have a memory leak, as that would indicate that the segment tree is malformed. For safety (if this PR is chosen over https://github.com/newrelic/newrelic-php-agent/pull/801), I can add a null check with an error log.

bduranleau-nr: I think adding a `null` check is a good idea if this PR is selected over the other.

ZNeumann:
Approach https://github.com/newrelic/newrelic-php-agent/pull/801 was preferred.

PR #782, opened 2023-11-21
tests: add EXPECT_PHP_PACKAGES to integration tests 

mfulb:
This PR adds test cases for php package version testing. The basic idea is to take a simple existing test for a framework or library and add an EXPECT_PHP_PACKAGES section. This will trigger special behavior in the integration_runner that will look for existing packages and then compare this to the php packages harvest generated by the agent.

The form of the EXPECT_PHP_PACKAGES is:
```
EXPECT_PHP_PACKAGES
command=composer show -f json
supported_packages=../../../include/supported_php_packages.json
```

The `command` is whatever needs to be run to generate a list of installed packages in JSON.

The test needs to know what the expected packages are - there are two options:
 - The `supported_packages` option points at a JSON file which is just a list of all the packagist style names for packages we expect to detect.  The integration_runner will run the `command` given above and filter the list of packages reported to those given in the supported JSON file. The file `includes/supported_json_packages.json` is included in the PR with a list of packages the agent could detect at the time of the PR.
 - The `expected_packages` options accepts a list of package names that the agent is expected to detect.

Ideally the `supported_packages` approach would be best as it would let the tests auto-detect the expected packages.  But in early testing we hit cases where the agent was not detecting all the expected packages (the test case wasn't using all the installed packages so some were not detected by the agent).  So using `expected_packages` to limit the expected to just what the test uses was added.

There are packages for which the agent cannot determine the package version - only that the package exists.  The `options=package_name_only` option is used to tell the integration_runner there is no version info expected.  This works for many cases - but there are cases where one of the detected packages has a version and another does not.  The suggestion for future work is to change this option to something like `package_name_only=<package list>`, where `<package list>` is a list of packages which are expected to only return package names.  This would require reworking the integration_runner to implement.


src/newrelic/integration/php_packages.go
```diff
@@ -0,0 +1,241 @@
...
+
+// composer entire JSON
+type ComposerJSON struct {
+	Installed []ComposerPackage
+}
+
+//
+// Given a JSON harvest payload, extract the PHP packages
+//
+// Params 1 : JSON byte string containing upload_module_updates endpoint data
```
hahuja2, line 55:
```suggestion
// Params 1 : JSON byte string containing update_loaded_modules endpoint data
```

hahuja2 (approved):

src/newrelic/integration/test.go
```diff
@@ -493,6 +498,74 @@ func (t *Test) comparePayload(expected json.RawMessage, pc newrelic.PayloadCreat
...
+	actualPackages, err := GetPhpPackagesFromData(audit)
+	if nil != err {
+		t.Fatal(err)
+		return
+	}
+
+	fmt.Printf("harvest.PhpPackages %+v\n", harvest.PhpPackages)
+	fmt.Printf("actualPackages %+v\n", actualPackages)
+	fmt.Printf("expectedPackages %+v\n", expectedPackages)
+
```
mfulb, line 536:
```suggestion
	fmt.Printf("harvest.PhpPackages %+v\n", harvest.PhpPackages)
	fmt.Printf("actualPackages %+v\n", actualPackages)
	fmt.Printf("expectedPackages %+v\n", expectedPackages)

```

src/newrelic/integration/test.go
```diff
@@ -493,6 +498,70 @@ func (t *Test) comparePayload(expected json.RawMessage, pc newrelic.PayloadCreat
...
+		expectedPackages, err = expectedPkgsCollection.GatherInstalledPackages()
+		if nil != err {
+			t.Fatal(err)
+			return
+		}
+		if nil == expectedPackages || 0 == len(expectedPackages) {
+			t.Fatal(fmt.Errorf("EXPECTED_PHP_PACKAGES used but no packaged detected in environment!"))
+		}
+	} else {
+		expectedPackages = nil
```
mfulb, line 519:
```suggestion
               // no configuration given for package (no EXPECT_PHP_PACKAGES in test case) so don't run test
               return
```

lavarou (approved):

PR #784, opened 2023-11-23
refactor: reduce # of wp plugin/theme name calculation

lavarou:
Improve agent's performance by figuring out the plugin/theme name at the point an action or filter is added and store it in the wraprec rather than doing it at runtime (which is slow, even with memoization in `wordpress_file_metadata` hashmap).

### DIFF
--- a/agent/fw_wordpress.c
+++ b/agent/fw_wordpress.c
@@ -296,7 +296,9 @@ cache_and_return:
 }
 
 NR_PHP_WRAPPER(nr_wordpress_wrap_hook) {
+#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO
   zend_function* func = NULL;
+#endif
   char* plugin = NULL;
 
   NR_UNUSED_SPECIALFN;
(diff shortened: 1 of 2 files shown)


agent/fw_wordpress.c
```diff
@@ -602,10 +608,23 @@ NR_PHP_WRAPPER(nr_wordpress_add_filter) {
...
     zval* callback = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS);
     /* the callback here can be any PHP callable. nr_php_wrap_generic_callable
    * checks that a valid callable is passed */
-    nr_php_wrap_generic_callable(callback, nr_wordpress_wrap_hook);
+    callback_wraprec = nr_php_wrap_generic_callable(callback, nr_wordpress_wrap_hook);
+
+    // We can cheat here: wraprecs on callables are always transient, so if
+    // there's a wordpress_plugin_theme set we know it's from this transaction,
+    // and we don't have any issues around a possible multi-tenant setup.
+    if (callback_wraprec && NULL == callback_wraprec->wordpress_plugin_theme) {
```
mfulb, line 620: It isn't clear to me how this is an optimization - is `callback_wraprec->wordpress_plugin_theme` going to be set for the rest of a request after the first `add_filter` is wrapped?

lavarou: Yes, `callback_wraprec->wordpress_plugin_theme` is going to be set for the rest of a request after the first `add_filter` is wrapped.

bduranleau-nr (approved):

mfulb (approved):

PR #786, opened 2023-11-27
refactor: limit wordpress instrumentation

lavarou:
Improve agent's performance when WordPress hook tracking is disabled with [`newrelic.framework.wordpress.hooks`](https://docs.newrelic.com/docs/apm/agents/php-agent/configuration/php-agent-configuration/#inivar-framework-wordpress-hooks) set to `false`. There's no need to wrap WordPress 'hook' functions when WordPress hooks are to not be instrumented. In such case, instrumentation is limited to transaction naming only.

### DIFF
--- a/agent/fw_wordpress.c
+++ b/agent/fw_wordpress.c
@@ -481,18 +481,18 @@ NR_PHP_WRAPPER(nr_wordpress_exec_handle_tag) {
   NR_UNUSED_SPECIALFN;
   (void)wraprec;
 
-  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);
+  if (nrlikely(0 != NRINI(wordpress_hooks))) {
+    NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);
 
-  tag = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+    tag = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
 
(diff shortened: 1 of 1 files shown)


agent/fw_wordpress.c
```diff
@@ -481,18 +481,18 @@ NR_PHP_WRAPPER(nr_wordpress_exec_handle_tag) {
   NR_UNUSED_SPECIALFN;
   (void)wraprec;
 
-  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_WORDPRESS);
+  if (nrlikely(0 != NRINI(wordpress_hooks))) {
```
ZNeumann, line 484:
```suggestion
  if (0 != NRINI(wordpress_hooks)) {
```
I don't think we want to override the branch predictor here. Within a given run, this will always evaluate to the same true/false, so the branch predictor will be able to hit with 100% after a pass or so. Otherwise, we are always taking a hit when hooks are off.

ZNeumann (approved):

bduranleau-nr (approved):

agent/fw_wordpress.c
lavarou: > we are always taking a hit when hooks are off.

@ZNeumann This line is more of an assert, than anything - this code path (`nr_wordpress_exec_handle_tag` function) will never execute if hooks instrumentation is disabled. It is guaranteed by this line: https://github.com/newrelic/newrelic-php-agent/blob/1fbb0d121afd80145c114beeac94f32ef6cf78a1/agent/fw_wordpress.c#L599
Are you saying it is still better not to give this hint?

ZNeumann: Ah, since the function is never executed in this case, it makes no difference to me whether this `nrlikely` exists

PR #787, opened 2023-11-28
refactor: improve wordpress instrumentation performance

bduranleau-nr:


agent/fw_wordpress.c
lavarou, line 1: This file needs #796 applied.

agent/fw_wordpress.c
```diff
@@ -428,13 +379,21 @@ static char* nr_wordpress_clean_tag(const zval* tag TSRMLS_DC) {
     return NULL;
   }
```
lavarou, line 380: This check is not needed - `tag` is guaranteed not to be `NULL` [here](https://github.com/newrelic/newrelic-php-agent/blob/bd37e27f12817d771240a416caef01571a94670a/agent/fw_wordpress.c#L435) and [here](https://github.com/newrelic/newrelic-php-agent/blob/bd37e27f12817d771240a416caef01571a94670a/agent/fw_wordpress.c#L522).

lavarou (approved):
Looks good. None of my remarks need to be resolved.

PR #789, opened 2023-11-30
refactor: suggestion to #765

lavarou:
A couple of suggestions to #765:
1. Save time by using pre-calculated string lengths vs calling strlen whenever possible
2. Move `nr_file_basename` as `nr_wordpress_strip_php_suffix` back to `fw_wordpress.c`

### DIFF
--- a/agent/fw_wordpress.c
+++ b/agent/fw_wordpress.c
@@ -149,24 +149,27 @@ static nr_matcher_t* nr_wordpress_theme_matcher() {
 
 char* nr_php_wordpress_plugin_match_matcher(const char* filename) {
   char* plugin = NULL;
-  plugin = nr_matcher_match(nr_wordpress_plugin_matcher(), filename);
-  plugin = nr_file_basename(plugin, nr_strlen(plugin));
+  int plugin_len;
+  plugin = nr_matcher_match_ex(nr_wordpress_plugin_matcher(), filename, nr_strlen(filename), &plugin_len);
+  plugin = nr_file_basename(plugin, plugin_len);
   nr_matcher_destroy(&NRPRG(wordpress_plugin_matcher));
   return plugin;
 }
 
 char* nr_php_wordpress_theme_match_matcher(const char* filename) {
   char* theme = NULL;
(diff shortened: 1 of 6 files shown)


axiom/util_matcher.c
```diff
@@ -31,38 +38,46 @@ void nr_matcher_destroy(nr_matcher_t** matcher_ptr) {
...
+  prefix = nr_calloc(1, sizeof(matcher_prefix));
+  prefix->len = nr_strlen(str);
+  while (prefix->len > 0 && '/' == str[prefix->len - 1]) {
+    prefix->len--;
   }
 
-  prefix_lc = nr_malloc(prefix_len + 2);
-  for (i = 0; i < prefix_len; i++) {
-    prefix_lc[i] = nr_tolower(prefix[i]);
+  prefix->len += 1;
```
bduranleau-nr, line 55: Why the addition of this increment to len? It looks like the previous logic could be carried over the same without this line

lavarou: The `'+1'` is for the ending `'/'`. Previous code did `nr_malloc(prefix_len+2)` on line 48, which wasn't well explained but I managed to figure it out: +2 = +1 for the `'/'` +1 for `'\0'`. New code is still not explained very well, but does the same thing: `prefix->len += 1;` on line 55 is for the `'/'`, whereas `nr_malloc(prefix->len+1)` on line 56 does +1 for the `'\0'`

agent/fw_wordpress.c
```diff
@@ -206,7 +210,7 @@ static char* nr_wordpress_plugin_from_function(zend_function* func TSRMLS_DC) {
                      NRP_PHP(NRPRG(wordpress_tag)));
     return NULL;
   }
-  filename_len = nr_strlen(filename);
+  filename_len = func->op_array.filename->len; // temp hack - should be nr_php_function_filename_len(func);
```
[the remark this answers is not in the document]
lavarou: Yes. Done with f6b21c8c29475ecf1fd5164cf4863e202ecf5aa3

PR #793, opened 2023-12-04
tests: add wordpress transaction naming tests

lavarou:
Add tests that verify that the agent names WordPress web transaction as an 'Action' with the name generated from the template used to generate the page when WordPress hooks are enabled as well as disabled.


tests/integration/frameworks/wordpress/test_wordpress_transaction_name_hooks_off.php
```diff
@@ -0,0 +1,51 @@
...
+$wp_filters = [];
+
+function add_filter($hook_name, $callback) {
+  global $wp_filters;
+  $wp_filters[$hook_name] = $callback;
+}
+
+function apply_filters($hook_name, $value, ...$args) {
+  global $wp_filters;
+  return call_user_func_array($wp_filters[$hook_name], array($value, $args));
```
lavarou, line 41: Line 41 emulates [this](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-hook.php#L323-L327).

tests/integration/frameworks/wordpress/test_wordpress_transaction_name_hooks_off.php
```diff
@@ -0,0 +1,51 @@
...
+
+/*EXPECT_ERROR_EVENTS null */
+
+// Mock WordPress hooks; only a single callback for a given hook can be defined
+$wp_filters = [];
+
+function add_filter($hook_name, $callback) {
+  global $wp_filters;
+  $wp_filters[$hook_name] = $callback;
+}
```
lavarou, line 37: Lines 32-37 emulate [this](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-hook.php#L323-L327) and [this](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-hook.php#L87-L90).

tests/integration/frameworks/wordpress/test_wordpress_transaction_name_hooks_off.php
```diff
@@ -0,0 +1,51 @@
...
+/*INI
+newrelic.framework = wordpress
+newrelic.framework.wordpress.hooks = false
+*/
+
+/*ENVIRONMENT
+REQUEST_METHOD=GET
+*/
+
+/*EXPECT_METRICS_EXIST
```
ZNeumann, line 24: Is there an easy way to also check that certain metrics DON'T exist? For example, we want to confirm that `Framework/WordPress/Hook/template_include` isn't here.

ZNeumann (approved):

tests/integration/frameworks/wordpress/test_wordpress_transaction_name_hooks_off.php
bduranleau-nr: I think this specific test would require `EXPECT_METRICS` instead for that reason

lavarou: > Is there an easy way to also check that certain metrics DON'T exist? For example, we want to confirm that Framework/WordPress/Hook/template_include isn't here.

I was also pondering on this idea. @ZNeumann Would you prefer `EXPECT_METRICS_DONT_EXIST` or `EXPECT_METRICS_EXIST` with `!` in front of the metric name to negate the check?

ZNeumann: No strong preference

lavarou: @bduranleau-nr @ZNeumann a2218e0525353e5d52668fddd5744abe282a926f adds `EXPECT_METRICS_DONT_EXIST` and ff796c78abba214eedfbdd8f9c2e9d9f50210910 uses it to enhance test expectation.

ZNeumann (approved):

bduranleau-nr (approved):

PR #795, opened 2023-12-05
feat(daemon): improve logging obfuscation

ZNeumann:
Follows the ideas set out in https://github.com/newrelic/go-agent/pull/828/files

### DIFF
--- a/src/daemon/main.go
+++ b/src/daemon/main.go
@@ -363,8 +363,18 @@ func main() {
 	}
 
 	log.Infof("%s", banner(cfg))
+	// Used to not log out the proxy, which has potentially secret credentials
+	var proxy bool = false
 	for i := range os.Args {
-		log.Debugf("ARGV[%d]: %s", i, os.Args[i])
+		if proxy {
+			log.Debugf("ARGV[%d]: **REDACTED**", i)
+			proxy = false
+		} else {
+			log.Debugf("ARGV[%d]: %s", i, os.Args[i])
+			if (os.Args[i] == "--proxy") {
+				proxy = true
+			}
+		}
 	}
 	log.Debugf("process role is %v", cfg.Role)
 
--- a/src/daemon/main.go
+++ b/src/daemon/main.go
@@ -371,7 +371,7 @@ func main() {
 			proxy = false
 		} else {
 			log.Debugf("ARGV[%d]: %s", i, os.Args[i])
-			if (os.Args[i] == "--proxy") {
+			if ("--proxy" == os.Args[i]) {
 				proxy = true
 			}
 		}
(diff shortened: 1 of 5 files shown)


src/daemon/main.go
```diff
@@ -363,8 +363,18 @@ func main() {
...
 	log.Infof("%s", banner(cfg))
+	// Used to not log out the proxy, which has potentially secret credentials
+	var proxy bool = false
 	for i := range os.Args {
-		log.Debugf("ARGV[%d]: %s", i, os.Args[i])
+		if proxy {
+			log.Debugf("ARGV[%d]: **REDACTED**", i)
+			proxy = false
+		} else {
+			log.Debugf("ARGV[%d]: %s", i, os.Args[i])
```
hahuja2, line 373: Should this `log.Debugf()` be here since the proxy check happens after that line?

ZNeumann: Argv[] splits the parameter name and value into 2 separate entries. So the proxy check is looking for "--proxy" and then the next loop iteration will be the actual proxy value.

hahuja2 (approved):

bduranleau-nr:
Can we add a daemon unit test that verifies this functionality is working as expected?

ZNeumann:
> Can we add a daemon unit test that verifies this functionality is working as expected?

added in https://github.com/newrelic/newrelic-php-agent/pull/795/commits/bca0610da39a99b7df5bc5087558c1f27d2fccb9

bduranleau-nr (approved):

src/newrelic/collector/client.go
```diff
@@ -402,26 +428,13 @@ func (c *clientImpl) Execute(cmd *RpmCmd, cs RpmControls) RPMResponse {
...
-		}
-		if nil == audit {
-			audit = cmd.Data
-		}
-	}
-
 	url := cmd.url(false)
 	cleanURL := cmd.url(true)
 
-	log.Audit("command='%s' url='%s' payload={%s}", cmd.Name, url, audit)
```
lavarou, line 424: Is logging to audit file removed because log file will already have the same info but with URL obfuscated? However, isn't the purpose of the audit file to include all communication with the collector?

ZNeumann: Yes, I removed it because it is already included in the log file. I made the decision to not allow the full URL ever, in order to ensure full compliance.

lavarou: Then wouldn't:
```
log.Audit("command='%s' url='%s' payload={%s}", cmd.Name, cleanUrl, audit)
```
be better?

ZNeumann: re-added in [93d2ddc](https://github.com/newrelic/newrelic-php-agent/pull/795/commits/93d2ddc631d1f7d0d1c9d9163b8eb66130961be5)

lavarou (approved):

PR #797, opened 2023-12-06
feat: add PHP 8.3 support

bduranleau-nr:
Adds non-OAPI PHP 8.3 support to the agent.


CRC-Mismatch:
I guess someone with permissions needs to run the [_make_agent](https://github.com/newrelic/newrelic-php-agent/actions/workflows/make-agent.yml) action for both `amd64` and `arm64` platforms, then re-run the checks :no_mouth: 

Either that, or (judging by no history being shown) someone with access to the private DockerHub repo needs to build and push the 4 new tags...

.github/workflows/test-agent.yml
```diff
@@ -248,7 +248,7 @@ jobs:
           LIBC: ${{matrix.platform}}
           PLATFORM: linux/${{matrix.arch}}
           AGENT_CODE: ${{github.workspace}}/php-agent
-          IMAGE_VERSION: ${{vars.MAKE_PHP_VERSION}}
+          IMAGE_VERSION: v2
```
mfulb, line 251: Is this a temporary change - not clear why this is being hard-coded.

lavarou: This workflow needs to change anyway in order to support running integration tests on arm64 and be more like run-integration-tests workflow, so I'm actually fine either way.

bduranleau-nr: This is necessary to get the PR runner to use the correct builder image. Changing this on a repo level would break all other PRs. Once this is merged another PR will be put up to revert the change.

lavarou (commented):
I think PHP 8.3 needs to be added [here](https://github.com/newrelic/newrelic-php-agent/blob/c56f4a4c74173f63df12a3237981eea16f0325d6/make/php_versions.mk#L6) as well.

.github/workflows/test-agent.yml
lavarou: I agree with @bduranleau-nr assessment.

bduranleau-nr:
> I think PHP 8.3 needs to be added [here](https://github.com/newrelic/newrelic-php-agent/blob/c56f4a4c74173f63df12a3237981eea16f0325d6/make/php_versions.mk#L6) as well.

@lavarou Fixed in a3741c7464fa82878dd0ab2519656fded9fcc241

mfulb (approved):

lavarou (approved):

lucassabreu (approved):

PR #665, opened 2023-05-05
feat(agent): Drupal 9.4+ hooks instrumentation

ZNeumann:

PR #798, opened 2023-12-08
refactor: improve performance of wrap_generic_callable

lavarou:
`nr_php_wrap_generic_callable` is used to create transient wraprecs for callables which **_don't require to be named_**. This assumption makes it possible to stay DRY and reuse `nr_php_zval_to_function` in lieu of complex pattern matching for the type of callable. This reduces agent's performance overhead as it now uses `nr_php_wrap_callable` always instead of sometimes using `nr_php_wrap_user_function_with_transience(IS_TRANSIENT)`. This is an alternative to #665 which obsoletes and removes #679 and makes #770 no longer needed.

TL:DR; The agent's performance overhead when `nr_php_wrap_user_function_with_transience(IS_TRANSIENT)` is used comes from the need to walk the linked list of named wraprecs in `nr_php_add_custom_tracer_named` [here](https://github.com/newrelic/newrelic-php-agent/blob/5a1ef2bcd7ed1e5ffde87c9175cce8e41abb8268/agent/php_user_instrument.c#L413-L426).

### DIFF
--- a/agent/php_wrapper.c
+++ b/agent/php_wrapper.c
@@ -7,6 +7,42 @@
 #include "php_user_instrument.h"
 #include "php_wrapper.h"
 #include "util_logging.h"
+static char* Z_TYPE_STR(zval* zv) {
+
+  switch (Z_TYPE_P(zv)) {
+    case IS_NULL:
+      return "IS_NULL";
+
+    case IS_STRING:
+      return "IS_STRING";
+
+    case IS_LONG:
+      return "IS_LONG";
+
+#if ZEND_MODULE_API_NO >= ZEND_7_0_X_API_NO /* PHP7+ */
+    case IS_TRUE:
+      return "IS_TRUE";
+
+    case IS_FALSE:
+      return "IS_FALSE";
+#else
+    case IS_BOOL:
+      return "IS_BOOL";
+#endif /* PHP7 */
+
+    case IS_DOUBLE:
+      return "IS_DOUBLE";
+
+    case IS_OBJECT:
+      return "IS_OBJECT";
+
+    case IS_ARRAY:
+      return "IS_ARRAY";
+
+    default:
+      return "UNKNOWN";
+  }
+}
 
 nruserfn_t* nr_php_wrap_user_function(const char* name,
                                       size_t namelen,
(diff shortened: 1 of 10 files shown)


ZNeumann (approved):

agent/php_wrapper.c
```diff
@@ -7,20 +7,47 @@
...
+    case IS_OBJECT:
+      return "IS_OBJECT";
+
+    case IS_ARRAY:
+      return "IS_ARRAY";
+
+    default:
+      return "UNKNOWN";
+  }
+}
```
lavarou, line 45: I'm not sure about this...

ZNeumann: I like it alright, though I think it would be beneficial to rename the function. With its current name, I would assume it is provided by the PHP internals API. It might also be worth mentioning that this is (I think) currently only used for logging. Maybe put "log" somewhere in the name so that somebody has to justify a rename for a more important use-case.

ZNeumann (approved):

agent/php_wrapper.c
lavarou: I wonder if it might be better to make `nr_format_zval_for_debug` function external to `agent/php_execute.c`. I sometimes wish it was as it makes debugging so much easier...

bduranleau-nr: I agree with @ZNeumann's comment regarding the naming convention of this function.

lavarou: @ZNeumann @bduranleau-nr What do you think about the idea of making `nr_format_zval_for_debug` function external to `agent/php_execute.c` and using it instead of adding `Z_TYPE_STR`?

ZNeumann: works for me

bduranleau-nr: agree 👍

bduranleau-nr (approved):

agent/php_wrapper.c
lavarou: @bduranleau-nr, @ZNeumann: See e3938763a55965361a99778e52e94578dc382b8f.

agent/php_wrapper.c
```diff
@@ -68,79 +59,22 @@ nruserfn_t* nr_php_wrap_callable(zend_function* callable,
...
-  if (NULL != name) {
-    nrl_verbosedebug(NRL_INSTRUMENT,
-                   "Failed to wrap callable: %s", ZEND_STRING_VALUE(name));
-  } else {
-    nrl_verbosedebug(NRL_INSTRUMENT, "Failed to wrap callable with unknown name");
+  if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+    char strbuf[NR_WRAPPER_DEBUG_STRBUFSZ];
+    nr_format_zval_for_debug(callable, strbuf, 0, NR_WRAPPER_DEBUG_STRBUFSZ - 1,
+                             0);
+    nrl_verbosedebug(NRL_INSTRUMENT, "Failed to cast to callable zval=%s",
```
mfulb, line 75: Could this possibly output a string value which shouldn't appear in a log file?

tests/integration/frameworks/drupal/skipif.inc
```diff
@@ -0,0 +1,9 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+if (version_compare(PHP_VERSION, "8.2", ">=")) {
```
mfulb, line 7: Is this related to the changes in this PR?

lavarou: WordPress and Drupal instrumentation is using code paths modified by this PR. There already existed WordPress integration tests. However there were no Drupal integration tests. This PR adds Drupal integration tests to increase coverage of the Drupal instrumentation and code paths modified by this PR.

mfulb: Is there a similar Drupal test for PHP >= 8.2?

mfulb: After discussion it was determined the test for PHP >= 8.2 is because the `invoke_all` Drupal API call is not available in Drupal versions which run on PHP >= 8.2.  The test using `invoke_all_with` is run for those PHP versions.

mfulb (approved):

tests/integration/frameworks/drupal/skipif.inc
lavarou: Yes there is. Older version of Drupal (the one that does not run on PHPs 8.2+) use one method of invoking custom plugins and this method is tested with `tests/integration/frameworks/drupal/test_module_invoke_all.php`. Newer version of Drupal (the one that runs on PHPs 7.4+) use another method of invoking custom plugins and this method is tested with `tests/integration/frameworks/drupal/test_invoke_all_with.php`.

PR #799, opened 2023-12-09
feat: instrument wordpress plugins' callbacks on demand

lavarou:
Improve agent's performance by adding two new toggles around WordPress instrumentation:
- `newrelic.framework.wordpress.plugins` - indicates if WordPress hooks callback functions, implemented in WordPress core, plugins and themes, are to be instrumented.
- `newrelic.framework.wordpress.hooks_threshold` - sets the WordPress hook's execution duration threshold above which the hook execution will be captured; used when WordPress hooks callback functions are not instrumented.

By default, WordPress hooks callback functions are not instrumented and `newrelic.framework.wordpress.hooks_threshold` is set at 1ms. This allows to identify which hooks take up most time during request processing. If more details are needed, i.e. which plugins are the slowest, `newrelic.framework.wordpress.plugins` need to be set to `true`.


agent/fw_wordpress.c
```diff
@@ -215,6 +215,23 @@ static void free_wordpress_metadata(void* metadata) {
...
 
+static inline void nr_wordpress_hooks_create_metric(nr_segment_t* segment,
+                                       const char* hook_name) {
+  // need to capture 'now' because segment has not finished yet and therefore
+  // can't use segment->stop_time
+  nrtime_t now = nr_txn_now_rel(segment->txn);
+  nrtime_t duration = nr_time_duration(segment->start_time, now);
+  if (duration >= NRINI(wordpress_hooks_threshold)) {
+    // only create metrics for hooks above threshold
+    nr_wordpress_create_metric(segment, NR_WORDPRESS_HOOK_PREFIX, hook_name);
```
ZNeumann, line 226: If the segment duration is long enough to be considered "interesting", will we possibly have a hook metric and a hook segment with different durations? That might confuse users, even if they only differ by a fraction of a second.

ZNeumann (approved):

agent/fw_wordpress.c
```diff
@@ -215,6 +215,23 @@ static void free_wordpress_metadata(void* metadata) {
   nr_free(metadata);
 }
 
+static inline void nr_wordpress_hooks_create_metric(nr_segment_t* segment,
+                                       const char* hook_name) {
```
lavarou, line 219: Needs:
```suggestion
                                       const char* hook_name) {
  if (NULL == segment) {
    return;
  }
```

bduranleau-nr:
Our current integration tests that test WordPress Hooks/Plugin functionality probably need to be updated to account for this new paradigm with INI settings along the lines of `newrelic.framework.wordpress.plugins = true` and `newrelic.framework.wordpress.hooks_threshold = 0` in order to see consistent integration test results.

lavarou:
> Our current integration tests that test WordPress Hooks/Plugin functionality probably need to be updated to account for this new paradigm with INI settings along the lines of newrelic.framework.wordpress.plugins = true and newrelic.framework.wordpress.hooks_threshold = 0 in order to see consistent integration test results.

I have a few tests ready up my sleeve. However, I want to add them either in #802, since that PR changes defaults, or in a new, separate PR after all [wp-optimization-beta3](https://github.com/newrelic/newrelic-php-agent/labels/wp-optimization-beta3) PRs land on [`wp-performance-improvements`](https://github.com/newrelic/newrelic-php-agent/tree/wp-performance-improvements)` branch.

bduranleau-nr (approved):

agent/fw_wordpress.c
```diff
@@ -215,6 +215,26 @@ static void free_wordpress_metadata(void* metadata) {
...
+  if (NULL == segment) {
+    return;
+  }
+  // need to capture 'now' because segment has not finished yet and therefore
+  // can't use segment->stop_time
+  nrtime_t now = nr_txn_now_rel(segment->txn);
```
mfulb, line 225: Probably unlikely - but I think if `segment->txn` was NULL then `now` would be set to `0` and `duration` would always be greater than the `wordpress_hooks_threshold` and so this would create a whole lot of metrics.  Maybe a check for if `NULL == segment->txn` is needed?

agent/php_nrini.c
```diff
@@ -2206,6 +2206,22 @@ STD_PHP_INI_ENTRY_EX("newrelic.framework.wordpress.hooks",
                      zend_newrelic_globals,
                      newrelic_globals,
                      nr_on_off_dh)
+STD_PHP_INI_ENTRY_EX("newrelic.framework.wordpress.hooks_threshold",
+                     "1ms",
+                     NR_PHP_REQUEST,
+                     nr_time_mh,
```
mfulb, line 2212: If an invalid valid is given I think `nr_time_mh` will return `0`.  This will set the threshold to `0` and lots of metric will be generated.

agent/fw_wordpress.c
lavarou: I don't think the `NULL == segment->txn` check is necessary here. `nr_time_duration(start, stop)` returns 0 if `start` (`segment->start_time`) is greater than `stop` (`now`). So if `now` is 0, because `segment->txn` is `NULL`, `duration` will be set to 0 and, unless `NRINI(wordpress_hooks_threshold)` is 0, no metric will be generated.

agent/php_nrini.c
lavarou: @mfulb Good catch! However, `nr_time_mh` is already used in other places where 0 will cause additional data to be generated by the agent. See [here](https://github.com/newrelic/newrelic-php-agent/blob/533f7bd0f4c4775c090d9754072a194fdb0b5ddb/agent/php_nrini.c#L2430-L2439), [here](https://github.com/newrelic/newrelic-php-agent/blob/533f7bd0f4c4775c090d9754072a194fdb0b5ddb/agent/php_nrini.c#L2440-L2449). What would you suggest be done about it?

mfulb: For this particular one I would create a `nr_time_wordpress_hooks_mh` that enforces the default if an invalid value is given.  Perhaps it can be made to leverage the current `nr_time_mm` for the value parsing.

lavarou: Also see #802, which will change the default value for `newrelic.framework.wordpress.plugins` from `false` to `true` (and such settings obsoletes `newrelic.framework.wordpress.hooks_threshold`) as explained by here: https://github.com/newrelic/newrelic-php-agent/pull/802#issuecomment-1856585843.

agent/fw_wordpress.c
```diff
@@ -215,6 +215,23 @@ static void free_wordpress_metadata(void* metadata) {
...
+  nrtime_t duration = nr_time_duration(segment->start_time, now);
+  if (duration >= NRINI(wordpress_hooks_threshold)) {
+    // only create metrics for hooks above threshold
+    nr_wordpress_create_metric(segment, NR_WORDPRESS_HOOK_PREFIX, hook_name);
+  } else {
+    nrl_verbosedebug(NRL_INSTRUMENT,
+                     "hook=%s, stop_time=" NR_TIME_FMT
+                     " start_time=" NR_TIME_FMT " duration=" NR_TIME_FMT "us",
+                     hook_name, now, segment->start_time, duration);
+  }
```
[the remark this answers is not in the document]
lavarou: Removed in 1a89ba544e63fd58b52cb0c5ea32c05b0675d7ab.

agent/php_nrini.c
mfulb: This appears to be an issue that needs a more comprehensive handling since it affects several INI values.  For this PR the existing behavior seems adequate.

mfulb (approved):

PR #801, opened 2023-12-14
feat: faster segment de-parenting via index tracking

ZNeumann:
Segment order does not matter, as the collector reconstructs the tree with timing and parenting info. As such, we can remove segments with O(1) speed instead of O(N).

This PR adds a new file to facilitate this, as the child management files need to know about the segment internals to properly track indexing. This file is added to avoid circular dependencies.

This is an alternative approach than https://github.com/newrelic/newrelic-php-agent/pull/780 and only 1 of the 2 should be merged.


axiom/nr_segment_children.h
```diff
@@ -165,36 +142,41 @@ static inline bool nr_segment_children_remove(nr_segment_children_t* children,
...
+    // Swap'n'Pop
+    temp = children->packed.elements[end];
+    temp->child_ix = ix;
+    children->packed.elements[ix] = temp;
+    children->packed.count -= 1;
+
   } else {
-    size_t index;
+    size_t index = child->child_ix;
+    nr_segment_t* temp;
```
bduranleau-nr, line 161: There's an opportunity here for a slight simplification of code since both the `if` and `else` statements require the `child->child_ix` value as well as the `nr_segment_t* temp` variable, so you could extract them out of the conditional statements and avoid code duplication. This would have the knock-on effect of also unifying the naming convention between the conditionals (`ix` vs `index`)

bduranleau-nr (approved):
Prefer this solution to #780

axiom/nr_segment_children.h
ZNeumann: Unfortunately, "ISO C90 forbids mixed declarations and code", so we cannot extract the definition of index to outside this structure, only the declaration. (Because the definition must happen after the `null` check).

I can move `size_t index` and` nr_segment_t* temp` to the top of the function if that is desired, or just unify the naming of "ix" vs "index".

bduranleau-nr: Probably not worth the trouble.

ZNeumann:
tests/integration/ini/test_transaction_tracer_max_segments_with_datastore.php5.php has been made non-deterministic with these changes. Investigating.

axiom/nr_segment_types.h
```diff
@@ -0,0 +1,207 @@
...
+} nr_segment_typed_attributes_t;
+
+typedef struct _nr_segment_t {
+  nr_segment_type_t type;
+  nrtxn_t* txn;
+
+  /* Tree related stuff. */
+  nr_segment_t* parent;
+  nr_segment_children_t children;
+  size_t child_ix;
```
lavarou, line 152: Introduction of this new field can be easily missed by reviewer amidst the code refactor. `child_ix` is an index into segment's parent children array (either segment's `parent->children.vector` or segment's `parent->children.packed.elements`).

lavarou:
> tests/integration/ini/test_transaction_tracer_max_segments_with_datastore.php5.php has been made non-deterministic with these changes. Investigating.

@ZNeumann The order of segments in transaction trace in test expectation vs actual trace is different. This may be caused by reparenting no longer re-ordering segments, right?

ZNeumann:
> > tests/integration/ini/test_transaction_tracer_max_segments_with_datastore.php5.php has been made non-deterministic with these changes. Investigating.
> 
> @ZNeumann The order of segments in transaction trace in test expectation vs actual trace is different. This may be caused by reparenting no longer re-ordering segments, right?

This conversation resulted in [31522da](https://github.com/newrelic/newrelic-php-agent/pull/801/commits/31522da6fe8ba66919b920257869143554e213ba). To summerize: previously, it did not matter which of the 3 "short" segments in this test were dropped, because the agent maintained order, and therefore the "long" datastore segment would always be last. Now, order is not maintained by way of swap'n'pop removal from the children list. Therefore, for determinism in the test, we need to know which spans will be sampled out (where here, because they all have the same priority, this sampling is determined by execution time). So 3 different length "short" segments enable this determinism.

axiom/nr_segment_types.h
ZNeumann: removed the refactor in favor of forward declared getters/setters in [3b715b4](https://github.com/newrelic/newrelic-php-agent/pull/801/commits/3b715b452a87bd96c81f5d7e51589bb2d173f67a)

axiom/nr_segment.h
```diff
@@ -647,4 +647,22 @@ extern void nr_segment_record_exception(nr_segment_t* segment,
                                         const char* error_message,
                                         const char* error_class);
 
+/*
+ * Purpose : Gets the child_ix of a segment.
+ *
+ * Paramas : 1. The pointer to the segment
+ *
+ * Returns : the child_ix, or 0 if passed NULL
+ *           (Note: 0 is a valid ix)
```
lavarou, line 656: I suggest changing the return type to `ssize_t` and return `-1` if passed `NULL`.

axiom/nr_segment.h
```diff
@@ -141,6 +141,7 @@ typedef struct _nr_segment_t {
   /* Tree related stuff. */
   nr_segment_t* parent;
   nr_segment_children_t children;
+  size_t child_ix;
```
lavarou, line 144:
```suggestion
  size_t child_ix; /* index of this segment in its parent->children vector */
```

ZNeumann:
> Doesn't `tests/integration/ini/test_transaction_tracer_max_segments_with_datastore.php` require similar change?

fixed in [2d74cd9](https://github.com/newrelic/newrelic-php-agent/pull/801/commits/2d74cd97ae2dd9865bb0390f12cef46ad9416f3a)

axiom/nr_segment.h
ZNeumann: I think I would rather the primary onux be on the caller to not pass NULL here (or know if they pass NULL). Otherwise, every time this function is called there will need to be a cast from `ssize_t` to `size_t`, after a `-1` check of course. So ultimately, it is extra code and in my opinion less readable.

lavarou: Returning a valid value in error condition may result in an unexpected behavior and this does not seem right. With `ssize_t` as return type the compiler will enforce the correct use of the returned value. I prefer trading off the need to consciously end explicitly casting over unexpected behavior.

axiom/nr_segment_children.h
```diff
@@ -165,36 +171,41 @@ static inline bool nr_segment_children_remove(nr_segment_children_t* children,
   }
 
   if (children->is_packed) {
-    size_t i;
+    size_t ix = (size_t)nr_segment_get_child_ix(child);
```
lavarou, line 174:
```suggestion
    size_t ix = (size_t)nr_segment_get_child_ix(child); // safe to cast, child != NULL asserted earlier
```

axiom/nr_segment_children.h
```diff
@@ -165,36 +171,41 @@ static inline bool nr_segment_children_remove(nr_segment_children_t* children,
...
-    }
+    // Swap'n'Pop
+    temp = children->packed.elements[end];
+    nr_segment_set_child_ix(temp, ix);
+    children->packed.elements[ix] = temp;
+    children->packed.count -= 1;
+
   } else {
-    size_t index;
+    size_t index = (size_t)nr_segment_get_child_ix(child);
```
lavarou, line 189:
```suggestion
    size_t index = (size_t)nr_segment_get_child_ix(child); // safe to cast, child != NULL asserted earlier
```

lavarou (approved):
Nice and clean! Great work! 👏

PR #802, opened 2023-12-14
feat: instrument wordpress core callbacks on demand

lavarou:
Improve agent's performance by adding a new toggle around WordPress instrumentation:

* `newrelic.framework.wordpress.core` - indicates if WordPress hooks callback functions, implemented in WordPress core, are to be instrumented.

By default WordPress hook callbacks from WordPress core are not
instrumented because it increases agent's overhead. It is however
possible to enable this instrumentation with INI setting.


agent/fw_wordpress.c
```diff
@@ -610,20 +612,22 @@ NR_PHP_WRAPPER(nr_wordpress_add_filter) {
...
+        // We can cheat here: wraprecs on callables are always transient, so if
+        // there's a wordpress_plugin_theme set we know it's from this
+        // transaction, and we don't have any issues around a possible
+        // multi-tenant setup.
+        if (NULL != callback_wraprec) {
+          // Unlike Drupal, we don't free the wordpress_plugin_theme, since we
+          // know it's transient anyway, and we only set the field if it was
+          // previously NULL.
+          callback_wraprec->wordpress_plugin_theme = wordpress_plugin_theme;
+        }
```
ZNeumann, line 629: I am a little confused as to what the comment is explaining in relation to the code. To me it seems like the comment is indicating that we will be checking `callback_wraprec->wordpress_plugin_theme` for `NULL` before setting it, because otherwise it has already been done.

lavarou: Good eye! This check `NULL == callback_wraprec->wordpress_plugin_theme` in line 624 (previously line 620) was removed by a mistake during re-base!

lavarou: Addressed with 263f6563669612bcfcd778d3ec78a61ab1c5d87b.

bduranleau-nr (approved):

mfulb:
Whats tests exist to check if this config option is working properly when `TRUE` and `FALSE`?  The existing tests this PR modifies sets the option to `TRUE` in order for the tests to work properly.  Aside from that is this something that ought to be tested here or in multiverse?

lavarou:
> Whats tests exist to check if this config option is working properly when TRUE and FALSE? The existing tests this PR modifies sets the option to TRUE in order for the tests to work properly. Aside from that is this something that ought to be tested here or in multiverse?

Good point! It was also pointed out [here](https://github.com/newrelic/newrelic-php-agent/pull/799#issuecomment-1867010377). As I explained [here](https://github.com/newrelic/newrelic-php-agent/pull/799#issuecomment-1867071569), additional tests are required. I will add them either to this PR, once all its parent PRs are merged, or to another, separate PR.

lavarou:
>> Whats tests exist to check if this config option is working properly when TRUE and FALSE? The existing tests this PR modifies sets the option to TRUE in order for the tests to work properly. Aside from that is this something that ought to be tested here or in multiverse?
> Good point! It was also pointed out https://github.com/newrelic/newrelic-php-agent/pull/799#issuecomment-1867010377. As I explained https://github.com/newrelic/newrelic-php-agent/pull/799#issuecomment-1867071569, additional tests are required. I will add them either to this PR, once all its parent PRs are merged, or to another, separate PR.

@bduranleau-nr @mfulb 496d3aaad2ecf3e44b8a2e6a2e735430022436d1 adds a few tests that improve wordpress instrumentation test coverage. Please take a look.

tests/integration/frameworks/wordpress/test_wordpress_slow_hooks_only.php
```diff
@@ -0,0 +1,74 @@
...
+/* WordPress mock app */
+require_once __DIR__.'/mock-wordpress-app.php';
+
+function wp_core_slow_action_1(...$args) {
+  usleep(2010); // must be greater than newrelic.framework.wordpress.hooks_threshold
+  return;
+}
+
+function wp_core_fast_action_1(...$args) {
+  usleep(1010); // must be less than newrelic.framework.wordpress.hooks_threshold
```
bduranleau-nr, line 63: I would be a little concerned that this test may occasionally fail depending on the system/environment with less than 1ms buffer before hitting the 2ms threshold. Potentially consider lowering this value to reduce this risk.

ZNeumann (approved):

tests/integration/frameworks/wordpress/test_wordpress_slow_hooks_only.php
mfulb, line 63: I would write this test differently:
- A callback that takes 100ms
- A callback that takes 10ms

I would set the threshold at 50ms so we can confirm the longest test is instrumented and the shorter one is not.

If system latency causes the shorter one to take > 50ms then change them to 1000ms and 10ms and set the threshold at 500ms.

mfulb: I'd make the callbacks take much longer so that is less a concern - say 1000ms and 10ms and set the threshold to 500ms.

mfulb (approved):
Nice example of how to write and document integration tests.

tests/integration/frameworks/wordpress/test_wordpress_slow_hooks_only.php
lavarou: @mfulb @bduranleau-nr Does a74caf46b74427fff5d078b7c29dce608b372089 look better?

bduranleau-nr: LGTM

mfulb: Yeah that looks good.

PR #812, opened 2024-01-12
feat(agent): Always link error traces using transaction ID

mfulb:
This PR adds the transaction ID to all error traces, even when CAT and DT are disabled.
This is added in two places:
   - the "CatGUID" field of Error Traces
   - as a "guid" attribute in the intrinsic attributes of both Error Traces and Transaction Errors.
 
It takes advantage of the fact that a random transaction ID is generated when all transactions are started independent of the DT/CAT settings.

Many integration tests were updated to verify the transaction ID field and "guid" attributes is now populated.

Additional tests were created to verify the value of the transaction ID is correct.  This required adding some internal testing APIs which give access to internal agent state needed for the integration test.  These internal testing APIs are not built into release artifacts.


agent/php_api_internal.c
```diff
@@ -382,6 +383,57 @@ PHP_FUNCTION(newrelic_get_trace_json) {
   nr_set_destroy(&fas_metadata.active_segments);
 }
 
+PHP_FUNCTION(newrelic_get_error_json) {
```
ZNeumann, line 386: Is the only purpose for this in integration testing?

mfulb: Yes - it is so the test can verify the transaction id injected into the error trace JSON has the correct value.  Since this value changes from test to test I added a way to get the value used for a given test run.

tests/integration/errors/test_error_linking_dt_on_cat_off.php
```diff
@@ -0,0 +1,60 @@
...
+  newrelic_notice_error(new RuntimeException('Hi!'));
+}
+
+throw_it();
+
+/* capture error trace json and verify transaction GUID */
+$json = newrelic_get_error_json();
+$payload = json_decode($json, true);
+$guid = newrelic_get_transaction_guid();
+tap_equal($guid, $payload[5], "Error trace includes txn guid");
```
lavarou, line 60: Doesn't this require corresponding `/*EXPECT Error trace includes txn guid - ok*/`?

mfulb: Yep - just discovered this and had this commit already:
[7eabda6](https://github.com/newrelic/newrelic-php-agent/pull/812/commits/7eabda6754030cb3fe1f8b4dbf424d50273fa237)

ZNeumann (approved):

axiom/cmd_txndata_transmit.c
```diff
@@ -52,12 +53,19 @@ char* nr_txndata_error_to_json(const nrtxn_t* txn) {
   user_attributes = nr_attributes_user_to_obj(txn->attributes,
                                               NR_ATTRIBUTE_DESTINATION_ERROR);
 
-  json = nr_error_to_daemon_json(txn->error, txn->name, agent_attributes,
-                                 user_attributes, txn->intrinsics,
-                                 txn->request_uri);
+  // add guid to aid error linking ui
+  // make copy of txn->intrisics to not cause it to be modified for other 
+  // potential uses during conversion to flatbuffer
+  intrinsics_attributes = nro_copy(txn->intrinsics);
```
lavarou, line 59: So the worry is that something inside `nr_error_to_daemon_json` will depend on `guid` in `txn->intrinsics` or that something outside of `nr_txndata_error_to_json` will depend on `guid` in `txn->intrinsics`? And the `nro_copy` strategy was used because `util_object.h` does not provide `nro_del_hash_string` API, right?

mfulb: I didn't want to modify `txn->instrinsics` for code that ran after `nr_error_to_daemon_json`.  Probably more out of caution of causing the least impact with these changes.  It seems most other agents send "guid" in intrinsics so perhaps it is being a bit cautious.

lavarou (approved):
LGTM

PR #817, opened 2024-01-24
tests: remove broken exception handling unit tests

bduranleau-nr:
PHP embed SAPI, used by unit tests, demonstrates memory issues when PHP code executed by way of `nr_php_call` or `tlib_php_request_eval` throws a PHP Exception. This functionality is best suited to be exercised via integration test.


lavarou (approved):
I agree: this functionality is best tested with integration tests.

ZNeumann (approved):

PR #824, opened 2024-01-31
refactor(daemon): GOPATH to Go Modules migration

bduranleau-nr:
This is a follow up to @mfulb's effort [here](https://github.com/newrelic/newrelic-php-agent/pull/732) to convert Go source to modules.

Previously, Golang had operated on a src model, where a GOPATH would point to the relevant Go application code. This model required structuring the project a certain way, with named directories like `src`, `bin`, and `pkg`, as well as requiring `vendor` directories to track and pin dependency versions between checkouts/environments.

Golang deprecated the `src` methodology in favor of a module model. Under this paradigm, Go projects are arranged around a `go.mod` file (or multiple files for multiple modules) and may contain as many packages as desired, while still following basic rules requiring at least one `main` package with a `main` function in order to run.

To best support this, it has been determined to refactor the daemon into a more appropriate model following many of the widely accepted Go language best practices: 
- [Effective Go](https://go.dev/doc/effective_go)
- [Using Go Modules](https://go.dev/blog/using-go-modules)
- [Migrating to Go Modules](https://go.dev/blog/migrating-to-go-modules)

Since the New Relic PHP Agent daemon is not a library, only one module is required to represent the entire daemon. Using the guidance above, we have decided to name this `github.com/newrelic/newrelic-php-agent/daemon`.

The directory structure of the daemon has also changed significantly. Since we no longer want to use `src`, that directory has been renamed to a more descriptive `daemon`. Additionally, packages have been migrated to more descriptive and intuitive directories. 
- For the executable code, those packages can now be found under `daemon/cmd`
- The library of internal source code has been moved to `daemon/internal`
- JSON test data files now reside under `daemon/test-data`

Additionally, we have decided to remove the `vendors` folder from the git repository, since modules allows handling of version pinning without the need for dependency code to be checked into the repo.

In order to support the typical way we develop (especially on vscode-style IDEs), a `go.work` file has been added under the root `newrelic-php-agent` directory. This instructs the Go toolchain to look in the `daemon/` directory for the relevant module, since Go expects the editor to be opened in the same directory as the `go.mod` file by default.

Since even the latest version of Debian (bookworm) only has Go 1.19 as the latest available in the apt repo, the logic to install golang has been separated and manually handled via wget and linking the go binary for the current supported version (1.21.1 as of this PR).

The Daemon logic previously handled in the root Makefile has been extracted to a separate `daemon/Makefile`, where the root makefile now acts as a passthrough.

03/05 Rebase Related Fixes:
- 2ea417d4006d9df102133c6e1e4e9cb96f5da277
- 580963cfca69454ea6da3be8dc00931c32a21458


files/Dockerfile
```diff
@@ -71,18 +71,14 @@ RUN apt-get update && apt-get install -y \
   zip && apt-get clean
 
 #
-# If the debian version is jessie or stretch, we need to install go manually;
-# otherwise, we just install the golang package from the repository.
-# go 1.11.6 matches the version that buster uses.
+# The Debian package repository is woefully behind the latest version of go
+# in order to download and run the latest version we support, we have to install
+# go manually
```
lavarou, line 76:
```suggestion
# Download and install Go
```

daemon/Makefile
```diff
@@ -0,0 +1,94 @@
...
+	$(GO) clean $(GOFLAGS) -i ./...
+
+
+.PHONY: go-minimum-version
+go-minimum-version:
+	@if $(GO) version | awk '/go1.(1[1-9]|[2-9][0-9])[\. ]/ { exit 0 } { exit 1 }'; then \
+		true; \
+	else \
+		echo -n 'Go 1.11 or newer required; found '; $(GO) version; false; \
+	fi
```
lavarou, line 56: Is it still needed? Doesn't Go version in `go.mod` suffice?

bduranleau-nr: Removed in b9907cdcd023619f675e5cf2cd12a9ca3818e4db

daemon/Makefile
```diff
@@ -0,0 +1,85 @@
...
+
+# Default go invocation command
+GO := go
+
+# module path
+GO_MODULE := github.com/newrelic/newrelic-php-agent/daemon
+
+# Pass relevant flags from secrets.mk && version.mk to the version and secrets packages
+GOFLAGS := -ldflags '-X $(GO_MODULE)/internal/newrelic/version.Number=$(AGENT_VERSION) 							\
+					-X $(GO_MODULE)/internal/newrelic/version.Commit=$(GIT_COMMIT) 								\
```
mfulb, line 23: This looks like it needs formatting but I'm having trouble making a suggestion via GH.

mfulb (approved):
Very nice work Brian!

daemon/Makefile
bduranleau-nr: This is due to tabs being differently aligned in my editor vs GH. Try checking out the PR and viewing in your editor- if it looks bad in the editor, we can change it. But if it's just GH, we're probably better off leaving it. In my editor this is all aligned.

lavarou (approved):

PR #825, opened 2024-01-31
refactor: condense exception handling instrumentation

ZNeumann:
Move exception handler instrumentation from fcall_begin to fcall_end. This requires a change of how we are obtaining the fields of the exception, because calling `nr_php_call` during the handling of an exception does not play nicely with PHP.


agent/php_agent.c
```diff
@@ -87,6 +90,32 @@ zval* nr_php_get_zval_object_property(zval* object,
...
+    return 0;
+  }
+
+  name = (char*)nr_alloca(nr_strlen(cname) + 1);
+  nr_strcpy(name, cname);
+
+  if (nr_php_is_zval_valid_object(exception)) {
+    if (nr_php_error_zval_is_exception(exception)) {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
+      ce = zend_get_exception_base(Z_OBJ_P(exception));
```
ZNeumann, line 108: This is inline with what the php source code did when we called via `nr_php_call(getTrace)`, etc. This is needed to maintain consistent results within our tests. Without getting the base class entry, certain values are incorrect for either errors/exceptions.

agent/php_agent.h
```diff
@@ -137,9 +137,22 @@ typedef unsigned int nr_php_object_handle_t;
  *
  * Returns : The specified element or NULL if it was not found.
  */
-extern zval* nr_php_get_zval_object_property(zval* object,
+extern zval* nr_php_get_zval_object_property(zval* exception,
```
ZNeumann, line 140:
```suggestion
extern zval* nr_php_get_zval_object_property(zval* object,
```

agent/php_error.c
```diff
@@ -407,14 +412,13 @@ static zval* nr_php_error_exception_stack_trace(zval* exception TSRMLS_DC) {
...
 static char* nr_php_error_exception_file(zval* exception TSRMLS_DC) {
-  zval* file_zv = nr_php_call(exception, "getFile");
+  zval* file_zv = nr_php_get_zval_object_property(exception, "file");
   char* file = NULL;
 
   if (nr_php_is_zval_valid_string(file_zv)) {
     file = nr_strndup(Z_STRVAL_P(file_zv), Z_STRLEN_P(file_zv));
   }
 
-  nr_php_zval_free(&file_zv);
```
ZNeumann, line 417: When we obtained these values through nr_php_call, a copy was made that we needed to free. With directly calling get_*_property, there is no copy

agent/php_error.c
```diff
@@ -387,12 +387,17 @@ void nr_php_error_install_exception_handler(TSRMLS_D) {
 static zval* nr_php_error_exception_stack_trace(zval* exception TSRMLS_DC) {
   zval* trace;
 
-  trace = nr_php_call(exception, "getTrace");
-  if (!nr_php_is_zval_valid_array(trace)) {
-    nr_php_zval_free(&trace);
+  trace = nr_php_get_zval_base_exception_property(exception, "trace");
+  Z_ADDREF_P(trace);
```
ZNeumann, line 391: Adding a ref count here was desired because the zval is directly returned. In other cases where the zval is not returned, the ref count is left alone for performance

agent/php_execute.c
```diff
@@ -2014,11 +1991,18 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
 
   if (wraprec && wraprec->is_exception_handler) {
     /*
-     * An exception handler sets the transaction exception in fcall_begin
-     * and does not have a return value, like an fcall_end during an
-     * uncaught exception. This code path is here to simplify and
-     * explicitly enumerate the possible cases.
+     * Before starting the error handler segment, put the error it handled on
```
ZNeumann, line 1994:
```suggestion
     * After running the error handler segment, put the error it handled on
```

agent/php_agent.c
```diff
@@ -87,6 +90,32 @@ zval* nr_php_get_zval_object_property(zval* object,
   return 0;
 }
 
+zval* nr_php_get_zval_base_exception_property(zval* exception,
+                                              const char* cname) {
+  char* name;
+  zend_class_entry* ce;
+
+  if ((0 == exception) || (0 == cname) || (0 == cname[0])) {
```
bduranleau-nr, line 98: If we're checking if a pointer is `NULL`, we should be explicit in using `NULL` as the comparison value, rather than `0`, which can easily be misinterpreted for a numerical value check rather than a `NULL` check. In the same way, if we're returning a pointer value, we should stick to the convention of returning an explicit `NULL` where applicable rather than a `0`.

agent/php_agent.c
bduranleau-nr: Can you add the above context in a code comment?

agent/php_error.c
bduranleau-nr: Would it make more sense to check if the returned `trace` value is `NULL` before incrementing the refcount? How does `Z_ADDREF_P` and `Z_DELREF_P` behave when fed a `NULL` value?

agent/php_agent.c
ZNeumann: Good call, these lines are just copied from the similar function directly above on line https://github.com/newrelic/newrelic-php-agent/pull/825/files#diff-3ae60d50007d698aae55659ea75350561bf514cf5b585e8c86676f26aff6dd04R71

Will change both

agent/php_execute.c
```diff
@@ -2014,11 +1991,18 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
...
-     * An exception handler sets the transaction exception in fcall_begin
-     * and does not have a return value, like an fcall_end during an
-     * uncaught exception. This code path is here to simplify and
-     * explicitly enumerate the possible cases.
+     * After running the error handler segment, put the error it handled on
+     * the transaction. The choice of E_ERROR for the error level is
+     * basically arbitrary, but matches the error level PHP uses if there
+     * isn't an exception handler, so this should give more consistency for
+     * the user in terms of what they'll see with and without an exception
+     * handler installed.
```
lavarou, line 1999:
```suggestion
     * After running the exception handler segment, create an error from
     * the exception it handled, and save the error in the transaction.
     * The choice of E_ERROR for the error level is basically arbitrary,
     * but matches the error level PHP uses if there isn't an exception
     * handler, so this should give more consistency for the user in terms
     * of what they'll see with and without an exception handler installed.
```

agent/php_agent.c
ZNeumann: fixed in [18f7088](https://github.com/newrelic/newrelic-php-agent/pull/825/commits/18f708807d20e7056fcaa7d1f80a14a8a985878a)

agent/php_agent.c
ZNeumann: added in [18f7088](https://github.com/newrelic/newrelic-php-agent/pull/825/commits/18f708807d20e7056fcaa7d1f80a14a8a985878a)

agent/php_agent.h
```diff
@@ -133,13 +133,26 @@ typedef unsigned int nr_php_object_handle_t;
...
 
+/*
+ * Purpose : Extract the named entity from a PHP exception zval.
+ *           This determines if the passed exception is a PHP error
+ *           or a PHP exception, and extracts accordingly.
+ *
+ * Params  : 1. The exception to extract the property from.
+ *           2. The name of the exception property.
+ *
+ * Returns : The specified element or NULL if it was not found.
```
lavarou, line 151: Who owns returned `zval*`? I.e.: does it need to be freed by the caller?

agent/php_agent.c
```diff
@@ -87,6 +90,37 @@ zval* nr_php_get_zval_object_property(zval* object,
...
+                                              const char* cname) {
+  char* name;
+  zend_class_entry* ce;
+
+  if ((NULL == exception) || (NULL == cname) || (0 == cname[0])) {
+    return 0;
+  }
+
+  name = (char*)nr_alloca(nr_strlen(cname) + 1);
+  nr_strcpy(name, cname);
```
lavarou, line 103: Why is this `strcpy` needed here?

ZNeumann: I am unsure. The copy is present in the sibling function that I based this one around, and I didn't want to risk changing it. I can remove the copy from both and see if it passes the tests...

agent/php_agent.h
ZNeumann: added a clarifying comment in [34e9324](https://github.com/newrelic/newrelic-php-agent/pull/825/commits/34e932403b8b9ac8e8543188a582cf5c9e2d8ddd)

agent/php_agent.c
ZNeumann: It still passes the CI, which I think is in line with our expectations. [e76d8e3](https://github.com/newrelic/newrelic-php-agent/pull/825/commits/e76d8e3b38a801ca6c407cabad0e13278dc1bfd1)

bduranleau-nr (approved):

lavarou (approved):
Nice cleanup!

PR #826, opened 2024-02-02
tests: simplify EXPECTED_PHP_PACKAGES "null" value processing

mfulb:
This change determines if the string "null" (with leading and trailing whitespace) was given as the `EXPECT_PHP_PACKAGES` stanza body and if so sets it to "null" to simplify later processing.

Changes needed to test PR #818.

### DIFF
--- a/src/newrelic/integration/parse.go
+++ b/src/newrelic/integration/parse.go
@@ -13,7 +13,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"unicode"
 )
@@ -286,8 +285,7 @@ func parseConfig(test *Test, content []byte) error {
 
 func parsePhpPackages(test *Test, content []byte) error {
 	// clean up provided string data if it is "null"
-	nullRE := regexp.MustCompile(`^\s*null\s*`)
-	if nullRE.Match(content) {
+	if "null" == strings.TrimSpace(string(content)) {
 		test.phpPackagesConfig = []byte("null")
 		return nil
 	}
--- a/src/newrelic/integration/parse.go
+++ b/src/newrelic/integration/parse.go
@@ -288,7 +288,7 @@ func parsePhpPackages(test *Test, content []byte) error {
 	// clean up provided string data if it is "null"
 	nullRE := regexp.MustCompile(`^\s*null\s*`)
 	if nullRE.Match(content) {
-		test.phpPackagesConfig = "null"
+		test.phpPackagesConfig = []byte("null")
 		return nil
 	}
 
--- a/src/newrelic/integration/parse.go
+++ b/src/newrelic/integration/parse.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"unicode"
 )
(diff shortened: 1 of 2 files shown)


src/newrelic/integration/parse.go
```diff
@@ -284,6 +285,13 @@ func parseConfig(test *Test, content []byte) error {
 }
 
 func parsePhpPackages(test *Test, content []byte) error {
+	// clean up provided string data if it is "null"
+	nullRE := regexp.MustCompile(`^\s*null\s*`)
+	if nullRE.Match(content) {
+		test.phpPackagesConfig = []byte("null")
+		return nil
+	}
```
bduranleau-nr, line 293: if the only need is to trim leading and trailing whitespaces, Go's `strings` package already has a means to do this without the need for regex.
Ex:
```
s := string(content[:])
```
In this use case `s` should now contain a string representation of `content` without leading or trailing whitespaces.

src/newrelic/integration/test.go
```diff
@@ -513,26 +513,32 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
...
-		test_package_name_only = expectedPkgsCollection.config.package_name_only
-		if test_package_name_only {
-			t.AddNote("Tested package name only")
-		}
-		expectedPackages, err = expectedPkgsCollection.GatherInstalledPackages()
-		if nil != err {
-			t.Fatal(err)
-			return
+
+		expect_null_pkgs := "null" == string(t.phpPackagesConfig)
```
bduranleau-nr, line 517: I would suggest using camelCase for this variable name rather than underscores to maintain Go [best practices](https://go.dev/doc/effective_go#mixed-caps) (although I recognize this has been historically violated in the daemon code)

src/newrelic/integration/parse.go
bduranleau-nr: See [this function](https://github.com/newrelic/newrelic-php-agent/pull/826/files#diff-0cca25c50dc0bef5ff2f46b4a27543cb92519898088734a6db5c2de837515846R276-R279) for an example of how you can apply this here

bduranleau-nr: using `TrimSpace` as opposed to regex follows the existing code convention from the functions above, but also makes it so we don't have to [import](https://github.com/newrelic/newrelic-php-agent/pull/826/files#diff-0cca25c50dc0bef5ff2f46b4a27543cb92519898088734a6db5c2de837515846R16) the `regexp` package for this one case. From a developer standpoint, I would also argue that `TrimSpace` is clearer as to the limited scope of what the function is doing at a glance than regex which can perform much more advanced operations that require analyzing the specific regex code.

src/newrelic/integration/test.go
mfulb: I made a [commit](https://github.com/newrelic/newrelic-php-agent/pull/827/commits/1ac0d2da1148414f8460ec067be15c5739a9edad) to PR #827 which refactors the introduced variables by all of the code for `EXPECT_PHP_PACKAGES` to follow Go coding style.

mfulb: This [commit](https://github.com/newrelic/newrelic-php-agent/pull/826/commits/0c2ef65713a469769d2fe0b4871e9899b799cc74) fixes the style issue in this PR.

src/newrelic/integration/parse.go
mfulb: This [commit](https://github.com/newrelic/newrelic-php-agent/pull/826/commits/bf047013d599219a33fc42e64e122862de63e633) changes the parsing to use `TrimSpace` and is consistent with the other usage in this file.

bduranleau-nr (approved):

hahuja2 (approved):

PR #828, opened 2024-02-06
fix(axiom): handle locale for dt w3c priority

bduranleau-nr:
Account for locale differences when generating the DT w3c tracestate header. 
Mitigates a risk against unexpected locale settings using `,` rather than `.` for the decimal priority value breaking the PHP Agent's DT w3c handling.

### DIFF
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -41,7 +41,7 @@ static inline void set_dt_field(char** field, const char* value) {
 }
 
 /*
- * Purpose      Format double priority numbers to string.
+ * Purpose      Format trace's priority for tracestate in W3C header.
  *
  * @param       value   double
  * @return      char*   priority string buffer
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -1224,7 +1224,7 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
 
   priority_buf = nr_priority_double_to_str(priority);
   if (NULL == priority_buf) {
-    nrl_warning(NRL_CAT, "Failed to allocate priority buffer");
+    nrl_verbosedebug(NRL_CAT, "Failed to allocate priority buffer");
   }
 
   trace_context_header = nr_formatf(
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -1211,6 +1211,10 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
 
   nr_free(priority_buf);
 
+  nrl_verbosedebug(NRL_CAT,
+                   "Outbound W3C TraceState Context Header generated: %s",
+                   trace_context_header);
+
   return trace_context_header;
 }
 
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -1204,16 +1204,12 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
 
   priority_buf = nr_priority_double_to_str(priority);
 
-  trace_context_header
-      = nr_formatf("%s@nr=0-0-%s-%s-%s-%s-%s-%s-%" PRId64, trusted_account_key,
-                   account_id, app_id, NRBLANKSTR(span_id), NRBLANKSTR(txn_id),
-                   sampled, priority_buf, (nr_get_time() / NR_TIME_DIVISOR_MS));
+  trace_context_header = nr_formatf(
+      "%s@nr=0-0-%s-%s-%s-%s-%s-%s-%" PRId64, trusted_account_key, account_id,
+      app_id, NRBLANKSTR(span_id), NRBLANKSTR(txn_id), sampled,
+      NRBLANKSTR(priority_buf), (nr_get_time() / NR_TIME_DIVISOR_MS));
 
   nr_free(priority_buf);
-
-  nrl_verbosedebug(NRL_CAT,
-                   "Outbound W3C TraceState Context Header generated: %s",
-                   NRSAFESTR(trace_context_header));
 
   return trace_context_header;
 }
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -1213,7 +1213,7 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
 
   nrl_verbosedebug(NRL_CAT,
                    "Outbound W3C TraceState Context Header generated: %s",
-                   trace_context_header);
+                   NRSAFESTR(trace_context_header));
 
   return trace_context_header;
 }
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -18,6 +18,7 @@
 #include "util_time.h"
 #include "util_strings.h"
 #include "util_logging.h"
+#include "util_number_converter.h"
 
 /*
  * Purpose : Helper function to assign a string value to a field.  This function
@@ -1164,6 +1165,8 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
   const char* app_id;
   char* trace_context_header = NULL;
   char* sampled = "0";
+  double priority;
+  char* priority_buf = NULL;
 
   if (nrunlikely(NULL == dt)) {
     return NULL;
@@ -1197,11 +1200,16 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
     sampled = "1";
   }
 
-  trace_context_header = nr_formatf(
-      "%s@nr=0-0-%s-%s-%s-%s-%s-%.6f-%" PRId64, trusted_account_key, account_id,
-      app_id, NRBLANKSTR(span_id), NRBLANKSTR(txn_id), sampled,
-      nr_distributed_trace_get_priority(dt),
-      (nr_get_time() / NR_TIME_DIVISOR_MS));
+  priority = nr_distributed_trace_get_priority(dt);
+
+  priority_buf = nr_priority_double_to_str(priority);
+
+  trace_context_header
+      = nr_formatf("%s@nr=0-0-%s-%s-%s-%s-%s-%s-%" PRId64, trusted_account_key,
+                   account_id, app_id, NRBLANKSTR(span_id), NRBLANKSTR(txn_id),
+                   sampled, priority_buf, (nr_get_time() / NR_TIME_DIVISOR_MS));
+
+  nr_free(priority_buf);
 
   return trace_context_header;
 }
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -41,6 +41,27 @@ static inline void set_dt_field(char** field, const char* value) {
   }
 }
 
+/*
+ * Purpose      Format double priority numbers to string.
+ *
+ * @param       value   double
+ * @return      char*   priority string buffer
+ */
+static char* nr_priority_double_to_str(nr_sampling_priority_t value) {
+  char* buf = NULL;
+
+  buf = nr_formatf(NR_PRIORITY_DT_FMT, value);
+
+  for (int i = 0; i < nr_strlen(buf); i++) {
+    if (',' == buf[i]) {
+      buf[i] = '.';
+      break;
+    }
+  }
+
+  return buf;
+}
+
 nr_distributed_trace_t* nr_distributed_trace_create(void) {
   nr_distributed_trace_t* dt;
   dt = (nr_distributed_trace_t*)nr_zalloc(sizeof(nr_distributed_trace_t));
--- a/axiom/nr_distributed_trace.c
+++ b/axiom/nr_distributed_trace.c
@@ -18,7 +18,6 @@
 #include "util_time.h"
 #include "util_strings.h"
 #include "util_logging.h"
-#include "util_number_converter.h"
 
 /*
  * Purpose : Helper function to assign a string value to a field.  This function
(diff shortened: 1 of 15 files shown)


axiom/tests/test_distributed_trace.c
```diff
@@ -9,6 +9,7 @@
 #include "nr_txn.h"
 #include "nr_distributed_trace_private.h"
 #include "util_memory.h"
+#include "locale.h"
```
lavarou, line 12: Shouldn't this be:
```suggestion
#include <locale.h>
```

axiom/tests/test_distributed_trace.c
```diff
@@ -1351,6 +1352,19 @@ static void test_create_trace_state_header(void) {
...
 
+  /*
+   * Test: locale is set to use `,` instead of `.` for decimal values
+   */
+  setlocale(LC_NUMERIC, "pl_PL");
+  dt->priority = 0.123456;
+  expected = "777@nr=0-0-1234-9876-123456789-meatball!-0-0.123456-";
+  result
+      = nr_distributed_trace_create_w3c_tracestate_header(dt, span_id, txn_id);
+  tlib_pass_if_not_null("agent should handle different locale correctly",
```
lavarou, line 1363:
```suggestion
  tlib_pass_if_not_null("locale should not affect priority formatting", 
```

axiom/tests/test_distributed_trace.c
bduranleau-nr: Yup, fixed

axiom/nr_distributed_trace.c
```diff
@@ -1197,11 +1200,20 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
...
+  trace_context_header
+      = nr_formatf("%s@nr=0-0-%s-%s-%s-%s-%s-%s-%" PRId64, trusted_account_key,
+                   account_id, app_id, NRBLANKSTR(span_id), NRBLANKSTR(txn_id),
+                   sampled, priority_buf, (nr_get_time() / NR_TIME_DIVISOR_MS));
+
+  nr_free(priority_buf);
+
+  nrl_verbosedebug(NRL_CAT,
+                   "Outbound W3C TraceState Context Header generated: %s",
+                   trace_context_header);
```
lavarou, line 1216: `nr_formatf` can return NULL, so I'd suggest to be ready for that like this:
```suggestion
  nrl_verbosedebug(NRL_CAT,
                   "Outbound W3C TraceState Context Header generated: %s",
                   NRSAFESTR(trace_context_header));
```

bduranleau-nr: In that vein probably best to also add a `NULL` check to the create tracestate header logic as well

lavarou: [`newrelic.special=debug_cat`](https://github.com/newrelic/newrelic-php-agent/blob/72a524cd2b0717b55d77504f935ecd246ea27cc3/agent/scripts/newrelic.ini.private.template#L51) exists to limit debug logging and only log tracing related verbose debug messages when it's set. Shouldn't this debug log be also only logged when `debug_dt` special is set? I'd also be fine with not adding `debug_dt` but re-using `debug_cat`.

bduranleau-nr: I can add a check for the `debug_cat` special flag if we only want to log this when that flag is set. A `debug_dt` option would be nice but is a little out of scope for this PR and should probably be it's own effort.

bduranleau-nr: Actually, we don't have access to the Global macros that make accessing the transaction possible here- `NRPRG` and `NRTXN` live under `agent/php_newrelic.h`, and this specific function does not receive the transaction object as an argument.

bduranleau-nr: I can lift up the debug call out of this function and place it in the caller, which does have the txn object `nr_txn_create_w3c_tracestate_header`

bduranleau-nr: see changes here 6b6d6749622af92aceaf95c5f58ad3c9e12d5b18

axiom/nr_txn.c
```diff
@@ -3426,7 +3432,8 @@ static void nr_txn_add_log_event(nrtxn_t* txn,
     event_dropped = true;
   } else {
     /* event passed log level filter so add it */
-    e = log_event_create(log_level_name, log_message, timestamp, context_attributes, txn, app);
+    e = log_event_create(log_level_name, log_message, timestamp,
+                         context_attributes, txn, app);
```
lavarou, line 3436: Stage the file and then run `git clang-format` to format only parts that were changed and stage again. This flow avoids reformatting the whole file which makes it harder to review.

axiom/nr_distributed_trace.c
```diff
@@ -1164,6 +1165,8 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
   const char* app_id;
   char* trace_context_header = NULL;
   char* sampled = "0";
+  double priority;
```
lavarou, line 1168: The return value of `nr_distributed_trace_get_priority` is `nr_sampling_priority_t`:
```suggestion
  nr_sampling_priority_t priority;
```

axiom/util_number_converter.c
```diff
@@ -46,6 +47,22 @@ void nr_itoa(char* buf, size_t len, int x) {
...
+  for (int i = 0; i < PRIORITY_MAXLEN; i++) {
+    if (',' == buf[i]) {
+      buf[i] = '.';
+      break;
+    }
+  }
+
+  return buf;
+}
+
```
lavarou, line 65: This function does not belong to this module: it is too distributed tracing specific. It knows specifics about how 'priority' should be formatted for tracestate header, not how priority (which btw is of type `nr_sampling_priority_t`) as double should be formatted in general. Maybe priority should be formatted differently in other contexts? This function should become a helper (`static`) function in `nr_distributed_trace.c` module and specifically be used to format priority for tracestate header.

bduranleau-nr: > Maybe priority should be formatted differently in other contexts?

I think that bespoke handling of priority formatting is part of the reason for this bug. If we're printing priority it should be standardized to a specific format, which was my rationale for making this a utility function- that way if there are other places in the code that handle the priority -> string conversion incorrectly, we can directly reference this function. It would also serve as a template for any priority handling code going forward that I think may be lost if hidden in `nr_distributed_trace.c`. Declaring it as a utility function also allows us to directly call it in any axiom unit tests (or debug log messages) if need be outside of that file.

I agree that it's a little too specific perhaps to live in this util file and maybe should be relocated, but would still like to avoid making it `static` and exist solely in `nr_distributed_trace.c`.

axiom/nr_txn.c
bduranleau-nr: fixed in 36f308404aac646b13619b555d588463f5dc8fb1

axiom/util_number_converter.c
lavarou: My point is that priority in distributed tracing's New Relic's W3C tracestate needs a '.' as radix separator and 6 decimal digits - this is very specific functionality. Maybe priority needs to be formatted differently in other tracing header - I don't know all the use cases and formats. This use case is known and fairly well defined and is local to distributed tracing's New Relic's W3C tracestate.

bduranleau-nr: I'll do a quick sweep of the code and see if this function is useful anywhere else and implement your suggestion if not

axiom/tests/test_distributed_trace.c
```diff
@@ -1351,6 +1352,19 @@ static void test_create_trace_state_header(void) {
                         nr_strstr(result, expected));
   nr_free(result);
 
+  /*
+   * Test: locale is set to use `,` instead of `.` for decimal values
+   */
+  setlocale(LC_NUMERIC, "pl_PL");
```
mfulb, line 1358: This seems a good place to verify the locate switch was successful or fail this test case.

axiom/util_number_converter.c
mfulb: I think most priority fields in JSON are created via `nro_to_json()` for other usages in the code.  This used `%.5f` fwiw.

axiom/util_number_converter.c
```diff
@@ -46,6 +47,22 @@ void nr_itoa(char* buf, size_t len, int x) {
   nr_strlcpy(buf, p, len);
 }
 
+#define PRIORITY_MAXLEN 6
+char* nr_priority_double_to_str(double value) {
+  char* buf = NULL;
+
+  buf = nr_formatf("%.6f", value);
```
mfulb, line 54: This format should use `PRIORITY_MAXLEN` instead of a hard coded constant.

axiom/util_number_converter.c
mfulb: I think this would be a good helper function in `util_sampling.c` which acts as a conversion function from the native data type to the accepted string representation.

axiom/util_number_converter.c
mfulb: Should accept `nr_sampling_priority_t`.

ZNeumann (approved):

axiom/util_number_converter.c
bduranleau-nr: fixed here: 880b18e98ed259ea876d70bc91e43a018c0b949f

axiom/tests/test_distributed_trace.c
bduranleau-nr: attempted to add this test here: 535f58950e96d177603bb38511a78b5c283e8e3c
but it appears to be causing test failures on the PR runner despite passing successfully on my local machine: https://github.com/newrelic/newrelic-php-agent/actions/runs/7835406513/job/21380743765?pr=828

possibly a result of the images the PR runner depends on not having locales installed? 
Error message is a very vague `SIGNAL 11`

axiom/util_number_converter.c
bduranleau-nr: function declaration moved in this commit: 880b18e98ed259ea876d70bc91e43a018c0b949f
`debug_dt` flag added in this commit: 29a2c65d0e577e32ab2c14300173255e2e6773bc
format string addressed in this commit: f37e684cc8925737b8a5664750f179975032d937

axiom/tests/test_distributed_trace.c
bduranleau-nr: Changed test logic to call setlocale directly in the tlib test function and got a more meaningful output here: https://github.com/newrelic/newrelic-php-agent/actions/runs/7835578673/job/21381288423?pr=828#step:9:46
It looks like the builder images are incapable of changing the locale.

bduranleau-nr: For now I think we'll have to remove this test to allow the PR runner to pass while we bake locales into the builder images

bduranleau-nr: removed for now in 8668b955262e9463f790ebe58e2c005648636881

mfulb (approved):

axiom/nr_distributed_trace.c
```diff
@@ -1197,11 +1221,19 @@ char* nr_distributed_trace_create_w3c_tracestate_header(
     sampled = "1";
   }
 
+  priority = nr_distributed_trace_get_priority(dt);
+
+  priority_buf = nr_priority_double_to_str(priority);
+  if (NULL == priority_buf) {
+    nrl_warning(NRL_CAT, "Failed to allocate priority buffer");
```
lavarou, line 1228: Isn't warning too high of a log level here. There have been situations where excessive logging caused issues on customer systems. I'd suggest being more conservative:
```suggestion
    nrl_verbosedebug(NRL_CAT, "Failed to allocate priority buffer");
```

axiom/nr_distributed_trace.c
```diff
@@ -18,6 +18,7 @@
 #include "util_time.h"
 #include "util_strings.h"
 #include "util_logging.h"
+#include "util_number_converter.h"
```
lavarou, line 21: This header is no longer needed:
```suggestion
```

axiom/nr_distributed_trace.c
```diff
@@ -40,6 +41,27 @@ static inline void set_dt_field(char** field, const char* value) {
   }
 }
 
+/*
+ * Purpose      Format double priority numbers to string.
```
lavarou, line 45:
```suggestion
 * Purpose      Format trace's priority for tracestate in W3C header.
```

axiom/tests/test_distributed_trace.c
```diff
@@ -1351,6 +1352,18 @@ static void test_create_trace_state_header(void) {
...
+   * Test: locale is set to use `,` instead of `.` for decimal values
+   */
+  dt->priority = 0.123456;
+  expected = "777@nr=0-0-1234-9876-123456789-meatball!-0-0.123456-";
+  result
+      = nr_distributed_trace_create_w3c_tracestate_header(dt, span_id, txn_id);
+  tlib_pass_if_not_null("locale should not affect priority formatting", 
+                        nr_strstr(result, expected));
+  nr_free(result);
+  setlocale(LC_NUMERIC, "en_US");
```
lavarou, line 1365: Shouldn't https://github.com/newrelic/newrelic-php-agent/commit/8668b955262e9463f790ebe58e2c005648636881 remove this also?
```suggestion
```

axiom/nr_distributed_trace.c
bduranleau-nr: In this case I would be concerned if this value was `NULL` since it would indicate that the call to `nr_formatf` in the `nr_priority_double_to_str` function failed. This should rarely (if ever) happen and I think it would warrant a higher log level for that reason.

axiom/tests/test_distributed_trace.c
bduranleau-nr: actually I think it may be better to add the first call to setlocale back in so it can at least be tested locally via dockerfile... otherwise this test ought to be removed altogether

bduranleau-nr: like this: 164e8738b3665ec2ddaf209a2005d4c4862d56a8

lavarou (approved):

PR #838, opened 2024-02-26
fix(agent): fix Lumen detection

lavarou:
Remove 'bootstrap/app.php' from the list of signature files used to detect
Laravel - Laravel is detected with 'illuminate/foundation/application.php'
for all supported Laravel versions: 6, 7, 8, 9 and 10.
Using 'bootstrap/app.php' causes Laravel to be detected before Lumen can be
detected because 'bootstrap/app.php' is the second file loaded when Lumen
based app is handling the request.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -381,8 +381,6 @@ static const nr_framework_table_t all_frameworks[] = {
      NR_FW_LARAVEL}, /* 5.0.15-5.0.x */
     {"Laravel", "laravel", NR_PSTR("bootstrap/cache/compiled.php"), 0, nr_laravel_enable,
      NR_FW_LARAVEL}, /* 5.1.0-x */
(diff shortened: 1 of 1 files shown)


agent/php_execute.c
```diff
@@ -857,6 +855,9 @@ static nrframework_t nr_try_detect_framework(
       }
 
       nr_framework_log("detected framework", frameworks[i].framework_name);
+      nrl_verbosedebug(
+          NRL_FRAMEWORK, "framework '%s' detected with %s, which ends with %s",
+          frameworks[i].framework_name, filename, frameworks[i].file_to_check);
```
lavarou, line 860: I don't know if it's interesting to log this or not. I used it in tests to ensure the framework is detected with a file that I thought it would be detected.

zsistla (approved):

mfulb (approved):

ZNeumann (approved):

PR #846, opened 2024-03-06
fix(agent): change detection file for Magento

hahuja2:
This PR changes the detection file for Magento to ensure compatibility with PHP 8.2 and above because the current file that is used is not detected because of the empty file optimizations that were introduced in PHP 8.2


zsistla (approved):

lavarou (approved):
Thanks!

ZNeumann (approved):

PR #823, opened 2024-01-31
feat(agent): Add Yii v2 framework instrumentation

razvanphp:
- [x] Implement Yii v2 instrumentation for auto transaction naming
- [x] Implement non-web transactions (yiic CLI) — works out-of-the-box
- [x] Update documentation
- [x] Implement wrapper for [Yii2 ErrorHandler](https://www.yiiframework.com/doc/api/2.0/yii-web-errorhandler) (currently NewRelic is blind because of the custom PHP error_handler of the framework)
- [ ] Add integration tests (needed?)

Fixes #821

I've recompiled the agent locally with those changes and transaction names correctly show up now:

Screenshot 2024-01-31 at 07 23 53

Screenshot 2024-01-31 at 11 32 02

CLI uses the same hook method:

Screenshot 2024-01-31 at 11 16 34

PR #848, opened 2024-03-11
feat(agent): add support for Yii v2

### LINKED ISSUE #821: Support for PHP Framework Yii2
Hello,

I see that in the documentation of the agent it's stated that Yii 1.x support is built-in for automatic framework detection, but we've noticed that for Yii 2 this is not the case.

To give you some context about this, Yii 2.0 was released in Oct 2014 and it's been well maintained since then to be a reputable PHP framework used in (many) production deployments.

Yii 1.1 end of life was declared since 2023, but it still receives  security updates from the community till 2026.

Since the integration extensions are not so great and have low quality of code, was wondering if it would be possible to develop the framework detection in the Newrelic Agent for everybody, I'm sure the community would love that.

Can we support somehow with the testing and implementation? Are contributions welcome for this? I could probably do it myself with some guidance and testing support.

PS: we also sent this internally through support, Feature Request ID: NR-158719

Thank you in advance!
R

lavarou:
Implement Yii v2 instrumentation for auto transaction naming.

Original work by @razvanphp in #823. Fixes #821.


tests/integration/frameworks/yii/test_basic_cli.php
```diff
@@ -0,0 +1,22 @@
...
+*/
+
+/*EXPECT_METRICS_EXIST
+Supportability/framework/Yii2/detected
+OtherTransaction/Action/test-integration-cli-action
+*/
+
+/*EXPECT_ERROR_EVENTS null */
+
+/* WordPress mock app */
```
razvanphp, line 18: This should've been Yii 🙂 but no bigie, thanks for adding the tests for me!

agent/fw_yii.c
```diff
@@ -150,10 +152,44 @@ NR_PHP_WRAPPER(nr_yii2_runWithParams_wrapper) {
...
+end:
+  nr_php_arg_release(&exception);
+}
+NR_PHP_WRAPPER_END
+
 /*
  * Enable Yii2 instrumentation.
  */
 void nr_yii2_enable(TSRMLS_D) {
+  NRPRG(framework_version) = 2;
```
razvanphp: is this still useful?

tests/integration/frameworks/yii/test_basic_cli.php
lavarou: Good catch!

lavarou:
> One more question: should the error catching be tested too?

It should be tested and is tested, just with a tool different than integration tests. That tool is using the actual framework, not the mock. I'll think some more about adding a mock to test error handling in integration tests.

tests/integration/frameworks/yii/test_basic_cli.php
lavarou: Addressed with d059d1ccf7d83098043a3d08b5423c310f4a23be.

agent/fw_yii.c
```diff
@@ -86,21 +86,134 @@ NR_PHP_WRAPPER(nr_yii_runWithParams_wrapper) {
...
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_YII2);
+
+  this_var = nr_php_scope_get(NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  if (NULL == this_var) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "Yii2: improper this");
+    goto end;
+  }
+
+  unique_idz = nr_php_call(this_var, "getUniqueId");
```
ZNeumann, line 129: do we need to call `nr_php_zval_free` on this value?

razvanphp: I followed the same logic as above, where `classz = nr_php_call(this_var, "getController");` is not freed. Not sure if needed tho....

lavarou: None of the tests catch it but it does look like a memory leak. The zval returned by `nr_php_call` is allocated here: https://github.com/newrelic/newrelic-php-agent/blob/51c11405c1efb530fe60ea178398f9450198ec74/agent/php_call.c#L51 Doc strings clearly say it should be destroyed: https://github.com/newrelic/newrelic-php-agent/blob/51c11405c1efb530fe60ea178398f9450198ec74/agent/php_call.h#L20-L21 Patched with dc5fa2c89e28140e8a4a1acd802690778f137b53.

ZNeumann (approved):

lavarou:
> >One more question: should the error catching be tested too?
>
> It should be tested and is tested, just with a tool different than integration tests. That tool is using the actual framework, not the mock. I'll think some more about adding a mock to test error handling in integration tests.

Further testing uncovered some issues with exception handling when Yii's Exception Handler is disabled. This requires further investigation.

agent/fw_yii.c
```diff
@@ -86,21 +87,134 @@ NR_PHP_WRAPPER(nr_yii_runWithParams_wrapper) {
...
+  if (nr_php_is_zval_valid_string(unique_idz)) {
```
zsistla, line 131: Should we check for `nr_php_is_zval_non_empty_string` vs just valid?

razvanphp:
> Further testing uncovered some issues with exception handling when Yii's Exception Handler is disabled. This requires further investigation.

Can you elaborate on this? I tried it and default PHP error handler was used and errors were catched by the default newrelic implementation.

agent/fw_yii.c
lavarou: Good idea! See ca81b4d6f09cdc1ca3ac482fa6e316dc9dfbdc3a

agent/fw_yii.c
```diff
@@ -86,21 +87,141 @@ NR_PHP_WRAPPER(nr_yii_runWithParams_wrapper) {
...
+  if (NULL == exception || !nr_php_is_zval_valid_object(exception)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "%s: exception is NULL or not an object",
+                     __func__);
+    NR_PHP_WRAPPER_CALL;
+    goto end;
+  }
+
+  NR_PHP_WRAPPER_CALL;
+
+  if (NR_SUCCESS
```
zsistla, line 182: why does this need to happen after NR_PHP_WRAPPER_CALL. We don't need anything else that we get from running the function in wrapper_call. Can't we simplify the whole thing.
something like:

```
  exception = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
  if (NULL == exception || !nr_php_is_zval_valid_object(exception)) {
    nrl_verbosedebug(NRL_FRAMEWORK, "%s: exception is NULL or not an object",
                     __func__);
  } else {
  if (NR_SUCCESS
!= nr_php_error_record_exception(
          NRPRG(txn), exception, nr_php_error_get_priority(E_ERROR), true,
          "Uncaught exception ", &NRPRG(exception_filters) TSRMLS_CC)) {
    nrl_verbosedebug(NRL_FRAMEWORK, "%s: unable to record exception", __func__);
  }
}
  nr_php_arg_release(&exception);
NR_PHP_WRAPPER_CALL;
}
NR_PHP_WRAPPER_END

lavarou: 0b892a6211096b0bd3f716b30f9462e83a8c38e3?

zsistla (approved):

lavarou:
> > Further testing uncovered some issues with exception handling when Yii's Exception Handler is disabled. This requires further investigation.
> 
> Can you elaborate on this? I tried it and default PHP error handler was used and errors were catched by the default newrelic implementation.

There were two problems which caused the default newrelic implementation not to record errors correctly. One has been resolved with #851 and another with #877.

PR #849, opened 2024-03-11
feat(tests): Adds override for package version

mfulb:
Adds a "override_versions_file" option for EXPECT_PHP_PACKAGE that allows specifying overrides for given expected versions detected.  This is useful for when a package did not modify the internal data structure used to by the agent to detect its version, causing the package test to fail.

The override file should be in the same directory as the test_php_package.php test and is a JSON file implementing a simple dictionary like:

{
    "4.13.0": "4.12.0"
}

In this case this specifies if the packge version "4.13.0" is detected then override this expectation with the version "4.12.0".  In this example the upstream package bumped up to 4.13.0 but did not change the internal version variable so the agent still detects it as 4.12.0.

### DIFF
--- a/daemon/internal/newrelic/integration/php_packages.go
+++ b/daemon/internal/newrelic/integration/php_packages.go
@@ -31,11 +31,12 @@ type PhpPackagesCollection struct {
 // PHP packages config describes how to collect the JSON for the packages installed
 // for the current test case
 type PhpPackagesConfiguration struct {
-	path              string
-	command           string
-	supportedListFile string
-	expectedPackages  []string
-	packageNameOnly   []string
+	path                 string
+	command              string
+	supportedListFile    string
+	overrideVersionsFile string
+	expectedPackages     []string
+	packageNameOnly      []string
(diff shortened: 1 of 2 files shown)


ZNeumann (approved):

src/newrelic/integration/php_packages.go
```diff
@@ -177,13 +179,17 @@ func NewPhpPackagesCollection(path string, config []byte) (*PhpPackagesCollectio
 		}
 	}
 
+	// option file containing overrides for expected package versions
```
zsistla, line 182: Please expand the comment to show the format expected in the file.

zsistla:
Nice update! can the integration runner docs be updated too?

mfulb:
> Nice update! can the integration runner docs be updated too?

I will update the PHP packages doc in multiverse - do we have another place for the integration runner docs now the internal agent repo is not in use?

zsistla (approved):
Thanks for the enhancements!

Issue #843, opened 2024-03-05
Thousands of E_WARNING "Failed to open stream: No such file or directory" after upgrading from 10.17.0.7 to 10.18.0.8


ruudk:
## Description

We bumped from 10.17.0.7 to 10.18.0.8 and noticed lots of E_WARNING errors being reported:

> include(/var/www/app/var/cache/prod/pools/system/9HGRhXbDyv/N/J/lJlB08wWBKnILB1B6L9w): Failed to open stream: No such file or directory

## Steps to Reproduce

I don't know

## Expected Behavior

No warning.

## Relevant Logs / Console output

```
include(/var/www/app/var/cache/prod/pools/system/9HGRhXbDyv/N/J/lJlB08wWBKnILB1B6L9w): Failed to open stream: No such file or directory

 in Symfony\Component\Cache\Adapter\PhpFilesAdapter::doFetch called at /var/www/app/vendor/symfony/cache/Traits/AbstractAdapterTrait.php (193)
 in Symfony\Component\Cache\Adapter\AbstractAdapter::getItem called at /var/www/app/vendor/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php (200)
 in Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getMetadataFor called at /var/www/app/vendor/doctrine/orm/src/EntityManager.php (318)
 in Doctrine\ORM\EntityManager::getClassMetadata called at /var/www/app/vendor/doctrine/orm/src/UnitOfWork.php (3410)
 in Doctrine\ORM\UnitOfWork::getSingleIdentifierValue called at /var/www/app/vendor/doctrine/orm/src/AbstractQuery.php (434)
 in Doctrine\ORM\AbstractQuery::processParameterValue called at /var/www/app/vendor/doctrine/orm/src/AbstractQuery.php (479)
 in Doctrine\ORM\AbstractQuery::processArrayParameterValue called at /var/www/app/vendor/doctrine/orm/src/AbstractQuery.php (415)
 in Doctrine\ORM\AbstractQuery::processParameterValue called at /var/www/app/vendor/doctrine/orm/src/Query.php (458)
 in Doctrine\ORM\Query::resolveParameterValue called at /var/www/app/vendor/doctrine/orm/src/Query.php (402)
 in Doctrine\ORM\Query::processParameterMappings called at /var/www/app/vendor/doctrine/orm/src/Query.php (318)
[...]

PR #850, opened 2024-03-12
fix(agent): Add missing errors decode and prioritization

zsistla:
Add remaining error codes listed in https://github.com/php/php-src/blob/master/Zend/zend_errors.h for completeness in decoding and prioritizing errors.

Added/modified tests accordingly.


airtonzanon:
I think with this change, flagging the `Error.Class` of different types of errors and their own class and not as `Error`, solves the problem on #843, no?

tests/integration/errors/test_E_USER_ERROR.php
```diff
@@ -5,17 +5,7 @@
...
-*/
-
-/*INI
-display_errors=1
-log_errors=0
-*/
-
-/*EXPECT_REGEX
-^\s*(PHP )?Fatal error:\s*Sample E_USER_ERROR in .*? on line [0-9]+\s*$
+The agent should capture and report E_USER_ERROR events.
```
mfulb, line 8: Why were the INI and EXPECT_REGEX sections removed here?

tests/integration/errors/test_E_ERROR.php7.4.php
```diff
@@ -6,6 +6,7 @@
 
 /*DESCRIPTION
 The agent should capture and report fatal errors.
+An E_ERROR is an exception and is handled by the agent's exception logic.
```
lavarou, line 9: I do not understand this comment. This test verifies that the agent is creating error traces when PHP code throws an exception that is neither caught nor handled.

tests/integration/errors/test_E_USER_ERROR.php
```diff
@@ -5,17 +5,7 @@
...
-stack trace.
-*/
-
-/*INI
-display_errors=1
-log_errors=0
-*/
-
-/*EXPECT_REGEX
-^\s*(PHP )?Fatal error:\s*Sample E_USER_ERROR in .*? on line [0-9]+\s*$
```
lavarou, line 18: Why was this removed?

agent/php_error.c
```diff
@@ -466,30 +477,51 @@ static char* nr_php_error_exception_message(zval* exception TSRMLS_DC) {
   return message;
 }
 
-static const char* get_error_type_string(int type) {
+const char* nr_get_error_type_string(int type) {
```
lavarou, line 480: The name of this function seems too generic. There's not enough context. Usually the convention for exported functions is to prefix them with:
- `nr_` to indicate it is New Relic space
- <filename> to indicate where the code for the function is; this gives the context of what a given function is about

Therefore I suggest updating it to `nr_php_error_get_type_string`.

tests/integration/errors/test_E_USER_ERROR.php
zsistla: good catch, I'll revert.

tests/integration/errors/test_E_USER_ERROR.php
zsistla: Please see: [8c169db](https://github.com/newrelic/newrelic-php-agent/pull/850/commits/8c169db45c7bdb8b26d19dae3cfa28a6d2e40813)

tests/integration/errors/test_E_ERROR.php7.4.php
zsistla: I'll clarify the comment.  It was meant to convey why the error class is 'Exception` instead of `E_ERROR`

please see: [8c169db](https://github.com/newrelic/newrelic-php-agent/pull/850/commits/8c169db45c7bdb8b26d19dae3cfa28a6d2e40813)

tests/integration/errors/test_E_USER_ERROR.php
zsistla: please see: [8c169db](https://github.com/newrelic/newrelic-php-agent/pull/850/commits/8c169db45c7bdb8b26d19dae3cfa28a6d2e40813)

agent/php_error.c
zsistla: Please see: [188cf39](https://github.com/newrelic/newrelic-php-agent/pull/850/commits/188cf391b2cc618dc4346a42ffc44a00d7967f1b)

zsistla:
> I think with this change, flagging the `Error.Class` of different types of errors and their own class and not as `Error`, solves the problem on #843, no?

HI @airtonzanon thanks for participating in the repo discussions.
This fixes the E_DEPRECATED and E_USER_DEPRECATED being reported with error class `Error`.
The other issues in 843 center around the increase in all errors being reported that were previously not reported if they were handled by error handlers other than PHPs standard error handler.  We are still investigating paths forward with that issue.

tests/integration/errors/test_E_ERROR.php7.4.php
lavarou: I still don't get it. The code literally throws an `Exception` [here](https://github.com/newrelic/newrelic-php-agent/blob/7490863d49ad9d62b5279cdf1903732eb8b69ead/tests/integration/errors/test_E_ERROR.php7.4.php#L71) so why `E_ERROR` is mentioned? I see that `E_ERROR` is used as value for `error.class` attribute in PHP < 7.4 version of this test. Maybe the `ce->name` (which is used as the value of `error.class` attribute [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_error.c#L676)) for PHPs 7.4+ was updated in PHP from `E_ERROR` to `Exception` for this type of an exception?

zsistla: Yes, was trying to add additional context for the next person doing a dive into error handling about why E_ERROR no longer shows up as an error.class despite the same test which previously generated an E_ERROR error.class.  

In the [pre7.4 tests](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/integration/errors/test_E_ERROR.php7.0.1.php#L29), we reported E_ERROR for the exact same test and now we record the exception.

Both agent and PHP made improvements with exceptions and agent will now report it as an Exception. E_ERROR and E_RECOVERABLE_ERROR are stored as a severity of an exception ( accessed via $exception->getSeverity() )

lavarou: I really don't think this statement is correct:
> An E_ERROR is an uncaught exception and the expected error.class 
is recorded as Exception since it is recorded in the agent's uncaught 
exception logic and not via the normal error handling logic.

Error handling logic is used for `trigger_error`, not for `throw new Exception()`. The `error.class: Exception` is probably because PHP changed what it reports as class entry name for this type of an exception.

zsistla: The "Exception" seen as the name is the name of the exception like "Exception", "RuntimeException" or even any custom exception customers have extended.

zsistla: trigger_error is only available with non-fatal errors; therefore error handling logic is intact.

zsistla: E_ERROR is a fatal PHP error that then is handled with uncaught exception logic in the agent.

In one case (non fatal errors) we populate error.class with the `type` of ERROR.
For fatal errors we populate error.class with the `name` of the exception.

zsistla: See [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1419-L1425) for specifics of why E_ERROR is treated differently.

```
      /*
       * The choice of E_ERROR for the error priority is basically arbitrary, but
       * matches the error level PHP uses if there isn't an exception handler,
       * so this should give more consistency for the user in terms of what
       * they'll see with and without an exception handler installed.
       */
```

E_ERROR will show if there is no exception handler, but in the absence of an exception handler but if this code is called there must be an exception handler, so E_ERROR (the type) will not show and we will use the name for error.class.

lavarou: Ok. So after looking at the code and running a few tests I see that the `Exception` thrown [here](https://github.com/newrelic/newrelic-php-agent/blob/7490863d49ad9d62b5279cdf1903732eb8b69ead/tests/integration/errors/test_E_ERROR.php7.4.php#L71) is treated as an error in PHPs < 7.4 and therefore `nr_php_error_cb` is used to record this exception [here](https://github.com/newrelic/newrelic-php-agent/blob/7490863d49ad9d62b5279cdf1903732eb8b69ead/agent/php_error.c#L592), and therefore `E_ERROR` is used for PHPs < 7.4 as `error.class` [here](https://github.com/newrelic/newrelic-php-agent/blob/7490863d49ad9d62b5279cdf1903732eb8b69ead/agent/php_error.c#L632). Starting with PHP 7.4, PHP started treating this exception as exception and `nr_php_error_cb` is no longer called by PHP.

agent/php_error.h
```diff
@@ -112,6 +112,15 @@ extern PHP_FUNCTION(newrelic_exception_handler);
  */
 extern int nr_php_error_get_priority(int type);
 
+/*
+ * Purpose : Converts a PHP error type into a readable string.
+ *
+ * Params  : 1. The error type.
+ *
+ * Returns : The PHP error string.
```
lavarou, line 120:
```suggestion
 * Returns : The PHP error type as constant, static string (must not be freed).
```

lavarou (approved):

tests/integration/errors/test_E_ERROR.php7.4.php
zsistla: PHP agent will never record an Uncaught error of type E_ERROR.
PHP agent has this check here to detect of an error of type E_ERROR was passed: https://github.com/newrelic/newrelic-php-agent/blob/v10.10.0.1/agent/php_error.c#L420-L436
If it was, we always ignore it, by design, since E_ERRORS are fatal and need to be handled by the nr exception handler.

Let's take the case of `test_E_ERROR.php7.4.php`. It generates the following fatal error.
Fatal error: Uncaught Exception in /usr/local/src/newrelic-php-agent/tests/integration/errors/test_E_ERROR.php7.4.php:81

When PHP first starts to process an error encountered in execution, it first checks to see if it there a fatal error in order to record the uncaught exception: https://github.com/php/php-src/blob/master/Zend/zend.c#L1427
If there is, and it is a fatal error like E_ERROR (https://github.com/php/php-src/blob/master/Zend/zend.c#L1427),
It will call zend_exception_error: https://github.com/php/php-src/blob/master/Zend/zend.c#L1442
If it's a parse or compile error, it will go ahead and call both zend_observer_error_notify and zend_error_cb  https://github.com/php/php-src/blob/master/Zend/zend_exceptions.c#L909-L910, but in our case, so we keep going.


Eventually, it works it way here where you see this portion (Uncaught Exception in /usr/local/src/newrelic-php-agent/tests/integration/errors/test_E_ERROR.php7.4.php:81) of our example test message getting created with the name (The name could be various like DivisionByZeroError, TypeError, Exception, RuntimeException, etc.):
https://github.com/php/php-src/blob/master/Zend/zend_exceptions.c#L940

Afterwards it calls zend_error_va, passing in the exception severity as the type, in this case the severity is E_ERROR: https://github.com/php/php-src/blob/master/Zend/zend_exceptions.c#L952
This calls both zend_observer_error_notify and then zend_error_cb.
But agent overwrote zend_error_cb, so it calls nr_php_error_cb.

We end up here: https://github.com/newrelic/newrelic-php-agent/blob/v10.18.0.8/agent/php_error.c#L528
We note the E_ERROR we note the "Uncaught" in the message string so we ignore it and pass execution back onto the regular zend_error_cb execution because we know we area already going to handle it with our newrelic_exception_handler which is called on exceptions: https://github.com/newrelic/newrelic-php-agent/blob/v10.18.0.8/agent/php_error.c#L241

zend_error_cb points here https://github.com/php/php-src/blob/master/main/main.c#L1302
And we get the "Fatal error" portion of our msg: Fatal error: Uncaught Exception in /usr/local/src/newrelic-php-agent/tests/integration/errors/test_E_ERROR.php7.4.php:81



Which is a long way to say:

> With PHP 7.4+, E_ERROR are fatal errors triggered by exceptions and are no longer handled using nr_php_error_cb with uses the error type as the error.class.  Instead, they are handled by newrelic_exception_handler which uses the exception name as the error.class.

lavarou: This makes sense. Thank you for taking the time to explain this.

PR #851, opened 2024-03-13
fix(agent): revert to using `zend_error_cb` hook for error reporting

ZNeumann:
Remove unwanted error reporting by overwriting zend_error_cb instead of using OAPI.


lavarou (approved):
LGTM

zsistla (approved):

php4fan:
Could this be the cause of https://github.com/newrelic/newrelic-php-agent/issues/872 somehow? I started observing that regression after updating from 10.13 to 10.19

PR #857, opened 2024-03-18
Change 10.18 codename back to ulmus (Elm)

NiklasBr:


lavarou:
Hi @NiklasBr! Thanks for your feedback. It has been incorporated in https://github.com/newrelic/newrelic-php-agent/pull/858.

PR #866, opened 2024-03-20
refactor(agent): remove oapi 'clean' callback

ZNeumann:
The "clean" callback is not needed. In order to get rid of it, we can just call the "after" callback instead, simplifying the API and future wrappers. To achieve this removal from our current code, some auditing is required. Every single use of `zval *func_return_value` must be first checked for `NULL`. Furthermore, in wrappers that used to have a "clean" callback, it must be ensured that the logic that was present in the "clean" callback is still executed in the "after" callback, even when the return value is `NULL`.


agent/php_newrelic.h
```diff
@@ -117,7 +117,7 @@ extern zend_module_entry newrelic_module_entry;
 
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
-#define NR_GET_RETURN_VALUE_PTR &func_return_value
+#define NR_GET_RETURN_VALUE_PTR func_return_value ? &func_return_value : NULL
```
zsistla, line 120: With this addition,[ all usages](https://github.com/newrelic/newrelic-php-agent/blob/9caf27b1bcd6ce58da28365b7a517180ad4a4d53/agent/fw_slim.c#L64) need to be doublechecked for null handling.

What is returned in case of an exception?  Is func_return_value null?  or is it pointing to a zval that has a value of NULL?

ZNeumann: I have manually checked all usages, though another pair of eyes could be useful. Furthermore, there are cases where the instrumentation is directly using `func_return_value` instead of calling `NR_GET_RETURN_VALUE_PTR`.

In the case of an exception, `func_return_value == NULL`

agent/fw_drupal.c
```diff
@@ -327,16 +327,6 @@ NR_PHP_WRAPPER(nr_drupal_http_request_after) {
 }
 NR_PHP_WRAPPER_END
 
-NR_PHP_WRAPPER(nr_drupal_http_request_clean) {
```
ZNeumann, line 330: With this removal, it is failing some drupal integration tests, because the tests were changed for PHP 8.0+ to not create an external segment after an exception. A simple `if` can be added to the "after" instrumentation if this behavior is still desired. Or do we want to go back to what the agent was seemingly previously doing?

(Also, this old "clean" function was leaving a dangling segment on the transaction. Not a memory leak because the transaction handles its own segment memory, but scary nonetheless)

ZNeumann: Fixed in [9081863](https://github.com/newrelic/newrelic-php-agent/pull/866/commits/9081863a7cd3554ae948f4e3118e91d8fc1b5b29)

agent/php_newrelic.h
zsistla: doublechecked these cases:

```
agent/lib_predis.c
607:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;

agent/fw_laravel.c
1099:  route = NR_GET_RETURN_VALUE_PTR;

agent/fw_magento2.c
148:  response = NR_GET_RETURN_VALUE_PTR;
194:  retval_ptr = NR_GET_RETURN_VALUE_PTR;
294:  retval_ptr = NR_GET_RETURN_VALUE_PTR;
319:  retval_ptr = NR_GET_RETURN_VALUE_PTR;

agent/php_newrelic.h
120:#define NR_GET_RETURN_VALUE_PTR func_return_value ? &func_return_value : NULL
122:#define NR_GET_RETURN_VALUE_PTR nr_php_get_return_value_ptr(TSRMLS_C)

agent/fw_mediawiki.c
169:  return_value = NR_GET_RETURN_VALUE_PTR;

agent/fw_drupal8.c
211:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
268:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
367:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
401:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
551:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;

agent/fw_wordpress.c
626:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
690:    zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;

agent/lib_zend_http.c
341:  retval_ptr = NR_GET_RETURN_VALUE_PTR;

agent/fw_drupal.c
293:  if (NULL == NR_GET_RETURN_VALUE_PTR) {
310:        = nr_drupal_http_request_get_response_header(NR_GET_RETURN_VALUE_PTR);
313:        = nr_drupal_http_request_get_response_code(NR_GET_RETURN_VALUE_PTR);
395:  return_value = NR_GET_RETURN_VALUE_PTR;

agent/fw_slim.c
66:  retval_ptr = NR_GET_RETURN_VALUE_PTR;

agent/fw_laravel_queue.c
764:  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
```

all cases either explicitly check for NULL or the value is passed to a function that checks it for NULL.

zsistla:
@ZNeumann , this looks close to ready.
Could you please 
1)rebase 
2)update any new before_after_cleans that were added since the PR was last updated

tests/integration/external/drupal7/test_bad_params_integer_headers.php8.php
```diff
@@ -25,21 +25,22 @@
...
+    [{"name":"Supportability/Logging/Metrics/PHP/enabled"},        [1, "??", "??", "??", "??", "??"]],
+    [{"name":"Errors/OtherTransaction/php__FILE__"},               [1, 0, 0, 0, 0, 0]],
+    [{"name":"Errors/all"},                                        [1, 0, 0, 0, 0, 0]],
+    [{"name":"Errors/allOther"},                                   [1, 0, 0, 0, 0, 0]],
+    [{"name":"OtherTransaction/all"},                              [1, "??", "??", "??", "??", "??"]],
+    [{"name":"OtherTransaction/php__FILE__"},                      [1, "??", "??", "??", "??", "??"]],
+    [{"name":"OtherTransactionTotalTime"},                         [1, "??", "??", "??", "??", "??"]],
+    [{"name":"OtherTransactionTotalTime/php__FILE__"},             [1, "??", "??", "??", "??", "??"]],
+    [{"name":"Supportability/framework/Drupal/forced"},            [1, 0, 0, 0, 0, 0]],
+    [{"name":"Supportability/Logging/LocalDecorating/PHP/disabled"}, [1, "??", "??", "??", "??", "??"]]
```
bduranleau-nr, line 38: Are these expectations changed at all or are the elements just reordered?

PR #868, opened 2024-03-27
feat(agent): Add supportability metrics for packages that provide major version

hahuja2:
This PR creates a new supportability metric that will provide information on what major version of the package is being used. The packages that provide the major version and will send up this supportability metric are: 
 - Drupal
 - Guzzle
 - Laravel
 - Monolog
 - Predis
 - Slim
 - Wordpress.


mfulb:
I think all versions of Guzzle and Monolog will give you a major version.  Probably other frameworks potentially will give us just major info.

agent/fw_drupal8.c
```diff
@@ -685,6 +685,8 @@ void nr_drupal_version() {
   if (nr_php_is_zval_valid_string(zval_version)) {
     char* version = Z_STRVAL_P(zval_version);
     nr_txn_add_php_package(NRPRG(txn), "drupal/core", version);
+    nr_fw_support_add_package_supportability_metric(NRPRG(txn), "drupal/core",
```
mfulb, line 688: Perhaps "drupal/core" would be a good candidate for a C macro to remove duplication of this constant and so the compiler would catch if the macro was spelled wrong and flag the error.  Whereas if the constant string here was ever modified incorrectly the compiler would still work but we'd send the wrong string.

Applies for all these string constants for the package "name" in this PR.

agent/fw_support.c
```diff
@@ -52,3 +52,41 @@ void nr_fw_support_add_logging_supportability_metric(nrtxn_t* txn,
...
+  }
+
+  char* metname = NULL;
+  char major_version[4] = {0};
+
+  /* The below for loop checks if the major version of the package is more than
+   * one digit and keeps looping until a '.' is encountered or one of the
+   * conditions is met.
+   */
+  for (int i = 0; package_version[i] && i < 4; i++) {
```
mfulb, line 71: The `i` variable should probably be `unsigned` since you aren't trying to

hahuja2:
> I think all versions of Guzzle and Monolog will give you a major version.

Good point, I have added supportability metrics for Guzzle and Monolog by major version also: [5248b97](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/5248b973b49710db11c228a9fd3ef42adb7b4c4c)

agent/fw_drupal8.c
hahuja2: Good idea, added: [1fb5879](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/1fb5879e914e31768f86c12a0ce99907a70f65a2)

agent/fw_support.c
```diff
@@ -52,3 +52,42 @@ void nr_fw_support_add_logging_supportability_metric(nrtxn_t* txn,
...
+      strncpy(major_version, package_version, i);
+      major_version[i] = '\0';
+      break;
+    }
+  }
+
+  if ('\0' == major_version[0] && '\0' != package_version[0]) {
+    strncpy(major_version, package_version, 1);
+    major_version[1] = '\0';
+  }
```
ZNeumann, line 82: This code is making some assumptions about the version format that I am unsure we can make. For example, if the version string were simply "10", my reading of this code is that we would extract "1". Is there a reason it cannot just be something like this?
```
for (int i = 0; package_version[i] && i < 4; i++) {
    if ('.' == package_version[i]) {
      break;
    }
  major_version[i] = package_version[i];
  }
```

hahuja2: Thank you for catching this! Fixed: [4b61257](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/4b612579ea2e1dfc9bb1e5a588cb8713e982e1c9)

agent/tests/test_fw_support.c
```diff
@@ -51,6 +70,42 @@ static void test_fw_supportability_metrics(void) {
...
+                                                  LIBRARY_MAJOR_VERSION_3);
+  tlib_pass_if_not_null("happy path test 3: package metric created",
+                        nrm_find(txn->unscoped_metrics, PACKAGE_METRIC
+                                 "/" LIBRARY_MAJOR_VERSION_3 "/detected"));
+
+  nr_fw_support_add_package_supportability_metric(txn, LIBRARY_NAME,
+                                                  LIBRARY_MAJOR_VERSION_4);
+  tlib_pass_if_not_null(
+      "happy path test 4: package metric created",
+      nrm_find(txn->unscoped_metrics, PACKAGE_METRIC "/1/detected"));
```
zsistla, line 95: the `LIBRARY_MAJOR_VERSION_4` define should be here.

agent/tests/test_fw_support.c
```diff
@@ -13,8 +13,15 @@ tlib_parallel_info_t parallel_info
 
 static void test_fw_supportability_metrics(void) {
 #define LIBRARY_NAME "php-package"
+#define LIBRARY_MAJOR_VERSION "7"
+#define LIBRARY_MAJOR_VERSION_2 "10"
+#define LIBRARY_MAJOR_VERSION_3 "100"
+#define LIBRARY_MAJOR_VERSION_4 "1.23"
+#define LIBRARY_MAJOR_VERSION_5 "12.34"
+#define LIBRARY_MAJOR_VERSION_6 "123.45"
```
zsistla, line 21: Nice test additions!

maybe something starting with a zero too?
like `0.4.5`?

agent/fw_support.c
hahuja2: Good idea, added: [c8f52a4](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/c8f52a4e074994c11b02109aa755bafdf672d4e2)

agent/tests/test_fw_support.c
hahuja2: Good idea, I added more tests: [97c639e](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/97c639e0e64aeb8e5e4a6d4db7c25de4aa928c0e). Instead of doing `NRINI(force_framework) = (nrframework_t)LIBRARY_NAME;`, I instead did `NRINI(force_framework) = true;` which allows us to test the forced code path without having to typecast `LIBRARY_NAME` since it doesn't exist in `nrframework_t`.

agent/fw_laravel.c
```diff
@@ -959,7 +961,9 @@ NR_PHP_WRAPPER(nr_laravel_application_construct) {
 
   if (NRINI(vulnerability_management_package_detection_enabled)) {
     // Add php package to transaction
-    nr_txn_add_php_package(NRPRG(txn), "laravel/framework", version);
+    nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
+    nr_fw_support_add_package_supportability_metric(
+        NRPRG(txn), PHP_PACKAGE_NAME, version);
```
zsistla, line 966: shouldn't we add the metric even if vm is disabled?

agent/fw_slim.c
```diff
@@ -120,7 +122,10 @@ NR_PHP_WRAPPER(nr_slim_application_construct) {
   version = nr_php_get_object_constant(this_var, "VERSION");
   
   // Add php package to transaction
-  nr_txn_add_php_package(NRPRG(txn), "slim/slim", version);
+  nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
```
zsistla, line 125: is this always added, even if vm is disabled?

agent/fw_wordpress.c
```diff
@@ -809,7 +810,9 @@ void nr_wordpress_version() {
   if (SUCCESS == result) {
     if (nr_php_is_zval_valid_string(&retval)) {
       char* version = Z_STRVAL(retval);
-      nr_txn_add_php_package(NRPRG(txn), "wordpress", version);
+      nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
```
zsistla, line 813: doesn't the add_php_package need to be in an `if vm` block?
Probably would have been easier to pass `vulnerability_management_package_detection_enabled` and have the logic to check for enabled/disabled in the `nr_txn_add_php_package` function to reduce code duplication of everyone checking for it.

agent/lib_guzzle6.c
```diff
@@ -352,8 +354,13 @@ NR_PHP_WRAPPER_START(nr_guzzle6_client_construct) {
 
   if (NRINI(vulnerability_management_package_detection_enabled)) {
     char* version = nr_php_get_object_constant(this_var, "VERSION");
+    if (NULL == version) {
+      version = nr_php_get_object_constant(this_var, "MAJOR_VERSION");
+    }
     // Add php package to transaction
-    nr_txn_add_php_package(NRPRG(txn), "guzzlehttp/guzzle", version);
+    nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
```
zsistla, line 361: doesn't only this function call need to be in the `if (NRINI(vulnerability_management_package_detection_enabled))` block?
We want version supportability metric whether vm is on or off.

agent/fw_laravel.c
hahuja2: Good point, fixed: [647d424](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/647d424ffb824368d14f803f6812df43a75be043)

agent/fw_slim.c
hahuja2: Fixed: [647d424](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/647d424ffb824368d14f803f6812df43a75be043)

agent/fw_wordpress.c
hahuja2: Fixed: [647d424](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/647d424ffb824368d14f803f6812df43a75be043). I decided to not pass `vulnerability_management_package_detection_enabled` in the `nr_txn_add_php_package` function and leave it like this for now because it requires `php_agent.h` which is in the agent directory and would require `#include "../agent/php_agent.h"`, which is leading to compilation errors.

agent/lib_guzzle6.c
hahuja2: Fixed: [647d424](https://github.com/newrelic/newrelic-php-agent/pull/868/commits/647d424ffb824368d14f803f6812df43a75be043)

agent/fw_wordpress.c
zsistla: Yes, it's a little tricky getting `agent` info to `axiom`, but in this case, isn't it just a bool that could be passed in?
Then the bool checked in axiom whether to add package or not?

zsistla: It's beyond the scope of this PR to change that, but think about if it could make sense to roll it into another PR.
At a minimum, the description [here](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_txn.h#L1164) should indicate it should only be called if vm is enabled.

zsistla (approved):

ZNeumann (approved):

agent/fw_support.c
```diff
@@ -76,8 +76,11 @@ void nr_fw_support_add_package_supportability_metric(
     }
   }
 
-  if ('\0' == major_version[0]) {
+  if ('\0' == major_version[0] && '\0' == package_version[0]) {
```
mfulb: I'm unclear on the logic here - it seems that if the supplied `package_version` is an empty string (or NULL) then we could short circuit out on line 60.

mfulb (approved):
Looks good!

PR #869, opened 2024-03-29
test(integration): Add additional custom error handler tests

zsistla:
Added additional tests to exercise a variety of return values from a custom error handler.


tests/integration/errors/test_error_handler_false.php
```diff
@@ -0,0 +1,96 @@
...
+}
+*/
+
+/*INI
+display_errors=1
+log_errors=0
+*/
+
+/*EXPECT_REGEX
+^\s*(PHP )?Deprecated:\s*Let this serve as a deprecation*
```
lavarou, line 26: Just an educational comment: with `INI` being `display_errors=1` and `log_errors=0` the `(PHP )?` part can be skipped from expected output.

tests/integration/errors/test_error_handler_implicit.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+    return false;
+}
+
+// set to the user defined error handler
+$old_error_handler = set_error_handler("errorHandlerOne");
+
+trigger_error("Let this serve as a deprecation", E_USER_DEPRECATED);  
+
+echo("Let this serve as a warning");
```
lavarou, line 46: I think this message has wrong connotations. This message is from a section of a program that didn't generate any errors, warnings or exception. This is from a normal execution of a program. Just seeing 'warning' in expected output, when actually nothing alarming happened raises an eyebrow.
```suggestion
echo("Hello from happy path!");
```

zsistla: Please see: [8d432b4](https://github.com/newrelic/newrelic-php-agent/pull/869/commits/8d432b4c8ef3e5066bb3badfb2d3a456b6cf8efc)

lavarou: Thank you! :)

tests/integration/errors/test_error_handler_with_fatal_error.php
```diff
@@ -0,0 +1,100 @@
...
+      {},
+      {}
+    ]
+  ]
+]
+*/
+
+function errorHandlerOne($errno, $errstr, $errfile, $errline)
+{
+  echo("Nothing to see here ever apparently.");
```
lavarou, line 84: I like this one! 👍

tests/integration/errors/test_error_handler_with_nonhandled_error_type.php
```diff
@@ -0,0 +1,123 @@
...
+//generate "Undefined property" error
+echo ($foo->propName);
+
+// generate "Attempt to read property" error
+$bar = false;
+echo ($bar->var);
+
+// generate "Undefined array key" error
+$missingOneArray = array(2=>'two', 4=>'four');
+echo $missingOneArray[1];
```
lavarou, line 118: So these lines will trigger errors that are handled by `errorHandlerOne`? Does it make sense to add expected output to this test expectation:
/*EXPECT_REGEX
Nothing to see here, suppressing undefined property
Nothing to see here, suppressing attempt to read property
Nothing to see here, suppressing undefined array key

Deprecated: Let this serve as a deprecation in.*
*/

tests/integration/errors/test_error_handler_with_strings.php7.php
lavarou, line 1: Isn't this case already covered with tests/integration/errors/test_error_handler_with_nonhandled_error_type.php?

lavarou (approved):
Thanks for adding these!

tests/integration/errors/test_error_handler_with_strings.php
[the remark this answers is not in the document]
zsistla: This actually relates to your other [question](https://github.com/newrelic/newrelic-php-agent/pull/869#discussion_r1544756696).

For that, I couldn't get regex to understand the handled "nothing to see here" msgs and the unhandled error message at the same time. 
So the tests are separate (strings handled and error nonhandled) for expect_regex purposes which cover the msgs individually, but combined here to verify it still works in tandem.

The more important thing is seeing the traced_errors and error_events in each test which show the correct behavior.

tests/integration/errors/test_error_handler_with_nonhandled_error_type.php
zsistla: See [this](https://github.com/newrelic/newrelic-php-agent/pull/869#discussion_r1544847790) comment.

tests/integration/errors/test_error_handler_with_strings.php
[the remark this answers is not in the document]
lavarou: I used this expect in the tests/integration/errors/test_error_handler_with_nonhandled_error_type.php test and it passed:
```
/*EXPECT_REGEX
Nothing to see here, suppressing undefined property
Nothing to see here, suppressing attempt to read property
Nothing to see here, suppressing undefined array key

Deprecated: Let this serve as a deprecation in.*
*/
```
Using that expect would reduce some code duplication and ensure same test coverage.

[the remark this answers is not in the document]
lavarou: But maybe you're right, maybe having them separately, and checking for traced errors separately is the right approach.

tests/integration/errors/test_error_handler_with_nonhandled_error_type.php
lavarou: I used this expect in the tests/integration/errors/test_error_handler_with_nonhandled_error_type.php test and it passed:

/*EXPECT_REGEX
Nothing to see here, suppressing undefined property
Nothing to see here, suppressing attempt to read property
Nothing to see here, suppressing undefined array key

Deprecated: Let this serve as a deprecation in.*
*/

tests/integration/errors/test_error_handler_with_strings.php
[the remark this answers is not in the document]
zsistla: Thanks for the snippet, I'm not sure what my issue with it was then :)

Please see: [a272070](https://github.com/newrelic/newrelic-php-agent/pull/869/commits/a272070fb05e861d03270670e3d9bb105275484a)
I added the regex but had to remove some tests that wouldn't work on PHP 7 (they aren't errors for PHP 7).

I still think it's a good to test separately then combined to also keep the tests more focused.

PR #871, opened 2024-04-03
fix: de-duplicate php package data

bduranleau-nr:
adds a filter for package data already sent during the current connection instance to prevent sending duplicate data between harvests.

### DIFF
--- a/daemon/internal/newrelic/app.go
+++ b/daemon/internal/newrelic/app.go
@@ -146,6 +146,7 @@ type App struct {
 	HarvestTrigger      HarvestTriggerFunc
 	LastActivity        time.Time
 	Rules               MetricRules
+	PhpPackages			map[PhpPackagesKey]struct{}
 }
 
 func (app *App) String() string {
@@ -180,6 +181,7 @@ func NewApp(info *AppInfo) *App {
 		info:               info,
 		HarvestTrigger:     nil,
 		LastActivity:       now,
+		PhpPackages: 		make(map[PhpPackagesKey]struct{}),
 	}
 }
 
@@ -337,4 +339,76 @@ func (app *App) Inactive(threshold time.Duration) bool {
 		panic(fmt.Errorf("invalid inactivity threshold: %v", threshold))
 	}
 	return time.Since(app.LastActivity) > threshold
+}
+
+// filter seen php packages data to avoid sending duplicates
+//
+// the `App` structure contains a map of PHP Packages the reporting
+// application has encountered.
+//
+// the map of packages should persist for the duration of the
+// current connection
+//
+// takes the `PhpPackages.data` byte array as input and unmarshals
+// into an anonymous interface array
+//
+// the JSON format received from the agent is:
+// 		[["package_name","version",{}],...]
+//
+// for each entry, assign the package name and version to the `PhpPackagesKey`
+// struct and use the key to verify data does not exist in the map. If the
+// key does not exist, add it to the map and the array of 'new' packages.
+//
+// convert the array of 'new' packages into a byte array representing
+// the expected data that should match input, minus the duplicates.
+func (app *App) filterPhpPackages(data []byte) []byte {
+	if data == nil {
+		return nil
+	}
+
+	var pkgKey PhpPackagesKey
+	var newPkgs []PhpPackagesKey
+	var x []interface{}
+
+	err := json.Unmarshal(data, &x)
+	if nil != err {
+		log.Errorf("failed to unmarshal php package json: %s", err)
+		return nil
+	}
+
+	for _, pkgJson := range x {
+		pkg, _ := pkgJson.([]interface{})
+		if len(pkg) != 3 {
+			log.Errorf("invalid php package json structure: %+v", pkg)
+			return nil
+		}
+		pkgKey = PhpPackagesKey{pkg[0].(string), pkg[1].(string)}
+		_, ok := app.PhpPackages[pkgKey]
+		if !ok {
+			app.PhpPackages[pkgKey] = struct{}{}
+			newPkgs = append(newPkgs, pkgKey)
+		}
+	}
+
+	if newPkgs == nil {
+		return nil
+	}
+
+	buf := &bytes.Buffer{}
+	buf.WriteString(`[`)
+	for _, pkg := range newPkgs {
+		buf.WriteString(`["`)
+		buf.WriteString(pkg.Name)
+		buf.WriteString(`","`)
+		buf.WriteString(pkg.Version)
+		buf.WriteString(`",{}],`)
+	}
+
+	resJson := buf.Bytes()
+
+	// swap last ',' character with ']'
+	resJson = resJson[:len(resJson) - 1]
+	resJson = append(resJson, ']')
+
+	return resJson
 }
--- a/daemon/internal/newrelic/app_test.go
+++ b/daemon/internal/newrelic/app_test.go
@@ -613,3 +613,41 @@ func TestMaxPayloadSizeInBytesFromConnectReply(t *testing.T) {
 		t.Errorf("parseConnectReply(something), got [%v], expected [%v]", c.MaxPayloadSizeInBytes, expectedMaxPayloadSizeInBytes)
 	}
 }
+
+func TestFilterPhpPackages(t *testing.T) {
+	app := &App{
+		PhpPackages: make(map[PhpPackagesKey]struct{}),
+	}
+	var nilData []byte = nil
+	emptyData := []byte(`[[{}]]`)
+	validData := []byte(`[["drupal","6.0",{}]]`)
+	moreValidData := []byte(`[["wordpress","7.0",{}],["symfony","5.1",{}]]`)
+	duplicateData := []byte(`[["drupal","6.0",{}]]`)
+
+	filteredData := app.filterPhpPackages(nilData)
+	if filteredData != nil {
+		t.Errorf("expected 'nil' result on 'nil' input, got [%v]", filteredData)
+	}
+
+	filteredData = app.filterPhpPackages(emptyData)
+	if filteredData != nil {
+		t.Errorf("expected 'nil' result on empty data input, got [%v]", filteredData)
+	}
+
+	expect := []byte(`[["drupal","6.0",{}]]`)
+	filteredData = app.filterPhpPackages(validData)
+	if string(filteredData) != string(expect) {
+		t.Errorf("expected [%v], got [%v]", string(expect), string(filteredData))
+	}
+
+	expect = []byte(`[["wordpress","7.0",{}],["symfony","5.1",{}]]`)
+	filteredData = app.filterPhpPackages(moreValidData)
+	if string(filteredData) != string(expect) {
+		t.Errorf("expected [%v], got [%v]", string(expect), string(filteredData))
+	}
+
+	filteredData = app.filterPhpPackages(duplicateData)
+	if filteredData != nil {
+		t.Errorf("expected 'nil', got [%v]", filteredData)
+	}
+}
--- a/daemon/internal/newrelic/php_packages.go
+++ b/daemon/internal/newrelic/php_packages.go
@@ -13,6 +13,11 @@ import (
 	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/log"
 )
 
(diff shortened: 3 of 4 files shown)


daemon/internal/newrelic/app.go
```diff
@@ -338,3 +340,75 @@ func (app *App) Inactive(threshold time.Duration) bool {
...
+		pkg, _ := pkgJson.([]interface{})
+		if len(pkg) != 3 {
+			log.Errorf("invalid php package json structure: %+v", pkg)
+			return nil
+		}
+		pkgKey = PhpPackagesKey{pkg[0].(string), pkg[1].(string)}
+		_, ok := app.PhpPackages[pkgKey]
+		if !ok {
+			app.PhpPackages[pkgKey] = struct{}{}
+			newPkgs = append(newPkgs, pkgKey)
```
zsistla, line 389: Is the newPkgs needed here?  Couldn't we just directly be writing to the `buf` variable instead?

bduranleau-nr: It's possible, but in my opinion not very clean. 
The buffer that outputs the result requires a very specific format that is easier to maintain if it's separated into it's own logic rather than injected into this for loop that's trying to parse out the data from the `byte` array.
Additionally, the logic to populate the buffer checks to verify that `newPkgs` is not `nil`, which is not possible using just the `buf` variable without some less than elegant logic.

zsistla: The tradeoff to consider is the additional overhead.

Couldn't you just check `buf` len to determine whether to return nil?

bduranleau-nr: In terms of performance, since this should only be happening once per harvest and only contain a handful of elements (the frameworks x versions we instrument) I don't anticipate this having any impact to overhead.
I prefer to keep the responsibility of things as atomic as possible. Currently this for loop is responsible for consuming the `byte` array and sorting out the "new" vs "old" packages. Constructing the resulting `byte` array is a separate task altogether, so I prefer to keep it decoupled from the parsing/sorting logic. This has the benefit of:
1. code readability- one block of code is responsible for one "job" rather than all jobs
2. ease of debugging- you can easily verify the result of the loop via checking the `newPkgs` variable, and the result of re-marshalling by checking the output of `buf`
3. ease of adding/changing functionality in the future- decoupling data parsing and data formatting isolates responsibilities makes modifying certain behavior or adding new behavior less invasive and more extensible in the long-term.

daemon/internal/newrelic/app.go
```diff
@@ -338,3 +340,75 @@ func (app *App) Inactive(threshold time.Duration) bool {
...
+//
+// for each entry, assign the package name and version to the `PhpPackagesKey`
+// struct and use the key to verify data does not exist in the map. If the
+// key does not exist, add it to the map and the array of 'new' packages.
+//
+// convert the array of 'new' packages into a byte array representing
+// the expected data that should match input, minus the duplicates.
+func (app *App) filterPhpPackages(data []byte) []byte {
+	if data == nil {
+		return nil
```
zsistla, line 366: When this returns nil, are we still harvesting the package data and sending update_loaded_modules with empty package info?

bduranleau-nr: Currently [this code](https://github.com/newrelic/newrelic-php-agent/blob/main/daemon/internal/newrelic/php_packages.go#L65) checks if `packages`, `packages.data` or `packages.numSeen` are `nil` or `0` prior to mashalling the data for the endpoint/collector and results in returning the empty payload `["Jars",[]]`. I've attempted to preserve this behavior for this PR.

zsistla: So majority of time, we'll be sending an empty payload?
Since it's not required to be sent as part of the harvest. if the payload is empty, should we still be sending it?
Otherwise, we'll still sending (most likely) an empty payload of the optional update_loaded_modules with every harvest that the backend will still have to process.

bduranleau-nr: If `newrelic.vulnerability_management.package_detection.enabled` is set to `true`, we should be sending payloads until we've managed to instrument all the supported libraries/packages, then the rest of the payloads will be empty. This is also on a per-process and per-connect basis, so it should reset at least once every day. We could update logic to not send anything if nothing is detected, but I think that would be best done in a separate PR since it modifies core packages code outside of this feature.

zsistla: Having it another PR is totally fine, but important to get done as part of this effort.  Is this sent on fast harvest or regular harvest?  Once a significant amount of customers are on this agent version, it would be sending an egregious amount of empty payloads every day for the backend to process and discard.

bduranleau-nr: Should only be sent on regular harvest cycle of 1 minute. Once this PR is merged I can put up another one adding the feature to skip reporting to this endpoint on `nil` `data`.

bduranleau-nr: On second pass through the code and this may actually not be the case- it seems like the `PhpPackages` struct implements the `PayloadCreator` interface which calls the `Empty` method (described above) when determining whether or not to send the harvest in `considerHarvestPayload`, which if `Empty` is true (aka `data` is `nil`), `considerHarvestPayload` simply returns before calling `harvestPayload`. The "empty" payload of `["Jars", []]` may only be for internal audit (logging) purposes.

zsistla: Thanks for addition tests to confirm behavior!!

[test: add test for nil package data for harvest](https://github.com/newrelic/newrelic-php-agent/pull/871/commits/6cc5b187f12e80c5c5d39da75398514e774de542)
[6cc5b18](https://github.com/newrelic/newrelic-php-agent/pull/871/commits/6cc5b187f12e80c5c5d39da75398514e774de542)

[test: add harvest test for valid package data](https://github.com/newrelic/newrelic-php-agent/pull/871/commits/8ecb0fdbb8e45ede21eb4419c1f29c47362114e8)
[8ecb0fd](https://github.com/newrelic/newrelic-php-agent/pull/871/commits/8ecb0fdbb8e45ede21eb4419c1f29c47362114e8)

zsistla (approved):

daemon/internal/newrelic/app.go
mfulb: It would probably be cleaner/more efficient if the agent sent package/version pairs instead of a JSON string.  But I agree that the number of packages per request is bounded by a small number and I suspect the vast majority of the time there will be no new packages so `newPkgs` will be `nil` most the time and the code below won't execute.  

Could be interesting to wrap the loop starting on line 379 with a timer to see how long it takes with different size JSON payloads just to put to rest any concern this could be time consuming with realistic payloads.

daemon/internal/newrelic/app.go
```diff
@@ -338,3 +340,75 @@ func (app *App) Inactive(threshold time.Duration) bool {
...
+		buf.WriteString(`["`)
+		buf.WriteString(pkg.Name)
+		buf.WriteString(`","`)
+		buf.WriteString(pkg.Version)
+		buf.WriteString(`",{}],`)
+	}
+
+	resJson := buf.Bytes()
+
+	// swap last ',' character with ']'
```
mfulb, line 409: What if instead of depending on the format of the JSON from the former block of code we change that block of code to do something like:
```
first = true
for _, pkg := range newPkgs {
  if !first {
     buf.WriteString(`,`)
  }
  <code to create JSON for this pkg>
  first = false
}
```

Just thinking if the code generates the desired format from the start it makes the rest of the code simpler.

daemon/internal/newrelic/app_test.go
```diff
@@ -613,3 +613,41 @@ func TestMaxPayloadSizeInBytesFromConnectReply(t *testing.T) {
...
+func TestFilterPhpPackages(t *testing.T) {
+	app := App{
+		PhpPackages: make(map[PhpPackagesKey]struct{}),
+	}
+	var nilData []byte = nil
+	emptyData := []byte(`[[{}]]`)
+	validData := []byte(`[["drupal","6.0",{}]]`)
+	moreValidData := []byte(`[["wordpress","7.0",{}],["symfony","5.1",{}]]`)
+	duplicateData := []byte(`[["drupal","6.0",{}]]`)
+
```
mfulb, line 626: I think adding a case like

```
similarData := []byte(`[["drupal","9.0",{}]]`)
```
would be good to test it can handle same package, different version.

daemon/internal/newrelic/app.go
```diff
@@ -146,6 +146,7 @@ type App struct {
 	HarvestTrigger      HarvestTriggerFunc
 	LastActivity        time.Time
 	Rules               MetricRules
+	PhpPackages			map[PhpPackagesKey]struct{}
```
ZNeumann, line 149: I think gofmt needs to be run

daemon/internal/newrelic/app.go
mfulb: The more I think about it I like the idea of making `PhpPackagesKey` a `struct` in the `protocol.fbs` and just passing it over from the agent that way which can be trivially decoded.

daemon/internal/newrelic/app.go
bduranleau-nr: This is gofmted. The rest of the struct uses spaces instead of tabs. My change uses tabs (which is [correct](https://go.dev/doc/effective_go#formatting) according to Go).

daemon/internal/newrelic/app.go
bduranleau-nr: This is outside the scope of this PR.

daemon/internal/newrelic/app.go
bduranleau-nr: I prefer not to introduce extra logic unnecessarily. This works, and for the amount of data we expect here, is trivial in terms of speed and space. This is something that should probably be addressed at the agent level to not send JSON at all in another PR.

ZNeumann (approved):

daemon/internal/newrelic/app.go
lavarou: I'm sorry but this is not true. Tabs are used for indent, not structures formatting. Moreover, `gofmt` shows that these lines (among a few others) need to be reformatted:
```
gofmt -d internal/newrelic/app.go 
diff internal/newrelic/app.go.orig internal/newrelic/app.go
--- internal/newrelic/app.go.orig
+++ internal/newrelic/app.go
@@ -146,7 +146,7 @@
        HarvestTrigger      HarvestTriggerFunc
        LastActivity        time.Time
        Rules               MetricRules
-       PhpPackages                     map[PhpPackagesKey]struct{}
+       PhpPackages         map[PhpPackagesKey]struct{}
 }
 
 func (app *App) String() string {
@@ -181,7 +181,7 @@
                info:               info,
                HarvestTrigger:     nil,
                LastActivity:       now,
-               PhpPackages:            make(map[PhpPackagesKey]struct{}),
+               PhpPackages:        make(map[PhpPackagesKey]struct{}),
        }
 }
```

bduranleau-nr: 24fbebf465ba0ffd6769482348db43b12d01d1b9

daemon/internal/newrelic/app_test.go
bduranleau-nr: 24f21ffae9b741538acf0e4317779b373ac5a9b6

daemon/internal/newrelic/app.go
```diff
@@ -338,3 +340,75 @@ func (app *App) Inactive(threshold time.Duration) bool {
...
+		return nil
+	}
+
+	for _, pkgJson := range x {
+		pkg, _ := pkgJson.([]interface{})
+		if len(pkg) != 3 {
+			log.Errorf("invalid php package json structure: %+v", pkg)
+			return nil
+		}
+		pkgKey = PhpPackagesKey{pkg[0].(string), pkg[1].(string)}
```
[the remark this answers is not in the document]
bduranleau-nr: It will return `nil`. See this added test case 749f2980e84b8755796760241afb68f74bbb681e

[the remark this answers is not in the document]
bduranleau-nr: added type checking logic just in case 375dc80d93e67269c77225bd086248d54cc9c2bf

PR #876, opened 2024-04-11
fix(agent): Fix error reporting when EH_THROW is enabled

zsistla:
This PR 
1)checks if EH_THROW(which signals to PHP to throw an exception from the error) is toggled and if so does not record an error because if uncaught it will be handled by the exception handler and if caught there's no need to record it.
2) add tests to verify fix

More details:  
/*
   * For a the following error codes:
   * E_WARNING || E_CORE_WARNING || E_COMPILE_WARNING || E_USER_WARNING
   * PHP triggers an exception if EH_THROW is toggled on and then immediately
   * returns after throwing the exception. See for more info:
   * https://github.com/php/php-src/blob/master/main/main.c In that case, we
   * should not handle it, but we should exist and let the exception handler
   * deal with it; otherwise, we could record an error even if an exception is
   * caught. */

In the case where a php E_WARNING error was triggered.  We intercept the error from PHP and record it then pass execution back to PHP.

But, for instance, in the particular case that deals with a filehandling which toggles EH_THROW on here: https://github.com/php/php-src/blob/master/ext/fileinfo/fileinfo.c#L176
PHP then handles the error but for a certain number of error codes, PHP triggers an exception if EH_THROW is toggled on: https://github.com/php/php-src/blob/master/main/main.c#L1254C1-L1259C24
and then immediately returns after throwing the exception.

This leads to the case in the issue where the exception that was triggered is caught, but we recorded the error....


agent/php_error.c
```diff
@@ -561,12 +561,29 @@ static int nr_php_should_record_error(int type, const char* format TSRMLS_DC) {
...
+    }
+  }
+
   errprio = nr_php_error_get_priority(type);
 
   if (0 == errprio) {
     return 0;
   }
-
+
```
ZNeumann, line 586:
```suggestion

```

ZNeumann (approved):

agent/php_error.c
```diff
@@ -561,12 +561,28 @@ static int nr_php_should_record_error(int type, const char* format TSRMLS_DC) {
...
+   * PHP triggers an exception if EH_THROW is toggled on and then immediately
+   * returns after throwing the exception. See for more info:
+   * https://github.com/php/php-src/blob/master/main/main.c In that case, we
+   * should not handle it, but we should exist and let the exception handler
+   * deal with it; otherwise, we could record an error even if an exception is
+   * caught.
+   */
+  if (EG(error_handling) == EH_THROW) {
+    if ((E_WARNING == type) || (E_CORE_WARNING == type)
+        || (E_COMPILE_WARNING == type) || (E_USER_WARNING == type)) {
```
lavarou, line 576: The list of error types for which the exception is thrown differs between PHPs - PHP 7.4: https://github.com/php/php-src/blob/PHP-7.4.17/main/main.c#L1229-L1258, PHP 8.0: https://github.com/php/php-src/blob/PHP-8.0.13/main/main.c#L1231-L1251. Shouldn't this list be conditionally compiled depending on PHP version?

tests/integration/errors/test_EH_THROW_errors_caught_exception.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+For a certain number of error codes, PHP triggers an exception if EH_THROW is toggled on.
```
lavarou, line 8:
```suggestion
For a certain number of error types, PHP triggers an exception if EH_THROW is toggled on.
```

tests/integration/errors/test_EH_THROW_errors_uncaught_exception.php
```diff
@@ -0,0 +1,119 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+For a certain number of error codes, PHP triggers an exception if EH_THROW is toggled on.
+Verify we don't record the error if the exception is thrown and NOT caught.
```
lavarou, line 9:
```suggestion
For a certain number of error types, PHP triggers an exception if EH_THROW is toggled on.
Verify we record the error if the exception is thrown and NOT caught.
```

agent/php_error.c
zsistla: Please see: [ed0456a](https://github.com/newrelic/newrelic-php-agent/pull/876/commits/ed0456a7013e7b1c4735a219fd8295e168c5941a)

tests/integration/errors/test_EH_THROW_errors_caught_exception.php
```diff
@@ -0,0 +1,76 @@
...
+For a certain number of error types, PHP triggers an exception if EH_THROW is toggled on.
+Verify we don't record the error if the exception is thrown and caught.
+There should be no traced errors, error events, or errors attached to span events.
+*/
+
+/*SKIPIF
+<?php
+if (version_compare(PHP_VERSION, "7.0", "<")) {
+  die("skip: PHP < 7.0 not supported\n");
+}
```
lavarou, line 17: Is this skipif needed? PHPs < 7.0 are no longer supported and tests are not run on those PHPs.

zsistla: please see: [cc91192](https://github.com/newrelic/newrelic-php-agent/pull/876/commits/cc91192420f66b05f72c7891d084da0f169867ee)

agent/php_error.c
```diff
@@ -561,12 +561,39 @@ static int nr_php_should_record_error(int type, const char* format TSRMLS_DC) {
...
+#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO
+/*
+ * Cover one more error condition that was falling through to default
+ * https://github.com/php/php-src/blob/PHP-7.4.17/main/main.c#L1248
+ */
+    if (E_RECOVERABLE_ERROR == type) {
+      return 0;
+    }
+#endif
+  }
```
lavarou, line 590: This will work, but this code's performance can be improved upon. Because of how Zend defines error types, they can be used as bits in a bit mask - see [here](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_errors.h#L48). Therefore I suggest taking advantage of that, and make the error type comparison once not 4 or 5 times. Defining these two macros:
```
#if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO
#define E_ERRORS_THROWING_EXCEPTIONS (E_WARNING | E_CORE_WARNING | E_COMPILE_WARNING | E_USER_WARNING | E_RECOVERABLE_ERROR)
#else
#define E_ERRORS_THROWING_EXCEPTIONS (E_WARNING | E_CORE_WARNING | E_COMPILE_WARNING | E_USER_WARNING)
#endif

#define E_THROWS_EXCEPTION(e) ((E_ERRORS_THROWING_EXCEPTIONS & e) == e)
```
simplifies the code to:
```suggestion
  if (EG(error_handling) == EH_THROW && E_THROWS_EXCEPTION(type)) {
    // let the exception handler deal with this error
    return 0;
  }
```
which, not only is more performant, but also more readable.

lavarou (approved):
Nice find!

PR #877, opened 2024-04-12
fix(agent): improve exception handler instrumentation for PHPs 8.0+

lavarou:
When user exception handler is installed it can call restore_exception_handler,
and that will reset is_exception_handler flag in the wraprec which will cause
the agent not to use the correct code path after user exception handler returns.
The information that `nr_php_instrument_func_end` is handling return from user
exception handler needs to be stored in the segment associated with the user
exception handler.


agent/php_execute.c
```diff
@@ -2027,7 +2035,7 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
 
   wraprec = segment->wraprec;
 
-  if (wraprec && wraprec->is_exception_handler) {
+  if (wraprec && segment->is_exception_handler) {
```
zsistla, line 2038: does it need the wraprec check in the `if` statement anymore?

lavarou: It probably doesn't. `segment->is_exception_handler` can only be set if wraprec for user exception handler has been found in `nr_php_instrument_func_begin`.

lavarou: Simplified the check in 5c7f58da1820dad00dfd8c0125f08467328fb8d7.

tests/integration/errors/test_oapi_uncaught_handled_exception_04.php
```diff
@@ -0,0 +1,180 @@
...
+  throw new RuntimeException('Expected unexpected happened');
+}
+
+function call_throw_it() {
+  throw_it();
+}
+
+set_exception_handler('user_exception_handler_01');
+set_exception_handler('user_exception_handler_02');
+//restore_exception_handler();
```
ZNeumann, line 178:
```suggestion
```

tests/integration/errors/test_oapi_uncaught_handled_exception_04.php
```diff
@@ -0,0 +1,180 @@
...
+]
+*/
+
+
+/*EXPECT_REGEX
+Fatal error: Uncaught RuntimeException: Not able to handle, throwing another exception from handler
+*/
+
+function user_exception_handler_01(Throwable $ex) {
+  //restore_exception_handler();
```
ZNeumann, line 158:
```suggestion
```

zsistla (approved):

zsistla:
Do you have an idea what is causing the 8.3 failure?

```
FAIL - tests/integration/errors/test_oapi_uncaught_handled_exception_04.php
regex error:
expected:
Fatal error: Uncaught RuntimeException: Not able to handle, throwing another exception from handler

actual:
01 Handled uncaught exception
```

lavarou:
> Do you have an idea what is causing the 8.3 failure?
> 
> ```
> FAIL - tests/integration/errors/test_oapi_uncaught_handled_exception_04.php
> regex error:
> expected:
> Fatal error: Uncaught RuntimeException: Not able to handle, throwing another exception from handler
> 
> actual:
> 01 Handled uncaught exception
> ```

@zsistla Good eye! This means that PHP 8.3 started to behave like I would expect all PHPs to behave, that is honor the exception handler's stack. That was the purpose of this test! I remember reading about it somewhere... I need to split this test into two tests: for PHPs < 8.3 and for PHPs >= 8.3.

lavarou:
> > Do you have an idea what is causing the 8.3 failure?
> > ```
> > FAIL - tests/integration/errors/test_oapi_uncaught_handled_exception_04.php
> > regex error:
> > expected:
> > Fatal error: Uncaught RuntimeException: Not able to handle, throwing another exception from handler
> > 
> > actual:
> > 01 Handled uncaught exception
> > ```
> 
> @zsistla Good eye! This means that PHP 8.3 started to behave like I would expect all PHPs to behave, that is honor the exception handler's stack. That was the purpose of this test! I remember reading about it somewhere... I need to split this test into two tests: for PHPs < 8.3 and for PHPs >= 8.3.

@zsistla This is really broken in PHP. Even not all versions of PHP 8.3 behave the same: https://3v4l.org/ARJ9A

lavarou:
> > > Do you have an idea what is causing the 8.3 failure?
> > > ```
> > > FAIL - tests/integration/errors/test_oapi_uncaught_handled_exception_04.php
> > > regex error:
> > > expected:
> > > Fatal error: Uncaught RuntimeException: Not able to handle, throwing another exception from handler
> > > 
> > > actual:
> > > 01 Handled uncaught exception
> > > ```
> > 
> > 
> > @zsistla Good eye! This means that PHP 8.3 started to behave like I would expect all PHPs to behave, that is honor the exception handler's stack. That was the purpose of this test! I remember reading about it somewhere... I need to split this test into two tests: for PHPs < 8.3 and for PHPs >= 8.3.
> 
> @zsistla This is really broken in PHP. Even not all versions of PHP 8.3 behave the same: https://3v4l.org/ARJ9A

Tried to address this with 4646c9c8c3150da1c5383a9177cf8212dc0ef14b.

tests/integration/errors/test_oapi_uncaught_handled_exception_04.php835.php
```diff
@@ -0,0 +1,176 @@
...
+newrelic.span_events_enabled=1
+newrelic.code_level_metrics.enabled = 0
+display_errors=1
+log_errors=0
+*/
+
+
+/*SKIPIF
+<?php
+if (version_compare(PHP_VERSION, "8.3", "<")) {
```
zsistla, line 26: test is named php835, but this just checks for 8.3?

zsistla:
Not a blocker, but why do the text cases need `_oapi_` in the name?  It doesn't seem to meaningfully add information about the test case.

lavarou:
> Not a blocker, but why do the text cases need `_oapi_` in the name? It doesn't seem to meaningfully add information about the test case.

I included `_oapi_` in the name because they only exercise `_oapi_` codepaths. I'm open to suggestions how to better convey that information without including `_oapi_` in the name.

tests/integration/errors/test_oapi_uncaught_handled_exception_04.php835.php
zsistla: The 835 naming convention was perfectly fine, but the 835 test only had this check:

```
<?php
if (version_compare(PHP_VERSION, "8.3", "<")) {
  die("skip: PHP < 8.3.0 not supported\n");
}
```

shouldn't it be 

```<?php
if (version_compare(PHP_VERSION, "8.3.5", "<")) {
  die("skip: PHP < 8.3.5 not supported\n");
}
```

zsistla: I assumed 835 meant php 8.3.5 and above.

zsistla: Maybe instead of in the name, update the description inside the test explaining why the skipif and which versions we expect it to run on.
Then you can have an 83a, 83b, etc to show different variants of the test for 8.3?

lavarou: Thanks for the suggestions. See 501325667d95cc2a820141ae8ffbd0168fb51bbe.

lavarou:
> > > Not a blocker, but why do the text cases need `_oapi_` in the name? It doesn't seem to meaningfully add information about the test case.
> > 
> > I included `_oapi_` in the name because they only exercise `_oapi_` codepaths. I'm open to suggestions how to better convey that information without including `_oapi_` in the name.
>
> For 8+ there are only oapi code paths, we didn't include oapi on any other 8+ only tests or non-oapi on any pre8 tests.  just seemed redundant.

Removed `_oapi` part from test filename in e8669dd8b092309e20fe13a1aa8fe521fed89637.

zsistla:
LGTM!

PR #878, opened 2024-04-15
fix(agent): fix mongodb instrumentation for PHP 8.0+

bduranleau-nr:
- split mongodb instrumentation into before/after inline with OAPI paradigm.
- remove broken integration tests in favor of Multiverse tests


agent/lib_mongodb.c
```diff
@@ -173,7 +175,205 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
 }
 NR_PHP_WRAPPER_END
 
-void nr_mongodb_enable(TSRMLS_D) {
+#else
+
+NR_PHP_WRAPPER(nr_mongodb_operation_before) {
+    (void)wraprec;
+    nr_segment_start(NRPRG(txn), NULL, NULL);
+}
```
zsistla, line 183: i think this needs to have an autosegment like here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/lib_predis.c#L762

bduranleau-nr: It isn't clear to me from any of our code or docs what the purpose of `auto_segment` is- I'd like to understand this better before I implement around it. Do you have any resources on what `auto_segment` is intended for?

bduranleau-nr: see 26cdcf283b07ec9283a68c3c7ee469f027a66c09

lavarou: Looks good.

agent/lib_mongodb.c
```diff
@@ -173,7 +175,205 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
+
+void nr_mongodb_enable() {
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
+    && !defined OVERWRITE_ZEND_EXECUTE_DATA
+
+    nr_php_wrap_user_function_before_after_clean_extra(
+        NR_PSTR("MongoDB\\Operation\\Aggregate::execute"),
+        nr_mongodb_operation_before,
+        nr_mongodb_operation_after,
+        NULL,
```
ZNeumann, line 259: Because "before" is always starting a segment, we will have dangling segments if "after" is never called (for example when an exception occurred). The remedies for this are either 1) create a "clean" function that closes the segment or 2) add the "after" function as the "clean" argument. This means that the "after" function needs to verify the return value and any arguments it uses.

I recommend 2 because that needs to be done anyway for https://github.com/newrelic/newrelic-php-agent/pull/866

bduranleau-nr: fixed in 0b911e485feb13860c0eb6db8bfd458405272b8d

agent/lib_mongodb.c
```diff
@@ -173,7 +175,205 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
+    instance.database_name = Z_STRVAL_P(database);
+  }
+
+  server = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
+  nr_mongodb_get_host_and_port_path_or_id(server, &instance.host,
+                                          &instance.port_path_or_id);
+
+  nr_segment_datastore_end(&auto_segment, &params);
+
+leave:
```
ZNeumann, line 240: If we `goto leave`, we end up with a dangling segment. This shouldn't be too bad, as the segment storage is within a transaction, so when the transaction ends the segment is cleaned up. Do we want to end the segment with limited information if we exit early?

I am dealing with this exact question with our drupal instrumentation right now

bduranleau-nr: addressed here 57364ba6937aef21bacb8c2e885eb670b32b796c

agent/php_wrapper.c
```diff
@@ -73,6 +73,26 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean(
...
+  nruserfn_t* wraprec = nr_php_add_custom_tracer_named(name, namelen);
+
+  wraprec->extra = extra;
+
+  nr_php_wraprec_add_before_after_clean_callbacks(name, namelen, wraprec,
+                                                  before_callback,
+                                                  after_callback,
+                                                  clean_callback);
+
+  return wraprec;
```
lavarou, line 93: This avoids duplicating logic of `nr_php_wrap_user_function_before_after_clean` and ensures `wraprec->extra` is set only if `wraprec` is not `NULL`:
```suggestion
  nruserfn_t* wraprec = nr_php_wrap_user_function_before_after_clean(
      name, namelen, before_callback, after_callback, clean_callback);
  if (nrunlikely(NULL == wraprec)) {
    nrl_warning(NRL_INSTRUMENT, "%s: unable to wrap '%s'", __func__,
                NRSAFESTR(name));
    return wraprec;
  }

  wraprec->extra = extra;
  return wraprec;
```

agent/lib_mongodb.c
```diff
@@ -173,7 +175,162 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
 }
 NR_PHP_WRAPPER_END
 
-void nr_mongodb_enable(TSRMLS_D) {
+#else
+
+NR_PHP_WRAPPER(nr_mongodb_operation_before) {
+  (void)wraprec;
+  nr_segment_t* segment;
+  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
```
lavarou, line 183: If [this check](https://github.com/newrelic/newrelic-php-agent/blob/64c6306ea27dc9f8474daae940f8fd63054ab2c6/agent/lib_mongodb.c#L211-L222) would be performed in `nr_mongodb_operation_before` and would fail, `nr_mongodb_operation_before` wouldn't create a new segment that is discarded in `nr_mongodb_operation_after`. I don't know which approach results in smaller agent's overhead. Has this been evaluated and therefore [this check](https://github.com/newrelic/newrelic-php-agent/blob/64c6306ea27dc9f8474daae940f8fd63054ab2c6/agent/lib_mongodb.c#L211-L222) is only done in `nr_mongodb_operation_after`?

lavarou (commented):
I have a few questions:
* https://github.com/newrelic/newrelic-php-agent/pull/878/files#r1600397303
* https://github.com/newrelic/newrelic-php-agent/pull/878/files#r1600405720
* https://github.com/newrelic/newrelic-php-agent/pull/878/files#r1600265944

And one suggestion:
* https://github.com/newrelic/newrelic-php-agent/pull/878/files#r1600264589

agent/php_wrapper.c
bduranleau-nr: f3164c6d31bcd5a4ea17d20d6c4d43983ed3018d

agent/php_wrapper.c
```diff
@@ -73,6 +73,26 @@ nruserfn_t* nr_php_wrap_user_function_before_after_clean(
   return wraprec;
 }
 
+nruserfn_t* nr_php_wrap_user_function_before_after_clean_extra(
```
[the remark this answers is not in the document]
bduranleau-nr: I'm sure it could be, but given that we're essentially just wrapping existing functionality of `nr_php_wrap_user_function_before_after_clean` I wouldn't consider it actually testing much and probably not worth the effort. If you look at the existing tests for `nr_php_wrap_user_function_before_after_clean` in `test_php_wrapper.c` you can see how involved tests involving this family of functions are.

agent/lib_mongodb.c
bduranleau-nr: These blocks of code are largely recycled from the non-OAPI implementation and modified to suit the OAPI paradigm.
- https://github.com/newrelic/newrelic-php-agent/blob/64c6306ea27dc9f8474daae940f8fd63054ab2c6/agent/lib_mongodb.c#L113
> Has this been evaluated and therefore [this check](https://github.com/newrelic/newrelic-php-agent/blob/64c6306ea27dc9f8474daae940f8fd63054ab2c6/agent/lib_mongodb.c#L211-L222) is only done in nr_mongodb_operation_after?

I haven't evaluated performance implications of this. I doubt we have the capability to do so to the granular degree that would be required here with confidence. Are you suggesting to duplicate the check and have it run not only in before but after as well? We would also have to handle not initializing the segment in both before in a way that after could be aware of it, then modify its behavior accordingly.

I don't believe that this case gets hit very often. There's no historical explanation or conversation over why it exists that I can find, and it may be that it should just be removed. All the [functions](https://github.com/mongodb/mongo-php-library/tree/master/src/Operation) we wrap should satisfy the `MongoDB\\Operation\\Executable` interface check, so I'm not sure if this actually ever gets triggered or if it was a relic of the past.

agent/lib_mongodb.c
```diff
@@ -173,7 +175,162 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
+  nr_datastore_instance_t instance = {
+      .host = NULL,
+      .port_path_or_id = NULL,
+      .database_name = NULL,
+  };
+  nr_segment_datastore_params_t params = {
+    .datastore = {
+      .type = NR_DATASTORE_MONGODB,
+    },
+    .operation = nr_strdup (wraprec->extra),
```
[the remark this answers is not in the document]
bduranleau-nr: `nr_strdup` is a straightforward way to circumvent the compiler error that will result from trying to cast a `const char*` to `char*` and discard the `const` qualifier.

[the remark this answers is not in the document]
lavarou: Straightforward but inefficient - it adds memory allocation which is not needed.

agent/php_wrapper.c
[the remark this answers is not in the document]
lavarou: After f3164c6d31bcd5a4ea17d20d6c4d43983ed3018d I'm ok with not adding unit tests for this function.

agent/lib_mongodb.c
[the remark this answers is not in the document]
bduranleau-nr: Do you have a suggestion that avoids the compilation error?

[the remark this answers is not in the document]
lavarou: ~~Stealing~~Borrowing idea from [here](https://github.com/newrelic/newrelic-php-agent/blob/v10.21.0.11/agent/php_agent.c#L1042-L1043):
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wcast-qual"
  nr_segment_datastore_params_t params = {
    .datastore = {
      .type = NR_DATASTORE_MONGODB,
    },
    .operation = (char *)wraprec->extra,
    .instance = &instance,
    .callbacks = {
      .backtrace = nr_php_backtrace_callback,
    },
  };
#pragma GCC diagnostic pop
```
Of course this also requires removing the code freeing `params.operation` here: https://github.com/newrelic/newrelic-php-agent/blob/6f0d244e78486ed1b318febcdcbba58510331d28/agent/lib_mongodb.c#L248

agent/lib_mongodb.c
lavarou: I guess the likelihood of the test failing is small. I'm ok with leaving the code as is.

agent/lib_mongodb.c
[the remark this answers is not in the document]
bduranleau-nr: a163988219b373e48acdda5aa6aaa67b7571d886

lavarou (dismissed):

agent/lib_mongodb.c
```diff
@@ -173,7 +175,166 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
 NR_PHP_WRAPPER_END
 
-void nr_mongodb_enable(TSRMLS_D) {
+#else
+
+NR_PHP_WRAPPER(nr_mongodb_operation_before) {
+  (void)wraprec;
+  nr_segment_t* segment;
+  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  segment->wraprec = auto_segment->wraprec;
```
zsistla, line 184: `nr_segment_start` can [return a null](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_segment.c#L166) so we should check for null before trying to access the wraprec.

agent/lib_mongodb.c
```diff
@@ -173,7 +175,166 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
 }
 NR_PHP_WRAPPER_END
 
-void nr_mongodb_enable(TSRMLS_D) {
+#else
+
+NR_PHP_WRAPPER(nr_mongodb_operation_before) {
+  (void)wraprec;
+  nr_segment_t* segment;
```
zsistla, line 182:
```suggestion
  nr_segment_t* segment = NULL;
```

bduranleau-nr: 59e8ec2a240efc89b44251c7681e6081c562850c

agent/lib_mongodb.c
```diff
@@ -173,7 +175,166 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
+      .host = NULL,
+      .port_path_or_id = NULL,
+      .database_name = NULL,
+  };
+
+  // tell the compiler to ignore the cast from const char * to char *
+  // to save having to do a strdup operation
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-qual"
+  nr_segment_datastore_params_t params = {
```
zsistla, line 207: Probably the other [nr_segment_datastore_params_t](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_segment_datastore.h#L17) values (like collection and database_name) should be explicitly initialized to NULL here.

agent/lib_mongodb.c
bduranleau-nr: added here 59e8ec2a240efc89b44251c7681e6081c562850c but after looking into this I can't find any instance where we bother checking this value for `NULL` in any other framework instrumentation. Drupal, Predis, even the non-OAPI mongodb implementation all assume this value won't be `NULL`.

agent/lib_mongodb.c
bduranleau-nr: `database_name` is already set to `NULL` in the `instance` variable declaration above.
`collection` set to `NULL` here f2759015630151982029c4db88d6d71c3229e311

agent/lib_mongodb.c
zsistla: Seems like a gap introduced when we needed to set wraprec. Not checking could cause issues if a txn is ended midway.

Looking at some examples...

in old mongodb instrumentation, it didn't need a check because it didn't access segment and only passed it to segment_end which does the null check: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/lib_mongodb.c#L163

drupal:
trying to access wraprec, should have a check before trying: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_drupal.c#L261

here, it doesn't need a check because we aren't trying to access segment elements like wraprec, and  the segment is only used when we call [segment_end](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_segment.c#L708) which does check for null.

same thing in https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_internal_instrument.c#L527
If we don't access it and simply pass the segment to nr_segment_end or nr_php_pdo_end_segment_sql or nr_segment_datastore_end, all those check for null, so we don't need to check.

but if we want to access it, a null check is needed like here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_internal_instrument.c#L1311

zsistla (approved):
Great job getting this working.

lavarou (approved):

Issue #875, opened 2024-04-10
MySQLi statement execute ArgumentCountError ("number of parameters")


iaibai:
Since upgrading to PHP 8.2, we've been getting frequent "ArgumentCountError: The number of variables must match the number of parameters in the prepared statement" errors when executing mysqli prepared statements. Although there are some really strange circumstances around this error, we can recreate it.

We _are_ binding the correct number of variables using bind_param, but for some reason as you'll see from the stack trace the Error is thrown by execute().

This is preventing us upgrading to PHP 8.2.

## Description

This Error is thrown regularly on certain statements (about a dozen) in our codebase, and as you'll see from the recreation steps there are quite a few factors that seem to affect whether a statement will error.

A full stack trace of the error looks like the following:

```
Fatal error: Uncaught ArgumentCountError: The number of variables must match the number of parameters in the prepared statement in /test.php:29
Stack trace:
#0 [internal function]: mysqli_stmt->bind_param('s', '2017-04-10 00:0...')
#1 /test.php(29): mysqli_stmt->execute()
#2 {main}
  thrown in /test.php on line 29
```

We first experienced this on Amazon Linux 2023 PHP 8.2 (installed newrelic using the tar archive), but we have recreated this on Amazon Linux 2 PHP 8.1 (installing through rpm), and locally in a docker instance based on php:8.1-cli (again using tar).

We have recreated the Error on 10.16.0.5, 10.19.0.9, and 10.10.0.1.

The errors stop when we set `newrelic.enabled = false`, or when we remove our license key from newrelic.ini. And of course, the error does not happen when we haven't installed newrelic PHP agent.

## Steps to Reproduce

In summary, install PHP >=8.1, with mysqli, enable New Relic and then prepare two statements that have different number of arguments.

Here is our current test script:

```
<?php 
 
$dbConn = new mysqli( 
    "db", 
    "root", 
    "changeme", 
    "dbtest" 
); 
 
justPrepare($dbConn); 
 
$ts = strtotime("today - 7 years"); 
$stringTs = date('Y-m-d H:i:s', $ts); 
 
$stmt = $dbConn->prepare(" SELECT q.id 
FROM query_test_table q 
CROSS JOIN query_test_table AS q2 
WHERE MD5(`q`.`key`) = MD5(?) 
; 
"); 
if (! $stmt) { 
    throw new Exception("Prepare Failed : " . $dbConn->error); 
} 
$stmt->bind_param('s', $stringTs); 
[...]

PR #881, opened 2024-04-24
fix(agent): Fixes warning from mysqli when explaining slow SQL queries (GH issue 875)

mfulb:
In GH issue #875 it was reported a warning is generated when using prepared statements with `mysqli`.  The issue included excellent instructions on how to reproduce the issue.  This PR addresses the cause of the warning.  

To understand the cause of the warning requires understanding how the agent handles `mysqli::stmt` objects.  When a SQL statement is prepared via mysqli the agent will store the SQL string in a global hashmap called `mysqli_queries` *if* the SQL is considered explainable (determined via `nr_php_explain_mysql_query_is_explainable()`), otherwise it is just ignored.  Later when a slow SQL query is detected the SQL string is retrieved and used to explain the query.

An additional factor is when a `mysqli::stmt` object is released it goes into a free pool PHP keeps which allows it to quickly hand out an allocated section of memory when a new object is created.  Each object has a `object ID` (referenced as a `handle` in the agent for the relevant code).  When an the allocated object is reused from the free pool it retains the same `object ID`.

The agent used the `object ID` (`handle`) as the key for storing the prepared SQL string for explainable SQL strings.  Now what if an explainable query comes through, the string is stored, and then that object is released.  A new query is prepared with a string that is NOT explainable and gets the object container with the same `object ID` as the explainable one just released.  The string stored in `mysqli_queries` using that `object ID` as the key is now stale.  If the new, unexplainable query is slow then the agent will pull this stale string and try to run an explain with it - leading to the warning.

The fix is to clear any query string for a given `object ID` if the query string is unexplainable.

Integration tests were added (based on the reproduction case) that test for a regression of this fix.

### DIFF
--- a/agent/php_mysqli.c
+++ b/agent/php_mysqli.c
@@ -353,6 +353,31 @@ nr_status_t nr_php_mysqli_query_set_query(nr_php_object_handle_t handle,
   return NR_SUCCESS;
 }
 
+nr_status_t nr_php_mysqli_query_clear_query(nr_php_object_handle_t handle) {
+  zval* metadata = NULL;
+
+  /* If a metadata entry exists then clear the "query" tag from it.
+   *  If an entry does not exist then nothing needs to be done.
+   */
+  metadata = nr_php_mysqli_query_find(handle);
+  if (NULL == metadata) {
+    return NR_FAILURE;
+  }
+
+  /* Clear the "query" element */
+  nr_php_zend_hash_del(Z_ARRVAL_P(metadata), "query");
+
+  /*
+   * Since the query is cleared so must the bind parameters, so let's get rid of
+   * whatever's here. We'll ignore the return values, since if the keys don't
+   * already exist no harm is done.
+   */
+  nr_php_zend_hash_del(Z_ARRVAL_P(metadata), "bind_args");
+  nr_php_zend_hash_del(Z_ARRVAL_P(metadata), "bind_format");
+
+  return NR_SUCCESS;
+}
+
 int nr_php_mysqli_zval_is_link(const zval* zv TSRMLS_DC) {
   if (NULL == zv) {
     return 0;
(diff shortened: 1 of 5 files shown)


agent/php_mysqli.c
```diff
@@ -353,6 +353,31 @@ nr_status_t nr_php_mysqli_query_set_query(nr_php_object_handle_t handle,
...
 
+nr_status_t nr_php_mysqli_query_clear_query(nr_php_object_handle_t handle) {
+  zval* metadata = NULL;
+
+  /* If a metadata entry exists then clear the "query" tag from it.
+   *  If an entry does not exist then nothing needs to be done.
+   */
+  metadata = nr_php_mysqli_query_find(handle);
+  if (NULL == metadata) {
+    return NR_FAILURE;
```
lavarou, line 364: Is this an error condition (the comment above suggests it is not)? Does `nr_php_mysqli_query_clear_query` need to return any value? It's return value is not checked in `nr_php_prepared_prepare_general` where it's used.

lavarou (approved):
Thanks! +1 for great tests' descriptions!

agent/php_mysqli.c
mfulb: A good question - was just mirroring the API for the `set()` function.  In the current use case of this function it is not important to know if any data was actually cleared - just that if it exists it was cleared.  It is possible the `handle` has never been used before so there would be no data to clear.  So this particular case (lines 362-364) is not a failure but a consequence of a reasonable, possible state of the agent.

But the `NR_FAILURE` return code does give the caller the context that there was no existing data for the `handle` which might be useful in a future, different use?

ZNeumann:
I would prefer if the PR title explained the fix in addition to the issue number, rather than just refer to the issue

ZNeumann (approved):

mfulb:
> I would prefer if the PR title explained the fix in addition to the issue number, rather than just refer to the issue

Is this better?

agent/php_mysqli.c
lavarou: This PR can be merged with the code as is.

PR #883, opened 2024-04-25
fix(agent): end guzzle's external segment for rejected request

### LINKED ISSUE #320: Guzzle timeouts are not recognized as externals.
## Description
When using Guzzle 6 to make external calls it was noticed some of the calls are not recognized as externals and it was narrowed down to those that result in timeouts.
## Steps to Reproduce

The following script will cause the issue:

```
<?php
require_once "vendor/autoload.php";

use GuzzleHttp\Promise;
use GuzzleHttp\Client;

$client = new Client();
sleep(3);
$promises = [
    "req1" => $client->getAsync('https://deelay.me/5000/http://www.example.com', ['timeout' => 3]),
    "req2" => $client->getAsync('http://www.google.com')
];

$results = GuzzleHttp\Promise\settle($promises)->wait();

echo $results['req1']['state'] . "\n";
echo $results['req2']['state'] . "\n";
?>
```

The async request "req1" will timeout and the external segment will not show up in the application trace information.

If the delay time passed to "http://deelay.me" is reduced to 500, for example, the "req1" request will show up as an external segment because it completes before the 3 second timeout specified.

## Expected Behavior

A async guzzle request which times out will show up as an external segment with a status representative of the time out and failure for the request to complete.   For example, when a async request with curl_multi time outs the http response status is set to 0.   

## Your Environment

The issue was recreated in the following environment:
Ubuntu 18.04
PHP 7.3.16
guzzle 6.5.5

Issue was reported for a system using:
Ubuntu 18.04
PHP 7.3.16
Guzzle: 6.3.3
cURL version: 7.58.0

lavarou:
Modify the onRejected handler to end guzzle's external segment even when
there's no response. This allows to mark the span as an external span and
capture the uri and method of the request. Additionally when sync requests is
rejected, onRejected callback is not called via `PromiseInterface::then` and
needs to be called manually.

Fixes #320.

### DIFF
--- a/agent/lib_guzzle6.c
+++ b/agent/lib_guzzle6.c
@@ -447,6 +447,7 @@ void nr_guzzle6_enable(TSRMLS_D) {
       "namespace newrelic\\Guzzle6;"
 
       "use Psr\\Http\\Message\\RequestInterface;"
+      "use GuzzleHttp\\Promise\\PromiseInterface;"
 
       "if (!function_exists('newrelic\\Guzzle6\\middleware')) {"
       "  function middleware(callable $handler) {"
@@ -467,8 +468,16 @@ void nr_guzzle6_enable(TSRMLS_D) {
        */
       "      $rh = new RequestHandler($request);"
       "      $promise = $handler($request, $options);"
-      "      $promise->then([$rh, 'onFulfilled'], [$rh, 'onRejected']);"
-
+      "      if (PromiseInterface::REJECTED == $promise->getState()) {"
+      /*
+                Special case for sync request. When sync requests is rejected,
+                onRejected callback is not called via `PromiseInterface::then`
+                and needs to be called manually.
+       */
+      "        $rh->onRejected($promise);"
+      "      } else {"
+      "        $promise->then([$rh, 'onFulfilled'], [$rh, 'onRejected']);"
+      "      }"
       "      return $promise;"
       "    };"
       "  }"
--- a/agent/lib_guzzle6.c
+++ b/agent/lib_guzzle6.c
@@ -112,14 +112,10 @@ static void nr_guzzle6_requesthandler_handle_response(zval* handler,
   nr_segment_external_params_t external_params = {.library = "Guzzle 6"};
   zval* request;
   zval* method;
-  zval* status;
+  zval* status = NULL;
 
   if (NR_FAILURE
       == nr_guzzle_obj_find_and_remove(handler, &segment TSRMLS_CC)) {
-    return;
-  }
-
-  if (!nr_php_psr7_is_response(response TSRMLS_CC)) {
     return;
   }
 
@@ -133,30 +129,32 @@ static void nr_guzzle6_requesthandler_handle_response(zval* handler,
     return;
   }
 
-  /*
-   * Get the X-NewRelic-App-Data response header. If there isn't one, NULL is
-   * returned, and everything still works just fine.
-   */
-  external_params.encoded_response_header
-      = nr_php_psr7_message_get_header(response, X_NEWRELIC_APP_DATA TSRMLS_CC);
-
-  if (NRPRG(txn) && NRTXN(special_flags.debug_cat)) {
-    nrl_verbosedebug(
-        NRL_CAT, "CAT: outbound response: transport='Guzzle 6' %s=" NRP_FMT,
-        X_NEWRELIC_APP_DATA, NRP_CAT(external_params.encoded_response_header));
-  }
-
-  status = nr_php_call(response, "getStatusCode");
-
-  if (nr_php_is_zval_valid_integer(status)) {
-    external_params.status = Z_LVAL_P(status);
-  }
-
   method = nr_php_call(request, "getMethod");
 
   if (nr_php_is_zval_valid_string(method)) {
     external_params.procedure
         = nr_strndup(Z_STRVAL_P(method), Z_STRLEN_P(method));
+  }
+
+  if (NULL != response && nr_php_psr7_is_response(response TSRMLS_CC)) {
+    /*
+    * Get the X-NewRelic-App-Data response header. If there isn't one, NULL is
+    * returned, and everything still works just fine.
+    */
+    external_params.encoded_response_header
+        = nr_php_psr7_message_get_header(response, X_NEWRELIC_APP_DATA TSRMLS_CC);
+
+    if (NRPRG(txn) && NRTXN(special_flags.debug_cat)) {
+      nrl_verbosedebug(
+          NRL_CAT, "CAT: outbound response: transport='Guzzle 6' %s=" NRP_FMT,
+          X_NEWRELIC_APP_DATA, NRP_CAT(external_params.encoded_response_header));
+    }
+
+    status = nr_php_call(response, "getStatusCode");
+
+    if (nr_php_is_zval_valid_integer(status)) {
+      external_params.status = Z_LVAL_P(status);
+    }
   }
 
   nr_segment_external_end(&segment, &external_params);
@@ -291,6 +289,12 @@ static PHP_NAMED_FUNCTION(nr_guzzle6_requesthandler_onrejected) {
     return;
   }
 
+  this_obj = NR_PHP_USER_FN_THIS();
+  if (NULL == this_obj) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "%s: cannot obtain 'this'", __func__);
+    return;
+  }
+
   /*
    * See if this is an exception that we can get a response from. We're going
    * to look for BadResponseException because, although it inherits from
@@ -305,18 +309,13 @@ static PHP_NAMED_FUNCTION(nr_guzzle6_requesthandler_onrejected) {
    */
   if (!nr_php_object_instanceof_class(
           exc, "GuzzleHttp\\Exception\\BadResponseException" TSRMLS_CC)) {
+    nr_guzzle6_requesthandler_handle_response(this_obj, NULL);
     return;
   }
 
   response = nr_php_call(exc, "getResponse");
   if (NULL == response) {
     nrl_verbosedebug(NRL_FRAMEWORK, "%s: error calling getResponse", __func__);
-    return;
-  }
-
-  this_obj = NR_PHP_USER_FN_THIS();
-  if (NULL == this_obj) {
-    nrl_verbosedebug(NRL_FRAMEWORK, "%s: cannot obtain 'this'", __func__);
     return;
   }
 
--- a/daemon/cmd/integration_runner/main.go
+++ b/daemon/cmd/integration_runner/main.go
@@ -245,6 +245,16 @@ func catRequest(w http.ResponseWriter, r *http.Request) {
 	w.Write(body)
 }
 
+func delayRequest(w http.ResponseWriter, r *http.Request) {
+	duration := r.URL.Query().Get("duration")
+	io.WriteString(w, "waiting...")
+	d, err := time.ParseDuration(duration)
+	if nil != err {
+		d = 0
+	}
+	time.Sleep(d)
+}
+
 func init() {
 	//setup typed flags
 	flag.Var(&flagLicense, "license", "use a license key other than the hard coded default. Supports @filename syntax for loading from files.")
@@ -303,6 +313,7 @@ func main() {
 		io.WriteString(w, "Hello world!")
 	})
 	mux.HandleFunc("/cat", catRequest)
+	mux.HandleFunc("/delay", delayRequest)
 	addr := "127.0.0.1:" + strconv.Itoa(*flagExternalPort)
 	srv := &http.Server{Addr: addr, Handler: mux}
 	ln, err := net.Listen("tcp", addr)
--- a/daemon/cmd/integration_runner/main.go
+++ b/daemon/cmd/integration_runner/main.go
@@ -375,6 +375,7 @@ func main() {
 
 	// Env vars common to all tests.
 	ctx.Env["EXTERNAL_HOST"] = externalHost
+	os.Setenv("EXTERNAL_HOST", externalHost)
 
 	handler, err := startDaemon("unix", *flagPort, flagSecurityToken.String(), flagSecuityPolicies.String())
 	if err != nil {
--- a/daemon/cmd/integration_runner/main.go
+++ b/daemon/cmd/integration_runner/main.go
@@ -375,7 +375,6 @@ func main() {
 
 	// Env vars common to all tests.
 	ctx.Env["EXTERNAL_HOST"] = externalHost
-	os.Setenv("EXTERNAL_HOST", externalHost)
 
 	handler, err := startDaemon("unix", *flagPort, flagSecurityToken.String(), flagSecuityPolicies.String())
 	if err != nil {
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -684,7 +684,7 @@ func (t *Test) Compare(harvest *newrelic.Harvest) {
 	}
 
 	if nil != t.metricsExist {
-		for _, name := range strings.Split(strings.TrimSpace(string(t.metricsExist)), "\n") {
+		for _, name := range strings.Split(strings.TrimSpace(string(t.subEnvVars(t.metricsExist))), "\n") {
 			name = strings.TrimSpace(name)
 			actual := strings.Replace(name, "__FILE__", t.Path, -1)
 			if !harvest.Metrics.Has(actual) {
@@ -694,7 +694,7 @@ func (t *Test) Compare(harvest *newrelic.Harvest) {
 	}
 
 	if nil != t.metricsDontExist {
-		for _, name := range strings.Split(strings.TrimSpace(string(t.metricsDontExist)), "\n") {
+		for _, name := range strings.Split(strings.TrimSpace(string(t.subEnvVars(t.metricsDontExist))), "\n") {
 			name = strings.TrimSpace(name)
 			actual := strings.Replace(name, "__FILE__", t.Path, -1)
 			if harvest.Metrics.Has(actual) {
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -289,9 +289,9 @@ func (t *Test) subEnvVars(in []byte) []byte {
 	re := regexp.MustCompile("ENV\\[.*?\\]")
 	return re.ReplaceAllFunc(in, func(match []byte) []byte {
 		k := string(match[4 : len(match)-1])
-		v, present := os.LookupEnv(k)
+		v, present := t.Env[k]
 		if !present {
-			v = t.Env[k]
+			v = os.Getenv(k)
 		}
 		return []byte(v)
 	})
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -199,17 +199,8 @@ func merge(a, b map[string]string) map[string]string {
 	return merged
 }
 
-// merge e into Test.Env but don't overwrite existing keys
-func (t *Test) mergeEnv(e map[string]string) {
-	for k, v := range e {
-		if _, present := t.Env[k]; !present {
-			t.Env[k] = v
-		}
-	}
-}
-
 func (t *Test) MakeRun(ctx *Context) (Tx, error) {
-	t.mergeEnv(ctx.Env)
+	t.Env = merge(ctx.Env, t.Env)
 	settings := merge(ctx.Settings, t.Settings)
 	settings["newrelic.appname"] = t.Name
 
@@ -237,7 +228,7 @@ func (t *Test) MakeSkipIf(ctx *Context) (Tx, error) {
 		return nil, nil
 	}
 
-	t.mergeEnv(ctx.Env)
+	t.Env = merge(ctx.Env, t.Env)
 	settings := merge(ctx.Settings, t.Settings)
 	settings["newrelic.appname"] = "skipif"
 
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -199,10 +199,17 @@ func merge(a, b map[string]string) map[string]string {
 	return merged
 }
 
-func (t *Test) MakeRun(ctx *Context) (Tx, error) {
-	for k, v := range ctx.Env {
-		t.Env[k] = v
+// merge e into Test.Env but don't overwrite existing keys
+func (t *Test) mergeEnv(e map[string]string) {
+	for k, v := range e {
+		if _, present := t.Env[k]; !present {
+			t.Env[k] = v
+		}
 	}
+}
+
+func (t *Test) MakeRun(ctx *Context) (Tx, error) {
+	t.mergeEnv(ctx.Env)
 	settings := merge(ctx.Settings, t.Settings)
 	settings["newrelic.appname"] = t.Name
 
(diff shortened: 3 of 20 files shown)


daemon/cmd/integration_runner/main.go
```diff
@@ -375,6 +375,7 @@ func main() {
 
 	// Env vars common to all tests.
 	ctx.Env["EXTERNAL_HOST"] = externalHost
+	os.Setenv("EXTERNAL_HOST", externalHost)
```
mfulb, line 378: Why is it necessary to add this?

tests/integration/external/guzzle6/test_timeout_async.php
```diff
@@ -20,6 +20,9 @@
 newrelic.code_level_metrics.enabled = 0
 */
 
+/*EXPECT_METRICS_EXIST
+External/127.0.0.1/all
```
mfulb, line 24: I'm unclear on why the metric and segment name below would contain "127.0.0.1" as the host but the `http.url` is captured with `ENV[EXTERNAL_HOST]` which I'm guessing is a different value as it necessitated adding support for pulling in env vars for `EXPECT_SPAN_EVENTS_LIKE`.

daemon/cmd/integration_runner/main.go
lavarou: In order to use it in `EXPECT_SPAN_EVENTS_LIKE` [here](https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/tests/integration/external/guzzle7/test_timeout_sync.php#L47).

tests/integration/external/guzzle6/test_timeout_async.php
lavarou: `EXTERNAL_HOST` is set [here](https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/daemon/cmd/integration_runner/main.go#L324) and includes port number, which, unless provided via [`-external_port` flag](https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/daemon/cmd/integration_runner/main.go#L55-L57), is assigned at random (see [here](https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/daemon/cmd/integration_runner/main.go#L317-L319) and [doc](https://pkg.go.dev/net#Listen)). I was wondering if I should split off the port part into `EXTERNAL_HOST_PORT` in order to be able to use `ENV[EXTERNAL_HOST]` in `EXPECT_METRICS_EXIST`. However this would require me to change a lot of other tests which assume that `EXTERNAL_HOST` has the port number included. Thoughts?

daemon/cmd/integration_runner/main.go
lavarou: Setting `ctx.Env["EXTERNAL_HOST"]` is not enough for `EXTERNAL_HOST` the env var to be available in the runtime environment for `os.Getenv`, used in [`SubEnvVars`](https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/daemon/internal/newrelic/integration/test.go#L291), to find it.

tests/integration/external/guzzle6/test_timeout_async.php
mfulb: That is another consideration - I'm more wondering if `EXTERNAL_HOST` is actually "127.0.0.1" in all cases or not - and if the "127.0.0.1" in the metric line might not need to have an env substitution as well?  Also should the IP/host in the metric always be the same as in the `http.url`?

daemon/cmd/integration_runner/main.go
mfulb: The code appears to associate the `ctx.Env` with each test run and is used as the env for the PHP which is spawned.

daemon/cmd/integration_runner/main.go
```diff
@@ -245,6 +245,16 @@ func catRequest(w http.ResponseWriter, r *http.Request) {
 	w.Write(body)
 }
 
+func delayRequest(w http.ResponseWriter, r *http.Request) {
+	duration := r.URL.Query().Get("duration")
+	io.WriteString(w, "waiting...")
```
mfulb, line 250: Is this output necessary?

lavarou: It is not required for the guzzle tests.

tests/integration/external/guzzle6/test_timeout_async.php
lavarou: > I'm more wondering if `EXTERNAL_HOST` is actually "127.0.0.1" in all cases or not - and if the "127.0.0.1" in the metric line might not need to have an env substitution as well?

It is not always 127.0.0.1: https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/tests/integration/external/guzzle7/test_transfer_exception_sync.php#L25

It is 127.0.0.1 for tests that make requests to endpoints exposed by `integration_runner`.

lavarou: Moreover, there are other tests that have 127.0.0.1 hardcoded in names but don't test "http.url" attributes. E.g.: https://github.com/newrelic/newrelic-php-agent/blob/55ab2076d11456cef2aed9d39f5bd87885679473/tests/integration/external/curl_exec/test_dt_correct_span_is_external.php#L69.

daemon/cmd/integration_runner/main.go
mfulb:
```suggestion
	io.WriteString(w, "waiting...")
```

mfulb: That suggestion is reasonable?

daemon/cmd/integration_runner/main.go
mfulb: It seems currently the `ctx.Env` is not included in the variables the `SubEnvVar` considers.  That might be an enhancement that would remove the need to make this addition.

lavarou: @mfulb 800b45a43ae46043ecdf1f1f0692de7af3fff4ed was needed for `ctx.Env` to be available to env var substitution.

lavarou: @mfulb https://github.com/newrelic/newrelic-php-agent/commit/800b45a43ae46043ecdf1f1f0692de7af3fff4ed wasn't enough - `Test.Env` still didn't have env vars available to PHP. I had to do this: https://github.com/newrelic/newrelic-php-agent/pull/883/commits/e20f4b93b4e7bca34327ab2c8680cf59b84ad9b5. 5e111d2374a53e4383e713cd932e667b0c456fd3 showcases how this works.

tests/integration/external/guzzle6/test_timeout_async.php
mfulb: Thank you for exploring this further.   I'm happy with your explanation.

daemon/cmd/integration_runner/main.go
mfulb: After discussion it was clarified for me this output appears in the body of the response from the `/delay` endpoint and not to the console and so I am not asking for any changes.

daemon/internal/newrelic/integration/test.go
```diff
@@ -200,35 +200,39 @@ func merge(a, b map[string]string) map[string]string {
 }
 
 func (t *Test) MakeRun(ctx *Context) (Tx, error) {
-	env := merge(ctx.Env, t.Env)
+	for k, v := range ctx.Env {
```
mfulb, line 203: The `merge()` call in the previous code would give values in `t.Env` precedence over values in `ctx.Env`.  This new code does the opposite.
My suggestion would be to call `merge()` as before and store the result in `t.Env` to ensure  operation is the same as previously.

daemon/internal/newrelic/integration/test.go
```diff
@@ -285,10 +289,15 @@ func ScrubHost(in []byte) []byte {
 	return re.ReplaceAll(in, []byte("__HOST__"))
 }
 
-func SubEnvVars(in []byte) []byte {
+func (t *Test) subEnvVars(in []byte) []byte {
 	re := regexp.MustCompile("ENV\\[.*?\\]")
 	return re.ReplaceAllFunc(in, func(match []byte) []byte {
-		return []byte(os.Getenv(string(match[4 : len(match)-1])))
+		k := string(match[4 : len(match)-1])
+		v, present := os.LookupEnv(k)
```
mfulb, line 296: This appears to give the OS environment values precedence over `t.Env` when substitution occurs?  Is this desired or is it better to let the `/*ENVIRONMENT*/` values be able to override the OS environment?

daemon/internal/newrelic/integration/test.go
lavarou: @mfulb How about 5064b7c93267fae793fbef33760b1df37fc35d77?

daemon/internal/newrelic/integration/test.go
lavarou: @mfulb Very good point! I think values from `/*ENVIRONMENT*/` directive should take precedence over OS environment. How about https://github.com/newrelic/newrelic-php-agent/commit/5064b7c93267fae793fbef33760b1df37fc35d77?

daemon/cmd/integration_runner/main.go
lavarou: And possibly the last patch: 5064b7c93267fae793fbef33760b1df37fc35d77, which ensures that values from `/*ENVIRONMENT*/` directive take precedence over OS environment.

daemon/internal/newrelic/integration/test.go
```diff
@@ -285,10 +289,15 @@ func ScrubHost(in []byte) []byte {
...
+		if !present {
+			v = t.Env[k]
```
mfulb: What happens if the key is not in in or the process environment variables? Does this return nil?

daemon/internal/newrelic/integration/test.go
```diff
@@ -285,10 +285,15 @@ func ScrubHost(in []byte) []byte {
...
+func (t *Test) subEnvVars(in []byte) []byte {
 	re := regexp.MustCompile("ENV\\[.*?\\]")
 	return re.ReplaceAllFunc(in, func(match []byte) []byte {
-		return []byte(os.Getenv(string(match[4 : len(match)-1])))
+		k := string(match[4 : len(match)-1])
+		v, present := t.Env[k]
+		if !present {
+			v = os.Getenv(k)
+		}
+		return []byte(v)
```
mfulb, line 296: What happens if the key is not in in or the process environment variables? Does this return nil?

lavarou: If `t.Env` doesn't have the value assigned to `k`, and `k` is not defined in the environment, `v` will be empty - `""`.

More info at: https://go.dev/ref/spec#Index_expressions, https://pkg.go.dev/os#Getenv, https://go.dev/blog/maps

lavarou: I have added a test to illustrate this: 5a8e3f4046fe427d329e49ac3faddc9da34b8e6c.

mfulb (approved):
Really nice work Michal - this is a great PR!

agent/lib_guzzle6.c
```diff
@@ -468,8 +468,16 @@ void nr_guzzle6_enable(TSRMLS_D) {
        */
       "      $rh = new RequestHandler($request);"
       "      $promise = $handler($request, $options);"
-      "      $promise->then([$rh, 'onFulfilled'], [$rh, 'onRejected']);"
-
+      "      if (PromiseInterface::REJECTED == $promise->getState()) {"
+      /*
+                Special case for sync request. When sync requests is rejected,
```
zsistla, line 473:
```suggestion
                Special case for sync request. When sync request is rejected,
```

tests/integration/external/guzzle6/test_transfer_exception_sync.php
```diff
@@ -0,0 +1,90 @@
...
+    new \GuzzleHttp\Exception\TransferException()
+  ]));
+
+$client = new GuzzleHttp\Client([
+  'handler' => $stack,
+]);
+
+try {
+  $response = $client->send($request);
+} catch (\GuzzleHttp\Exception\TransferException $e) {
```
zsistla, line 86: In the tests that throw exceptions, could we have versions where the exception isn't caught to verify the exception is also recorded correctly?

tests/integration/external/guzzle6/test_transfer_exception_async.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that ends up throwing exception other than BadResponseException
+is marked as http and uri is captured.
```
zsistla, line 9:
```suggestion
is marked as http and uri is captured. Status code is marked as 0 since there was no response.
```

tests/integration/external/guzzle6/test_timeout_async.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that times out is marked as http and uri is captured.
```
zsistla, line 8:
```suggestion
Test that the span for an async external request that times out is marked as http and uri is captured. Status code is marked as 0.
```

tests/integration/external/guzzle6/test_bad_response_exception_async.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that ends up throwing BadResponseException exception
```
zsistla, line 8: copy/paste comment for all tests: `reqest` should be `request`

tests/integration/external/guzzle6/test_timeout_sync.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an sync external reqest that times out is marked as http and uri is captured.
```
zsistla, line 8:
```suggestion
Test that the span for an sync external request that times out is marked as http and uri is captured. Status code is marked as 0.
```

tests/integration/external/guzzle6/test_transfer_exception_sync.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for a sync external reqest that ends up throwing exception other than BadResponseException
+is marked as http and uri is captured.
```
zsistla, line 9:
```suggestion
is marked as http and uri is captured. Status code is marked as 0.
```

tests/integration/external/guzzle7/test_timeout_async.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that times out is marked as http and uri is captured.
```
zsistla, line 8:
```suggestion
Test that the span for an async external request that times out is marked as http and uri is captured. Status code is marked as 0.
```

tests/integration/external/guzzle7/test_timeout_sync.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an sync external reqest that times out is marked as http and uri is captured.
```
zsistla, line 8:
```suggestion
Test that the span for an sync external request that times out is marked as http and uri is captured. Status code is marked as 0.
```

tests/integration/external/guzzle7/test_transfer_exception_async.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that ends up throwing exception other than BadResponseException
+is marked as http and uri is captured.
```
zsistla, line 9:
```suggestion
is marked as http and uri is captured. Status code is marked as 0.
```

tests/integration/external/guzzle7/test_transfer_exception_sync.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for a sync external reqest that ends up throwing exception other than BadResponseException
+is marked as http and uri is captured.
```
zsistla, line 9:
```suggestion
is marked as http and uri is captured. Status code is marked as 0.
```

zsistla:
Really nicely done!  Thanks for addressing this issue.

Couple of tests comments, but otherwise, LGTM.

tests/integration/external/guzzle6/test_bad_response_exception_async.php
lavarou, line 8:
```suggestion
Test that the span for an async external request that ends up throwing BadResponseException exception
```

tests/integration/external/guzzle6/test_bad_response_exception_sync.php
```diff
@@ -0,0 +1,96 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for a sync external reqest that ends up throwing BadResponseException exception
```
lavarou, line 8:
```suggestion
Test that the span for a sync external request that ends up throwing BadResponseException exception
```

tests/integration/external/guzzle6/test_transfer_exception_async.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that ends up throwing exception other than BadResponseException
```
lavarou, line 8:
```suggestion
Test that the span for an async external request that ends up throwing exception other than BadResponseException
```

tests/integration/external/guzzle6/test_transfer_exception_sync.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for a sync external reqest that ends up throwing exception other than BadResponseException
```
lavarou, line 8:
```suggestion
Test that the span for a sync external request that ends up throwing exception other than BadResponseException
```

tests/integration/external/guzzle7/test_bad_response_exception_async.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that ends up throwing BadResponseException exception
```
lavarou, line 8:
```suggestion
Test that the span for an async external request that ends up throwing BadResponseException exception
```

tests/integration/external/guzzle7/test_bad_response_exception_sync.php
```diff
@@ -0,0 +1,96 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for a sync external reqest that ends up throwing BadResponseException exception
```
lavarou, line 8:
```suggestion
Test that the span for a sync external request that ends up throwing BadResponseException exception
```

tests/integration/external/guzzle7/test_transfer_exception_async.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for an async external reqest that ends up throwing exception other than BadResponseException
```
lavarou, line 8:
```suggestion
Test that the span for an async external request that ends up throwing exception other than BadResponseException
```

tests/integration/external/guzzle7/test_transfer_exception_sync.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that the span for a sync external reqest that ends up throwing exception other than BadResponseException
```
lavarou, line 8:
```suggestion
Test that the span for a sync external request that ends up throwing exception other than BadResponseException
```

tests/integration/external/guzzle6/test_transfer_exception_sync.php
lavarou: @zsistla Although exception handling is a different feature in the agent, I added a couple of tests which demonstrate that Guzzle's instrumentation doesn't affect agent's ability to record errors for uncaught exceptions. Is bed374a611dd7df68af857eb8e284b624f933f66 sufficient?

zsistla: Since we're creating the external span and guzzle is a special case with our middleware, it's good to check that errors are properly added for the two exception paths (BadResponseException and other)we cover here.
Thanks for adding the tests for guzzle7, they look good!  Can they be added for guzzle6 as well?

tests/integration/external/guzzle6/test_bad_response_exception_async.php
lavarou: Addressed in 13db0365f685a2a186d570757fee9dc15a7fdfd8.

tests/integration/external/guzzle6/test_transfer_exception_sync.php
lavarou: @zsistla Added in 0af94988bbfc48a010089a6b23d19a5efcd5e84f. 

Side note: Guzzle 6 is no longer supported by Guzzle. It reached its EOL on 2023-10-31 - see [Guzzle version guidance](https://github.com/guzzle/guzzle?tab=readme-ov-file#version-guidance).

PR #890, opened 2024-05-07
security(daemon): upgrade golang to 1.22

lavarou:


.github/workflows/test-agent.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     env:
       IMAGE_NAME: newrelic/nr-php-agent-builder
       IMAGE_TAG: make-go
-      IMAGE_VERSION: ${{vars.MAKE_GO_VERSION}}
+      IMAGE_VERSION: v3
```
mfulb, line 27: Is this intentional to hard code the version or is this a relic from testing?

lavarou: Yes, this is intentional and will be reverted once 'v3' of make-go variant of newrelic/nr-php-agent-builder passes testing.

zsistla (approved):

PR #892, opened 2024-05-08
feat(agent): add function to convert ini to env

bduranleau-nr:


agent/Makefile.frag
```diff
@@ -97,9 +97,10 @@ TEST_BINARIES = \
 	tests/test_pdo_mysql \
 	tests/test_pdo_pgsql \
 	tests/test_pgsql \
-        tests/test_php_error \
+	tests/test_php_error \
```
mfulb, line 100: Is this a white space change?

bduranleau-nr: Yes

agent/php_nrini.c
```diff
@@ -222,6 +223,53 @@
 
 ZEND_DECLARE_MODULE_GLOBALS(newrelic)
 
+#define NR_INI_PREFIX_LEN 9
+
+char* nr_ini_to_env(const char* ini_name) {
+  char* env_name = NULL;
+  char* buf = NULL;
+  char* ini_upper = NULL;
+  const char* NR_PREFIX_STR = "NEW_RELIC";
```
mfulb, line 232: I am not sure about using all caps for a local variable name.  I see what you are going for here.  If we want to go with all caps because this is a constant then can we move `NR_INI_PREFIX_LEN` into as a local as well defined as 

```
size_t NR_INI_PREFIX_LEN = nr_strlen(NR_PREFIX_STR);
```
I would like a solution where it isn't necessary to update the length if the string is ever changed.

agent/php_nrini.c
```diff
@@ -222,6 +223,53 @@
...
+  int ini_len = nr_strlen(ini_name);
+
+  // 'newrelic.' is 9 characters - anything less should be rejected.
+  if (NULL == ini_name || nr_strlen(ini_name) <= NR_INI_PREFIX_LEN) {
+    return NULL;
+  }
+
+  // ini value should start with the 'newrelic.' prefix
+  if (0 != nr_stridx(ini_name, "newrelic.")) {
```
mfulb, line 241: A more direct test is something like used in `php_environment.c::nr_php_process_environment_variable_to_nrobj()`:

```
  if (0 == nr_strncmp(ini_name, NR_PREFIX_STR, NR_INI_PREFIX_LEN)) {
```

agent/php_nrini.c
```diff
@@ -222,6 +223,53 @@
...
+    return NULL;
+  }
+
+  // algorithm:
```
mfulb, line 245: There is an additional requirement that consecutive underscores should be merged into a single underscore.  I don't think this function would handle that case.

agent/php_nrini.c
bduranleau-nr: > I am not sure about using all caps for a local variable name.

I elected to do it this way to cut down on what I'd call `#define` pollution- saves the effort of having to undef something that only has to be used once or in a single function. As for all caps, I like the idea of the nature of the variable (being a constant) being abundantly clear vs a non-const variable.

> If we want to go with all caps because this is a constant then can we move NR_INI_PREFIX_LEN into as a local as well

We can (and maybe should) do this- but the provided solution is not quite correct- `NR_INI_PREFIX_LEN` is meant to capture the length of the string `newrelic.`, not `NEW_RELIC`. It just happens that they are currently the same length, but as you point out, that may not be the case.

agent/php_nrini.c
```diff
@@ -222,6 +223,53 @@
...
+  // 4. snprintf append the string to 'NEW_RELIC_'
+  // 5. return result
+
+  ini_upper = nr_string_to_uppercase(ini_name);
+
+  buf = (char*)nr_malloc(ini_len - NR_INI_PREFIX_LEN + 1);
+  nr_strcpy(buf, ini_upper + NR_INI_PREFIX_LEN);
+
+  for (int i = 0; i < nr_strlen(buf); i++) {
+    if (buf[i] == '.') {
```
mfulb, line 258: If you change the allocation of `buf` to use `nr_zalloc()` and then change this check to 

```
if ('.' == buf[i] && (0 == i || '_' != buf[i-1])) {
```
then this will handle consecutive underscores properly.  By switching to `nr_zalloc()` then as you build the string it will always be zero terminated, accounting for this change might make the string shorter than the original.

agent/php_nrini.c
bduranleau-nr: This will not quite work because the `ini_name` in this case should be something with the format `newrelic.<field>.<etc>`, whereas `NR_PREFIX_STR` is the "env" version of the ini, and is set to `NEW_RELIC`.

agent/tests/test_php_nrini.c
```diff
@@ -0,0 +1,61 @@
...
+  tlib_pass_if_null("no value after prefix", res);
+
+  nr_free(res);
+
+  res = nr_ini_to_env(NULL);
+
+  tlib_pass_if_null("reject null values", res);
+
+  nr_free(res);
+}
```
mfulb, line 54: I'd add a case which leads to multiple underscores, like `newrelic._option_`.  This is an odd name that I don't think we'd ever use, but the spec says multiple underscores should be condensed to a single one.

agent/php_nrini.c
mfulb: Got it - I'd like the (corrected) change if you're good with it.

agent/php_nrini.c
mfulb: Ok I see I have confused the use of prefix for `newrelic.` and `NEW_RELIC`.  The current form is OK.

```

agent/php_nrini.c
bduranleau-nr: Cleaned up constants here: 611b2f0b54e66b70c3074b87182c058006ee88d2

agent/php_nrini.c
bduranleau-nr: reworked the algorithm to handle this edge case here: d58b00e68f8ab2587fa3cfee4d4cde5b64f44e88

agent/php_nrini.c
bduranleau-nr: incorporated this feedback here: d58b00e68f8ab2587fa3cfee4d4cde5b64f44e88

mfulb (approved):

PR #896, opened 2024-05-14
chore(agent): cleanup agent's verbosedebug log

lavarou:
Don't print `Stack depth: ...` messages. Stack depth can
can be inferred from the level of indentation used in `nr_php_show_exec` and
`nr_php_show_exec_return` - both use `nr_php_show_exec_indentation`, which
uses `NRPRG(php_cur_stack_depth)` to calculate indentation level.

### DIFF
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -2163,9 +2163,13 @@ void nr_php_observer_fcall_begin(zend_execute_data* execute_data) {
   int show_executes = NR_PHP_PROCESS_GLOBALS(special_flags).show_executes;
 
   if (nrunlikely(show_executes)) {
-    nrl_verbosedebug(NRL_AGENT,
-                     "Stack depth: %d after OAPI function beginning via %s",
-                     NRPRG(php_cur_stack_depth), __func__);
+    int show_stack_depth
+        = NR_PHP_PROCESS_GLOBALS(special_flags).show_execute_stack_depth;
+    if (nrunlikely(show_stack_depth)) {
+      nrl_verbosedebug(NRL_AGENT,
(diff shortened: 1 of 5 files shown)


agent/php_execute.c
```diff
@@ -2191,9 +2195,13 @@ void nr_php_observer_fcall_end(zend_execute_data* execute_data,
         = NR_PHP_PROCESS_GLOBALS(special_flags).show_execute_returns;
 
     if (nrunlikely(show_executes_return)) {
-      nrl_verbosedebug(NRL_AGENT,
-                       "Stack depth: %d before OAPI function exiting via %s",
-                       NRPRG(php_cur_stack_depth), __func__);
+      int show_stack_depth
```
zsistla, line 2198: Are we likely to save a little bit of overhead by not doing the assignment here?

This case [here](https://github.com/newrelic/newrelic-php-agent/pull/896/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R2194) 
especially is setting a variable with every single func call end.  Why not just use the global value?

so something like:
`if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).show_execute_stack_depth))`
instead of 
 `if (nrunlikely(show_stack_depth))`

lavarou: The CPU instructions generated by the compiler are the same regardless of the C code. I disassembled php_execute.o object on arm64 build from different variations of C and the CPU instructions were always the same.

zsistla: Thanks for check on this!

zsistla (approved):

ZNeumann (approved):

PR #906, opened 2024-05-24
feat(installer): Adds support for configuration via environment variables [DRAFT]

mfulb:


agent/php_nrini.c
```diff
@@ -7,6 +7,7 @@
 #include "php_globals.h"
 #include "php_hash.h"
 #include "php_internal_instrument.h"
+#include "php_nrini.h"
```
lavarou, line 10: This header is already included on line 12:
```suggestion
```

agent/newrelic-install.sh
```diff
@@ -420,6 +422,9 @@ fi
 #   If this is set, it must be set to a valid New Relic license key, and this
 #   key will be set in the daemon config file.
 #
+#   NOTE: If NR_CONFIG_WITH_ENVIRON is defined and NEW_RELIC_LICENSE is defined
+#         then the value of NEW_RELIC_LICENSE will be used.s
```
lavarou, line 426:
```suggestion
#         then the value of NEW_RELIC_LICENSE will be used instead of NR_INSTALL_KEY.
```

agent/newrelic-install.sh
```diff
@@ -1471,25 +1486,49 @@ install_agent_here() {
 # file exists in the target directory.
 #
     if [ -n "${pi_inidir_cli}" -a ! -f "${pi_inidir_cli}/newrelic.ini" ]; then
-      if sed -e "s/REPLACE_WITH_REAL_KEY/${nrkey}/" "${ilibdir}/scripts/newrelic.ini.template" > "${pi_inidir_cli}/newrelic.ini"; then
+      # check if doing old style injection of just "newrelic.license" or if injecting values from env vars
+      if [ -n "${NR_CONFIG_WITH_ENVIRON}" ]; then
+        if php "${ilibdir}/newrelic-install-inject-envvars.php" "${ilibdir}/scripts/newrelic.ini.template" "${pi_inidir_cli}/newrelic.ini"; then
```
lavarou, line 1491: Shouldn't starting the daemon be disabled here?

tests/newrelic-installer/verify_inject_envvars_script_injection.php
```diff
@@ -0,0 +1,84 @@
+<?PHP
```
lavarou, line 1:
```suggestion
<?php
```

tests/newrelic-installer/verify_inject_envvars_script_mapping.php
```diff
@@ -0,0 +1,61 @@
+<?PHP
```
lavarou, line 1:
```suggestion
<?php
```

lavarou (approved):

agent/php_nrini.c
```diff
@@ -3494,3 +3495,56 @@ int nr_php_ini_setting_is_set_by_user(const char* name) {
...
 #endif /* PHP7 */
 }
+
+static int newrelic_collect_ini_names(zend_ini_entry* ini_entry,
+                                      zval* name_array,
+                                      zend_hash_key* key NRUNUSED) {
+  char* ini_name = NULL;
+  char* env_name = NULL;
+
+  if (ini_entry->module_number != NR_PHP_PROCESS_GLOBALS(our_module_number)) {
```
bduranleau-nr, line 3505: are we safe to always assume that `ini_entry` will never be `NULL`?

agent/newrelic-install.sh
```diff
@@ -1471,25 +1486,49 @@ install_agent_here() {
...
+      elif sed -e "s/REPLACE_WITH_REAL_KEY/${nrkey}/" "${ilibdir}/scripts/newrelic.ini.template" > "${pi_inidir_cli}/newrelic.ini"; then
         logcmd chmod 644 "${pi_inidir_cli}/newrelic.ini"
         if [ -z "${NR_INSTALL_SILENT}" ]; then
           echo "      Install Status : ${pi_inidir_cli}/newrelic.ini created"
         fi
       else
         istat="failed"
+      fi
+      istat="failed"
+      if [ -n "${istat}" ]; then
```
bduranleau-nr, line 1508: if `istat` is set to `"failed"` on the line above, won't this statement always resolve to `true`? Can't we remove this conditional entirely in that case?

bduranleau-nr: Was this `istat="failed"` is intentional or was it accidentally copied from the `else` statement above?

PR #914, opened 2024-06-10
feat: deprecate PHPs 7.0 and 7.1

lavarou:
Have the newrelic-install script log warning when it detects deprecated PHPs.

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -1038,9 +1038,19 @@ for this copy of PHP. We apologize for the inconvenience.
 
   case "${pi_ver}" in
      7.0.*)
+      warning_message="${pdir}: Support for PHP '${pi_ver}' in the New Relic PHP agent is deprecated."
+      if [ -z "${NR_INSTALL_SILENT}" ]; then
+         echo $warning_message
+      fi
+      log $warning_message
       ;;
 
     7.1.*)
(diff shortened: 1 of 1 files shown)


agent/newrelic-install.sh
```diff
@@ -1038,9 +1038,19 @@ for this copy of PHP. We apologize for the inconvenience.
 
   case "${pi_ver}" in
      7.0.*)
+      warning_message="${pdir}: Support for PHP '${pi_ver}' in the New Relic PHP agent is deprecated."
```
zsistla, line 1041: Are the single quotes required around pi_ver?

lavarou: They're not required - it's purely a matter of visual preference. The only reason the quotes are there, is because that's the format of the message that was used when PHPs 5.5 and 5.6 were deprecated. Here's the comparison of the message with the quotes:
```
/usr/local/bin: Support for PHP '7.0.33' in the New Relic PHP agent is deprecated.
```
and without the quotes:
```
/usr/local/bin: Support for PHP 7.0.33 in the New Relic PHP agent is deprecated.
```

zsistla: Personally, I think it looks better without, but it's fine either way.

mfulb: I'd vote for removing the single quotes.

mfulb (approved):

PR #916, opened 2024-06-12
fix(agent): Fixes Doctrine 2.18+ detection

mfulb:
Addresses changes in Doctrine 2.18+ (but not 3.x) that affected detection.

There was a [commit](https://github.com/doctrine/orm/commit/073f2aa891249e582e63320cab34b035b591cf22) for Doctrine 2.18 that reworked the directory structure of the Doctrine library.  This PR adds another file for Doctrine to allow it to be auto-detected.


lavarou (approved):

zsistla (approved):
Thanks for investigating/fixing!

PR #917, opened 2024-06-13
fix(agent): Slim txn naming updates

zsistla:
1) the scheme for route run to correlate with proper OAPI naming was incorrectly listed and therefore txn naming precedence was incorrectly handled
2) this PR kept same naming scheme in route:run putting it before wrapper_call and saying ok to overwrite (instead of after and not ok to overwrite)
3) number 2 needed to incorporate fallback naming scheme 
4) added fallback naming scheme (since some errors and middleware intervention give the wrong txn name) using dispatcher:dispatch.

When there is an exception thrown in a route that is handled by the slim routing and error middleware, it is superseded by: Uncaught exception 'Slim\\Exception\\HttpNotFoundException' with message 'Not found.' and transformed into an http 404 error and is shown in the http.statuscode field as 404.  If the middleware isn't loaded, it is recorded as an uncaught exception and status code is 200.

Wrap function Slim\\Routing\\Dispatcher::dispatch
It will name set the txn_name in case we hit an error and Slim\\Routing\\Route::run isn't called.  It is set to be overwritable do that if we encounter Slim\\Routing\\Route::run, it can overwrite with the most accurate info (this is only relevant in the case of redirects).  This allows us to give the txn_name even in error cases when Slim\\Routing\\Route::run isn't called.

Corresponds with multiverse updates which passed against this branch.

### DIFF
--- a/agent/fw_slim.c
+++ b/agent/fw_slim.c
@@ -117,7 +117,7 @@ NR_PHP_WRAPPER_END
 
 /*
  * public function dispatch(string $method, string $uri): RoutingResults
- * This is fallback naming mechanism for Slim 4 routing when the Slime 4
+ * This is fallback naming mechanism for Slim 4 routing when the Slim 4
  * Slim\\Routing\\Route::run due to errors or middlware intervening.
  * In this case, `nr_txn_set_path` is called before `NR_PHP_WRAPPER_CALL` with
  * `NR_NOT_OK_TO_OVERWRITE` and as this corresponds to calling the first wrapped
--- a/agent/fw_slim.c
+++ b/agent/fw_slim.c
@@ -89,8 +89,8 @@ NR_PHP_WRAPPER_END
  * to a route.
  *
  * In this case, `nr_txn_set_path` is called before `NR_PHP_WRAPPER_CALL` with
- * `NR_OK_TO_OVERWRITE` and as this corresponds to calling the last wrapped 
- * function call of this type gets to name the txn; therefore needs a before 
+ * `NR_OK_TO_OVERWRITE` and as this corresponds to calling the last wrapped
+ * function call of this type gets to name the txn; therefore needs a before
  * call for OAPI.
  */
 NR_PHP_WRAPPER(nr_slim3_4_route_run) {
@@ -118,7 +118,8 @@ NR_PHP_WRAPPER_END
 /*
  * public function dispatch(string $method, string $uri): RoutingResults
  * This is fallback naming mechanism for Slim 4 routing when the Slim 4
- * Slim\\Routing\\Route::run due to errors or middlware intervening.
+ * Slim\\Routing\\Route::run does not run due middlware intervening on
+ * certain errors.
  * In this case, `nr_txn_set_path` is called before `NR_PHP_WRAPPER_CALL` with
  * `NR_NOT_OK_TO_OVERWRITE` and as this corresponds to calling the first wrapped
  * function in func_begin.
--- a/agent/fw_slim.c
+++ b/agent/fw_slim.c
@@ -81,28 +81,64 @@ NR_PHP_WRAPPER(nr_slim2_route_dispatch) {
 NR_PHP_WRAPPER_END
 
 /*
- * Wrap the \Slim\Route::run method, which is the happy path for Slim routing.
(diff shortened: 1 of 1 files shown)


agent/fw_slim.c
```diff
@@ -81,28 +81,64 @@ NR_PHP_WRAPPER(nr_slim2_route_dispatch) {
...
  * to a route.
  *
  * In this case, `nr_txn_set_path` is called after `NR_PHP_WRAPPER_CALL` with
- * `NR_OK_TO_OVERWRITE` and as this corresponds to calling the wrapped function
- * in func_end no change is needed to ensure OAPI compatibility as it will use
- * the default func_end after callback. This entails that the first wrapped
- * function call of this type gets to name the txn.
+ * `NR_OK_TO_OVERWRITE` before wrapper_call and as this corresponds to calling
+ * the last wrapped function call of this type gets to name the txn therefore
+ * needs a before call for OAPI.
```
lavarou, line 94: This comment doesn't seem right - it says two contradicting things: after `NR_PHP_WRAPPER_CALL` and before wrapper_call:

> ... `nr_txn_set_path` is called after `NR_PHP_WRAPPER_CALL` with
`NR_OK_TO_OVERWRITE` before wrapper_call

I think some of it is a left over from before this PR, when `nr_txn_set_path` was actually called *after* `NR_PHP_WRAPPER_CALL`. I think it should be changed to:
> ... `nr_txn_set_path` is called before `NR_PHP_WRAPPER_CALL` with
`NR_OK_TO_OVERWRITE`

agent/fw_slim.c
```diff
@@ -81,28 +81,64 @@ NR_PHP_WRAPPER(nr_slim2_route_dispatch) {
...
+    nr_free(txn_name);
+  }
+
   NR_PHP_WRAPPER_CALL;
+}
+NR_PHP_WRAPPER_END
+
+/*
+ * public function dispatch(string $method, string $uri): RoutingResults
+ * This is fallback naming mechanism for Slim 4 routing when the Slime 4
```
lavarou, line 120:
```suggestion
 * This is fallback naming mechanism for Slim 4 routing when the Slim 4
```

agent/fw_slim.c
```diff
@@ -81,28 +81,64 @@ NR_PHP_WRAPPER(nr_slim2_route_dispatch) {
...
+NR_PHP_WRAPPER(nr_slim4_route_dispatch) {
+  char* txn_name = NULL;
+  zval* route_name = NULL;
+
+  (void)wraprec;
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_SLIM);
+
+  /* Get the route name. The first arg is the method, 2nd arg is routename. */
+  route_name = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
lavarou, line 135:
```suggestion
  route_name = nr_php_arg_get(2, NR_EXECUTE_ORIG_ARGS);
```

agent/fw_slim.c
zsistla: good catch, thanks!
please see: [3ac4075](https://github.com/newrelic/newrelic-php-agent/pull/917/commits/3ac40759826e4ba64bdc45785549f922c167d83b)

agent/fw_slim.c
```diff
@@ -81,28 +81,64 @@ NR_PHP_WRAPPER(nr_slim2_route_dispatch) {
...
+  }
+
   NR_PHP_WRAPPER_CALL;
+}
+NR_PHP_WRAPPER_END
+
+/*
+ * public function dispatch(string $method, string $uri): RoutingResults
+ * This is fallback naming mechanism for Slim 4 routing when the Slime 4
+ * Slim\\Routing\\Route::run due to errors or middlware intervening.
```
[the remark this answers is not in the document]
zsistla: Please see [2e35374](https://github.com/newrelic/newrelic-php-agent/pull/917/commits/2e35374450fed26004c09868a243002388ecb60c)
which clarifies that sometimes route:run does not run at all when middleware handles certain error conditions.

lavarou (approved):
Thanks for fixing this!

ZNeumann (approved):

mfulb (approved):

PR #918, opened 2024-06-14
feat(agent): add support for csec php agent

anmol-ap:
Changes required for security agent :
Extract account id from the connection info
Add public method to get entity guid, account id, and other metadata required by security agent to connect to SE
Handling to read the security.agent.enabled and security.enabled flags and log the same [NR-276914](https://new-relic.atlassian.net/browse/NR-276914)
Add additional config required for security agent

[NR-276914]: https://new-relic.atlassian.net/browse/NR-276914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/csec_metadata.c
+++ b/agent/csec_metadata.c
@@ -1,0 +1,48 @@
+#include "csec_metadata.h"
+#include "util_strings.h"
+#include "php_hash.h"
+#include "php_api_internal.h"
+
+static void nr_csec_php_add_assoc_string_const(zval* arr,
+                                          const char* key,
+                                          const char* value) {
+  char* val = NULL;
+
+  if (NULL == arr || NULL == key || NULL == value) {
+    return;
+  }
+
+  val = nr_strdup(value);
+  nr_php_add_assoc_string(arr, key, val);
+  nr_free(val);
+}
+
+#ifdef TAGS
+void zif_newrelic_get_security_metadata(void); /* ctags landing pad only */
+void newrelic_get_security_metadata(void);     /* ctags landing pad only */
+#endif
+PHP_FUNCTION(newrelic_get_security_metadata) {
+
+  NR_UNUSED_RETURN_VALUE;
+  NR_UNUSED_RETURN_VALUE_PTR;
+  NR_UNUSED_RETURN_VALUE_USED;
+  NR_UNUSED_THIS_PTR;
+  NR_UNUSED_EXECUTE_DATA;
+
+  array_init(return_value);
+
+  nr_csec_php_add_assoc_string_const(return_value, KEY_ENTITY_NAME, nr_app_get_entity_name(NRPRG(app)));
+  nr_csec_php_add_assoc_string_const(return_value, KEY_ENTITY_TYPE, nr_app_get_entity_type(NRPRG(app)));
+  nr_csec_php_add_assoc_string_const(return_value, KEY_ENTITY_GUID, nr_app_get_entity_guid(NRPRG(app)));
+  nr_csec_php_add_assoc_string_const(return_value, KEY_HOSTNAME, nr_app_get_host_name(NRPRG(app)));
+  nr_csec_php_add_assoc_string_const(return_value, KEY_LICENSE, NRPRG(license).value);
+
+  if (NRPRG(app)) {
+    nr_csec_php_add_assoc_string_const(return_value, KEY_AGENT_RUN_ID, NRPRG(app)->agent_run_id);
+    nr_csec_php_add_assoc_string_const(return_value, KEY_ACCOUNT_ID, NRPRG(app)->account_id);
+    nr_csec_php_add_assoc_string_const(return_value, KEY_PLICENSE, NRPRG(app)->plicense);
+    int high_security = NRPRG(app)->info.high_security;
+    add_assoc_long(return_value, KEY_HIGH_SECURITY, (long)high_security);
+  }
+
+}
--- a/agent/csec_metadata.h
+++ b/agent/csec_metadata.h
@@ -1,0 +1,12 @@
+#include "php_agent.h"
+#include "util_hashmap.h"
+
+#define KEY_ENTITY_NAME "entity.name"
+#define KEY_ENTITY_TYPE "entity.type"
+#define KEY_ENTITY_GUID "entity.guid"
+#define KEY_HOSTNAME "hostname"
+#define KEY_AGENT_RUN_ID "agent.run.id"
+#define KEY_ACCOUNT_ID "account.id"
+#define KEY_LICENSE "license"
+#define KEY_PLICENSE "plicense"
+#define KEY_HIGH_SECURITY "high_security"
--- a/agent/php_nrini.c
+++ b/agent/php_nrini.c
@@ -515,6 +515,58 @@ static PHP_INI_MH(nr_high_security_mh) {
   return SUCCESS;
 }
 
+static PHP_INI_MH(nr_security_enabled_mh) {
+  int val;
+
+  (void)entry;
+  (void)NEW_VALUE_LEN;
+  (void)mh_arg1;
+  (void)mh_arg2;
+  (void)mh_arg3;
+  (void)stage;
+  NR_UNUSED_TSRMLS;
+
+  val = nr_bool_from_str(NEW_VALUE);
+
+  if (-1 == val) {
+    return FAILURE;
+  }
+
+  if (val) {
+    NR_PHP_PROCESS_GLOBALS(nr_security_enabled) = true;
+  } else {
+    NR_PHP_PROCESS_GLOBALS(nr_security_enabled) = false;
+  }
+
+  return SUCCESS;
+}
+
+static PHP_INI_MH(nr_security_agent_enabled_mh) {
+  int val;
+
+  (void)entry;
+  (void)NEW_VALUE_LEN;
+  (void)mh_arg1;
+  (void)mh_arg2;
+  (void)mh_arg3;
+  (void)stage;
+  NR_UNUSED_TSRMLS;
+
+  val = nr_bool_from_str(NEW_VALUE);
+
+  if (-1 == val) {
+    return FAILURE;
+  }
+
+  if (val) {
+    NR_PHP_PROCESS_GLOBALS(nr_security_agent_enabled) = true;
+  } else {
+    NR_PHP_PROCESS_GLOBALS(nr_security_agent_enabled) = false;
+  }
+
+  return SUCCESS;
+}
+
 static PHP_INI_MH(nr_preload_framework_library_detection_mh) {
   int val;
 
@@ -2007,6 +2059,18 @@ PHP_INI_ENTRY_EX("newrelic.high_security",
                  "0",
                  NR_PHP_SYSTEM,
                  nr_high_security_mh,
+                 0)
+
+PHP_INI_ENTRY_EX("newrelic.security.agent.enabled",
+                 "0",
+                 NR_PHP_SYSTEM,
+                 nr_security_agent_enabled_mh,
+                 0)
+
+PHP_INI_ENTRY_EX("newrelic.security.enabled",
+                 "0",
+                 NR_PHP_SYSTEM,
+                 nr_security_enabled_mh,
                  0)
 
 /*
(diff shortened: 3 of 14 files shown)


agent/csec_metadata.c
```diff
@@ -0,0 +1,48 @@
...
+  val = nr_strdup(value);
+  nr_php_add_assoc_string(arr, key, val);
+  nr_free(val);
+}
+
+#ifdef TAGS
+void zif_newrelic_get_security_metadata(void); /* ctags landing pad only */
+void newrelic_get_security_metadata(void);     /* ctags landing pad only */
+#endif
+PHP_FUNCTION(newrelic_get_security_metadata) {
```
lavarou, line 24: Does this function need to be called from PHP?

agent/scripts/newrelic.ini.template
```diff
@@ -1332,3 +1332,86 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
+; Type   : string
+; Scope  : system
+; Default: none
+; Info   : Sets the full path of the client certificate in PEM
+;          format. When set, this certificate will be used to
+;          authenticate the New Relic IAST Security Engine's url. If
+;          not set, the default certificate will be used. Make
+;          sure the file permissions are correct.
+;
+;newrelic.security.validator.client_ssl_cert_filepath = ""
```
lavarou, line 1417: Why these settings are added? newrelic-php-agent is neither parsing nor using any of these settings.

agent/php_nrini.c
```diff
@@ -2009,6 +2061,18 @@ PHP_INI_ENTRY_EX("newrelic.high_security",
...
+                 NR_PHP_SYSTEM,
+                 nr_security_agent_enabled_mh,
+                 0)
+
+PHP_INI_ENTRY_EX("newrelic.security.enabled",
+                 "0",
+                 NR_PHP_SYSTEM,
+                 nr_security_enabled_mh,
+                 0)
+
```
lavarou, line 2075: Does the newrelic-php-agent need these settings? They don't seem to be used anywhere...

agent/csec_metadata.c
anmol-ap: No but this function will be called by the Security Agent

agent/scripts/newrelic.ini.template
anmol-ap: newrelic-php-agent will be parsing newrelic.security.agent.enabled and newrelic.security.enabled settings and logging in the log file if these settings are disabled. That is consistent across other language agents also. And the other settings will be read by the Security agent and we do not want to introduce another config file.

agent/php_nrini.c
anmol-ap: newrelic.security.agent.enabled and newrelic.security.enabled settings are logged in the APM log file if these settings are disabled in the minit function

lavarou: Will the values of these settings be used by the security agent as well?

agent/scripts/newrelic.ini.template
lavarou: > newrelic-php-agent will be parsing newrelic.security.agent.enabled and newrelic.security.enabled settings and logging in the log file if these settings are disabled. That is consistent across other language agents also.

What is the meaning of these two flags? What is the difference between them?

> And the other settings will be read by the Security agent and we do not want to introduce another config file.

Since newrelic-php-agent is not parsing any of those, I don't think they belong to the `agent/scripts/newrelic.ini.template`, which serves as a quick-access documentation of newrelic-php-agent settings. These settings must be properly documented in official documentation available on docs.newrelic.com. If Security agent wants to also provide a quick-access documentation for ini settings controlling its behavior, it should include its own ini template in the distribution.

agent/csec_metadata.c
lavarou: If that's the case then this shouldn't be a PHP user function but a function exported by newrelic-php-agent shared object, callable from a program by the means of dynamic binding - see https://linux.die.net/man/3/dlsym.

lavarou: newrelic-php-agent already does similar thing - see [here](https://github.com/newrelic/newrelic-php-agent/blob/R10.21/agent/php_minit.c#L387).

anmol-ap: Thanks Michal, I will update the code and not expose this as a PHP user function. Thanks for the reference too.

agent/scripts/newrelic.ini.template
anmol-ap: > What is the meaning of these two flags? What is the difference between them?

The newrelic.security.agent.enabled is a flag for not even loading the agent. But since php security agent is not a dependency like in other language agents, this flag tells the security agent to not perform instrumentation. A complete bypass of the agent. But this has been kept to keep the config consistent across all language agents.

The newrelic.security.enabled flag tells the php security agent to perform instrumentation and all the tasks for the security agent.

> Since newrelic-php-agent is not parsing any of those, I don't think they belong to the

I understand this but IAST security agent is not a different agent for the customer. To the customer, there is a single agent which supports both APM and IAST. Therefore, we do not want a separate ini file for the security agent. Plus 2 settings mentioned above are already read and logged in APM agent.

agent/php_nrini.c
anmol-ap: Yes, these values are to be used by the security agent as well as mentioned [here](https://github.com/newrelic/newrelic-php-agent/pull/918#discussion_r1645443921) in comment.

agent/scripts/newrelic.ini.template
```diff
@@ -1332,3 +1332,86 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
 ;
 ;newrelic.vulnerability_management.package_detection.enabled = true
+
+; Setting: newrelic.security.enabled
+; Type   : boolean
+; Scope  : system
+; Default: false
+; Info   : Indicates if attack detection security module is to be enabled
+;
+newrelic.security.enabled = true
```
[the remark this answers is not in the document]
anmol-ap: That is correct, by default the newrelic.security.enabled setting should be false. I will make this change.

anmol-ap:
Updated the readme file and the third party notices as per the legal review : [Legal Review](https://new-relic.atlassian.net/browse/NR-264094?focusedCommentId=774581)

agent/csec_metadata.c
anmol-ap: I tried the dynamic call using dlopen and dlsym, but seems like the values may not be accessible.
`nr_app_get_entity_name(NRPRG(app))` is one of the calls which is required, but `NRPRG` is the zend module's address space which does not seem accessible directly.

`fprintf(stderr, "%p : newrelic_globals.app\n", NRPRG(app));`

gives `(nil)` when called from dlopen'ed binary while from inside a valid address is printed.

Lets connect over call to finalise a solution.

agent/scripts/newrelic.ini.template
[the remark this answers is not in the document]
anmol-ap: Done [newrelic.security.enabled should be false by default](https://github.com/newrelic/newrelic-php-agent/pull/918/commits/19739b68728113633ddce38771e550718894c6b0)

agent/scripts/newrelic.ini.template
```diff
@@ -1332,3 +1332,86 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
 ;
 ;newrelic.vulnerability_management.package_detection.enabled = true
+
+; Setting: newrelic.security.enabled
+; Type   : boolean
+; Scope  : system
+; Default: false
+; Info   : Indicates if attack detection security module is to be enabled
+;
+;newrelic.security.enabled = true
```
lavarou, line 1342: The comment should match default:
```suggestion
;newrelic.security.enabled = false
```

agent/csec_metadata.c
lavarou: @anmol-ap Please take a look at #922 - it should have what's needed.

agent/scripts/newrelic.ini.template
anmol-ap: Works !

agent/csec_metadata.c
anmol-ap: Yes, the special purpose value for handle worked. Thanks.

agent/csec_metadata.h
```diff
@@ -0,0 +1,12 @@
...
+
+#define KEY_ENTITY_NAME "entity.name"
+#define KEY_ENTITY_TYPE "entity.type"
+#define KEY_ENTITY_GUID "entity.guid"
+#define KEY_HOSTNAME "hostname"
+#define KEY_AGENT_RUN_ID "agent.run.id"
+#define KEY_ACCOUNT_ID "account.id"
+#define KEY_LICENSE "license"
+#define KEY_PLICENSE "plicense"
+#define KEY_HIGH_SECURITY "high_security"
```
lavarou, line 12: Are all these values really required by csec-agent? Why? When does the csec-agent going to use these values?

lavarou: > Yes, the backend (Security Engine) uses these values to correctly map all the information for application, incidents, etc. csec-agent sends this information with the events.

I really doubt that backend will use `plicense` to map anything. Why is this particular value needed?

> If the value changes, as mentioned in the documentation for given method, same impact will be there. But we will not be missing any vulnerabilities due to this.

I do not understand your answer. If csec-agent is going to retrieve this information before `[newrelic_set_appname](https://docs.newrelic.com/docs/apm/agents/php-agent/php-agent-api/newrelic_set_appname/)` is called in PHP code, and send the 'event' to Security Engine after `[newrelic_set_appname](https://docs.newrelic.com/docs/apm/agents/php-agent/php-agent-api/newrelic_set_appname/)` is called, then the 'event' will not be correctly mapped. When does csec-agent retrieve the metadata and when does csec-agent generate the events sent to the backend? This is very important design consideration.

anmol-ap: We are logging some of these values, and that includes the license as well. Logging for license in particular has to be masked, and hence the reason for plicense.

At any point of time, when the apm agent is connected, entity.name is going to have a value along with entity.guid. When the newrelic_set_appname is called, csec-agent will be using that value on next reconnect cycle to the server. So, yes as soon as newrelic_set_appname is called, entity.name and entity.guid will not change for the csec-agent but on the next reconnect (this is planned at SE, the backend), the events will be mapped to new entity.name and entity.guid, this would reset the values for all the running and connected processes.

csec-agent tries to retrieve the metadata at the beginning of each request until it has data. It will wait for next reconnect cycle with server to see if anything has changed.

csec-agent generates the events based on the hooks and send the data immediately to the backend.

As has been detailed the intent of using newrelic_set_appname as early as possible in code, taking runtime values from APM agent would not affect the app's data at backend.

lavarou: > csec-agent tries to retrieve the metadata at the beginning of each request until it has data.

This is exactly why php-agent team has a problem with this. Not only there's potential for incorrect mapping of an event to an app, but also this depends on wether php-agent has been initialized before csec-agent, which we cannot see how it can be guaranteed. Why doesn't csec-agent retrieve the metadata only at the time it is needed, in the 'hook' you mention here:

> csec-agent generates the events based on the hooks and send the data immediately to the backend.

anmol-ap: csec-agent takes the connection information from the apm agent only. It does not generate any data until the apm agent has been connected. There is no way csec-agent can get entity.guid from any source other than APM agent.

anmol-ap: And without entity.guid, it does not connect to the backend

anmol-ap:
@ZNeumann @lavarou @bduranleau-nr 
Could you please review the PRs now. 

- The config has been moved to a separate config file at csec agent.
- `newrelic.security.agent.enabled` and `newrelic.security.enabled` are read in the APM agent for logging purpose.
- Metadata required by csec agent has been exposed by api.
- We are going ahead with separate installation approach as decided.

Related PRs :
[#933 ](https://github.com/newrelic/newrelic-php-agent/pull/933)
[#922 ](https://github.com/newrelic/newrelic-php-agent/pull/922)
[#918 ](https://github.com/newrelic/newrelic-php-agent/pull/933)

anmol-ap:
Rebased the branch.

PR #919, opened 2024-06-18
fix(scripts): fix rhel and alpine init scripts

ZNeumann:
Systemd systems may not have access to the functions file provided by initscripts. The code from initscripts is provided under the GNU v2 license and will not be pulled in.

Additionally, a header comment block must be added, using https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/initscrcomconv.html for specs

### DIFF
--- a/agent/scripts/init.rhel
+++ b/agent/scripts/init.rhel
@@ -240,7 +240,7 @@ EOF
         success
         echo
       else
-        echo " [ OK ]"
+        echo " [  OK  ]"
       fi
     fi
     return 0
--- a/agent/scripts/init.rhel
+++ b/agent/scripts/init.rhel
@@ -250,7 +250,7 @@ EOF
         failure
         echo
       else
-        echo " [ FAILED ]"
+        echo " [FAILED]"
       fi
     fi
     return 1
--- a/agent/scripts/init.rhel
+++ b/agent/scripts/init.rhel
@@ -10,6 +10,15 @@
 # processname: newrelic-daemon
 # config:      /etc/newrelic/newrelic.cfg
 #
+### BEGIN INIT INFO
+# Provides:          newrelic-daemon
+# Required-Start:    $all
+# Required-Stop:
+# Default-Start:     2 3 4 5
+# Default-Stop:	     0 1 6
+# Description: The New Relic Daemon is used by the New Relic PHP Agent to communicate
+#               with the New Relic Backend
+### END INIT INFO
 
 LANG=C
 NAME=newrelic-daemon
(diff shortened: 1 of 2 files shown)


agent/scripts/init.rhel
```diff
@@ -223,14 +236,18 @@ EOF
 
   if running ; then
     if [ -z "${NR_SILENT}" -a -z "${SILENT}" ]; then
-      success
-      echo
+      if [ $has_initd_functions -eq 1 ]; then
+        success
```
ZNeumann, line 240: There is an argument to be had on whether we should log success using our own functions if we do not have access to initd functions. I am leaning towards no. The reason being: [initd functions](https://github.com/fedora-sysv/initscripts/blob/main/etc/rc.d/init.d/functions#L561) have pre-checks and pos-executions that happen that our internal functions would be unable to replicate.

lavarou: I don't think this can stay as is. The init script must send some information back about it succeeding or failing. This is the behavior when `/etc/init.d/functions` file is present:
```
[root@centos9 ~]# /sbin/service newrelic-daemon start
Starting newrelic-daemon:                                  [  OK  ]
[root@centos9 ~]# /sbin/service newrelic-daemon stop
Stopping newrelic-daemon:                                  [  OK  ]
```

This is the behavior when `/etc/init.d/functions` file is not present:
```
[root@centos9 ~]# /sbin/service newrelic-daemon start
Starting newrelic-daemon: [root@centos9 ~]# /sbin/service newrelic-daemon stop
Stopping New Relic Daemon (newrelic-daemon) ... [root@centos9 ~]#
```

Notice that when `/etc/init.d/functions` file is not present, shell prompt appears on the same line as `Starting newrelic-daemon:` message.

ZNeumann: fixed in [f078212](https://github.com/newrelic/newrelic-php-agent/pull/919/commits/f07821222788bed4ee2230986f8d8e03733986b3)

agent/scripts/init.rhel
```diff
@@ -223,14 +236,22 @@ EOF
...
     return 0
   else
     if [ -z "${NR_SILENT}" -a -z "${SILENT}" ]; then
-      failure
-      echo
+      if [ $has_initd_functions -eq 1 ]; then
+        failure
+        echo
+      else
+        echo " [ FAILED ]"
```
lavarou, line 253:
```suggestion
        echo " [FAILED]"
```

agent/scripts/init.rhel
```diff
@@ -223,14 +236,22 @@ EOF
 
   if running ; then
     if [ -z "${NR_SILENT}" -a -z "${SILENT}" ]; then
-      success
-      echo
+      if [ $has_initd_functions -eq 1 ]; then
+        success
+        echo
+      else
+        echo " [ OK ]"
```
lavarou, line 243:
```suggestion
        echo " [  OK  ]"
```

lavarou (approved):
LGTM

PR #921, opened 2024-06-20
tests: enhance `EXPECT_METRICS_EXIST` directive

mfulb:
This PR adds an optional count to the `EXPECT_METRICS_EXIST` test - the format is:

```
/*EXPECT_METRICS_EXIST
<metric name>                                // OLD form - just tests metric is present w/o checking count
<metric name> , <count>               // NEW OPTIONAL form - tests that exactly <count> exist
*/
```
A `count` value of `"??"` is treated as a wildcard for count number.


daemon/internal/newrelic/integration/test.go
```diff
@@ -456,6 +458,122 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
...
+	for _, spec := range strings.Split(strings.TrimSpace(string(t.subEnvVars(t.metricsExist))), "\n") {
+		var err error
+		var count int64
+
+		fields := strings.Split(spec, ",")
+		name := fields[0]
+		name = strings.TrimSpace(name)
+
+		// -1 means just test for existence not an exact count
+		count = -1
```
zsistla, line 471: since count is set to -1 here already, do we need to set it additionally [here](https://github.com/newrelic/newrelic-php-agent/pull/921/files#diff-c7f1af7665ab0865f15e79eeba7e73a7295fe66e0ff9205b9781c7b0d7c75faeR478) and [here](https://github.com/newrelic/newrelic-php-agent/pull/921/files#diff-c7f1af7665ab0865f15e79eeba7e73a7295fe66e0ff9205b9781c7b0d7c75faeR486)?

mfulb: I went back and forth on this - I felt it might help readability to have the alternate condition in those two cases balanced by setting a default value.   The code will work fine without those however and I imagine the optimizer removes that code anyways.

zsistla (dismissed):

daemon/internal/newrelic/integration/test.go
```diff
@@ -456,6 +458,122 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
...
+		}
+
+		// expect array of len 4
+		v2, _ := x1.([]interface{})
+		if len(v2) != 4 {
+			t.Fatal(errors.New("metric event data json doesnt match expected format - expected 4 elements"))
+			return
+		}
+
+		// get array of actual metric from 4rd element
```
lavarou, line 536: Is it 3rd or 4th:
```suggestion
		// get array of actual metric from 4th element
```

lavarou (approved):
@mfulb Thanks for adding this useful feature!

P.S. I'm not a huge fan of long `if/else` type functions, and would prefer `compareMetricsExist` to read something like this:
```
func (t *Test) compareMetricsExist(harvest *newrelic.Harvest) {
	for _, spec := range strings.Split(strings.TrimSpace(string(t.subEnvVars(t.metricsExist))), "\n") {

		name, count := parseMetricsExistSpec(spec)
		if count == -1 {
			if !harvest.Metrics.Has(name) {
				t.Fail(fmt.Errorf("metric %s not found, actual metric table: %s", name, actualPretty.String()))
			}
		} else {
			if !cmpMetricsExistCount(harvest.Metrics, name, count) {
				t.Fail(fmt.Errorf("metric %s count %d mismatch, actual metric table: %s", name, count, actualPretty.String()))
			}
		}
	}
}
```
with all the complexities and `if/else` hidden away in `cmpMetricsExistCount` helper function(s). However, this is just a test tool code, so I won't complain.

zsistla (approved):

lavarou (approved):

daemon/internal/newrelic/integration/test.go
```diff
@@ -456,6 +458,123 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
...
+				// is it "??" - with or without quotes
+				match, _ := regexp.MatchString("^\"*\\?\\?\"*$", countStr)
+				if match {
+					count = -1
+				} else {
+					count, err = strconv.ParseInt(countStr, 10, 64)
+					if nil != err {
+						t.Fail(fmt.Errorf("EXPECT_METRICS_EXIST has unparsable count: %s", spec))
+					}
+				}
```
bduranleau-nr, line 485: Since `count` is already set to `-1` above, it feels like we can optimize this nested if block.
```suggestion
				if !match {
					count, err = strconv.ParseInt(countStr, 10, 64)
					if nil != err {
						t.Fail(fmt.Errorf("EXPECT_METRICS_EXIST has unparsable count: %s", spec))
					}
				}
```
This cuts out the need for the `else` statement while preserving the existing logic.

daemon/internal/newrelic/integration/test.go
```diff
@@ -456,6 +458,123 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
...
+			} else {
+				count = -1
+			}
```
bduranleau-nr, line 488: I think this `else` statement can be removed, since `count` is already initialized to `-1` above this conditional.
```suggestion
			}
```

daemon/internal/newrelic/integration/test.go
```diff
@@ -456,6 +458,123 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
...
+				// compare count
+				actualData := actual[i].([]interface{})[1]
+				actualCount := int64(math.Round(actualData.([]interface{})[0].(float64)))
+
+				metricPasses := true
+				if count == -1 {
+					if actualCount == 0 {
+						metricPasses = false
+					}
+				} else if actualCount != count {
```
bduranleau-nr, line 556: This could be simplified to cut down a nesting layer
```suggestion
				if count == -1 && actualCount == 0 {
					metricPasses = false
				} else if actualCount != count {
```

daemon/internal/newrelic/integration/test.go
```diff
@@ -456,6 +458,123 @@ func (t *Test) compareSpanEventsLike(harvest *newrelic.Harvest) {
...
+					metricPasses = false
+				}
```
bduranleau-nr, line 558: Update to my comment above...

Actually, the logic here feels a bit backwards. Instead of setting `metricPasses := true` and checking the failure cases, the more defensive approach would be to invert that logic and only set to `true` on the passing case.

Something like this:
```suggestion
				metricPasses := false
				if (count == -1 && actualCount > 0) || (actualCount == count) {
					metricPasses = true
				}
```
This way we explicitly narrow down all the "success" cases to the two conditions here.

daemon/internal/newrelic/integration/test.go
bduranleau-nr: Update: See my comment here for an even better way to approach this https://github.com/newrelic/newrelic-php-agent/pull/921/files#r1655462159

bduranleau-nr (approved):

lavarou (approved):

PR #929, opened 2024-06-27
feat(agent): optionally pad agent generated trace_ids

ZNeumann:
W3C compliant headers will have a 32 char trace id. We generate these 32 character ids when we create a header by appending 0's in front of our 16 char trace id.

This PR takes that same concept, but optionally does so for when we store the trace id, not just when we are sending a header.


axiom/nr_distributed_trace.c
```diff
@@ -489,8 +489,8 @@ void nr_distributed_trace_set_trace_id(nr_distributed_trace_t* dt,
   nr_free(dt->trace_id);
   if (trace_id) {
     int len = nr_strlen(trace_id);
-    if (len < 33) {
-      int padding = 33 - len;
+    if (len < 32) {
```
zsistla, line 492: It'd probably be helpful to make this value a DEFINE.

tests/integration/span_events/test_span_events_max_samples_stored5.php
```diff
@@ -5,20 +5,20 @@
  */
 
 /*DESCRIPTION
-This is the max that can be sent with DT only.  With span limit < 7000,
+This is the max that can be sent with DT only.  With span limit > 6800,
```
ZNeumann, line 8: Because the traceid is now larger, the maximum number of spans that we can send to the daemon in a single message is now smaller.

axiom/nr_distributed_trace.c
ZNeumann: done in [c1f11ed](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/c1f11edf870a58b7f87e93c2b0be908b2099f1d5)

lavarou (commented):
Does OTLP collector expect lowercase only letters in the trace id (similar to w3c traceparent header spec)?

axiom/nr_distributed_trace.c
```diff
@@ -488,7 +488,15 @@ void nr_distributed_trace_set_trace_id(nr_distributed_trace_t* dt,
 
   nr_free(dt->trace_id);
   if (trace_id) {
-    dt->trace_id = nr_strdup(trace_id);
+    int len = nr_strlen(trace_id);
+    if (len < NR_TRACE_ID_SIZE) {
+      int padding = NR_TRACE_ID_SIZE - len;
+      char* dest = (char*)(malloc)(NR_TRACE_ID_SIZE + 1);
```
lavarou, line 494: Why not using `nr_malloc` or `nr_zalloc`:
```suggestion
      char* dest = (char*)nr_malloc(NR_TRACE_ID_SIZE + 1);
```
`nr_(m|z)alloc` are guaranteed to succeed or crash. The code in this PR does not check  for `malloc`'s return value.

axiom/nr_distributed_trace.c
```diff
@@ -488,7 +488,15 @@ void nr_distributed_trace_set_trace_id(nr_distributed_trace_t* dt,
...
+      snprintf(dest, NR_TRACE_ID_SIZE+1, "%0*d%s", padding, 0, trace_id);
+      dt->trace_id = dest;
+    } else {
+      dt->trace_id = nr_strdup(trace_id);
+    }
```
lavarou, line 499: What if `trace_id` size is > 32? Is this an error or will OTLP accept such `trace_id`? Maybe it's not part of this PR...

ZNeumann: trace_id should never be more than 32 characters. Inbound header ids will always be 32 and agent created ids will always be the guid (16) before this PR. If somehow we have bigger trace ids, OTLP compatibility is the least of our worries, as we will create invalid W3C headers.

lavarou (commented):
Could tests for `nr_distributed_trace_set_trace_id` be added to `axiom/tests/test_distributed_trace.c`? There exist tests for `nr_txn_create_w3c_traceparent_header` in `axiom/tests/test_txn.c`  which ensures `nr_distributed_trace_create_w3c_traceparent_header` padding works. But there don't exists explicit tests for `nr_distributed_trace_set_trace_id` padding covering 3 cases:
- input is null
- input is < 32
- input is >= 32

axiom/nr_distributed_trace.c
lavarou: > Inbound header ids will always be 32

This assumption is of course verified somewhere, yes? Again - maybe not part of this PR, but 'inbound' and 'will always be' causes my eyebrow to raise 🤨

ZNeumann: It turns out that this is not even the code that handles inbound headers. This function is only called in 1 place, with that 1 place being when we are first creating a txn. The only argument ever passed in for the trace id is txn guid, so hypothetically the size comparison isn't even needed (though still should be included).

ZNeumann:
> Does OTLP collector expect lowercase only letters in the trace id (similar to w3c traceparent header spec)?

All trace_ids that we create within the agent are lowercase, so this should not be important.

axiom/nr_distributed_trace.h
```diff
@@ -12,6 +12,8 @@
 #include "util_time.h"
 #include "util_object.h"
 
+#define NR_TRACE_ID_SIZE 32
```
mfulb, line 15: I don't think this name is clear - what NR trace spec says it should be 32?  This is not a NR related specification.  Perhaps `W3C_TRACE_ID_SIZE`?

axiom/nr_distributed_trace.c
```diff
@@ -481,14 +481,23 @@ void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
 }
 
 void nr_distributed_trace_set_trace_id(nr_distributed_trace_t* dt,
```
mfulb, line 483: I wonder if this interface would be cleaner if the padding was a value passed into the function.  If it is `0` then no padding is done, otherwise it is padded to the value supplied.  As it stands it is baked into the function that only W3C padding is supported.  The caller could decide what type of padding is required and pass in the appropriate value.

axiom/nr_distributed_trace.c
mfulb: It would be nice to know when how we expect code to be exercised is wrong - though a verbosedebug log seems a tad heavy.

axiom/nr_txn.h
```diff
@@ -91,6 +91,8 @@ typedef struct _nrtxnopt_t {
 
   int distributed_tracing_enabled; /* Whether distributed tracing functionality
                                       is enabled */
+  bool distributed_tracing_use_full_trace_id; /* whether to always create a
```
mfulb, line 94: I wonder if these should refer to `W3C` and not `full`.

axiom/nr_distributed_trace.c
ZNeumann: I think that solution is over-engineered for the current problem. We have no reason to believe that there will ever be padding besides 16 prepended 0s.

mfulb:
Could we use `get_linking_metadata()` in some integration tests to verify the `trace.id` is 16 or 32 characters based on the setting of the INI value?

axiom/nr_distributed_trace.h
ZNeumann: > This is not a NR related specification.

Nor is it a W3C specification. The W3C specification is for headers, which we already set to 32 character is a different place in the code. The NR Collector will accept a trace-id up to 32 characters and will prepend 0s to anything less than 32 characters. As such, the NR traceid size is 32.

axiom/nr_txn.h
ZNeumann: see https://github.com/newrelic/newrelic-php-agent/pull/929#discussion_r1681414263

ZNeumann:
> Could we use `get_linking_metadata()` in some integration tests to verify the `trace.id` is 16 or 32 characters based on the setting of the INI value?

added in [5afcf9f](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/5afcf9f3f45075bf7bfa3fcde8711d4e799b3d66)

axiom/nr_distributed_trace.h
mfulb: The purpose of the PR is W3C compatible trace ids so it seems this is a W3C driven decision.

mfulb: Also the default length we create is 16 characters.  Is that not the "NR" trace id length by definition?

ZNeumann: The purpose is OTLP Collector compatible trace ids. And New Relic backend full trace id size is 32. 32 is the full size that our NR Collector expects.

axiom/nr_distributed_trace.c
```diff
@@ -481,14 +481,23 @@ void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
...
+                                       const char* trace_id,
+                                       bool do_padding) {
   if (NULL == dt) {
     return;
   }
 
   nr_free(dt->trace_id);
   if (trace_id) {
-    dt->trace_id = nr_strdup(trace_id);
+    int len = nr_strlen(trace_id);
```
lavarou, line 492: The call to `nr_strlen` is only needed when `do_padding` is `true`. There's no need to increase agent's overhead for default settings of the agent.

axiom/nr_distributed_trace.c
```diff
@@ -481,14 +481,23 @@ void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
...
+    if (do_padding && len < NR_TRACE_ID_SIZE) {
+      int padding = NR_TRACE_ID_SIZE - len;
+      char* dest = (char*)nr_malloc(NR_TRACE_ID_SIZE + 1);
+      snprintf(dest, NR_TRACE_ID_SIZE+1, "%0*d%s", padding, 0, trace_id);
```
lavarou, line 496: For performance reasons `snprintf` should be replaced with manual padding. Manual byte manipulation is almost an order of magnitude more performant as shown by this simple program:
```
$ cat main.c
#include <stdio.h>
#include <string.h>
#include "nr_time.h"

#define NR_TRACE_ID_SIZE 32
#define N_ITER (1000 * 1000)

int main(int argc, char* argv[]) {
	int i;
	nrtime_t start, stop, sprintf_duration, strcpy_duration;
	char buf[NR_TRACE_ID_SIZE+1];
	char *cp;
	char *trace_id = "fedcba9876543210";

       	start = nr_get_time();
	for (i = 0; i < N_ITER; i++) {
		snprintf(buf, NR_TRACE_ID_SIZE+1, "%0*d%s", 16, 0, trace_id);
	}
       	stop = nr_get_time();
	sprintf_duration = stop - start;
	printf("%d x sprintf used " NR_TIME_FMT "us\n", i, sprintf_duration);

       	start = nr_get_time();
	for (i = 0; i < N_ITER; i++) {
		cp = buf;
		for (int p = 0; p < 16; p++) {
			*cp++ = '0';
		}
		strcpy(cp, trace_id);
	}
       	stop = nr_get_time();
	strcpy_duration = stop - start;
	printf("%d x manual used " NR_TIME_FMT "us\n", i, strcpy_duration);
	printf("sprintf vs manual diff " NR_TIME_FMT "us\n", sprintf_duration-strcpy_duration);
$ gcc main.c && ./a.out
1000000 x sprintf used 58901us
1000000 x manual used 10504us
sprintf vs manual diff 48397us
```

ZNeumann: changed in [e2c51c2](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/e2c51c2595c8bd2620da4697790f957b9af4a4b4)

axiom/nr_distributed_trace.c
ZNeumann: changed in [e2c51c2](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/e2c51c2595c8bd2620da4697790f957b9af4a4b4)

axiom/nr_distributed_trace.c
```diff
@@ -481,14 +481,28 @@ void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
...
-    dt->trace_id = nr_strdup(trace_id);
+    if (do_padding) {
+      int len = nr_strlen(trace_id);
+      if (len < NR_TRACE_ID_SIZE) {
+        int padding = NR_TRACE_ID_SIZE - len;
+        char* dest = (char*)nr_malloc(NR_TRACE_ID_SIZE + 1);
+        for (int i=0; i < padding; i++) {
+          dest[i] = '0';
+        }
+        nr_strcpy(dest + padding, trace_id);
```
lavarou, line 500: `nr_strcpy` uses `strlen` (see [here](https://github.com/newrelic/newrelic-php-agent/blob/R10.22/axiom/util_strings.c#L154)) and therefore is less efficient than just `strcpy`. Please use `strcpy`.

axiom/nr_distributed_trace.c
```diff
@@ -481,14 +481,28 @@ void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
...
   if (NULL == dt) {
     return;
   }
 
   nr_free(dt->trace_id);
   if (trace_id) {
-    dt->trace_id = nr_strdup(trace_id);
+    if (do_padding) {
+      int len = nr_strlen(trace_id);
+      if (len < NR_TRACE_ID_SIZE) {
```
lavarou, line 494: If `len` will be equal to `NR_TRACE_ID_SIZE`, `dt->trace_id` will not be set.

ZNeumann: that's embarrassing. Good find, fixed in [48230b7](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/48230b7d98677955c32f66dbf6d022d5e0f6c685)

axiom/nr_distributed_trace.c
ZNeumann: changed in [8c5cf32](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/8c5cf329b62381992220378ad576d58752ed3b5e)

axiom/nr_distributed_trace.h
mfulb: As I understand the nomenclature "NR" is the namespace for this define - so the remaining part is "TRACE_ID_SIZE". Perhaps at a minimum we could have a comment to explain what this size is compared to `NR_GUID_SIZE`([code](https://github.com/newrelic/newrelic-php-agent/blob/6500c76c7515468e61ef634c2a466344bce6cf53/axiom/nr_guid.h#L16)) which also talks about connecting traces.

ZNeumann: added a comment in [215aa0e](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/215aa0e7ed2898a181b124004dd612da490a4e01)

axiom/nr_txn.h
lavarou: The name has been changed in https://github.com/newrelic/newrelic-php-agent/pull/929/commits/cf08ad0cb0f6b761ed4845355d68acd65a865397

agent/scripts/newrelic.ini.private.template
```diff
@@ -193,3 +193,12 @@
 ;          hooks instrumented.
 ;
 ;newrelic.framework.wordpress.hooks_skip_filename=""
+
+; Setting: newrelic.distributed_tracing.pad_trace_id
+; Type   : bool
+; Scope  : system
+; Default: false
+; Info   : Pads the trace id to 32 characters. This will increase
+;          the amount of data sent to and therefore billed by New Relic.
```
lavarou, line 202:
```suggestion
; Info   : If set to true, the internal trace_id generated by the agent will have the length of 32 characters. This will increase
;          the amount of data sent to and therefore billed by New Relic.
```

axiom/nr_distributed_trace.c
```diff
@@ -481,14 +481,30 @@ void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
...
+                                       const char* trace_id,
+                                       bool do_padding) {
   if (NULL == dt) {
     return;
   }
 
   nr_free(dt->trace_id);
   if (trace_id) {
-    dt->trace_id = nr_strdup(trace_id);
+    if (do_padding) {
```
lavarou, line 492:
```suggestion
    if (nrunlikely(do_padding)) {
```

axiom/nr_distributed_trace.c
lavarou: The new code does not change the existing behavior - whatever happened with trace_id > 32 will still happen.

axiom/nr_distributed_trace.c
```diff
@@ -1256,7 +1272,7 @@ char* nr_distributed_trace_create_w3c_traceparent_header(const char* trace_id,
    * 0's.
    */
   tmp = nr_string_to_lowercase(trace_id);
-  padding = 32 - nr_strlen(tmp);
+  padding = NR_TRACE_ID_SIZE - nr_strlen(tmp);
```
lavarou, line 1275: This should really say `NR_W3C_TRACE_ID_SIZE`. It just so happens that the values for `NR_W3C_TRACE_ID_SIZE` is 32 which is the size of the padded internal trace id. But if either of the specs changes, it shouldn't affect the other place.

axiom/nr_distributed_trace.h
```diff
@@ -242,9 +245,12 @@ extern void nr_distributed_trace_set_app_id(nr_distributed_trace_t* dt,
  *
  * Params  : 1. The distributed trace.
  *           2. The trace id.
+ *           3. Bool where true indicates to pad the trace_id
+ *              with 0s to make it 32 characters
```
lavarou, line 249:
```suggestion
 *           3. Bool where true indicates to pad the trace_id
 *              with '0' to make it 32 characters
```

axiom/nr_txn.h
```diff
@@ -91,6 +91,8 @@ typedef struct _nrtxnopt_t {
 
   int distributed_tracing_enabled; /* Whether distributed tracing functionality
                                       is enabled */
+  bool distributed_tracing_pad_trace_id; /* whether to always create a
+                                                 32 character trace id */
```
lavarou, line 95:
```suggestion
  bool distributed_tracing_pad_trace_id; /* whether to pad internally generated trace_id to 32 characters */
```

axiom/nr_distributed_trace.c
lavarou: After rebase the commit id is 68d8d5c6aedcb990d4596f00e412b1cceaa24fdc.

axiom/nr_distributed_trace.h
```diff
@@ -12,6 +12,9 @@
 #include "util_time.h"
 #include "util_object.h"
 
+/* This is the maximum size of trace id that the backend will accept */
+#define NR_TRACE_ID_SIZE 32
```
zsistla, line 16: Better name would be `NR_TRACE_ID_MAX_SIZE` since we know the nr trace id can also be 16.

axiom/nr_txn.h
lavarou: Addressed in ba0de49e8ddc123eb4f56f59bbc29e37e7a6d545.

axiom/nr_distributed_trace.h
lavarou: Addressed in 4f7f2a90f6e3d9c983a2031997099d95c5d22148.

agent/scripts/newrelic.ini.private.template
lavarou: Addressed in 0872a62e692b0f700ce788d9eba30a230487616a.

axiom/nr_distributed_trace.h
lavarou: Addressed in f58c3f9b05fec9ffcc9672e5759b4f76fb926725.

agent/scripts/newrelic.ini.private.template
```diff
@@ -193,3 +193,14 @@
 ;          hooks instrumented.
 ;
 ;newrelic.framework.wordpress.hooks_skip_filename=""
+
+; Setting: newrelic.distributed_tracing.pad_trace_id
+; Type   : bool
+; Scope  : system
```
lavarou, line 199: This says that the new, INI setting has the ‘SYSTEM’ scope: https://github.com/newrelic/newrelic-php-agent/blob/48316ddef44ffb3378a91771adc8fc513ef37bb1/agent/scripts/newrelic.ini.private.template#L199 However the implementation says otherwise: https://github.com/newrelic/newrelic-php-agent/blob/48316ddef44ffb3378a91771adc8fc513ef37bb1/agent/php_nrini.c#L3084 I don’t think this is correct - there shouldn’t be a discrepancy like that.

axiom/tests/test_txn.c
```diff
@@ -5969,7 +5994,7 @@ static void test_create_w3c_traceparent_header(void) {
   /*
    * Test : valid string random guid
    */
-  nr_distributed_trace_set_trace_id(txn.distributed_trace, "meatballs!");
+  nr_distributed_trace_set_trace_id(txn.distributed_trace, "meatballs!", true);
```
lavarou, line 5997: This test should have two versions for two possible values of `nrtxnopt_t.distributed_tracing_pad_trace_id`.

axiom/tests/test_txn.c
```diff
@@ -6725,7 +6750,7 @@ static void test_txn_accept_distributed_trace_payload_w3c(void) {
   TEST_TXN_ACCEPT_DT_PAYLOAD_RESET
   nr_distributed_trace_destroy(&txn.distributed_trace);
   txn.distributed_trace = nr_distributed_trace_create();
-  nr_distributed_trace_set_trace_id(txn.distributed_trace, "35ff77");
+  nr_distributed_trace_set_trace_id(txn.distributed_trace, "35ff77", false);
```
lavarou, line 6753: This test should have two versions for two possible values of `nrtxnopt_t.distributed_tracing_pad_trace_id`.

axiom/tests/test_txn.c
```diff
@@ -7545,7 +7570,7 @@ static void test_get_current_trace_id(void) {
   txn_id = nr_txn_get_guid(txn);
   trace_id = nr_txn_get_current_trace_id(txn);
   tlib_fail_if_null("txn id", txn_id);
-  tlib_pass_if_str_equal("txn_id == trace_id", txn_id, trace_id);
+  tlib_pass_if_str_equal("padded txn_id == trace_id", txn_id, trace_id);
```
lavarou, line 7573: Why has this test message been modified? This test should have two versions: one when txn is started with pad_trace_id set to false and another when pad_trace_id is set to true.

axiom/tests/test_txn.c
lavarou: Further evaluation revealed that it is unimportant what is the value for the third argument of `nr_distributed_trace_set_trace_id` function. See f71fca4a2ba3fdb2f17d5816ef3f9abc63f3cdd7 which provides explanations why.

axiom/tests/test_txn.c
lavarou: Further evaluation revealed that it is unimportant what is the value for the third argument of `nr_distributed_trace_set_trace_id` function. See f71fca4a2ba3fdb2f17d5816ef3f9abc63f3cdd7 which provides explanations why.

axiom/tests/test_txn.c
lavarou: f71fca4a2ba3fdb2f17d5816ef3f9abc63f3cdd7 fixes this unit test by providing better explanations of what is tested and two cases are tested: one with trace_id being padded and one with trace_id not being padded.

lavarou (approved):

agent/scripts/newrelic.ini.private.template
lavarou: Fixed in [f40d8da](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/f40d8da6d36061dab021aab96655c85cf3dd6025)

lavarou:
> Could tests for `nr_distributed_trace_set_trace_id` be added to `axiom/tests/test_distributed_trace.c`? There exist tests for `nr_txn_create_w3c_traceparent_header` in `axiom/tests/test_txn.c` which ensures `nr_distributed_trace_create_w3c_traceparent_header` padding works. But there don't exists explicit tests for `nr_distributed_trace_set_trace_id` padding covering 3 cases:
> 
> * input is null
> * input is < 32
> * input is >= 32

Adressed in [87b7a85](https://github.com/newrelic/newrelic-php-agent/pull/929/commits/87b7a856fb2ff4146ffbe241f033507d4a582bdf)

lavarou:
> https://github.com/newrelic/newrelic-php-agent/pull/929#discussion_r1657754748

This probably needs its own test(s).

zsistla (approved):

mfulb (approved):

PR #932, opened 2024-07-01
feat(agent): Add basic aws-sdk-php intrumentation

zsistla:
1) Detects the use of aws-sdk-php 3 library with a magic file 
2) wraps a function to extract version number
3) Sends a supportability metric for the library and package 
4) Sends PHP Package info for version
5) Multiverse tests were added in support of this

further updated to:
1. added unit tests
2. created a wrapped function that will create a metric for all detected service names. as soon as the client is initialized, the metric is automatically created.

Then the metric is generated as seen in the associated multiverse tests.

Please see the challenge with instrumenting for PHP 8.2+ and various options that were considered below.  
Option 2 is the currently chosen method due to lower risk, complexity, and overhead.

Challenges:
Php8.2+ and composer autoload leads to the main magic file being optimized directly to opcache. 

Options considered:
1) for PHP8.2, and only optimizable libraries, when encountering autoload.php files, ask the file what includes it added and check against only the optimizable library. Small overhead incurred when encountering an autoload file.

summary:
Magic file =     `{"AWS-SDK-PHP", NR_PSTR("aws-sdk-php/src/functions.php"), nr_aws_sdk_php_enable, true},`

Requires bool added to library structure:

```
typedef struct _nr_library_table_t {
  const char* library_name;
  const char* file_to_check;
  size_t file_to_check_len;
  nr_library_enable_fn_t enable;
  bool optimizable;
} nr_library_table_t;
```

```
Additional complexity in `nr_execute_handle_library
#if amber && ZEND_MODULE_API_NO >= ZEND_8_2_X_API_NO
/*
 * Only check if files have been optimized out for PHP 8.2+ and if we know the file is optimizable.
 * It's basically preloading things into the opcache without setting the preload flag that
 * we check to see if we need to check the opcache.
 * In this case, we need to check the PHP list of files that are included to see if our magic
 * file is there. We further limit overhead by only checking when it is an autoload file.
 */
 if ((libraries[i].optimizable) && (nr_striendswith(STR_AND_LEN(filename), "autoload.php", 12)) )  {
      zend_string *zstring_fname;
      
      if (NULL != &EG(included_files)) {
        ZEND_HASH_MAP_FOREACH_STR_KEY(&EG(included_files), zstring_fname) {
              if (nr_striendswith(ZEND_STRING_VALUE(zstring_fname), ZEND_STRING_LEN(zstring_fname) ,
                                  STR_AND_LEN(libraries[i].file_to_check))) {
                nrl_debug(NRL_INSTRUMENT, "detected library=%s",
                          libraries[i].library_name);
                
                nr_fw_support_add_library_supportability_metric(NRPRG(txn), libraries[i].library_name);
                if (NULL != libraries[i].enable) {
                  libraries[i].enable();
                }
                
  
              }  
          
    
        } ZEND_HASH_FOREACH_END();
 }
#endif // PHP 8.2+     
```
`

2) use a file that gets called later and only when AwsClient.php file  file is called.  It's called later and we'll miss some instrumentation, but if we're only ever going to be interested in `Client`  calls anyway, maybe that's ok?  Doesn't detect Sdk.php (optimized out) so when customers only use that or when they use it first, we will not instrument it.  This only detects when a Client is called to use a service so potentially misses out on other instrumentation and misses out when customers use the aws-sdk-php but use non-SDK way to interact with the service (possibly with redis/memcached). This way is definitely the least complex and lowest overhead and less complexity means lower risk as well.

3) Just directly add the wrappers to the hash map.  With potentially 50ish clients to wrap, this will add overhead to every hash map lookup.


agent/fw_hooks.h
```diff
@@ -45,6 +45,7 @@ extern void nr_zend_enable(TSRMLS_D);
 extern void nr_fw_zend2_enable(TSRMLS_D);
 
 /* Libraries. */
+extern void nr_aws_sdk_php_enable(TSRMLS_D);
```
lavarou, line 48:
```suggestion
extern void nr_aws_sdk_php_enable();
```

agent/lib_aws_sdk_php.h
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2024 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Functions relating to instrumenting AWS-SDK-PHP.
+ */
+#ifndef LIB_AWS_SDK_PHP_HDR
+#define LIB_AWS_SDK_PHP_HDR
+
+extern void nr_aws_sdk_php_enable(TSRMLS_D);
```
lavarou, line 10:
```suggestion
extern void nr_aws_sdk_php_enable();
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,59 @@
...
+
+NR_PHP_WRAPPER(nr_aws) {
+  (void)wraprec;
+  NR_PHP_WRAPPER_CALL;
+
+  aws_sdk_php_version();
+}
+NR_PHP_WRAPPER_END
+
+void nr_aws_sdk_php_enable(TSRMLS_D) {
```
lavarou, line 48:
```suggestion
void nr_aws_sdk_php_enable() {
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,59 @@
...
+  /* This is the base of all other Clients */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\AwsClient::__construct"),
+                            nr_aws TSRMLS_CC);
```
lavarou, line 51:
```suggestion
                            nr_aws);
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,59 @@
...
+  /* The Sdk class is where version resides and has other useful features */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\Sdk::__construct"), nr_aws TSRMLS_CC);
```
lavarou, line 53:
```suggestion
  nr_php_wrap_user_function(NR_PSTR("Aws\\Sdk::__construct"), nr_aws);
```

lavarou (commented):
I’m still thinking if both constructors `Aws\\AwsClient::__construct` and `Aws\\Sdk::__construct` need to be instrumented 🤔 Reason why I’m thinking about this is overhead. If only `Aws\\Sdk::__construct` was instrumented, then https://github.com/newrelic/newrelic-php-agent/blob/2c27a8cd99e658414b3d5281c3322154d544021a/agent/lib_aws_sdk_php.c#L25-L27 would be cheap, because `$this`, available through `nr_php_scope_get`, would be of `Aws\Sdk` type and therefore no class lookup (`class_entry = nr_php_find_class("aws\\sdk");`) would be necessary, and `VERSION` could be pulled directly from `$this` - see how Monolog’s API version is retrieved: https://github.com/newrelic/newrelic-php-agent/blob/2c27a8cd99e658414b3d5281c3322154d544021a/agent/lib_monolog.c#L376 

However, I’m not sure if `Aws\\Sdk::__construct` is called when only `Aws\\AwsClient::__construct` is used. Either way - I think using a dedicated wrapper for `Aws\\Sdk::__construct` to retrieve version without class lookup is warranted.

zsistla:
> I’m still thinking if both constructors `Aws\\AwsClient::__construct` and `Aws\\Sdk::__construct` need to be instrumented 🤔 Reason why I’m thinking about this is overhead. If only `Aws\\Sdk::__construct` was instrumented, then
> 
> https://github.com/newrelic/newrelic-php-agent/blob/2c27a8cd99e658414b3d5281c3322154d544021a/agent/lib_aws_sdk_php.c#L25-L27
> 
> would be cheap, because `$this`, available through `nr_php_scope_get`, would be of `Aws\Sdk` type and therefore no class lookup (`class_entry = nr_php_find_class("aws\\sdk");`) would be necessary, and `VERSION` could be pulled directly from `$this` - see how Monolog’s API version is retrieved:
> https://github.com/newrelic/newrelic-php-agent/blob/2c27a8cd99e658414b3d5281c3322154d544021a/agent/lib_monolog.c#L376
> 
> However, I’m not sure if `Aws\\Sdk::__construct` is called when only `Aws\\AwsClient::__construct` is used. Either way - I think using a dedicated wrapper for `Aws\\Sdk::__construct` to retrieve version without class lookup is warranted.

Thanks for the comment.  After further research, `Aws\\Sdk::__construct` is not a good class to generate version info off of for PHP 8.2+ as it does not show up in basic testing.  Additionally, we are currently detecting based off of AwsClient.php so even if aws/sdk::construct did show up (which it didn't) we would ignore it until detection kicked in.  So `Aws\\Sdk::__construct` is undependable and should not be relied on. We only need to grab version info from one spot; otherwise, it's duplicate info and additional overhead.
`Aws\\ClientResolver::_apply_user_agent` is a good choice as it guarantees access to VERSION (although again, even though aws/sdk is used in this function and version is accessed, aws\sdk::construct still doesn't get called).

For additional commentary on this, please see [a36d635](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/a36d635cbabc48c2be756ac0eb3ecfb7682a48bd)
1. added unit tests
2 created to wrapped functions that
a) one creates a supportability metric based off of the class name of the wrapped function that points to it.
b) one generates version info and creates package and version metrics
3. Updated how version info was generated.

agent/lib_aws_sdk_php.c
zsistla: added in another commit [a36d635](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/a36d635cbabc48c2be756ac0eb3ecfb7682a48bd)

agent/lib_aws_sdk_php.c
zsistla: removed in another commit [a36d635](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/a36d635cbabc48c2be756ac0eb3ecfb7682a48bd)

agent/lib_aws_sdk_php.c
zsistla: removed in another commit [a36d635](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/a36d635cbabc48c2be756ac0eb3ecfb7682a48bd)

agent/lib_aws_sdk_php.h
zsistla: removed in another commit [a36d635](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/a36d635cbabc48c2be756ac0eb3ecfb7682a48bd)

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,166 @@
...
+    return;
+  }
+
+  /* Strip out the `\\` characters and normalize */
+  names = nr_strsplit(metric_name, "\\", 0 /* discard empty */);
+
+  buf[0] = '\0';
+  snprintf(buf, sizeof(buf), "%s%s", PHP_AWS_CLASS_PREFIX,
+           nro_get_array_string(names, 1, NULL));
+  for (int i = 2, n = nro_getsize(names); i <= n; i++) {
```
ZNeumann, line 84:
```suggestion
  for (int i = 2, n = nro_getsize(names); i < n; i++) {
```
Maybe I am misunderstanding the loop, but I think it should be this?

ZNeumann: Ah I am misunderstanding.

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,166 @@
...
+    const char* name = nro_get_array_string(names, i, NULL);
+    if (NULL != name) {
+      snprintf(buf, sizeof(buf), "%s_%s", buf, name);
```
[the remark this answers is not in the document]
zsistla: Of course!  Good catch, thanks!
how about: [d5834c6](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/d5834c6ddd5892a6562388c0a4cf4544987cd3be)
?

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,166 @@
...
+ * AwsClient.php file, this method will always be called when a client is
+ * created. Having separate functionality to extract from Aws/Sdk::__constructor
+ * is both not required and is redundant and causes additional overhead.
+ */
+extern void lib_aws_sdk_php_handle_version() {
+  char* version = NULL;
+  zval retval;
+  int result = FAILURE;
+
+  result = zend_eval_string("Aws\\Sdk::VERSION;", &retval, "Get AWS Version");
```
mfulb, line 48: Can't we have some guard code that verifies the class exists before referencing it using `class_exists()`?

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,166 @@
...
+  nrobj_t* names = NULL;
+
+  if (NULL == metric_name) {
+    return;
+  }
+  if (NULL == NRPRG(txn)) {
+    return;
+  }
+
+  /* Strip out the `\\` characters and normalize */
```
mfulb, line 80: Could you give an example of `metric_name` and the corresponding value of `buf` as I'm not sure what this operation is supposed to do.

agent/lib_aws_sdk_php.c
zsistla: We could, which would slightly add to the overhead, but the advantage of calling it from ClientResolver::_apply_user_agent is that it is guaranteed to already exist so it saves slightly on overhead.
We calculate this value in func end, so you see the [wrapped function ](https://new-relic.atlassian.net/browse/NR-287951)has already made the call to sdk::version, so if it didn't exist and were to fail, it would do so before it got to us.

agent/lib_aws_sdk_php.c
zsistla: metric name shows up as:
`Supportability/library/aws/aws-sdk-php/Aws_S3_S3Client`

without stripping the \\ chars, it would show up as:
`Supportability/library/aws/aws-sdk-php/Aws\\S3\\S3Client`

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -0,0 +1,176 @@
...
+  /*
+   * Aws/Sdk exists. Should return aws package metric with
+   * version
+   */
+  tlib_php_request_start();
+  declare_aws_sdk_class("Aws", "Sdk", LIBRARY_MAJOR_VERSION);
+  lib_aws_sdk_php_handle_version();
+  tlib_pass_if_not_null("version test 1: package metric created",
+                        nrm_find(NRPRG(txn)->unscoped_metrics, PACKAGE_METRIC
+                                 "/" LIBRARY_MAJOR_VERSION "/detected"));
```
mfulb, line 102: It is a matter of style - but if the test case input is represented by a `#define` it seems like it would be a good idea for the expected result would also have a similarly named `#define` so if the case is modified it is clear how to handle input and expected result.

agent/lib_aws_sdk_php.c
zsistla: Delimiter could be anything, but the `\\` didn't look good.

agent/tests/test_lib_aws_sdk_php.c
zsistla: Can you explain a little more? I'm not quite catching the meaning since the define is already used in the input and output, similar to what was done for the [fw_support tests](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/tests/test_fw_support.c#L74).
The input has a define [here](https://github.com/newrelic/newrelic-php-agent/pull/932/files/0e4f2819dc2fb21bca1867804c09c6a213f10fc0#diff-5cbabeb350f68acfaa71e353ed2387f1d2aa71b7b03201d820a283c3810e3aa2R98), and the output has the define [here](https://github.com/newrelic/newrelic-php-agent/pull/932/files/0e4f2819dc2fb21bca1867804c09c6a213f10fc0#diff-5cbabeb350f68acfaa71e353ed2387f1d2aa71b7b03201d820a283c3810e3aa2R102)

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,170 @@
...
+  buf[0] = '\0';
+
+  int cur_len = snprintf(buf, sizeof(buf), "%s%s", PHP_AWS_CLASS_PREFIX,
+                         nro_get_array_string(names, 1, NULL));
+  for (int i = 2, n = nro_getsize(names); i <= n; i++) {
+    const char* name = nro_get_array_string(names, i, NULL);
+    if (NULL != name) {
+      cur_len
+          += snprintf(buf + cur_len, MAX_LEN - cur_len, "%s%s", DELIM, name);
+    }
```
bduranleau-nr, line 92: Can we break out early here if `cur_len >= MAX_LEN`?

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,170 @@
...
+  char buf[MAX_LEN];
+  nrobj_t* names = NULL;
+  char* DELIM = "_";
+
+  if (NULL == metric_name) {
+    return;
+  }
+  if (NULL == NRPRG(txn)) {
+    return;
+  }
```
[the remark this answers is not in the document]
zsistla: We check it for NULL in line 73 and nr_strsplit handles checking for an empty string.

agent/lib_aws_sdk_php.c
mfulb: I guess as defense against an older or newer version of the sdk where those assumptions dont hold.  Otherwise there will be a surprise for customers.

agent/lib_aws_sdk_php.c
zsistla: good idea, please see: [3b34325](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/3b343255641d44ec4a4b6c5bab37499052b497f0)

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,170 @@
...
+  /* Use the Aws\Sdk::VERSION to determine the version */
+  if (NRINI(vulnerability_management_package_detection_enabled)) {
+    /* Add php package to transaction */
+    nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
+  }
+  nr_fw_support_add_package_supportability_metric(NRPRG(txn), PHP_PACKAGE_NAME,
+                                                  version);
+}
+
+extern void lib_aws_sdk_php_add_supportability_metric(const char* metric_name) {
```
ZNeumann, line 67: I am concerned about the performance of this function. It seemingly copies the name twice. Once into the object array during the split, and then once into the buffer. And this is executed on every single call to specific functions.

agent/lib_aws_sdk_php.c
zsistla: The older version (2.x) of the sdk is from 2016, so we won't support that anyway.
This function is dependable and will always get that info (it's what they use to build their headers for communication) so is likely to always be guaranteed to retrieve that information.

It does save us in overhead, but if you feel strongly that it outweighs the overhead gain I can add it.

agent/lib_aws_sdk_php.c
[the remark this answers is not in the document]
bduranleau-nr: > nr_strsplit handles checking for an empty string.

True, but it doesn't return if the string is empty. It looks like we could exit early on an empty string via something like
```suggestion
  if (NULL == metric_name || '\0' == metric_name[0]) {
    return;
  }
  if (NULL == NRPRG(txn)) {
    return;
  }
```

agent/lib_aws_sdk_php.c
zsistla: It operates only on calls to the Client's constructor, so theoretically shouldn't happen often, but optimization is always the goal.

options are:

1) We could either accept the `\\` in the name
2) instead of strplit we could iterate through the string and manually remove/replace.

Checkout commit for option 2 here: [7d774ef](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d774ef904c4da10b012c89cc46fce6041dc8d21)

agent/lib_aws_sdk_php.c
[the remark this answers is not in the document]
zsistla: made the change with some other string manipulation optimizations here: [7d774ef](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d774ef904c4da10b012c89cc46fce6041dc8d21)

agent/lib_aws_sdk_php.c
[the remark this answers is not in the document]
zsistla: This is out of date now, string manipulation is entirely rewritten: [7d774ef](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d774ef904c4da10b012c89cc46fce6041dc8d21)

bduranleau-nr (approved):

agent/lib_aws_sdk_php.c
zsistla: To future/past proof as well as harden against this being used to wrap against any other function besides the special one with assumed guarantees, the version extraction is reverted to [original](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/2c27a8cd99e658414b3d5281c3322154d544021a#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R21-R37).

Please see [10d00a4](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/10d00a4fbc4d339f55d36224a731cebbe3737272)

agent/lib_aws_sdk_php.c
ZNeumann: I like option 2. Maybe we could extract the code into an `nr_strreplace` function?

ZNeumann: Something else interesting from the commit above: The modification is in-place on the `wraprec->classname` field. If we determine that this doesn't have negative side-effects, this string modification would only need to happen once during wrapper creation. If, however, there are side effects, we probably want to do a single copy each time this is called.

zsistla: Modifying in place wouldn't work as we use that field for other things which was why manipulation was only happening on the copied version in the buffer.

After considering and testing more, I think I overthought this and an solving a problem that doesn't exist.
The metric naming details in agent-specs don't have anything against backslashes, and the backend properly shows it as `Supportability/library/aws/aws-sdk-php/Aws\S3\S3Client`

Leaving it as-is (without replacing the backslash) seems the most efficient.

After experimentation, option one seems better in this case.  It let's us simplify code readability,  reduce string copy/manipulation, and minimize the effect of enabling supportability metrics.

image

Please see [1148df3](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/1148df3439920382b561b616ba412a7b3a023c08)

agent/tests/test_lib_aws_sdk_php.c
zsistla: Please see tests in  [1148df3](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/1148df3439920382b561b616ba412a7b3a023c08) 

Is that more along the lines of what you meant?

ZNeumann (dismissed):

mfulb (approved):

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,165 @@
...
+#include "php_agent.h"
+#include "php_wrapper.h"
+#include "fw_hooks.h"
+#include "fw_support.h"
+#include "util_logging.h"
+#include "php_call.h"
+#include "lib_aws_sdk_php.h"
+
+#define PHP_PACKAGE_NAME "aws/aws-sdk-php"
+#define PHP_AWS_CLASS_PREFIX "Supportability/library/aws/aws-sdk-php/"
```
lavarou, line 19: The name of this macro is very misleading - it is not a class prefix but a metrics prefix. I'm not sure what the metrics that use this prefix are supposed to capture/measure, but it looks like it not the same thing as other metrics that use `Supportability/library` prefix, which is used in [`nr_fw_support_add_library_supportability_metric`](https://github.com/newrelic/newrelic-php-agent/blob/R10.22/agent/fw_support.c#L33-L43) to indicate that a library was detected [here](https://github.com/newrelic/newrelic-php-agent/blob/R10.22/agent/php_execute.c#L921-L922). The use of this macro suggests that the intention of this metric is to capture/measure what classes/aws services? are instrumented/detected?. There already exist `Supportability/InstrumentedFunction` family of metrics with an [API to create them](https://github.com/newrelic/newrelic-php-agent/blob/R10.22/axiom/nr_txn.c#L1641-L1646) that capture/measure what functions are instrumented. If capturing/measuring what classes are instrumented is important, why not follow the same naming pattern and use the prefix `Supportability/InstrumentedClass`? But it really seems duplicated info because there are already created `Supportability/InstrumentedFunction` for all instrumented classes. If capturing/measuring what AWS services are detected is important, then `Supportability/Aws/Services/{serviceName/service client class name?}/PHP/detected` seems more appropriate as the metric name pattern and `AWS_SERVICE_DETECTED_SUPPORTABILITY_METRIC_FMT` as macro name.
```suggestion
#define AWS_SERVICE_DETECTED_SUPPORTABILITY_METRIC_FMT "Supportability/AWS/Services/%s/PHP/detected"
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * Concerns about future/past proofing to the checking prioritized the following
+ * implementation vs using the eval method.
+ */
+extern void lib_aws_sdk_php_handle_version() {
+  zval* zval_version = NULL;
+  zend_class_entry* class_entry = NULL;
+  char* version = NULL;
+
+  class_entry = nr_php_find_class("aws\\sdk");
```
lavarou, line 60: On PHP 7.4, when using this sample code to construct `Aws\S3\S3Client`:
```
$s3Client = new Aws\S3\S3Client([
    'profile' => 'default',
    'region' => 'us-west-2'
]);
```
`class_entry` returned here is `NULL`. It is not `NULL` on PHP 8.2. I did not test with other PHPs.

lavarou: Further investigation reveals that on PHP 7.4 `aws/aws-sdk-php/src/Sdk.php` file with the definition of `\Aws\Sdk::Version` constant is loaded **after** `Aws\ClientResolver::_apply_user_agent` executes and that would explain why `nr_php_find_class("aws\\sdk");` returns `NULL`.

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,165 @@
...
+ * so only one function is needed to extract version.
+ *
+ * Aws\\ClientResolver::_apply_user_agent a reliable function as it is
+ * always called on client initialization since it is key to populating
+ * the request headers, and it loads Sdk by default.
+ *
+ * Concerns about future/past proofing to the checking prioritized the following
+ * implementation vs using the eval method.
+ */
+extern void lib_aws_sdk_php_handle_version() {
```
lavarou, line 55: `extern` keyword is not needed when defining the function:
```suggestion
void lib_aws_sdk_php_handle_version() {
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,165 @@
...
+  if (NRINI(vulnerability_management_package_detection_enabled)) {
+    /* Add php package to transaction */
+    nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
+  }
+  nr_fw_support_add_package_supportability_metric(NRPRG(txn), PHP_PACKAGE_NAME,
+                                                  version);
+  nr_php_zval_free(&zval_version);
+}
+
+extern void lib_aws_sdk_php_add_supportability_metric(const char* metric_name) {
```
lavarou, line 77: extern keyword is not needed when defining the function:
```suggestion
void lib_aws_sdk_php_add_supportability_metric(const char* metric_name) {
```

agent/lib_aws_sdk_php.h
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Functions relating to instrumenting AWS-SDK-PHP.
+ */
+#ifndef LIB_AWS_SDK_PHP_HDR
+#define LIB_AWS_SDK_PHP_HDR
+
+extern void nr_aws_sdk_php_enable();
+extern void lib_aws_sdk_php_handle_version();
+extern void lib_aws_sdk_php_add_supportability_metric(const char* metric_name);
```
lavarou, line 12: Since C doesn't have the concept of namespaces, in order to avoid potential symbol symbol clashes, it is good practice to include 'namespace' as a prefix in all external functions. Therefore these two functions should be prefixed with `nr_` - a conventional namespace for all global symbols in newrelic-php-agent:
```suggestion
extern void nr_lib_aws_sdk_php_handle_version();
extern void nr_lib_aws_sdk_php_add_supportability_metric(const char* metric_name);
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,165 @@
...
+void nr_aws_sdk_php_enable() {
+  /* This is will guarantee we can extract the version. */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\ClientResolver::_apply_user_agent"),
+                            nr_aws_version);
+  /* This is the base of all other Clients */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\AwsClient::__construct"),
+                            nr_aws_create_metric);
+  /* The Sdk class */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\Sdk::__construct"),
+                            nr_aws_create_metric);
```
lavarou, line 156: Why these two methods need to be wrapped? I thought we're interested in generating metrics for:
- aws-sdk-php@v3 package to understand usage of that package
- AWS services used by PHP applications to understand usage of AWS services by PHP applications

First objective is achieved (mostly - see [this comment](https://github.com/newrelic/newrelic-php-agent/pull/932/files#r1681801550)) with wrapping `Aws\\ClientResolver::_apply_user_agent` - `nr_aws_version` generates `Supportability/PHP/package/aws/aws-sdk-php/3/detected` metric.

Implementation of the second objective for S3 service is demonstrated with wrapping `Aws\\S3\\S3Client::__construct` - `nr_aws_create_metric` tries to generate this metric (however it would benefit from [my other suggestion](https://github.com/newrelic/newrelic-php-agent/pull/932/files#r1681470624)). What value does wrapping `Aws\AwsClient::__construct` and `Aws\Sdk::__construct` give? The metric generated by `nr_aws_create_metric` does not provide the name of AWS service - just the name of the class (`Aws\AwsClient` and `Aws\Sdk`).

agent/lib_aws_sdk_php.c
lavarou: There really is no convention about where the language should be in the supportability metric name. Some have language right after Supportablity prefix, others have them almost at the end. Because of that I don't have a strong preference one way or the other. `Supportability/AWS/Services/%s/PHP/detected` is as good as `Supportability/PHP/AWS/Services/%s/detected` to me.

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,165 @@
...
+
+  nr_php_wrap_user_function(NR_PSTR("Aws\\S3\\S3Client::__construct"),
+                            nr_aws_create_metric);
```
lavarou, line 159: Use `_extra` variant of `nr_php_wrap_user_function` to store service name in `wraprec` and use it in the metric name instead of class name:
```suggestion
nr_php_wrap_user_function_extra(NR_PSTR("Aws\\S3\\S3Client::__construct"),
                            nr_aws_create_metric, "S3");
```
This gives more control over how long the metric name can be and therefore allows for more predictable results of `lib_aws_sdk_php_add_supportability_metric` (it will less likely truncate anything). With this change, the implementation of `nr_aws_create_metric`  should be changed to:
```
NR_PHP_WRAPPER(nr_aws_create_metric) {
  if (NULL != wraprec && NULL != wraprec->extra) {
    lib_aws_sdk_php_add_supportability_metric(wraprec->extra);
  }

  NR_PHP_WRAPPER_CALL;
}
NR_PHP_WRAPPER_END
```

agent/lib_aws_sdk_php.c
zsistla: Please see: [7d98b70](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d98b70ca2e9ada5dc79739fefa63e535ae6220d)

Instrumented function mostly did the job, but would require wrapping 50ish Aws Clients variations (and would need to determine on case by case basis what to use, for instance, dynamodbclient::__construct does not show up for PHP 8.2+).

Now we wrap the function that the aws-sdk-php uses to determine the service whenever any client is created and create a metric based off the name we snag from that.  This automatically will create metrics for all clients.

metric prefix is `Supportability/PHP/AWS/Services/%s/`
(`detected` is not required. it's only needed for frameworks to differentiate with `forced`).

agent/lib_aws_sdk_php.c
zsistla: Please see: [7d98b70](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d98b70ca2e9ada5dc79739fefa63e535ae6220d)

agent/lib_aws_sdk_php.c
zsistla: Please see: [7d98b70](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d98b70ca2e9ada5dc79739fefa63e535ae6220d)

agent/lib_aws_sdk_php.h
zsistla: Please see: [7d98b70](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d98b70ca2e9ada5dc79739fefa63e535ae6220d)

agent/lib_aws_sdk_php.c
zsistla: Please see: [7d98b70](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d98b70ca2e9ada5dc79739fefa63e535ae6220d)

Both are removed as unneeded.

agent/lib_aws_sdk_php.c
zsistla: Please see: [7d98b70](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/7d98b70ca2e9ada5dc79739fefa63e535ae6220d)

Instrumented function mostly did the job, but would require wrapping 50ish Aws Clients variations (and would need to determine on case by case basis what to use, for instance, dynamodbclient::__construct does not show up for PHP 8.2+).

Now we wrap the function that the aws-sdk-php uses to determine the service whenever any client is created and create a metric based off the name we snag from that. This automatically will create metrics for all clients.

metric prefix is Supportability/PHP/AWS/Services/%s/
(detected is not required. it's only needed for frameworks to differentiate with forced).

agent/lib_aws_sdk_php.c
zsistla: Great catch, please see: [c27a257](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/c27a25719af98a9ba6e17638e33636b08cc9f670)


NR_PHP_WRAPPER_CALL needs to happen before checking the version number for PHP 7.x.
Since Aws\ClientResolver::_apply_user_agent uses Sdk::VERSION it should be loaded by the time it finishes.

Also updated the multiverse test_php_packages.php test to only call the client constructor and to have no reference or use of the Aws/Sdk class.

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+
+NR_PHP_WRAPPER(nr_aws_version) {
+  (void)wraprec;
+  NR_PHP_WRAPPER_CALL;
+  nr_lib_aws_sdk_php_handle_version();
+}
+NR_PHP_WRAPPER_END
+
+/*
+ * AwsClient::parseClass
```
lavarou, line 105: Leaving this for posterity: https://github.com/aws/aws-sdk-php/blob/3.316.3/src/AwsClient.php#L373-L387

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+ * all calls including aws\sdk calls are ignored.  Version detection will be
+ * tied to Aws/ClientResolver::_apply_user_agent which is ALWAYS called when
+ * dealing with aws clients.  It will not be computed from
+ * Aws/Sdk::__constructor which would at best be duplicate info and worst would
+ * never be ignored until a client is called.
+ */
+void nr_aws_sdk_php_enable() {
+  /* This will be used to extract the version. */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\ClientResolver::_apply_user_agent"),
+                            nr_aws_version);
```
lavarou, line 168: The name of this wrapper function is not clear enough. I really prefer the code to the talking:
```suggestion
  /* This will be used to extract the version. */
  nr_php_wrap_user_function(NR_PSTR("Aws\\ClientResolver::_apply_user_agent"),
                            nr_create_aws_sdk_version_metrics);
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+  /* Called when initializing all other Clients */
+  nr_php_wrap_user_function(NR_PSTR("Aws\\AwsClient::parseClass"),
+                            nr_aws_client_parse_class);
```
lavarou, line 171: Same comment as above - yes it is called when initializing all other Clients but what do we wrap it for? To create supportability metrics:
```suggestion
  /* Called when initializing all other Clients */
  nr_php_wrap_user_function(NR_PSTR("Aws\\AwsClient::parseClass"),
                            nr_create_aws_service_metric);
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+}
+
+void nr_lib_aws_sdk_php_add_supportability_service_metric(
+    const char* service_name) {
+  int MAX_LEN = 512;
+  char buf[MAX_LEN];
+
+  if (NULL == service_name || '\0' == service_name[0]) {
+    return;
+  }
```
lavarou, line 86: Let's not duplicate existing logic:
```suggestion
  if (nr_strempty(service_name)) {
    return;
  }
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+  if (NULL == NRPRG(txn)) {
+    return;
+  }
+  buf[0] = '\0';
+
+  snprintf(buf, MAX_LEN, "%s%s", PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX,
+           service_name);
```
lavarou, line 93: For efficiency reasons I suggest using strcpy/strcat instead of formatting function. Something like this should do the trick and will explicitly tell what is the maximum length of the AWS metric the agent will accept and report vs now I have to manually subtract length of `PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX` from 512:
```suggestion
#define AWS_SERVICE_NAME_LEN_MAX 64 // maximum name of aws service arbitrarily selected 64
  char *buf = nr_alloca(sizeof(PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX) + SERVICE_NAME_LEN_MAX);
  char *cp = buf;
  cp += nr_strcpy(cp, PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX);
  cp += nr_strlcpy(cp, service_name, SERVICE_NAME_LEN_MAX);
```

zsistla: This is is similar to how some other metrics are created and has the advantage of automatically truncating for us.

Using str_cpy really doesn't seem more efficient.
Calling [nr_strcpy](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/util_strings.c#L152) adds multiple checks and call to strlen.
Then [nr_strlcpy](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/util_strings.c#L167) again does more checks and strlen function call before doing a memcopy.

Plus the size of the prefix and name need to be tracked.

Seems like other metric name generators in agent use `nr_formatf` but that creates a string that needs to be freed or snprint like [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_api.c#L54).
snprintf seems to do what we want in this case.

lavarou: 🤦 I always forget that `nr_` versions of `strcpy` have extra overhead because they call `strlen`. However, raw `strncpy` is an order of magnitude faster than `sprintf` as demonstrated by this simple program:
```
$ cat main.c
#include <stdio.h>
#include <string.h>
#include "nr_time.h"

#define METRIC_NAME_PREFIX "Supportability/Aws/Service/"

int main(int argc, char* argv[]) {
	int i;
	nrtime_t start, stop, sprintf_duration, strcpy_duration;
	int MAX_LEN = 512;
	char buf[MAX_LEN];
	char *cp;
	char *service_name = "DynamoDb";

       	start = nr_get_time();
	for (i = 0; i < 1000 * 1000; i++) {
		buf[0] = '\0';
		snprintf(buf, MAX_LEN, "%s%s", METRIC_NAME_PREFIX, service_name);
	}
       	stop = nr_get_time();
	sprintf_duration = stop - start;
	printf("%d x sprintf used " NR_TIME_FMT "us\n", i, sprintf_duration);

       	start = nr_get_time();
	for (i = 0; i < 1000 * 1000; i++) {
		cp = buf;
		strcpy(cp, METRIC_NAME_PREFIX);
		cp += sizeof(METRIC_NAME_PREFIX) - 1;
		strncpy(cp, service_name, 64);
	}
       	stop = nr_get_time();
	strcpy_duration = stop - start;
	printf("%d x strcpy used " NR_TIME_FMT "us\n", i, strcpy_duration);
	printf("sprintf vs strcpy diff " NR_TIME_FMT "us\n", sprintf_duration-strcpy_duration);
}
$ gcc main.c && ./a.out
1000000 x sprintf used 48124us
1000000 x strcpy used 5195us
sprintf vs strcpy diff 42929us
```
We should really consider these types of optimizations whenever we have a chance. I strongly suggest optimizing this piece of code. Just because something has been done in the past doesn't mean it needs to be repeated.

lavarou: The reason why `sprintf` is so much slower is that it has to parse the format string and the extract the correct argument from var arg list and cast it to correct type and then copy the memory from one place to another applying correct format. `sprintf` is useful under some circumstances, but this is a very simple case, where the whole machinery of `sprintf` is not needed.

zsistla: Good points.  
Please see: [09e54ad](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/09e54adffaa12b2e422122439c6b678dbbccf230)

lavarou: This is better, but as you said earlier `nr_str*` wrappers around `str` functions use `strlen`. This is an inefficiency and can be avoided here. Please replace `nr_strcpy` with raw (not `nr_` wrapped) `strcpy` and compute new value for `cp` after `strcpy` call. The sample program demonstrates this.

zsistla: please see [4c7896f](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/4c7896f506a833099c4fe7b6f6dbd5ab3873dd8e)

lavarou: It's better but still can be improved. There's no need for this computation (especially call to `nr_strlen`) at runtime: `MAX_METRIC_NAME_LEN - nr_strlen(PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX)` on this line:
```
  strlcpy(
      cp, service_name,
      MAX_METRIC_NAME_LEN - nr_strlen(PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX));
```
This can be pre-computed at compile time:
```
#define MAX_METRIC_NAME_LEN 256
#define MAX_AWS_SERVICE_NAME_LEN (MAX_METRIC_NAME_LEN - sizeof(PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX))
...
strlcpy(cp, service_name, MAX_AWS_SERVICE_NAME_LEN);
```

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -0,0 +1,180 @@
...
+#include "php_agent.h"
+#include "lib_aws_sdk_php.h"
+#include "fw_support.h"
+
+tlib_parallel_info_t parallel_info
+    = {.suggested_nthreads = -1, .state_size = 0};
+
+#if ZEND_MODULE_API_NO > ZEND_7_1_X_API_NO
+
+#define PHP_AWS_CLASS_PREFIX "Supportability/library/aws/aws-sdk-php/"
```
lavarou, line 17:
```suggestion
```

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -0,0 +1,180 @@
...
+#define PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX \
+  "Supportability/PHP/AWS/Services/"
```
lavarou, line 19: To avoid code duplication, this macro should be defined in `lib_aws_sdk_php.h` and used in both modules: `lib_aws_sdk_php.c` and `test_lib_aws_sdk_php.c`.

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,183 @@
...
+  }
+  nr_fw_support_add_package_supportability_metric(NRPRG(txn), PHP_PACKAGE_NAME,
+                                                  version);
+  nr_php_zval_free(&zval_version);
+}
+
+void nr_lib_aws_sdk_php_add_supportability_service_metric(
+    const char* service_name) {
+  /* total MAX metric name length per agent-specs */
+  int MAX_METRIC_NAME_LEN = 256;
```
lavarou, line 82: This really should be a macro to avoid allocation on stack:
```suggestion
#define MAX_METRIC_NAME_LEN 256
```

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,183 @@
...
+  char* cp = NULL;
+
+  if (nr_strempty(service_name)) {
+    return;
+  }
+  if (NULL == NRPRG(txn)) {
+    return;
+  }
+
+  buf[0] = '\0';
```
lavarou, line 93: This is redundant - subsequent `strcpy` copies a constant string and will fill `buf[0]` as well as null terminate the string:
```suggestion
```

agent/lib_aws_sdk_php.c
zsistla: please see: [842d437](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/842d4377f3be22d2feca2dd1f32e429ff4bd6a40)

agent/lib_aws_sdk_php.c
zsistla: please see [842d437](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/842d4377f3be22d2feca2dd1f32e429ff4bd6a40)

agent/tests/test_lib_aws_sdk_php.c
zsistla: please see [842d437](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/842d4377f3be22d2feca2dd1f32e429ff4bd6a40)

agent/tests/test_lib_aws_sdk_php.c
zsistla: please see

agent/lib_aws_sdk_php.c
zsistla: nice suggestion!  please see [842d437](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/842d4377f3be22d2feca2dd1f32e429ff4bd6a40)

agent/tests/test_lib_aws_sdk_php.c
lavarou: Resolved in [842d437](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/842d4377f3be22d2feca2dd1f32e429ff4bd6a40)

agent/tests/test_lib_aws_sdk_php.c
lavarou: It's still there...

agent/lib_aws_sdk_php.h
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Functions relating to instrumenting AWS-SDK-PHP.
+ */
+#ifndef LIB_AWS_SDK_PHP_HDR
+#define LIB_AWS_SDK_PHP_HDR
+
+#define PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX \
+  "Supportability/PHP/AWS/Services/"
+#define MAX_METRIC_NAME_LEN 256
```
mfulb, line 12: Is this for any metric sent? Perhaps it would be better in `axiom/util_metrics.h`?

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+
+  if (nr_strempty(service_name)) {
+    return;
+  }
+  if (NULL == NRPRG(txn)) {
+    return;
+  }
+
+  cp = buf;
+  strcpy(cp, PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX);
```
mfulb, line 91: To be safest this ought to be a `strncpy` with `sizeof(PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX`)` as the max len.  Although unlikely, it could be that the string being added (through user error)  is longer than `MAX_METRIC_NAME_LEN` and this would stop a potential SEGV.

agent/lib_aws_sdk_php.c
```diff
@@ -0,0 +1,177 @@
...
+  cp += PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX_LEN - 1;
+  strlcpy(cp, service_name, MAX_AWS_SERVICE_NAME_LEN);
```
mfulb, line 93: I believe `strlcat()` is intended to handle this scenario without having to do the pointer math on `cp`.

agent/lib_aws_sdk_php.h
zsistla: The specs say it, but in looking at code, it's not explicitly handled elsewhere, so in order not to impact code elsewhere, it's only used here.

agent/lib_aws_sdk_php.c
mfulb: Upon review I see this is probably done for optimization.

agent/lib_aws_sdk_php.c
zsistla: This is only copying `#define PHP_AWS_SDK_SERVICE_NAME_METRIC_PREFIX \
  "Supportability/PHP/AWS/Services/"`

The extra handling doesn't seem needed unless it's possible that at some point we will change the prefix to be much much bigger.

mfulb: It was suggested to me an `assert()` here to verify the expectation that this string would fit would catch the failure case if it ever happened when testing the agent and not affect production use.

agent/lib_aws_sdk_php.c
zsistla: but then woulnd't strlcat have to determine the str length to know where to concat it?  This seems like it would be slightly more performant.

agent/tests/test_lib_aws_sdk_php.c
zsistla: it's either a ghost....

or i forgot to add the file to the earlier commit.
please see [41dfe38](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/41dfe38c47302107512a6c15f40edab91a9b7f2c)

agent/lib_aws_sdk_php.c
lavarou: > Upon review I see this is probably done for optimization.
Correct. I concur with @zsistla.

agent/tests/test_lib_aws_sdk_php.c
lavarou: It's gone now! :) 👍

agent/lib_aws_sdk_php.c
zsistla: I'm not sure this extra case is necessary.

[5005aaf](https://github.com/newrelic/newrelic-php-agent/pull/932/commits/5005aafeda575456a6158645a07d87c48bfa4e42)

If it were ever accidentally too big because we decided to make the prefix define over the limit, the test cases would fail/segfault.

lavarou: >I'm not sure this extra case is necessary because if it were ever accidentally too big because we decided to make the prefix define over the limit, the unit/multiverse test cases would fail/segfault.

I concur with the above statement. Also 5005aaf is not the `assert` that Michael was talking about. Regardless, I don't see the need for this extra check in this code.

lavarou (approved):
Nice job!

agent/lib_aws_sdk_php.c
mfulb: Yeah just leave it in its original state.

mfulb (approved):

PR #942, opened 2024-07-18
feat(agent): add additional arguments as error attributes for newrelic_notice_error

hahuja2:
This PR does the following:

- Create new function called `nr_txn_record_error_with_additional_attributes`, which takes in the additional parameters that are passed in `newrelic_notice_error` and adds them as error attributes.
- Add unit tests to test new functions
- Add integration tests to verify additional parameters are showing up as error attributes

### DIFF
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -173,29 +173,22 @@ PHP_FUNCTION(newrelic_notice_error) {
   }
 
   if (!determine) {
-    char* buf = nr_strndup(errormsgstr, errormsglen);
     char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
 
-    nr_txn_record_error(NRPRG(txn), priority, true, buf, errclass, stack_json);
+    nr_txn_record_error(NRPRG(txn), priority, true, errormsgstr, errclass,
+                        stack_json);
 
-    nr_free(buf);
     nr_free(stack_json);
 
     RETURN_TRUE;
   }
   if (determine) {
-    char* buf1 = nr_strndup(errormsgstr, errormsglen);
-    char* buf2 = nr_strndup(error_file, error_file_len);
-    char* buf3 = nr_strndup(error_context, error_context_len);
     char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
 
     nr_txn_record_error_with_additional_attributes(
-        NRPRG(txn), priority, true, buf1, errclass, buf2,
-        error_line, buf3, error_number, stack_json);
+        NRPRG(txn), priority, true, errormsgstr, errclass, error_file,
+        error_line, error_context, error_number, stack_json);
 
-    nr_free(buf1);
-    nr_free(buf2);
-    nr_free(buf3);
     nr_free(stack_json);
 
     RETURN_TRUE;
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -172,27 +172,19 @@ PHP_FUNCTION(newrelic_notice_error) {
     }
   }
 
-  if (!determine) {
-    char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
-
-    nr_txn_record_error(NRPRG(txn), priority, true, errormsgstr, errclass,
-                        stack_json);
-
-    nr_free(stack_json);
-
-    RETURN_TRUE;
-  }
   if (determine) {
     char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
-
     nr_txn_record_error_with_additional_attributes(
         NRPRG(txn), priority, true, errormsgstr, errclass, error_file,
         error_line, error_context, error_number, stack_json);
-
     nr_free(stack_json);
-
     RETURN_TRUE;
   }
+  char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
+  nr_txn_record_error(NRPRG(txn), priority, true, errormsgstr, errclass,
+                      stack_json);
+  nr_free(stack_json);
+  RETURN_TRUE;
 }
 
 /*
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -146,7 +146,7 @@ PHP_FUNCTION(newrelic_notice_error) {
     case 5:
       if (FAILURE
           == zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
-                                      ZEND_NUM_ARGS() TSRMLS_CC, "lssls!",
+                                      ZEND_NUM_ARGS() TSRMLS_CC, "ls!s!ls!",
                                       &error_number, &errormsgstr, &errormsglen,
                                       &error_file, &error_file_len, &error_line,
                                       &error_context, &error_context_len)) {
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -11,6 +11,7 @@
 #include "php_hash.h"
 #include "php_user_instrument.h"
 #include "fw_drupal_common.h"
+#include "nr_errors.h"
 #include "nr_rum.h"
 #include "util_logging.h"
 #include "util_memory.h"
@@ -71,16 +72,17 @@ PHP_FUNCTION(newrelic_notice_error) {
   const char* errclass = "NoticedError";
   char* errormsgstr = NULL;
   nr_string_len_t errormsglen = 0;
-  zend_long error_number = 0;
-  char* error_file = NULL;
-  nr_string_len_t error_file_len = 0;
-  zend_long error_line = 0;
+  char* user_error_message = NULL;
+  nr_string_len_t user_error_message_len = 0;
+  zend_long user_error_number = 0;
+  char* user_error_file = NULL;
+  nr_string_len_t user_error_file_len = 0;
+  zend_long user_error_line = 0;
   zval* exc = NULL;
-  nr_string_len_t error_context_len = 0;
-  char* error_context = NULL;
+  nr_string_len_t user_error_context_len = 0;
+  char* user_error_context = NULL;
   int priority = 0;
-  zval* ignore = NULL;
-  bool determine = false;
+  bool five_params = false;
 
   NR_UNUSED_RETURN_VALUE;
   NR_UNUSED_RETURN_VALUE_PTR;
@@ -134,11 +136,11 @@ PHP_FUNCTION(newrelic_notice_error) {
     case 2:
       if (FAILURE
           == zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
-                                      ZEND_NUM_ARGS() TSRMLS_CC, "zo!", &ignore,
+                                      ZEND_NUM_ARGS() TSRMLS_CC, "s!o!", &user_error_message, &user_error_message_len,
                                       &exc)) {
         nrl_debug(NRL_API,
                   "newrelic_notice_error: invalid two arguments: expected "
-                  "exception as second argument");
+                  "string as first argument and exception as second argument");
         RETURN_NULL();
       }
       break;
@@ -147,13 +149,13 @@ PHP_FUNCTION(newrelic_notice_error) {
       if (FAILURE
           == zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
                                       ZEND_NUM_ARGS() TSRMLS_CC, "ls!s!ls!",
-                                      &error_number, &errormsgstr, &errormsglen,
-                                      &error_file, &error_file_len, &error_line,
-                                      &error_context, &error_context_len)) {
+                                      &user_error_number, &user_error_message, &user_error_message_len,
+                                      &user_error_file, &user_error_file_len, &user_error_line,
+                                      &user_error_context, &user_error_context_len)) {
         nrl_debug(NRL_API, "newrelic_notice_error: invalid five arguments");
         RETURN_NULL();
       }
-      determine = true;
+      five_params = true;
       break;
 
     default:
@@ -172,11 +174,14 @@ PHP_FUNCTION(newrelic_notice_error) {
     }
   }
 
-  if (determine) {
+  if (five_params) {
+    nr_user_error_t* user_error = nr_user_error_create(
+        user_error_message, user_error_number, user_error_file, user_error_line,
+        user_error_context);
     char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
-    nr_txn_record_error_with_additional_attributes(
-        NRPRG(txn), priority, true, errormsgstr, errclass, error_file,
-        error_line, error_context, error_number, stack_json);
+    nr_txn_record_error_with_additional_attributes(NRPRG(txn), priority, true,
+                                                   user_error_message, errclass,
+                                                   stack_json, user_error);
     nr_free(stack_json);
     RETURN_TRUE;
   }
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -71,14 +71,16 @@ PHP_FUNCTION(newrelic_notice_error) {
   const char* errclass = "NoticedError";
   char* errormsgstr = NULL;
   nr_string_len_t errormsglen = 0;
-  zend_long ignore1 = 0;
-  char* ignore2 = 0;
-  nr_string_len_t ignore3 = 0;
-  zend_long ignore4 = 0;
+  zend_long error_number = 0;
+  char* error_file = NULL;
+  nr_string_len_t error_file_len = 0;
+  zend_long error_line = 0;
   zval* exc = NULL;
-  zval* ignore5 = NULL;
+  nr_string_len_t error_context_len = 0;
+  char* error_context = NULL;
   int priority = 0;
   zval* ignore = NULL;
+  bool determine = false;
 
   NR_UNUSED_RETURN_VALUE;
   NR_UNUSED_RETURN_VALUE_PTR;
@@ -144,12 +146,14 @@ PHP_FUNCTION(newrelic_notice_error) {
     case 5:
       if (FAILURE
           == zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
-                                      ZEND_NUM_ARGS() TSRMLS_CC, "lsslz!",
-                                      &ignore1, &errormsgstr, &errormsglen,
-                                      &ignore2, &ignore3, &ignore4, &ignore5)) {
+                                      ZEND_NUM_ARGS() TSRMLS_CC, "lssls!",
+                                      &error_number, &errormsgstr, &errormsglen,
+                                      &error_file, &error_file_len, &error_line,
+                                      &error_context, &error_context_len)) {
         nrl_debug(NRL_API, "newrelic_notice_error: invalid five arguments");
         RETURN_NULL();
       }
+      determine = true;
       break;
 
     default:
@@ -168,13 +172,30 @@ PHP_FUNCTION(newrelic_notice_error) {
     }
   }
 
-  {
+  if (!determine) {
     char* buf = nr_strndup(errormsgstr, errormsglen);
     char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
 
     nr_txn_record_error(NRPRG(txn), priority, true, buf, errclass, stack_json);
 
     nr_free(buf);
+    nr_free(stack_json);
+
+    RETURN_TRUE;
+  }
+  if (determine) {
+    char* buf1 = nr_strndup(errormsgstr, errormsglen);
+    char* buf2 = nr_strndup(error_file, error_file_len);
+    char* buf3 = nr_strndup(error_context, error_context_len);
+    char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
+
+    nr_txn_record_error_with_additional_attributes(
+        NRPRG(txn), priority, true, buf1, errclass, buf2,
+        error_line, buf3, error_number, stack_json);
+
+    nr_free(buf1);
+    nr_free(buf2);
+    nr_free(buf3);
     nr_free(stack_json);
 
     RETURN_TRUE;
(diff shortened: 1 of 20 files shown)


agent/php_api.c
```diff
@@ -177,6 +181,23 @@ PHP_FUNCTION(newrelic_notice_error) {
...
+    char* stack_json = nr_php_backtrace_to_json(NULL TSRMLS_CC);
+
+    nr_txn_record_error_with_additional_attributes(
+        NRPRG(txn), priority, true, buf1, errclass, buf2,
+        error_line, buf3, error_number, stack_json);
+
+    nr_free(buf1);
+    nr_free(buf2);
+    nr_free(buf3);
+    nr_free(stack_json);
```
ZNeumann, line 199: What is the purpose of copying these strings only to immediately free them? Is there a reason why we cannot just directly pass in `errormsgstr`, `error_file`, and `error_context`?

axiom/nr_errors.c
```diff
@@ -97,6 +170,25 @@ void nr_error_destroy(nr_error_t** error_ptr) {
...
+  if (0 == error) {
+    return;
+  }
+  nr_free(error->message);
+  nr_free(error->klass);
+  nr_free(error->error_file);
+  nr_free(error->error_context);
+  nr_free(error->span_id);
+  nr_free(error->stacktrace_json);
+  nr_realfree((void**)error_ptr);
```
ZNeumann, line 189: Based on the name of this function, I would expect it to only destroy the additional params, and thus still require the calling of `nr_error_destroy` afterward to fully free the error.

axiom/nr_errors_private.h
```diff
@@ -25,6 +25,11 @@ struct _nr_error_t {
   int priority;          /* Error priority - lowest to highest */
   char* message;         /* Error message */
   char* klass;           /* Error class */
+  char* error_file;      /* Error file */
+  int error_line;        /* Error line */
+  char* error_context;   /* Error context */
+  int error_no;          /* Error number */
+  int option;            /* Error option */
```
ZNeumann, line 32: As far as I can tell, these are specifically user-provided through the API. It would be useful to specify that in the name of the variables, or at least their comment.

ZNeumann, line 32: So am I correct in that the `option` variable indicates whether the other user-provided variables are populated? Could we not just check, for example, `error_file` for null behind an API like `nr_error_has_user_attributes` rather than having this confusing variable? As far as I can tell, these attributes are either all-or-nothing.

axiom/nr_txn.c
```diff
@@ -1638,6 +1641,109 @@ void nr_txn_record_error(nrtxn_t* txn,
   nr_free(span_id);
 }
 
+void nr_txn_record_error_with_additional_attributes(
```
ZNeumann, line 1644: There is a large amount of code duplication in this API. It would be nice to create a more unified API such that we don't have multiple copies of large functions.

agent/php_api.c
hahuja2: Good point, fixed: [b8e6402](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/b8e64022e5b029a057aece820b663e4b444e556b)

axiom/nr_errors.c
hahuja2: To reduce code duplication, I have removed this function and added the additional logic in `nr_error_destroy` inside of a if statement: [f917ced](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/f917ced797ecc2b0f2f560487081e73bf4cf47a1)

axiom/nr_errors_private.h
hahuja2: Good point, I have updated the comments: [9f3e7e6](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/9f3e7e652eeeaef0e19c45a1aacbd5ba6891d161)

hahuja2: Agree, the `option` variable isn't necessary and I have removed it: [9f3e7e6](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/9f3e7e652eeeaef0e19c45a1aacbd5ba6891d161)

axiom/nr_txn.c
hahuja2: Good point, I have added helper functions for `nr_txn_record_error`, `nr_segment_set_error`, and `nr_error_create` to reduce code duplication: [23823f9](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/23823f9ac52055e35d00951a9af83aadaeb756fe) and [f917ced](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/f917ced797ecc2b0f2f560487081e73bf4cf47a1)

agent/php_api.c
```diff
@@ -144,12 +146,14 @@ PHP_FUNCTION(newrelic_notice_error) {
     case 5:
       if (FAILURE
           == zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET,
-                                      ZEND_NUM_ARGS() TSRMLS_CC, "lsslz!",
-                                      &ignore1, &errormsgstr, &errormsglen,
-                                      &ignore2, &ignore3, &ignore4, &ignore5)) {
+                                      ZEND_NUM_ARGS() TSRMLS_CC, "lssls!",
```
mfulb, line 149: The format for the parameters changes so the last value is a 's' instead of 'z' - what is the rationale behind this change?

agent/php_api.c
```diff
@@ -168,13 +172,23 @@ PHP_FUNCTION(newrelic_notice_error) {
     }
   }
 
-  {
-    char* buf = nr_strndup(errormsgstr, errormsglen);
+  if (!determine) {
```
mfulb, line 175: This code seems to be able to be reduced as there are common elements to each possible path - also use an `else` instead of having two separate `if` statements testing logically opposite conditions.

axiom/nr_errors.c
```diff
@@ -14,37 +14,68 @@
...
+                            const char* error_file,
+                            int error_line,
+                            const char* error_context,
+                            int error_no,
                             const char* stacktrace_json,
                             const char* span_id,
                             nrtime_t when) {
   nr_error_t* error;
 
-  if (0 == message) {
```
mfulb, line 25: Why were these guards removed?

axiom/nr_errors.c
```diff
@@ -14,37 +14,68 @@
...
-    return 0;
-  }
-
   error = (nr_error_t*)nr_zalloc(sizeof(nr_error_t));
   error->priority = priority;
   error->when = when;
   error->message = nr_strdup(message);
   error->klass = nr_strdup(klass);
   error->stacktrace_json = nr_strdup(stacktrace_json);
+  if (NULL != error_file && NULL != error_context) {
```
mfulb, line 35: If only one of these two values is defined do we still want to create the associated field?

axiom/nr_errors.c
```diff
@@ -14,37 +14,68 @@
...
+                                              const char* message,
+                                              const char* klass,
+                                              const char* error_file,
+                                              int error_line,
+                                              const char* error_context,
+                                              int error_no,
+                                              const char* stacktrace_json,
+                                              const char* span_id,
+                                              nrtime_t when) {
+  if (0 == message || 0 == klass || 0 == error_file || 0 == error_context
```
mfulb, line 71: Please change usages in this file of `0` to `NULL` if used to test a pointer value.  The usage of `0` is from old code and we do not use that form now.

axiom/nr_errors.c
```diff
@@ -94,6 +153,10 @@ void nr_error_destroy(nr_error_t** error_ptr) {
   nr_free(error->klass);
   nr_free(error->span_id);
   nr_free(error->stacktrace_json);
+  if ((NULL != error->error_file) && (NULL != error->error_context)) {
```
mfulb, line 156: `nr_free` should handle this case.

axiom/nr_errors.h
```diff
@@ -38,6 +38,38 @@ extern nr_error_t* nr_error_create(int priority,
...
+ *              The PHP agent uses PHP's predefined error constant names.
+ *              Examples are "E_ERROR" and "E_WARNING".
+ *           4. Error file provided by user.
+ *           5. Error line provided by user.
+ *           6. Error context provided by user.
+ *           7. Error number provided by user.
+ *           8. String containing stack trace in JSON format.
+ *           9. Span ID
+ *           10. When the error occurred.
+ *
```
mfulb, line 58: Would be good to explain if all are required or if there are certain combinations allowed.

axiom/nr_errors.h
```diff
@@ -54,6 +86,35 @@ extern const char* nr_error_get_message(const nr_error_t* error);
  * Returns : The klass of the error or NULL on failure.
  */
 extern const char* nr_error_get_klass(const nr_error_t* error);
+
+/*
+ * Purpose : Get the error file of an error.
+ *
+ * Returns : The error file of the error or NULL on failure.
```
mfulb, line 93: Is it a 'failure' or simply it is not defined?

axiom/nr_segment.c
```diff
@@ -1170,23 +1170,59 @@ void nr_segment_record_exception(nr_segment_t* segment,
...
+                                        int error_line,
+                                        char* error_context,
+                                        int error_no) {
   if (NULL == segment->error) {
     segment->error = nr_zalloc(sizeof(nr_segment_error_t));
   }
 
   nr_free(segment->error->error_message);
   nr_free(segment->error->error_class);
+  if (NULL != error_file && NULL != error_context) {
```
mfulb, line 1186: This code seems like it could leak.  If the caller doesn't define these values we don't free the current values.  Below the current value is set to `NULL` if the caller doesn't define a value, thus losing a reference to the memory?

axiom/nr_segment.c
```diff
@@ -1170,23 +1170,59 @@ void nr_segment_record_exception(nr_segment_t* segment,
...
+    nr_free(segment->error->error_file);
+    nr_free(segment->error->error_context);
+  }
 
   segment->error->error_message
       = error_message ? nr_strdup(error_message) : NULL;
   segment->error->error_class = error_class ? nr_strdup(error_class) : NULL;
+  if (NULL != error_file && NULL != error_context) {
```
mfulb, line 1194: I'm not clear on the behavior of this function if the segment had all fields defined previously and then this function is called and `error_file` and `error_context` are `NULL`.  It seems the previous values would be retained.  That could be confusing as it seems there is no way to clear those fields once set.

axiom/nr_segment.c
```diff
@@ -1170,23 +1170,59 @@ void nr_segment_record_exception(nr_segment_t* segment,
...
 }
 
-void nr_segment_set_error(nr_segment_t* segment,
-                          const char* error_message,
-                          const char* error_class) {
-  if ((NULL == segment) || (NULL == error_message && NULL == error_class)) {
-    return;
-  }
-
+static void nr_segment_set_error_helper(nr_segment_t* segment,
```
mfulb, line 1173: I would recommend putting a comment on each of these static helper functions throughout this file to make it clear no NULL checking is done.  It is unlikely any new code would call these directly but I think it is a good safety guard to make sure a coder is aware of this.

Do you think we need unit tests for these helper functions throughout this file?

axiom/nr_segment.h
```diff
@@ -117,6 +117,10 @@ typedef struct _nr_segment_metric_t {
 typedef struct _nr_segment_error_t {
   char* error_message; /* The error message that will appear on a span event. */
   char* error_class;   /* The error class that will appear on a span event. */
+  char* error_file;
```
mfulb, line 120: Can you add a comment for each of these fields like the existing ones have?

axiom/nr_txn.c
mfulb: Could you remove `nr_txn_record_error_helper` and then just have `nr_txn_record_error` call `nr_txn_record_error_with_additional_attributes` with the unused fields filled with nothing values?

axiom/nr_txn.c
```diff
@@ -2479,6 +2536,13 @@ nr_analytics_event_t* nr_error_to_event(const nrtxn_t* txn) {
   nro_set_hash_string(params, "error.class", nr_error_get_klass(txn->error));
   nro_set_hash_string(params, "error.message",
                       nr_error_get_message(txn->error));
+  if (nr_error_get_file(txn->error) && nr_error_get_context(txn->error)) {
```
mfulb, line 2539: Is it ever possible we'd only one to write a subset of these 4 or do they either all exist or not exist?

axiom/nr_txn.h
```diff
@@ -471,6 +471,17 @@ extern nr_status_t nr_txn_record_error_worthy(const nrtxn_t* txn, int priority);
 #define NR_TXN_ALLOW_RAW_EXCEPTION_MESSAGE \
   "Message removed by New Relic security settings"
 
+extern void nr_txn_record_error_with_additional_attributes(nrtxn_t* txn,
```
mfulb, line 474: Could you document the record error functions (both forms)?

axiom/tests/test_txn.c
```diff
@@ -8098,6 +8098,126 @@ static void test_segment_record_error(void) {
...
+  nr_distributed_trace_set_sampled(txn->distributed_trace, true);
+  txn->options.allow_raw_exception_messages = 1;
+
+  /* No error attributes added if error collection isn't enabled */
+  txn->options.err_enabled = 0;
+  nr_txn_record_error_with_additional_attributes(
+      txn, 1, true, "low priority message", "low priority class", "random.php",
+      150, "random context", 256, "[\"A\",\"B\"]");
+  tlib_pass_if_null("No segment error created", segment->error);
+  txn->options.err_enabled = 1;
```
mfulb, line 8128:
```suggestion

  /* Enable error collection */
  txn->options.err_enabled = 1;
```

axiom/tests/test_txn.c
```diff
@@ -8098,6 +8098,126 @@ static void test_segment_record_error(void) {
...
+  segment = nr_segment_start(txn, NULL, NULL);
+  nr_distributed_trace_set_sampled(txn->distributed_trace, true);
+  txn->options.allow_raw_exception_messages = 1;
+
+  /* No error attributes added if error collection isn't enabled */
+  txn->options.err_enabled = 0;
+  nr_txn_record_error_with_additional_attributes(
+      txn, 1, true, "low priority message", "low priority class", "random.php",
+      150, "random context", 256, "[\"A\",\"B\"]");
+  tlib_pass_if_null("No segment error created", segment->error);
```
mfulb, line 8127: Shouldn't `txn->error` also be checked that no error was created?

tests/integration/attributes/test_transaction_additional_params2.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
mfulb, line 1: I would add tests passing invalid values to both forms of this API.  For example, passing `NULL` or passing an integer or object where a string is expected, etc.   Make sure a mistake calling from PHP will not take the agent down.

agent/php_api.c
hahuja2: This has been changed to a 's' because in the [docs API](https://docs.newrelic.com/docs/apm/agents/php-agent/php-agent-api/newrelic_notice_error/#syntax), it mentions that `error_context` is a string which is the 5th parameter for this function.

axiom/nr_errors.c
hahuja2: These were removed because these guards are being checked before this helper function is called. An example of this is in [nr_error_create](https://github.com/newrelic/newrelic-php-agent/pull/942/files#diff-6f14a71dbed1d482f4f6006b000174fbd0e41f5138cdb61664ea50e9a1aa7947R54).

axiom/tests/test_segment.c
```diff
@@ -2868,6 +2874,12 @@ static void test_segment_set_error_attributes(void) {
   tlib_pass_if_str_equal("error.class", "error.class 1",
                          segment.error->error_class);
 
+  nr_segment_set_error_with_additional_params(&segment, "error.message", "error.class", "error.file", 125, "rand3", 100);
```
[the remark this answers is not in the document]
hahuja2: If the user passes in an empty string, then the attribute will still show up. The attribute will not show up if `NULL` is passed in.

axiom/nr_errors.c
hahuja2: After discussion, it has been decided that if one attribute is passed in as `NULL`, the rest of the attributes should still show up. I will create a commit for this.

axiom/nr_segment.c
hahuja2: After discussion, it has been decided that if `NULL` is passed in, the attribute should not show up. I will modify the code and create a commit for this.

agent/php_api.c
hahuja2: Fixed: [66a069d](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/66a069dd06cd14f1dc72f87f940a794d8180053a)

axiom/nr_errors.c
hahuja2: Fixed: [ace780f](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/ace780f45742b8df90d08494c340bda271feb495)

axiom/nr_errors.c
hahuja2: Fixed: [c058df8](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/c058df86277f602e0f73e37e0a0a86e805a2d7d4)

axiom/nr_errors.h
hahuja2: Good point, it should instead say that it was not defined, fixed: [9fad7fa](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/9fad7fae54e3fa1a4a2d2dfecf7493e61ee1f2a3)

axiom/nr_errors.h
hahuja2: Added: [9fad7fa](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/9fad7fae54e3fa1a4a2d2dfecf7493e61ee1f2a3)

axiom/nr_segment.c
hahuja2: I have modified this function where all strings are freed now, so the previous values will not be retained. New values will only be added if they are not `NULL`: [ed191c0](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/ed191c0b4ee2cff587aa3fcee1219c9133e586f7), [bc0a82e](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/bc0a82e066d75632bad2a84be5fa90b5dc42d5ed)

axiom/nr_segment.h
hahuja2: Added: [eb64129](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/eb6412921bd5a5b301a6f76ffde87aa581ce664e)

axiom/nr_txn.h
hahuja2: Added: [c5a40cc](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/c5a40cc3cabb46abb889e5ba51355091dd73f2e2)

axiom/tests/test_segment.c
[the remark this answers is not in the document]
hahuja2: Added: [06430a2](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/06430a26c136c81d986351c1714e6ae814e9e731)

axiom/tests/test_txn.c
hahuja2: Good point, added: [42a7a81](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/42a7a81482af738ad7a4602ad9babe4cab6becaa)

axiom/nr_txn.c
hahuja2: Good idea, I have removed all of the helper functions: [e177f40](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/e177f40a8b868410e5e1b728df7b5df8356ce1f4), [ed191c0](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/ed191c0b4ee2cff587aa3fcee1219c9133e586f7), [788c5a0](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/788c5a0e6330c7c788b2a6c5b028391319b7af17)

axiom/tests/test_txn.c
```diff
@@ -8098,6 +8098,126 @@ static void test_segment_record_error(void) {
...
+  tlib_pass_if_int_equal("Correct segment error.no", 256,
+                         segment->error->error_no);
+  tlib_pass_if_str_equal("txn error message matches segment error message",
+                         segment->error->error_message,
+                         nr_error_get_message(txn->error));
+  tlib_pass_if_str_equal("txn error class matches segment error class",
+                         segment->error->error_class,
+                         nr_error_get_klass(txn->error));
+
+  /* Multiple errors on the same segment */
```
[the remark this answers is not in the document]
hahuja2: Added: [42a7a81](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/42a7a81482af738ad7a4602ad9babe4cab6becaa)

axiom/nr_segment.c
hahuja2: I have removed all of the helper functions: [e177f40](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/e177f40a8b868410e5e1b728df7b5df8356ce1f4), [ed191c0](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/ed191c0b4ee2cff587aa3fcee1219c9133e586f7), [788c5a0](https://github.com/newrelic/newrelic-php-agent/pull/942/commits/788c5a0e6330c7c788b2a6c5b028391319b7af17)

mfulb:
PR #960 supersedes the work in this PR.  The conclusion was that the agent spec does not support these current parameters currently as attributes, and much of the information is available from the stack trace or by adding the desired information to the error message passed to the API.

PR #947, opened 2024-08-07
feat(agent): Adds supportability metrics for PHP and agent version

mfulb:
Adds supportability metrics for the agent and PHP version.

I did clean up some ZTS macros as well but I think it is obvious what is a real change and what isnt.

### DIFF
--- a/agent/Makefile.frag
+++ b/agent/Makefile.frag
@@ -49,12 +49,15 @@ $(PHP_MODULES): .libs/deps.mk
 newrelic.la: $(PHP_AXIOM)/libaxiom.a
 
 #
-# The version number is needed by php_newrelic.c as a static string literal,
+# The version number is needed by php_newrelic.c and php_txn.c as a static string literal,
 # so it can be placed in the module entry.
 #
 include ../make/version.mk
 php_newrelic.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
 php_newrelic.lo: ../VERSION
+
+php_txn.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
+php_txn.lo: ../VERSION
 
 #
 # Unit tests!
@@ -260,6 +263,13 @@ ifeq (/opt/nr/lamp/lib,$(findstring /opt/nr/lamp/lib,$(PHP_EMBED_LIBRARY)))
 		TEST_NEWRELIC_SHARED_LIBADD := $(TEST_NEWRELIC_SHARED_LIBADD) -pthread
 	endif
 endif
+
+#
+# Need agent version for test_txn
+#
+#NEWRELIC_CFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
+tests/test_txn.o: EXTRA_CFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
+tests/test_txn.o: ../VERSION
 
 #
 # Used when linking test binaries.
--- a/agent/Makefile.frag
+++ b/agent/Makefile.frag
@@ -49,10 +49,13 @@ $(PHP_MODULES): .libs/deps.mk
 newrelic.la: $(PHP_AXIOM)/libaxiom.a
 
 #
-# The version number is needed by php_newrelic.c and php_txn.c as a static string literal,
+# The version number is needed by several source files as a static string literal,
 # so it can be placed in the module entry.
 #
 include ../make/version.mk
+php_api_internal.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
+php_api_internal.lo: ../VERSION
+
 php_newrelic.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
 php_newrelic.lo: ../VERSION
 
--- a/agent/Makefile.frag
+++ b/agent/Makefile.frag
@@ -53,8 +53,6 @@ newrelic.la: $(PHP_AXIOM)/libaxiom.a
 # so it can be placed in the module entry.
 #
 include ../make/version.mk
-php_api_internal.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
-php_api_internal.lo: ../VERSION
 
 php_newrelic.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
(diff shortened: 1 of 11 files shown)


agent/php_api_internal.c
```diff
@@ -103,6 +103,27 @@ PHP_FUNCTION(newrelic_get_hostname) {
...
+
+  if (FAILURE == zend_parse_parameters_none()) {
+    RETURN_FALSE;
+  }
+
+#ifdef NR_VERSION
+  version = NR_VERSION;
+#else
+  version = "unknown";
+#endif
```
lavarou, line 123: When would `NR_VERSION` not be defined? Why is this conditional construct needed?

agent/php_txn.c
```diff
@@ -669,6 +669,62 @@ static void nr_php_txn_send_metrics_once(nrtxn_t* txn TSRMLS_DC) {
...
+
+  if (NULL == txn) {
+    return;
+  }
+
+#ifdef NR_VERSION
+  version = NR_VERSION;
+#else
+  version = "unknown";
+#endif
```
lavarou, line 715: When would NR_VERSION not be defined? Why is this conditional construct needed?

agent/php_api_internal.c
mfulb: The check is a guard against an unexpected situation where it is undefined.  Perhaps the compilation should just fail in that scenario.

mfulb: Fixed in commit [7ea6c93](https://github.com/newrelic/newrelic-php-agent/pull/947/commits/7ea6c93be430eeadd4671cc76923a6b17e78c5f8).

agent/php_txn.c
mfulb: Fixed in commit [7ea6c93](https://github.com/newrelic/newrelic-php-agent/pull/947/commits/7ea6c93be430eeadd4671cc76923a6b17e78c5f8).

agent/Makefile.frag
```diff
@@ -49,13 +49,19 @@ $(PHP_MODULES): .libs/deps.mk
...
 
 #
-# The version number is needed by php_newrelic.c as a static string literal,
+# The version number is needed by several source files as a static string literal,
 # so it can be placed in the module entry.
 #
 include ../make/version.mk
+php_api_internal.lo: CPPFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
+php_api_internal.lo: ../VERSION
+
```
lavarou, line 58: Should be reverted as part of 7ea6c93be430eeadd4671cc76923a6b17e78c5f8

agent/Makefile.frag
```diff
@@ -261,6 +267,13 @@ ifeq (/opt/nr/lamp/lib,$(findstring /opt/nr/lamp/lib,$(PHP_EMBED_LIBRARY)))
 	endif
 endif
 
+#
+# Need agent version for test_txn
+#
+#NEWRELIC_CFLAGS += -DNR_VERSION="\"$(AGENT_VERSION)\""
```
lavarou, line 273: The flag defining `NR_VERSION` macro is appended to `EXTRA_CFLAGS` and there should be no need for this comment:
```suggestion
```

agent/Makefile.frag
mfulb: Good catch - fixed in [fe93413c](https://github.com/newrelic/newrelic-php-agent/pull/947/commits/fe93413c374d0700f79d6122d336f31f2c3383b1).

agent/Makefile.frag
mfulb: Good catch - fixed in [fe93413c](https://github.com/newrelic/newrelic-php-agent/pull/947/commits/fe93413c374d0700f79d6122d336f31f2c3383b1).

agent/php_txn.c
```diff
@@ -669,6 +669,58 @@ static void nr_php_txn_send_metrics_once(nrtxn_t* txn TSRMLS_DC) {
...
+
+  if (NULL == txn) {
+    return;
+  }
+
+  if (nr_strempty(version)) {
+    return;
+  }
+
+  metric_name = nr_formatf("Supportability/PHP/AgentVersion/%s", version);
```
zsistla, line 683: since we are just appending and not putting version in the middle of the string, can't we just use nr_strcat?

agent/php_txn.c
```diff
@@ -669,6 +669,58 @@ static void nr_php_txn_send_metrics_once(nrtxn_t* txn TSRMLS_DC) {
...
+
+  if (NULL == txn) {
+    return;
+  }
+
+  if (nr_strempty(version)) {
+    return;
+  }
+
+  metric_name = nr_formatf("Supportability/PHP/Version/%s", version);
```
zsistla, line 699: php version and agent version are things that will be known at minit time.
can't we just compute it once at start up, make it a global variable, then create a metric as needed instead of recomputing a known value on every txn?

daemon/cmd/integration_runner/main.go
```diff
@@ -376,6 +376,15 @@ func main() {
 	// Env vars common to all tests.
 	ctx.Env["EXTERNAL_HOST"] = externalHost
 
+	// Store PHP and agent versions
```
lavarou, line 379: This comment suggests it is some sort of a database operation. The comment on line 376 and the code itself is enough to understand what is happening here:
```suggestion
```

lavarou (dismissed):

zsistla (dismissed):

agent/php_txn.c
mfulb: Based on a suggestion by zsistla the agent version metric name is now computed at compile time - added in commit [db1ee0dc8](https://github.com/newrelic/newrelic-php-agent/pull/947/commits/db1ee0dc88fb8f658d3dec7f08bcc4b8793c81f0).
My thought on the php version metric name creation is that these are very fast operations (profiling shows 10's of thousands per msec) so the addition of this one (which happens only once per request at the end of the request, after PHP has serviced the request) is not enough to merit changing its creation.

agent/php_txn.c
mfulb: Switching to `nr_strcat` would still require measuring the string lengths and allocating a buffer.  There may be some improvement in speed - but these `sprintf` ops take 10s of microseconds so it is not going to be noticeable in latency or CPU since it occurs once per request and after the PHP app is done.

lavarou (approved):

zsistla (approved):

PR #956, opened 2024-08-27
ci: fix issue response

bduranleau-nr:
Fix gh command failure when no github repository can be found.

### DIFF
--- a/.github/workflows/issue-comment.yml
+++ b/.github/workflows/issue-comment.yml
@@ -23,12 +23,6 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.token }}
     steps:
-      - name: Checkout newrelic-php-agent code
-        uses: actions/checkout@v4
-        with:
-          path: php-agent
-          repository: ${{ inputs.origin }}/newrelic-php-agent
-          ref: ${{ inputs.ref }}
       - name: respond to issue
         run: |
-          gh issue comment ${{ inputs.issue-number }} --body "${{ inputs.message }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-comment.yml
```diff
@@ -23,6 +23,12 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.token }}
     steps:
+      - name: Checkout newrelic-php-agent code
+        uses: actions/checkout@v4
+        with:
+          path: php-agent
```
lavarou, line 29: Will this be enough for `gh` to pick up which repo it needs to operate with? Wouldn't it be better instead of checkout step to provide repo as an argument to `gh` command (`-R` - see https://cli.github.com/manual/gh_issue_comment)?

.github/workflows/issue-comment.yml
```diff
@@ -23,6 +23,12 @@ jobs:
...
+          repository: ${{ inputs.origin }}/newrelic-php-agent
+          ref: ${{ inputs.ref }}
       - name: respond to issue
         run: |
           gh issue comment ${{ inputs.issue-number }} --body "${{ inputs.message }}"
```
lavarou, line 34: I have not tried this but it should do the job without the checkout step:
```suggestion
          gh issue -R ${{ github.repository }} comment ${{ inputs.issue-number }} --body "${{ inputs.message }}"
```

bduranleau-nr: Looks promising, let's give it a shot

bduranleau-nr: changed here f96ac63b57cb2bb020c059c9873d402ad35456f7

lavarou (approved):

PR #958, opened 2024-08-28
feat(agent): Memcached instance metrics with host name

ZNeumann:
These metrics can be used by the backend to create AWS relationship maps with memcached

### DIFF
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -1541,7 +1541,6 @@ NR_INNER_WRAPPER(memcached_add_server) {
   nr_datastore_instance_t* instance = NULL;
   char* instance_metric = NULL;
   int zcaught = 0;
-  nr_segment_t* segment = nr_segment_start(NRPRG(txn), NULL, NULL);
 
   if (SUCCESS
       == zend_parse_parameters_ex(
@@ -1551,12 +1550,12 @@ NR_INNER_WRAPPER(memcached_add_server) {
     instance = nr_php_memcached_create_datastore_instance(host, port);
     instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
                                instance->host, instance->port_path_or_id);
+    nrm_force_add(NRPRG(txn)->unscoped_metrics, instance_metric, 0);
     nr_free(instance);
-    nr_segment_add_metric(segment, instance_metric, false);
+    nr_free(instance_metric);
   }
   zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
                                      INTERNAL_FUNCTION_PARAM_PASSTHRU);
-  nr_segment_end(&segment);
   if (zcaught) {
     zend_bailout();
     /* NOTREACHED */
@@ -1569,7 +1568,6 @@ NR_INNER_WRAPPER(memcached_add_servers) {
   nr_datastore_instance_t* instance = NULL;
   char* instance_metric = NULL;
   int zcaught = 0;
-  nr_segment_t* segment = nr_segment_start(NRPRG(txn), NULL, NULL);
 
   if (SUCCESS
       == zend_parse_parameters_ex(
@@ -1583,8 +1581,9 @@ NR_INNER_WRAPPER(memcached_add_servers) {
           instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
           instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
                                        instance->host, instance->port_path_or_id);
-          nr_segment_add_metric(segment, instance_metric, false);
+          nrm_force_add(NRPRG(txn)->unscoped_metrics, instance_metric, 0);
           nr_free(instance);
+          nr_free(instance_metric);
         }
       }
       ZEND_HASH_FOREACH_END();
@@ -1593,7 +1592,6 @@ NR_INNER_WRAPPER(memcached_add_servers) {
   zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
                                      INTERNAL_FUNCTION_PARAM_PASSTHRU);
 
-  nr_segment_end(&segment);
   if (zcaught) {
     zend_bailout();
     /* NOTREACHED */
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -802,14 +802,15 @@ static void nr_php_instrument_datastore_operation_call(
 
 /* 
  * Like nr_php_instrument_datastore_operation_call() but does not
- * call the underlying internal function
+ * call the underlying internal function. This allows multiple
+ * datastore metrics to be creating during a single underlying
+ * PHP call.
  */
 static void nr_php_instrument_datastore_operation(
     nr_datastore_t datastore,
     const char* operation,
     nr_datastore_instance_t* instance,
-    bool instance_only,
-    INTERNAL_FUNCTION_PARAMETERS) {
+    bool instance_only) {
   nr_segment_t* segment = NULL;
   nr_segment_datastore_params_t params = {
       .datastore = {
@@ -1602,8 +1603,7 @@ NR_INNER_WRAPPER(memcached_multi_connect_function) {
       if (NULL != host) {
         instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
         nr_php_instrument_datastore_operation(NR_DATASTORE_MEMCACHE,
-                                              NULL, instance, true,
-                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
+                                              NULL, instance, true);
       }
     }
     ZEND_HASH_FOREACH_END();
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -16,6 +16,7 @@
 #include "php_hash.h"
 #include "php_httprequest_send.h"
 #include "php_internal_instrument.h"
+#include "php_memcached.h"
 #include "php_mysql.h"
 #include "php_mysqli.h"
 #include "php_pdo.h"
@@ -1560,20 +1561,6 @@ NR_INNER_WRAPPER(memcache_function) {
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
                                              nr_wrapper->extra, NULL, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
-}
-
-static nr_datastore_instance_t* nr_php_memcached_create_datastore_instance(
-    const char* host,
-    zend_long port) {
-  nr_datastore_instance_t* instance = NULL;
-  if (port == 0) { // local socket
-    instance = nr_datastore_instance_create("localhost", host, NULL);
-  } else {
-    char* port_str = nr_formatf("%ld", (long)port);
-    instance  = nr_datastore_instance_create(host, port_str, NULL);
-    nr_free(port_str);
-  }
-  return instance;
 }
 
 NR_INNER_WRAPPER(memcached_connect_function) {
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -13,6 +13,7 @@
 #include "php_explain_mysqli.h"
 #include "php_file_get_contents.h"
 #include "php_globals.h"
+#include "php_hash.h"
 #include "php_httprequest_send.h"
 #include "php_internal_instrument.h"
 #include "php_mysql.h"
@@ -1529,6 +1530,66 @@ NR_INNER_WRAPPER(memcache_function) {
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
                                              nr_wrapper->extra, NULL,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
+}
+
+static nr_datastore_instance_t* nr_php_memcached_create_datastore_instance(
+    const char* host,
+    zend_long port) {
+  nr_datastore_instance_t* instance = NULL;
+  if (port == 0) { // local socket
+    instance = nr_datastore_instance_create("localhost", host, NULL);
+  } else {
+    char* port_str = nr_formatf("%ld", (long)port);
+    instance  = nr_datastore_instance_create(host, port_str, NULL);
+    nr_free(port_str);
+  }
+  return instance;
+}
+
+NR_INNER_WRAPPER(memcached_connect_function) {
+  char* host = NULL;
+  nr_string_len_t host_len = 0;
+  zend_long port = 0;
+  zend_long weight = 0;
+  nr_datastore_instance_t* instance = NULL;
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "s|ll", &host,
+          &host_len, &port, &weight)) {
+    if (NULL != host) {
+      instance = nr_php_memcached_create_datastore_instance(host, port);
+    }
+  }
+  nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
+                                             nr_wrapper->extra, instance,
+                                             INTERNAL_FUNCTION_PARAM_PASSTHRU);
+}
+
+NR_INNER_WRAPPER(memcached_multi_connect_function) {
+  zval* servers = NULL;
+  zval* server = NULL;
+  zend_ulong num_key = 0;
+  nr_php_string_hash_key_t* string_key = NULL;
+  nr_datastore_instance_t* instance = NULL;
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "a", &servers)) {
+    ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(servers), num_key, string_key, server) {
+      (void)num_key;
+      (void)string_key;
+      zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
+      zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
+      if (NULL != host) {
+        instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
+        nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
+                                                   nr_wrapper->extra, instance,
+                                                   INTERNAL_FUNCTION_PARAM_PASSTHRU);
+      }
+    }
+    ZEND_HASH_FOREACH_END();
+  }
 }
 
 /*
@@ -3098,6 +3159,8 @@ NR_OUTER_WRAPPER(memcached_set)
 NR_OUTER_WRAPPER(memcached_setbykey)
 NR_OUTER_WRAPPER(memcached_setmulti)
 NR_OUTER_WRAPPER(memcached_setmultibykey)
+NR_OUTER_WRAPPER(memcached_addserver)
+NR_OUTER_WRAPPER(memcached_addservers)
 
 NR_OUTER_WRAPPER(redis_append)
 NR_OUTER_WRAPPER(redis_bitcount)
@@ -3511,6 +3574,10 @@ void nr_php_generate_internal_wrap_records(void) {
                       memcache_function, 0, "set")
   NR_INTERNAL_WRAPREC("memcached::setmultibykey", memcached_setmultibykey,
                       memcache_function, 0, "set")
+  NR_INTERNAL_WRAPREC("memcached::addServer", memcached_addserver,
+                      memcached_connect_function, 0, "connect");
+  NR_INTERNAL_WRAPREC("memcached::addServers", memcached_addservers,
+                      memcached_multi_connect_function, 0, "connect");
 
   NR_INTERNAL_WRAPREC("redis::connect", redis_connect, redis_connect, 0,
                       "connect")
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -1574,14 +1574,20 @@ NR_INNER_WRAPPER(memcached_connect_function) {
   if (SUCCESS
       == zend_parse_parameters_ex(
           ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "s|ll", &host,
-          &host_len, &port, &weight)) {
-    if (NULL != host) {
-      instance = nr_php_memcached_create_datastore_instance(host, port);
+          &host_len, &port, &weight) &&
+      NULL != host) {
+    instance = nr_php_memcached_create_datastore_instance(host, port);
+    nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
+                                               NULL, instance, true,
+                                               INTERNAL_FUNCTION_PARAM_PASSTHRU);
+  } else {
+    zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
+                                       INTERNAL_FUNCTION_PARAM_PASSTHRU);
+    if (zcaught) {
+      zend_bailout();
+      /* NOTREACHED */
     }
   }
-  nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                             NULL, instance, true,
-                                             INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
 NR_INNER_WRAPPER(memcached_multi_connect_function) {
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -1600,16 +1600,19 @@ NR_INNER_WRAPPER(memcached_add_servers) {
   if (SUCCESS
       == zend_parse_parameters_ex(
           ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "a", &servers)) {
-    ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(servers), server) {
-      zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
-      zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
-      if (NULL != host && NULL != port) {
-        instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
-        nr_php_instrument_datastore_operation(NR_DATASTORE_MEMCACHE,
-                                              NULL, instance, true);
+    if (Z_TYPE_P(servers) == IS_ARRAY) {
+      ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(servers), server) {
+        zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
+        zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
+        if (NULL != host && NULL != port &&
+            Z_TYPE_P(host) == IS_STRING && Z_TYPE_P(port) == IS_LONG) {
+          instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
+          nr_php_instrument_datastore_operation(NR_DATASTORE_MEMCACHE,
+                                                NULL, instance, true);
+        }
       }
+      ZEND_HASH_FOREACH_END();
     }
-    ZEND_HASH_FOREACH_END();
   }
   zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
                                      INTERNAL_FUNCTION_PARAM_PASSTHRU);
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -1595,9 +1595,7 @@ NR_INNER_WRAPPER(memcached_multi_connect_function) {
   if (SUCCESS
       == zend_parse_parameters_ex(
           ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "a", &servers)) {
-    ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(servers), num_key, string_key, server) {
-      (void)num_key;
-      (void)string_key;
+    ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(servers), server) {
       zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
       zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
       if (NULL != host) {
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -800,35 +800,6 @@ static void nr_php_instrument_datastore_operation_call(
   }
 }
 
-/* 
- * Like nr_php_instrument_datastore_operation_call() but does not
- * call the underlying internal function. This allows multiple
- * datastore metrics to be creating during a single underlying
- * PHP call.
- */
-static void nr_php_instrument_datastore_operation(
-    nr_datastore_t datastore,
-    const char* operation,
-    nr_datastore_instance_t* instance,
-    bool instance_only) {
-  nr_segment_t* segment = NULL;
-  nr_segment_datastore_params_t params = {
-      .datastore = {
-          .type = datastore,
-      },
-      .operation = nr_strdup(operation),
-      .instance  = instance,
-      .instance_only = instance_only,
-      .callbacks = {
-          .backtrace = nr_php_backtrace_callback,
-      },
-  };
-
-  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
-  nr_segment_datastore_end(&segment, &params);
-  nr_free(params.operation);
-}
-
 /*
  * Handle
  *   bool mysqli_commit ( object $link [, int $flags=0, string $name] )
@@ -1570,7 +1541,9 @@ NR_INNER_WRAPPER(memcached_add_server) {
   zend_long port = 0;
   zend_long weight = 0;
   nr_datastore_instance_t* instance = NULL;
+  char* instance_metric = NULL;
   int zcaught = 0;
+  nr_segment_t* segment = nr_segment_start(NRPRG(txn), NULL, NULL);
 
   if (SUCCESS
       == zend_parse_parameters_ex(
@@ -1578,16 +1551,17 @@ NR_INNER_WRAPPER(memcached_add_server) {
           &host_len, &port, &weight) &&
       NULL != host) {
     instance = nr_php_memcached_create_datastore_instance(host, port);
-    nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                               NULL, instance, true,
-                                               INTERNAL_FUNCTION_PARAM_PASSTHRU);
-  } else {
-    zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
-                                       INTERNAL_FUNCTION_PARAM_PASSTHRU);
-    if (zcaught) {
-      zend_bailout();
-      /* NOTREACHED */
-    }
+    instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
+                               instance->host, instance->port_path_or_id);
+    nr_free(instance);
+    nr_segment_add_metric(segment, instance_metric, false);
+  }
+  zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
+                                     INTERNAL_FUNCTION_PARAM_PASSTHRU);
+  nr_segment_end(&segment);
+  if (zcaught) {
+    zend_bailout();
+    /* NOTREACHED */
   }
 }
 
@@ -1595,7 +1569,9 @@ NR_INNER_WRAPPER(memcached_add_servers) {
   zval* servers = NULL;
   zval* server = NULL;
   nr_datastore_instance_t* instance = NULL;
+  char* instance_metric = NULL;
   int zcaught = 0;
+  nr_segment_t* segment = nr_segment_start(NRPRG(txn), NULL, NULL);
 
   if (SUCCESS
       == zend_parse_parameters_ex(
@@ -1607,8 +1583,10 @@ NR_INNER_WRAPPER(memcached_add_servers) {
         if (NULL != host && NULL != port &&
             Z_TYPE_P(host) == IS_STRING && Z_TYPE_P(port) == IS_LONG) {
           instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
-          nr_php_instrument_datastore_operation(NR_DATASTORE_MEMCACHE,
-                                                NULL, instance, true);
+          instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
+                                       instance->host, instance->port_path_or_id);
+          nr_segment_add_metric(segment, instance_metric, false);
+          nr_free(instance);
         }
       }
       ZEND_HASH_FOREACH_END();
@@ -1617,6 +1595,7 @@ NR_INNER_WRAPPER(memcached_add_servers) {
   zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
                                      INTERNAL_FUNCTION_PARAM_PASSTHRU);
 
+  nr_segment_end(&segment);
   if (zcaught) {
     zend_bailout();
     /* NOTREACHED */
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -1538,8 +1538,6 @@ NR_INNER_WRAPPER(memcached_add_server) {
   nr_string_len_t host_len = 0;
   zend_long port = 0;
   zend_long weight = 0;
-  nr_datastore_instance_t* instance = NULL;
-  char* instance_metric = NULL;
   int zcaught = 0;
 
   if (SUCCESS
@@ -1547,12 +1545,7 @@ NR_INNER_WRAPPER(memcached_add_server) {
           ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "s|ll", &host,
           &host_len, &port, &weight) &&
       NULL != host) {
-    instance = nr_php_memcached_create_datastore_instance(host, port);
-    instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
-                               instance->host, instance->port_path_or_id);
-    nrm_force_add(NRPRG(txn)->unscoped_metrics, instance_metric, 0);
-    nr_datastore_instance_destroy(&instance);
-    nr_free(instance_metric);
+    nr_php_memcached_create_instance_metric(host, port);
   }
   zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
                                      INTERNAL_FUNCTION_PARAM_PASSTHRU);
@@ -1565,8 +1558,6 @@ NR_INNER_WRAPPER(memcached_add_server) {
 NR_INNER_WRAPPER(memcached_add_servers) {
   zval* servers = NULL;
   zval* server = NULL;
-  nr_datastore_instance_t* instance = NULL;
-  char* instance_metric = NULL;
   int zcaught = 0;
 
   if (SUCCESS
@@ -1576,14 +1567,9 @@ NR_INNER_WRAPPER(memcached_add_servers) {
       ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(servers), server) {
         zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
         zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
-        if (NULL != host && NULL != port &&
-            Z_TYPE_P(host) == IS_STRING && Z_TYPE_P(port) == IS_LONG) {
-          instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
-          instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
-                                       instance->host, instance->port_path_or_id);
-          nrm_force_add(NRPRG(txn)->unscoped_metrics, instance_metric, 0);
-          nr_datastore_instance_destroy(&instance);
-          nr_free(instance_metric);
+        if (nr_php_is_zval_valid_string(host) &&
+            nr_php_is_zval_valid_integer(port)) {
+          nr_php_memcached_create_instance_metric(Z_STRVAL_P(host), Z_LVAL_P(port));
         }
       }
       ZEND_HASH_FOREACH_END();
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -769,6 +769,7 @@ static void nr_php_instrument_datastore_operation_call(
     nr_datastore_t datastore,
     const char* operation,
     nr_datastore_instance_t* instance,
+    bool instance_only,
     INTERNAL_FUNCTION_PARAMETERS) {
   int zcaught = 0;
   nr_segment_t* segment = NULL;
@@ -778,6 +779,7 @@ static void nr_php_instrument_datastore_operation_call(
       },
       .operation = nr_strdup(operation),
       .instance  = instance,
+      .instance_only = instance_only,
       .callbacks = {
           .backtrace = nr_php_backtrace_callback,
       },
@@ -833,7 +835,7 @@ NR_INNER_WRAPPER(mysqli_commit) {
   }
 
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MYSQL,
-                                             "commit", instance,
+                                             "commit", instance, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
@@ -1528,7 +1530,7 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
  */
 NR_INNER_WRAPPER(memcache_function) {
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                             nr_wrapper->extra, NULL,
+                                             nr_wrapper->extra, NULL, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
@@ -1562,7 +1564,7 @@ NR_INNER_WRAPPER(memcached_connect_function) {
     }
   }
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                             nr_wrapper->extra, instance,
+                                             NULL, instance, true,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
@@ -1584,7 +1586,7 @@ NR_INNER_WRAPPER(memcached_multi_connect_function) {
       if (NULL != host) {
         instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
         nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                                   nr_wrapper->extra, instance,
+                                                   NULL, instance, true,
                                                    INTERNAL_FUNCTION_PARAM_PASSTHRU);
       }
     }
@@ -1617,7 +1619,7 @@ NR_INNER_WRAPPER(redis_connect) {
   }
 
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_REDIS,
-                                             nr_wrapper->extra, instance,
+                                             nr_wrapper->extra, instance, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
@@ -1676,7 +1678,7 @@ NR_INNER_WRAPPER(redis_function) {
   instance = nr_php_redis_retrieve_datastore_instance(this_obj TSRMLS_CC);
 
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_REDIS,
-                                             nr_wrapper->extra, instance,
+                                             nr_wrapper->extra, instance, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
@@ -2669,7 +2671,7 @@ NR_INNER_WRAPPER(ob_flush_common) {
  */
 NR_INNER_WRAPPER(mongodb_execute) {
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MONGODB,
-                                             "execute", NULL,
+                                             "execute", NULL, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
@@ -3574,10 +3576,10 @@ void nr_php_generate_internal_wrap_records(void) {
                       memcache_function, 0, "set")
   NR_INTERNAL_WRAPREC("memcached::setmultibykey", memcached_setmultibykey,
                       memcache_function, 0, "set")
-  NR_INTERNAL_WRAPREC("memcached::addServer", memcached_addserver,
-                      memcached_connect_function, 0, "connect");
-  NR_INTERNAL_WRAPREC("memcached::addServers", memcached_addservers,
-                      memcached_multi_connect_function, 0, "connect");
+  NR_INTERNAL_WRAPREC("memcached::addserver", memcached_addserver,
+                      memcached_connect_function, 0, 0);
+  NR_INTERNAL_WRAPREC("memcached::addservers", memcached_addservers,
+                      memcached_multi_connect_function, 0, 0);
 
   NR_INTERNAL_WRAPREC("redis::connect", redis_connect, redis_connect, 0,
                       "connect")
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -1573,7 +1573,7 @@ NR_INNER_WRAPPER(memcached_connect_function) {
 
   if (SUCCESS
       == zend_parse_parameters_ex(
-          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "s|ll", &host,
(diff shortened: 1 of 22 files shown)


agent/php_internal_instrument.c
```diff
@@ -1527,10 +1558,78 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
  */
 NR_INNER_WRAPPER(memcache_function) {
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                             nr_wrapper->extra, NULL,
+                                             nr_wrapper->extra, NULL, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+static nr_datastore_instance_t* nr_php_memcached_create_datastore_instance(
```
ZNeumann, line 1565: The redis equivalent of this function is in the file `agent/php_redis.c`. I am ambivalent on whether to create an equal `agent/php_memcached.c` for just this helper or to keep it here.

zsistla: I would vote for creating a php_memcached.c even if it's just for one function (for now) for future readability.

ZNeumann: refactored in [a0d261f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/a0d261f86063ce6973e474993b0a79c6a26af544)

agent/php_internal_instrument.c
```diff
@@ -796,6 +800,34 @@ static void nr_php_instrument_datastore_operation_call(
   }
 }
 
+/* 
+ * Like nr_php_instrument_datastore_operation_call() but does not
+ * call the underlying internal function
+ */
```
bduranleau-nr, line 806: Can you include some context as to why we specifically don't want to call the underlying internal function?

agent/php_internal_instrument.c
```diff
@@ -796,6 +800,34 @@ static void nr_php_instrument_datastore_operation_call(
...
+      .operation = nr_strdup(operation),
+      .instance  = instance,
+      .instance_only = instance_only,
+      .callbacks = {
+          .backtrace = nr_php_backtrace_callback,
+      },
+  };
+
+  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  nr_segment_datastore_end(&segment, &params);
```
bduranleau-nr, line 827: This operation will probably result in an immediate (nanoseconds) segment duration. Is the rationale for this just that we have to send this data via a segment to the backend?

agent/php_internal_instrument.c
```diff
@@ -796,6 +800,34 @@ static void nr_php_instrument_datastore_operation_call(
...
+/* 
+ * Like nr_php_instrument_datastore_operation_call() but does not
+ * call the underlying internal function
+ */
+static void nr_php_instrument_datastore_operation(
+    nr_datastore_t datastore,
+    const char* operation,
+    nr_datastore_instance_t* instance,
+    bool instance_only,
+    INTERNAL_FUNCTION_PARAMETERS) {
```
bduranleau-nr, line 812: Is this argument to the function necessary? It doesn't seem like we're using any of the `INTERNAL_FUNCTION_PARAMETERS` (`execute_data`, `return_value`) here.

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1559,64 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "a", &servers)) {
+    ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(servers), num_key, string_key, server) {
+      (void)num_key;
+      (void)string_key;
+      zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
+      zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
+      if (NULL != host) {
+        instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
+        nr_php_instrument_datastore_operation(NR_DATASTORE_MEMCACHE,
```
bduranleau-nr, line 1604: I think I see- so the reason we removed the `call` from this function is because we need to loop through (potentially) multiple servers to get the instance data, then the wrapper function itself calls the multi_connect_function below.

agent/php_internal_instrument.c
ZNeumann: Yes, the datastore metric needs to be send to the backend, and the length here isn't entirely important for several reasons. 1) this datastore metric is only an instance metric, and is not associated with any operation. 2) the operation being wrapped, `add_servers`, is not a function that the customer will likely care about timings of

agent/php_internal_instrument.c
ZNeumann: Yes, because we need to create several datastore metrics during a single underlying PHP call. There didn't exist code previously for that use case.

agent/php_internal_instrument.c
```diff
@@ -3511,6 +3599,10 @@ void nr_php_generate_internal_wrap_records(void) {
                       memcache_function, 0, "set")
   NR_INTERNAL_WRAPREC("memcached::setmultibykey", memcached_setmultibykey,
                       memcache_function, 0, "set")
+  NR_INTERNAL_WRAPREC("memcached::addserver", memcached_addserver,
+                      memcached_connect_function, 0, 0);
+  NR_INTERNAL_WRAPREC("memcached::addservers", memcached_addservers,
+                      memcached_multi_connect_function, 0, 0);
```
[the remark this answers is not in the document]
ZNeumann: For the other memcached wrappers, this extra field is used to name the datastore operation for the datastore metric. For addServer(s), it is just creating an instance metric without any specific operation, and therefore we have no need to store that information

agent/php_internal_instrument.c
ZNeumann: removed in [c8c3637](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/c8c3637f08b6375f7c92902dada51b124ba977e6)

agent/php_internal_instrument.c
ZNeumann: further details added in [c8c3637](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/c8c3637f08b6375f7c92902dada51b124ba977e6)

agent/php_internal_instrument.c
bduranleau-nr: It sounds like we're going to be sending incorrect duration timings to the backend for this specific function call. I'm slightly worried that someone out there will care about these timings, but off the top of my head I can't think of a workaround that will generate all the segments we need in this manner without forcing a rewrite of our segment logic.

ZNeumann: It could call the metric creation function after the internal call and pass in the timing information to override the timings automatically generated by the segment start/end. It's not pretty, but would work

bduranleau-nr: Or potentially a small function that can loop over a set of segments that have already had timings set and update `segment->start_time` and `segment->stop_time` to appropriate values? Store `start` and `stop` as variables within the wrapper function then passing them as the values to the custom function?
I don't know if that would have any side-effects...

axiom/nr_segment_datastore.c
```diff
@@ -12,6 +12,36 @@
 #include "util_sql.h"
 #include "util_logging.h"
 
+static bool create_instance_metric(nr_segment_t* segment,
```
bduranleau-nr, line 15: could nix the `bool` return type entirely since it doesn't seem to get used in the below function calls

agent/php_internal_instrument.c
ZNeumann: But this does raise the important question: will these metrics -- even with correct timings -- confuse customers? There will be much longer instance metrics associated with all of the different datastore operations of the transaction, and then a few much shorter instance metrics at the start of each transaction that is not clear what they are

tests/integration/memcached/test_add_servers.php
```diff
@@ -0,0 +1,34 @@
...
+Datastore/instance/Memcached/host2/2
+Datastore/instance/Memcached/host3/11211
+*/
+
+require_once(realpath (dirname ( __FILE__ )) . '/../../include/helpers.php');
+require_once(realpath (dirname ( __FILE__ )) . '/../../include/tap.php');
+require_once(realpath (dirname ( __FILE__ )) . '/memcache.inc');
+
+$memcached = new Memcached();
+$memcached->addServers(array(
```
bduranleau-nr, line 30: Can you add a test case for the `0` port option that specifies a unix socket/localhost?

agent/php_internal_instrument.c
bduranleau-nr: Torn on this as well- would be good to see how this will be displayed in the UI to try and gauge the "confusion" factor

ZNeumann: A correction to my previous statement. These instance metrics (with the host and port) are not created for other database operations because we don't have access to the host and port at those times. So these instance metrics will ONLY be the short-lived metrics created when adding a server, and the longer database metrics will be separated under operation metrics.

axiom/nr_segment_datastore.c
ZNeumann: It's unused but I still think that returning the bool is important for the completeness of this function. I could see at some point adding a log message, for example, if this returns false.

tests/integration/memcached/test_add_servers.php
ZNeumann: added in [8a3ad31](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/8a3ad31b2cec45c97f685b7bf191cfec56cea406)

agent/php_memcached.h
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef PHP_MEMCACHED_HDR
+#define PHP_MEMCACHED_HDR
+
+#include "nr_datastore_instance.h"
+#include "php_includes.h"
+
```
lavarou, line 11: Please add a doc string that will describe what does this function does.

agent/php_memcached.c
```diff
@@ -0,0 +1,18 @@
+#include "php_memcached.h"
```
lavarou, line 1:
```suggestion
/*
 * Copyright 2020 New Relic Corporation. All rights reserved.
 * SPDX-License-Identifier: Apache-2.0
 */

#include "php_memcached.h"
```

agent/php_memcached.c
```diff
@@ -0,0 +1,18 @@
...
+#include "nr_datastore_instance.h"
+#include "php_agent.h"
+
+nr_datastore_instance_t* nr_php_memcached_create_datastore_instance(
+    const char* host,
+    zend_long port) {
+  nr_datastore_instance_t* instance = NULL;
+  if (port == 0) { // local socket
+    instance = nr_datastore_instance_create("localhost", host, NULL);
```
lavarou, line 10: 1. Why is `host` used for `port_path_or_id` argument when `port` is unknown?
2. Does `0 == port` if and only if `host` == `"localhost"`? Is there a case when `0 == port` but `host` != `localhost`?

agent/php_memcached.c
```diff
@@ -0,0 +1,18 @@
...
+  } else {
+    char* port_str = nr_formatf("%ld", (long)port);
```
lavarou, line 12: `nr_formatf` seems a bit heavy handed for `itoa` conversion. FYI: there exists `nr_itoa`. Also can `zend_long` be safely cast to `long`?

lavarou: Looking at other places (e.g. [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.1/agent/lib_predis.c#L187)), this seems ok.

agent/php_memcached.c
ZNeumann: The parameters are explained here: https://www.php.net/manual/en/memcached.addserver.php#refsect1-memcached.addserver-parameters

As far as I can tell, 0 port is always and only used for accessing a localhost socket filepath

agent/php_memcached.h
ZNeumann: added in [a7bbbe9](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/a7bbbe9f6788070cd01868488c69f3dba745ba3a)

agent/php_memcached.c
lavarou: Knowing that, consider changing the name of the argument to `host_or_socket_name`. This name would match the doc string as well as be less confusing when used in this context.

ZNeumann: done in [d0aabeb](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/d0aabeb7846be00f2b0cb74ebecf5953aa45a84a)

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,63 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+NR_INNER_WRAPPER(memcached_connect_function) {
+  char* host = NULL;
+  nr_string_len_t host_len = 0;
+  zend_long port = 0;
+  zend_long weight = 0;
+  nr_datastore_instance_t* instance = NULL;
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "s|ll", &host,
```
lavarou, line 1576:
```suggestion
          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "s|ll", &host,
```

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,63 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+          &host_len, &port, &weight)) {
+    if (NULL != host) {
+      instance = nr_php_memcached_create_datastore_instance(host, port);
```
lavarou, line 1579: Where does `instance` get freed?

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,63 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+    }
+  }
+  nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
+                                             NULL, instance, true,
+                                             INTERNAL_FUNCTION_PARAM_PASSTHRU);
```
lavarou, line 1584: Is there a point in calling `nr_php_redis_save_datastore_instance` when `instance` is `NULL`?

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,63 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+  zval* servers = NULL;
+  zval* server = NULL;
+  zend_ulong num_key = 0;
+  nr_php_string_hash_key_t* string_key = NULL;
+  nr_datastore_instance_t* instance = NULL;
+  int zcaught = 0;
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "a", &servers)) {
```
lavarou, line 1597:
```suggestion
          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "a", &servers)) {
```

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,63 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+    ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(servers), num_key, string_key, server) {
+      (void)num_key;
+      (void)string_key;
```
lavarou, line 1600: If the `num_key` and `string_key` are not needed/used, then use `ZEND_HASH_FOREACH_VAL`:
```suggestion
    ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(servers), server) {
```

agent/php_internal_instrument.c
```diff
@@ -796,6 +800,35 @@ static void nr_php_instrument_datastore_operation_call(
...
+    nr_datastore_t datastore,
+    const char* operation,
+    nr_datastore_instance_t* instance,
+    bool instance_only) {
+  nr_segment_t* segment = NULL;
+  nr_segment_datastore_params_t params = {
+      .datastore = {
+          .type = datastore,
+      },
+      .operation = nr_strdup(operation),
```
lavarou, line 819: Since this function is used only for memcached_multi_connect, this strdup and later free could be avoided because memcached_multi_connect doesn't have any operation associated with it.

agent/php_internal_instrument.c
ZNeumann: It gets freed during nr_php_instrument_datastore_operation_call (or nr_php_instrument_datastore_operation when the underlying call isn't needed).

It is a somewhat confusing code structure, but is replicated from the redis instrumentation

agent/php_internal_instrument.c
ZNeumann: No, but `instance` should never be `NULL` and is still eventually checked. `instance` is a string created from `host`, which we check for `NULL` above. So `instance` should only be `NULL` if an allocation failed. Given that the code is still safe if that happens, I am disinclined to add a check.

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,63 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "a", &servers)) {
+    ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(servers), num_key, string_key, server) {
+      (void)num_key;
+      (void)string_key;
+      zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
+      zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
+      if (NULL != host) {
```
lavarou, line 1601: Check port for `NULL` as well?

agent/php_memcached.h
```diff
@@ -0,0 +1,22 @@
...
+/*
+ * Purpose : Create a datastore instance metadata for a Memcached connection.
+ *
+ * Params  : 1. The memcached host or socket name as given to Memcached::addServer().
+ *           2. The memcached port as given as given to Memcached::addServer().
```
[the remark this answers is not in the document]
ZNeumann: The caller owns the return value, but in all cases that this function is currently used, the caller immediately passes ownership to a called function. doc string updated in [39bc069](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/39bc06913fb9ad7d97be8d67cabd8473e0585145)

axiom/nr_segment_datastore.c
lavarou: How does `bool` return value make this function more complete? This function is not checking anything (`is_...`, `check_...`), it is only causing side effects (update `segment`). If anything, it could return whether it succeeded (`NR_SUCCESS`) or failed (`NR_FAILURE`) to accomplish the task.

axiom/nr_segment_datastore.c
```diff
@@ -208,11 +216,14 @@ bool nr_segment_datastore_end(nr_segment_t** segment_ptr,
...
-  scoped_metric
-      = create_metrics(segment, duration, datastore_string, collection,
-                       operation, &datastore, params->instance);
-
-  nr_segment_set_name(segment, scoped_metric);
+  if (!params->instance_only) {
+    scoped_metric
+        = create_metrics(segment, duration, datastore_string, collection,
+                         operation, &datastore, params->instance);
+    nr_segment_set_name(segment, scoped_metric);
```
lavarou, line 223: Should the segment get the name in the case only instance metrics get created as well? Is the segment name important in such case at all?

axiom/nr_segment_datastore.c
lavarou, line 1: Please add unit tests for `nr_segment_datastore_end` to test the new 'bool' flag that affects what this function will do.

agent/php_memcached.c
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "php_memcached.h"
+#include "nr_datastore_instance.h"
+#include "php_agent.h"
+
+nr_datastore_instance_t* nr_php_memcached_create_datastore_instance(
```
lavarou, line 10: This looks like a good candidate for unit tests. At least test for invalid input.

agent/php_internal_instrument.c
ZNeumann: this call was moved inside the `if`'s scope in [155bc5f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/155bc5ff645557bf0d001f691b2982fbf51fcf48)

axiom/nr_segment_datastore.c
```diff
@@ -12,6 +12,36 @@
...
 
+static bool create_instance_metric(nr_segment_t* segment,
+                                   const char* product,
+                                   nr_segment_datastore_t* datastore,
+                                   const nr_datastore_instance_t* instance) {
+  /*
+   * If a datastore instance was provided, we need to add the relevant data to
+   * the segment and the relavant metrics.
+   */
+  nrtxn_t* txn = segment->txn;
```
mfulb, line 23: Check for `segment` being `NULL`?

axiom/nr_segment_datastore.c
```diff
@@ -12,6 +12,36 @@
...
+  char* instance_metric = NULL;
+  if (NULL == instance || 0 == txn->options.instance_reporting_enabled) {
```
mfulb, line 25: Check for `txn` being `NULL`?

axiom/nr_segment_datastore.c
```diff
@@ -12,6 +12,36 @@
...
+    return false;
+  }
+
+  if (txn->options.database_name_reporting_enabled) {
+    nr_datastore_instance_set_database_name(&datastore->instance,
+                                            instance->database_name);
+  }
+
+  instance_metric = nr_formatf("Datastore/instance/%s/%s/%s", product,
```
mfulb, line 34: Are we concerned about the performance of this versus constructing it by concatting pieces?

axiom/nr_segment_datastore.c
```diff
@@ -24,7 +54,6 @@ static char* create_metrics(nr_segment_t* segment,
   char* rollup_metric = NULL;
   char* scoped_metric = NULL;
   char* statement_metric = NULL;
-  char* instance_metric = NULL;
```
mfulb, line 27: I don't see how to add a comment for existing code - but on the line above this (line 52 when writing this)
`nrtxn_t* txn = segment->txn;`
I believe it would be important to add a `NULL` check for `segment`.

tests/integration/memcached/test_add_servers.php
mfulb: Can we also add a test where an empty array is passed to `addServers()`?

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,67 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                             nr_wrapper->extra, NULL,
+                                             nr_wrapper->extra, NULL, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+NR_INNER_WRAPPER(memcached_connect_function) {
+  char* host = NULL;
+  nr_string_len_t host_len = 0;
+  zend_long port = 0;
```
zsistla, line 1570:
```suggestion
  int zcaught = 0;
  zend_long port = 0;
```

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,67 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+      /* NOTREACHED */
+    }
+  }
+}
+
+NR_INNER_WRAPPER(memcached_multi_connect_function) {
+  zval* servers = NULL;
+  zval* server = NULL;
+  zend_ulong num_key = 0;
+  nr_php_string_hash_key_t* string_key = NULL;
```
zsistla, line 1597:
```suggestion
```

These values don't appear to be used.

axiom/nr_segment_datastore.c
```diff
@@ -50,28 +79,7 @@ static char* create_metrics(nr_segment_t* segment,
...
-
-  if (txn->options.database_name_reporting_enabled) {
-    nr_datastore_instance_set_database_name(&datastore->instance,
-                                            instance->database_name);
-  }
-
-  instance_metric = nr_formatf("Datastore/instance/%s/%s/%s", product,
-                               instance->host, instance->port_path_or_id);
-  nr_segment_add_metric(segment, instance_metric, false);
-  nr_datastore_instance_set_host(&datastore->instance, instance->host);
```
zsistla, line 69: These lines seem separate from the rest that was extracted into `create_instance_metric` and seems like it should have stayed in original function.

agent/php_internal_instrument.c
```diff
@@ -796,6 +800,35 @@ static void nr_php_instrument_datastore_operation_call(
...
+      },
+      .operation = nr_strdup(operation),
+      .instance  = instance,
+      .instance_only = instance_only,
+      .callbacks = {
+          .backtrace = nr_php_backtrace_callback,
+      },
+  };
+
+  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
```
zsistla, line 827: Do we need to create a datastore segment?  Since it's not associated with any datastore calls, should it create the segment?  Would it be sufficient to simply send an instance metric?

tests/integration/memcached/test_add_servers.php
ZNeumann: added in [3cef8c3](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/3cef8c34afaa4358f2fa080e2d865b2c84390f76)

axiom/nr_segment_datastore.c
ZNeumann: changed to a status in [8460dfa](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/8460dfaa7a2ffa68a7b8f8b123ba48d2717102c6)

agent/php_internal_instrument.c
```diff
@@ -796,6 +800,35 @@ static void nr_php_instrument_datastore_operation_call(
   }
 }
 
+/* 
+ * Like nr_php_instrument_datastore_operation_call() but does not
+ * call the underlying internal function. This allows multiple
+ * datastore metrics to be creating during a single underlying
+ * PHP call.
```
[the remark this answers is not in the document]
ZNeumann: It functions exactly like `nr_php_instrument_datastore_operation_call()`, but does not call the underlying function. So I am not sure of the exact definition of "instrument" here, but if the other function instruments, then I think so does this one? But ultimately, yes, it exists to create datastore instance metrics.

agent/php_internal_instrument.c
ZNeumann: Because `operation` is `NULL` in this case, the impact is minimal. Because there is a potential use case for this function with an actual defined `operation`, I would rather leave it as is.

agent/php_internal_instrument.c
ZNeumann: port is a zend_long, which as far as I was aware cannot be `NULL`?

agent/php_internal_instrument.c
lavarou: I was thinking the same. I did some digging into the code and unfortunately it looks like separating instance metric creation from other pieces of `nr_segment_datastore_end` is rather involved (if not impossible). But I'm curious of @ZNeumann reasoning behind it.

axiom/nr_segment_datastore.c
ZNeumann: unit test here: [f01b0c0](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/f01b0c004e319a6a0b0ec94d6f78a8edb20ab375)

agent/php_internal_instrument.c
```diff
@@ -768,6 +770,7 @@ static void nr_php_instrument_datastore_operation_call(
     nr_datastore_t datastore,
     const char* operation,
     nr_datastore_instance_t* instance,
+    bool instance_only,
```
lavarou, line 773: This just feels really awkward that `nr_php_instrument_datastore_operation_call` is used solely to create instance metrics. The name of this function strongly suggests that it is interested in datastore operation, whereas the `bool` flag forces this suggestion to be discarded. For what this PR tries to do, I'd expect new function (or functions): `nr_php_create_datastore_instance_metric_call` (`nr_php_create_datastore_instance_metric`) and this function remain unchanged. Current implementation seems to work, however reading the code requires additional mental power.

agent/php_internal_instrument.c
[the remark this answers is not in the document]
lavarou: This function is written to be generic, but it's currently only used in one special case - memcached add servers - which is not even datastore operation. I have similar reservation pertaining to its name as I expressed [earlier](https://github.com/newrelic/newrelic-php-agent/pull/958/files#r1777637152).

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,66 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
  */
 NR_INNER_WRAPPER(memcache_function) {
   nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_MEMCACHE,
-                                             nr_wrapper->extra, NULL,
+                                             nr_wrapper->extra, NULL, false,
                                              INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+NR_INNER_WRAPPER(memcached_connect_function) {
```
lavarou, line 1567: This function has nothing to do with any type of connection, it is a wrapper for [`Memcached::addServer`](https://www.php.net/manual/en/memcached.addserver.php). I suggest keeping the name close to what it instruments:
```suggestion
NR_INNER_WRAPPER(memcached_add_server) {
```

agent/php_internal_instrument.c
ZNeumann: I initially did try to create this metric without also doing the datastore segment. I agree that it would be very difficult to do, and -- considering this seems to be the only current use case for such a paradigm -- decided against it

agent/php_internal_instrument.c
ZNeumann: fixed in [c2396ae](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/c2396ae73c4e70af513e62416c3e42dd2ac8e6f1)

agent/php_internal_instrument.c
ZNeumann: fixed in [c2396ae](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/c2396ae73c4e70af513e62416c3e42dd2ac8e6f1)

agent/php_internal_instrument.c
```diff
@@ -1527,10 +1560,66 @@ NR_INNER_WRAPPER(mysqli_stmt_prepare) {
...
+    zcaught = nr_zend_call_old_handler(nr_wrapper->oldhandler,
+                                       INTERNAL_FUNCTION_PARAM_PASSTHRU);
+    if (zcaught) {
+      zend_bailout();
+      /* NOTREACHED */
+    }
+  }
+}
+
+NR_INNER_WRAPPER(memcached_multi_connect_function) {
```
lavarou, line 1594: This function has nothing to do with any type of connection, it is a wrapper for [Memcached::addServers](https://www.php.net/manual/en/memcached.addservers.php). I suggest keeping the name close to what it instruments:
```suggestion
NR_INNER_WRAPPER(memcached_addservers_function) {
```

agent/php_memcached.c
ZNeumann: added in [9ba4554](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/9ba4554faa8946ae0a2258192b48a9e388ec0f67)

axiom/nr_segment_datastore.c
ZNeumann: the segment is checked for NULL by the caller

axiom/nr_segment_datastore.c
ZNeumann: `txn` is checked for `NULL` by the caller. `datastore` is a local variable of the caller.

axiom/nr_segment_datastore.c
ZNeumann: `segment` is checked for `NULL` by the caller.

agent/tests/test_memcached.c
```diff
@@ -0,0 +1,95 @@
...
+      nr_php_memcached_create_datastore_instance("host.name", 0));
+
+  assert_datastore_instance_equals_destroy(
+      "host and port",
+      &((nr_datastore_instance_t){
+          .host = "host.name",
+          .database_name = "unknown",
+          .port_path_or_id = "6379",
+      }),
+      nr_php_memcached_create_datastore_instance("host.name", 6379));
```
lavarou, line 82: Add a test case for `nr_php_memcached_create_datastore_instance(NULL, 0)`

axiom/nr_segment_datastore.c
ZNeumann: refactored in [e2d7a78](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/e2d7a786bd73da46eef4fb0de581839e38e8e17f)

agent/php_internal_instrument.c
lavarou: It is `zval *` as defined on line 1605:
```
1605: zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
```
`zval *` can be `NULL`.

agent/tests/test_memcached.c
ZNeumann: added in [d36a469](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/d36a469047758f9464e27f2af5da80a033400b27)

agent/php_internal_instrument.c
ZNeumann: renamed in [bc833b7](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/bc833b7c0cda48d97bccc0dfd2a26ec622502253)

agent/php_internal_instrument.c
ZNeumann: renamed in [bc833b7](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/bc833b7c0cda48d97bccc0dfd2a26ec622502253)

agent/php_internal_instrument.c
ZNeumann: Ah, I was looking at add_server, where it is a zend_long. Added the NULL check here: [faf547f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/faf547f126dbc7e99f7305539129004f944e6a62)

axiom/nr_segment_datastore.c
mfulb: We don't have a good convention currently to alert when we can depend on a function to accept `NULL` or not.  In this case it is a local static function so perhaps it is tightly coupled to the calling code in the same file and the `NULL` check could be relaxed.  I am just concerned in general that it is possible for new code to call these functions and cause a segfault if they also do not enforce the NULL check.  I've always leaned towards robustness at all levels versus the very slight overhead of redundant checking.  Not sure how to resolve this for this PR or the agent code in general.

tests/integration/memcached/test_add_servers.php
```diff
@@ -0,0 +1,35 @@
...
+                       array("host1", 1),
+                       array("host2", 2),
+                       array("host3", 11211)));
+$memcached->addServers(array());
```
lavarou, line 34: Can following tests be added as well:
- with non-array
- with an array of array that only has one element - host name a as a string
- with an array of array that only has one element - host name as a number
- with an array of array that has two elements - different combos
- with an array of array that has more than 3 elements

tests/integration/memcached/test_add_servers.php
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+/*INI
+*/
+
+/*EXPECT_METRICS_EXIST
+Datastore/instance/Memcached/host1/1
+Datastore/instance/Memcached/host2/2
+Datastore/instance/Memcached/host3/11211
+*/
```
lavarou, line 23: This test should not generate errors, correct? If so, please add this expectation:
```suggestion
*/

/*EXPECT_ERROR_EVENTS null */
```

tests/integration/memcached/test_socket.php
```diff
@@ -0,0 +1,45 @@
...
+    [{"name":"OtherTransaction/all"},                                 [1, "??", "??", "??", "??", "??"]],
+    [{"name":"OtherTransaction/php__FILE__"},                         [1, "??", "??", "??", "??", "??"]],
+    [{"name":"OtherTransactionTotalTime"},                            [1, "??", "??", "??", "??", "??"]],
+    [{"name":"OtherTransactionTotalTime/php__FILE__"},                [1, "??", "??", "??", "??", "??"]],
+    [{"name":"Supportability/Logging/Forwarding/PHP/enabled"},        [1, "??", "??", "??", "??", "??"]],
+    [{"name":"Supportability/Logging/Metrics/PHP/enabled"},           [1, "??", "??", "??", "??", "??"]],
+    [{"name":"Supportability/Logging/LocalDecorating/PHP/disabled"},  [1, "??", "??", "??", "??", "??"]]
+  ]
+]
+*/
```
lavarou, line 37: Please replace with `EXPECT_METRICS_EXIST`.

tests/integration/memcached/test_add_servers.php
ZNeumann: As of [e141481](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/e141481475416099da72d1cab770d748844640e8), we no longer expect no errors

tests/integration/memcached/test_add_servers.php
ZNeumann: I have added some additional tests (and more robust type handling in the code) here [e141481](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/e141481475416099da72d1cab770d748844640e8)

This cannot be tested with a non-array, as that crashes PHP

tests/integration/memcached/test_add_servers.php
lavarou: I still don't see `/*EXPECT_ERROR_EVENTS null */` in this test.

tests/integration/memcached/test_add_servers.php
lavarou: > This cannot be tested with a non-array, as that crashes PHP

Could you add commented test case for that with your explanation, please?

tests/integration/memcached/test_add_servers.php
ZNeumann: With the added test cases, PHP has warnings about argument types begin incorrect, so there are now errors expected.

agent/php_internal_instrument.c
ZNeumann: very good idea. I have made the change here [3ed5bc5](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/3ed5bc54bb0d90bc792ed32cd650cfb87c7de972) and cleaned out the now-unused code here [ffb0711](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/ffb0711d8e562d4e1ece30c0dd9b557bc49d04ef)

tests/integration/memcached/test_add_servers.php
lavarou: If that's the case, then please separate the test cases that don't generate errors from the ones that do and add correct `/*EXPECT_ERROR_EVENTS` expectations.

agent/php_internal_instrument.c
lavarou: This comment is obsolete as of 3ed5bc54bb0d90bc792ed32cd650cfb87c7de972

axiom/nr_segment_datastore.c
lavarou: This comment is obsolete as of 3ed5bc54bb0d90bc792ed32cd650cfb87c7de972

agent/php_internal_instrument.c
lavarou: This thread seems obsolete after 3ed5bc54bb0d90bc792ed32cd650cfb87c7de972.

agent/php_internal_instrument.c
lavarou: This thread is obsolete as of 3ed5bc54bb0d90bc792ed32cd650cfb87c7de972.

axiom/nr_segment_datastore.c
lavarou: This is obsolete as of 3ed5bc54bb0d90bc792ed32cd650cfb87c7de972.

axiom/nr_segment_datastore.c
```diff
@@ -211,7 +211,6 @@ bool nr_segment_datastore_end(nr_segment_t** segment_ptr,
   scoped_metric
       = create_metrics(segment, duration, datastore_string, collection,
                        operation, &datastore, params->instance);
-
```
lavarou, line 214: Could you reformat this file in such a way that it doesn't show up in diff (`git checkout origin/dev -- axiom/nr_segment_datastore.c && git add `axiom/nr_segment_datastore.c && `git commit -m "restore original state" axiom/nr_segment_datastore.c` should do it)?

ZNeumann: fixed in [2140f54](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/2140f543ceee3f31462b975ad05328d1eeb4653d)

tests/integration/memcached/test_add_servers.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should report instance metrics when multiple servers are
+added at once
```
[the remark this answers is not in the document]
ZNeumann: not adding the link, but added the class::method in [ffddd54](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/ffddd547e266668dba56be35cffe1a8ce4ea20a7)

tests/integration/memcached/test_socket.php
ZNeumann: changed in [ac0d299](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/ac0d299523f5c0011a71a6f4572b5059cb587583)

tests/integration/memcached/test_add_servers.php
ZNeumann: split in [ffddd54](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/ffddd547e266668dba56be35cffe1a8ce4ea20a7)

agent/tests/test_memcached.c
```diff
@@ -0,0 +1,104 @@
...
+          .database_name = "unknown",
+          .port_path_or_id = "unknown",
+      }),
+      nr_php_memcached_create_datastore_instance(NULL, 0));
+}
+
+void test_main(void* p NRUNUSED) {
+#if defined(ZTS) && !defined(PHP7)
+  void*** tsrm_ls = NULL;
+#endif /* ZTS && !PHP7 */
```
bduranleau-nr, line 97: Is this block necessary? We don't ship ZTS anymore so this should never evaluate to true.

ZNeumann: removed in [24f3e6e](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/24f3e6ec8873335d356382e618399e67367ee24d)

bduranleau-nr (dismissed):

agent/php_internal_instrument.c
```diff
@@ -1531,6 +1533,71 @@ NR_INNER_WRAPPER(memcache_function) {
...
+  nr_datastore_instance_t* instance = NULL;
+  char* instance_metric = NULL;
+  int zcaught = 0;
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "s|ll", &host,
+          &host_len, &port, &weight) &&
+      NULL != host) {
+    instance = nr_php_memcached_create_datastore_instance(host, port);
```
zsistla, line 1550: does creating it as an `instance` format the host and/or path in a special way?
Can we just pass host and port directly to the [nr_format](https://github.com/newrelic/newrelic-php-agent/pull/958/files#diff-d66d558f126dbc42fe98d5332fa13d281fe925d4e01b7608a8dc39c56dce778eR1551) command?  ten we wouldn't have to create/free the `instance`.

zsistla: Oh, I see, [it is a little bit of special checking](https://github.com/newrelic/newrelic-php-agent/pull/958/files#diff-50b3bcc41a255bfafa2a5b71bdf4d9ec0d8165f881ddeb9bbeb0118f2d31c843R10).

zsistla (approved):

agent/php_internal_instrument.c
```diff
@@ -1531,6 +1533,71 @@ NR_INNER_WRAPPER(memcache_function) {
...
+    instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
+                               instance->host, instance->port_path_or_id);
+    nrm_force_add(NRPRG(txn)->unscoped_metrics, instance_metric, 0);
+    nr_free(instance);
```
zsistla, line 1554: [nr_free](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/util_memory.h#L159) only frees the instance pointer, it doesn't free the strduped fields it was pointing to.

This needs [nr_datastore_instance_destroy](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/nr_datastore_instance.c#L37) to nicely cleanup all the strings that were strduped.

ZNeumann: addressed in [18fff7a](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/18fff7abc952385e4eea2925f631906fc7463e3b) and [c40cbd4](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/c40cbd42f74ca295924f55fbd6a9deab996b2085)

zsistla (dismissed):

agent/php_internal_instrument.c
```diff
@@ -1531,6 +1533,71 @@ NR_INNER_WRAPPER(memcache_function) {
...
+  zval* servers = NULL;
+  zval* server = NULL;
+  nr_datastore_instance_t* instance = NULL;
+  char* instance_metric = NULL;
+  int zcaught = 0;
+
+  if (SUCCESS
+      == zend_parse_parameters_ex(
+          ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "a", &servers)) {
+    if (Z_TYPE_P(servers) == IS_ARRAY) {
```
lavarou, line 1575: Ensure that `servers` is not `NULL` before dereferencing it in `Z_TYPE_P` macro:
```suggestion
    if (NULL != servers && Z_TYPE_P(servers) == IS_ARRAY) {
```

agent/tests/test_memcached.c
```diff
@@ -0,0 +1,100 @@
...
+      nr_php_memcached_create_datastore_instance("", 6379));
+
+  assert_datastore_instance_equals_destroy(
+      "localhost socket",
+      &((nr_datastore_instance_t){
+          .host = system_host_name,
+          .database_name = "unknown",
+          .port_path_or_id = "localhost",
+      }),
+      nr_php_memcached_create_datastore_instance("localhost", 0));
```
lavarou, line 55: How is this test different from [this one](https://github.com/newrelic/newrelic-php-agent/blob/c40cbd42f74ca295924f55fbd6a9deab996b2085/agent/tests/test_memcached.c#L21-L28)? Use of `"localhost"` as the socket is confusing and it does not test any additional code paths than [this one](https://github.com/newrelic/newrelic-php-agent/blob/c40cbd42f74ca295924f55fbd6a9deab996b2085/agent/tests/test_memcached.c#L21-L28).

agent/tests/test_memcached.c
```diff
@@ -0,0 +1,100 @@
...
+
+tlib_parallel_info_t parallel_info
+    = {.suggested_nthreads = -1, .state_size = 0};
+
+static char* system_host_name;
+
+static void test_create_datastore_instance(void) {
+  /*
+   * Test : Normal operation.
+   */
```
lavarou, line 20: I'm not sure which tests this comment applies to first one or all? If only the first one, then each subsequent should have a similar comment, which IMHO is redundant because the first argument passed to `assert_datastore_instance_equals_destroy` serves the same purpose. If all tests, then it is not accurate, because the last test that passes `NULL` as hostname does not seem Normal operation.
```suggestion

```

lavarou: The comment would be useful to group tests by the value of `port` passed to `nr_php_memcached_create_datastore_instance`. I.e. if `port == 0`, then first argument of `nr_php_memcached_create_datastore_instance` is used for `port_path_or_id` in the datastore instance created. If `port != 0` then first argument is used for `host` and second argument is used as `port_path_or_id`.

agent/tests/test_memcached.c
lavarou, line 1: This test should test the following cases for arguments passed to `nr_php_memcached_create_datastore_instance`:
Normal operation:
- "host.domain", 6379 - normal operation with remote host
- "/tmp/memcached.sock", 0 - normal operation with socket on localhost
Invalid inputs:
- "", 6379 - empty host, valid port results in ???
- NULL, 6379 - null host, valid port results in ???
- "", 0 - empty host, empty port results in ???
- NULL, 0 - null host, empty port results in ???'

Can you think of any more test cases?

tests/integration/memcached/test_basic.php
```diff
@@ -46,6 +46,7 @@
     [{"name":"Datastore/allOther"},                                   [15, "??", "??", "??", "??", "??"]],
     [{"name":"Datastore/Memcached/all"},                              [15, "??", "??", "??", "??", "??"]],
     [{"name":"Datastore/Memcached/allOther"},                         [15, "??", "??", "??", "??", "??"]],
+    [{"name":"Datastore/instance/Memcached/memcached/11211"},          [1, "??", "??", "??", "??", "??"]],
```
lavarou, line 49: Where do the values of `memcached` as host and `11211` come from?

tests/integration/memcached/test_socket.php
```diff
@@ -0,0 +1,31 @@
...
+
+/*EXPECT_ERROR_EVENTS null */
+
+require_once(realpath (dirname ( __FILE__ )) . '/../../include/helpers.php');
+require_once(realpath (dirname ( __FILE__ )) . '/../../include/tap.php');
+require_once(realpath (dirname ( __FILE__ )) . '/memcache.inc');
+
+$memcached = new Memcached();
+$memcached->addServer("my_socket", 0);
```
lavarou, line 30: Could this be tested with an actual path to the socket? I.e. have the first argument include `/` (forward slash) directory separator?

tests/integration/memcached/test_basic.php
ZNeumann: memcached is set up in the docker-compose [here](https://github.com/newrelic/newrelic-php-agent/blob/main/docker-compose.yaml#L80) and [here](https://github.com/newrelic/newrelic-php-agent/blob/main/docker-compose.yaml#L33). The default port is 11211, but we set the environment variable [here](https://github.com/newrelic/newrelic-php-agent/blob/main/tests/include/config.php#L28).

lavarou: If this information is available in the environment, then please update these tests to use environment variables instead of hardcoding these values. Here's an example of what I mean:
```suggestion
    [{"name":"Datastore/instance/Memcached/ENV[MEMCACHE_HOST]/ENV[MEMCACHE_PORT]"},          [1, "??", "??", "??", "??", "??"]],
```

lavarou (changes requested):
Please address all comments and refactor the code.

agent/php_internal_instrument.c
```diff
@@ -1531,6 +1533,71 @@ NR_INNER_WRAPPER(memcache_function) {
...
+        zval* host = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 0);
+        zval* port = nr_php_zend_hash_index_find(Z_ARRVAL_P(server), 1);
+        if (NULL != host && NULL != port &&
+            Z_TYPE_P(host) == IS_STRING && Z_TYPE_P(port) == IS_LONG) {
+          instance = nr_php_memcached_create_datastore_instance(Z_STRVAL_P(host), Z_LVAL_P(port));
+          instance_metric = nr_formatf("Datastore/instance/Memcached/%s/%s",
+                                       instance->host, instance->port_path_or_id);
+          nrm_force_add(NRPRG(txn)->unscoped_metrics, instance_metric, 0);
+          nr_datastore_instance_destroy(&instance);
+          nr_free(instance_metric);
```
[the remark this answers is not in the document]
lavarou: Once the helper function is factored out, it can be unit tested. Please add unit tests for the helper function.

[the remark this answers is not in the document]
ZNeumann: Adding unit tests to this helper function would be both cumbersome and unnecessary. The creation of the instance is already tested. nr_formatf is already tested. Adding a metric to a txn is already tested. The entire block is already tested via integration tests, and there are no additional code paths to test.

tests/integration/memcached/test_basic.php
ZNeumann: MEMCACHE_PORT is a PHP variable, not an environment variable, and as far as I can tell there is therefore no easy way to replace the hard coded value? Because it is hard coded in many places, is the default, and is unlikely to ever change, I am inclined to leave it as is

tests/integration/memcached/test_socket.php
ZNeumann: added in [8ec0268](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/8ec02684f0bb8726ae111d78d37e0dd492d662c5)

agent/tests/test_memcached.c
```diff
@@ -0,0 +1,100 @@
...
+      nr_php_memcached_create_datastore_instance("localhost", 0));
+
+  assert_datastore_instance_equals_destroy(
+      "localhost and port",
+      &((nr_datastore_instance_t){
+          .host = system_host_name,
+          .database_name = "unknown",
+          .port_path_or_id = "6379",
+      }),
+      nr_php_memcached_create_datastore_instance("localhost", 6379));
```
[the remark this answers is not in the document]
ZNeumann: organized and consolidated unit tests in [5597d7f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/5597d7fccf8657254c4c9d486612ef821b2754be)

agent/tests/test_memcached.c
ZNeumann: organized and consolidated unit tests in [5597d7f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/5597d7fccf8657254c4c9d486612ef821b2754be)

agent/tests/test_memcached.c
ZNeumann: organized and consolidated unit tests in [5597d7f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/5597d7fccf8657254c4c9d486612ef821b2754be)

agent/tests/test_memcached.c
ZNeumann: organized and consolidated unit tests in [5597d7f](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/5597d7fccf8657254c4c9d486612ef821b2754be)

agent/tests/test_memcached.c
ZNeumann: removed comment in [62b8889](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/62b888920b6a143f6c3d6dad923ad1fa61951af8)

agent/php_memcached.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#ifndef PHP_MEMCACHED_HDR
+#define PHP_MEMCACHED_HDR
+
+#include "nr_datastore_instance.h"
+#include "php_includes.h"
+
+/*
+ * Purpose : Create a datastore instance metadata for a Memcached connection.
```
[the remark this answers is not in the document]
ZNeumann: reworded in [1b227d3](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/1b227d32f7e7cd82f86796fde7cc02c7c1779e5f)

agent/php_internal_instrument.c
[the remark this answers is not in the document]
lavarou: Also checking how this function reacts to invalid input is something that unit test should verify.

tests/integration/memcached/test_basic.php
lavarou: I don't like the fact that port is hardcoded but changing this is not part of this PR. This is good enough.

agent/php_memcached.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+nr_datastore_instance_t* nr_php_memcached_create_datastore_instance(
+    const char* host_or_socket,
+    zend_long port);
+
+/*
+ * Purpose : Create a memcached instance metric
+ *
+ * Params  : 1. The memcached host or socket name as given to Memcached::addServer().
+ *              Must be non-null.
```
lavarou, line 28: Since this function is a public API, this cannot be guaranteed and therefore this function needs to account for invalid input.

tests/integration/memcached/test_socket.php
```diff
@@ -0,0 +1,31 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should report instance metrics when multiple servers are
+added at once
```
lavarou, line 9: This is not accurate.

ZNeumann: fixed in [3364eb0](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/3364eb0b8bbf0e984add9030588f82ff3475ed05)

agent/php_memcached.h
ZNeumann: removed this comment in [c8bdfb7](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/c8bdfb78b275c8e6ff9faff01a709d1c5853bc00). The function can handle NULL

agent/php_internal_instrument.c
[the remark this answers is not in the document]
ZNeumann: added tests, including NULL tests, in [6e720c9](https://github.com/newrelic/newrelic-php-agent/pull/958/commits/6e720c9b372aa04176e4e3aaa031681157f21680)

zsistla (approved):

agent/php_internal_instrument.c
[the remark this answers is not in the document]
lavarou: Thank you.

PR #960, opened 2024-09-09
fix(agent): Fixes newrelic_notice_error() API for PHP 8+

mfulb:
Starting with PHP 8.0 the `set_error_handler()` callback function signature drops the `$errcontext` final parameter so it only contains 4 parameters.  This caused using `newrelic_notice_error()` as a callback handler for PHP 8+ to not work as the API call did not accept only 4 arguments.  This PR adds support for this function signature.  There are also numerous tests added that will hopefully caught this kind of issue in the future.

### DIFF
--- a/tests/integration/api/notice_error/test_good_1_arg_string.php
+++ b/tests/integration/api/notice_error/test_good_1_arg_string.php
@@ -6,7 +6,7 @@
 
 /*DESCRIPTION
 The agent should record a traced error when newrelic_notice_error is
-called with 1 argument which is an string.
+called with 1 argument which is a string.
 */
 
 /*EXPECT_TRACED_ERRORS
--- a/tests/integration/api/notice_error/test_good_1_arg_string.php
+++ b/tests/integration/api/notice_error/test_good_1_arg_string.php
@@ -1,0 +1,101 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should record a traced error when newrelic_notice_error is
+called with 1 argument which is an string.
+*/
+
+/*EXPECT_TRACED_ERRORS
+[
+  "?? agent run id",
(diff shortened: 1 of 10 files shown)


tests/integration/api/notice_error/test_good_1_arg_string.php
```diff
@@ -0,0 +1,99 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should record a traced error when newrelic_notice_error is
+called with 1 argument which is an string.
```
zsistla, line 9:
```suggestion
called with 1 argument which is a string.
```

tests/integration/api/notice_error/test_good_4_args_error_handler.php
```diff
@@ -0,0 +1,118 @@
...
+        "spanId": "??"
+      },
+      {},
+      {}
+    ]
+  ]
+]
+*/
+
+/*EXPECT_SPAN_EVENTS_LIKE
```
zsistla, line 66: Might make sense to add an EXPECT_ANALYTICS_EVENTS block to verify the attributes were added there as well.

hahuja2 (dismissed):

bduranleau-nr (dismissed):

bduranleau-nr (approved):

tests/integration/api/notice_error/test_good_1_arg_string.php
```diff
@@ -0,0 +1,162 @@
...
+      {},
+      {
+        "errorType": "NoticedError",
+        "errorMessage": "Exception 1 arg string"
+      }
+    ]
+  ]
+]
+*/
+
```
hahuja2, line 155:
```suggestion
```

hahuja2 (dismissed):

hahuja2 (approved):

zsistla (approved):

PR #962, opened 2024-09-12
feat(agent): use composer for vuln mgmt package info

lavarou:
If possible, use [Composer's runtime API](https://getcomposer.org/doc/07-runtime.md) to collect information about PHP packages used by the application for [New Relic Vulnerability Management](https://docs.newrelic.com/docs/vulnerability-management/overview/). This feature is disabled by default and can be enabled by setting `newrelic.vulnerability_management.composer_detection.enabled` to `true`.


agent/lib_composer.c
```diff
@@ -0,0 +1,257 @@
...
+  char* code = NULL;
+  zval retval;
+  int result = -1;
+
+  if (nr_execute_handle_autoload_composer_is_initialized()) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: already initialized", __func__);
+    return NR_SUCCESS;
+  }
+
+#if 0
```
zsistla, line 45: What is this code block supposed to do if it's always false?

agent/lib_composer.c
```diff
@@ -0,0 +1,257 @@
...
+  code = nr_formatf("include_once '%s/composer/InstalledVersions.php';",
+                    vendor_path);
+#endif
+
+  result = zend_eval_string(code, &retval, "newrelic\\init_composer_api");
+  if (result != SUCCESS) {
+    nrl_verbosedebug(NRL_INSTRUMENT,
+                     "%s: zend_eval_string(%s) failed, result=%d", __func__,
+                     code, result);
+    return NR_FAILURE;
```
zsistla, line 76: If we return here, code doesn't get freed.
I'm not sure what the state of retval is and if it also needs a zval_dtor if the eval_string fails.

zsistla:
What version(s) of composer will this work with?

agent/lib_composer.c
```diff
@@ -0,0 +1,257 @@
...
+              "%s - unable to initialize Composer runtime API - package info "
+              "unavailable",
+              __func__);
+    return;
+  }
+
+  nrl_verbosedebug(NRL_INSTRUMENT, "%s - Composer runtime API available",
+                   __func__);
+
+#if 1
```
zsistla, line 129: For debugging?

daemon/internal/newrelic/integration/php_packages.go
```diff
@@ -371,6 +371,18 @@ func (pkgs *PhpPackagesCollection) GatherInstalledPackages() ([]PhpPackage, erro
...
+		version := ""
+		for _, line := range lines {
+			//fmt.Printf("line is |%s|\n", line)
+			splitLine := strings.Split(line, "=>")
+			if 2 == len(splitLine) {
+				name := strings.TrimSpace(splitLine[0])
+				version = strings.TrimSpace(splitLine[1])
+				pkgs.packages = append(pkgs.packages, PhpPackage{name, version})
+			}
+		}
```
lavarou, line 385: This is added so that composer mock can be used to test packages payload.

agent/lib_composer.c
lavarou: The code in `#if 0` part is from the initial implementation of this feature. However I determined that all it does is what is in the `#else` part. I left this in for now so that I can easily switch the implementations to demonstrate that there's no difference in behavior regardless of the method used. The final version will only have the `#else` part.

agent/lib_composer.c
```diff
@@ -0,0 +1,257 @@
...
+
+  if (!nr_execute_handle_autoload_composer_file_exists(vendor_path,
+                                                       COMPOSER_MAGIC_FILE_3)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "'%s' not found in '%s'",
+                     COMPOSER_MAGIC_FILE_3, vendor_path);
+    return;
+  }
+
+  nrl_verbosedebug(NRL_FRAMEWORK, "detected composer");
+  NRPRG(txn)->composer_info.composer_detected = true;
```
lavarou, line 252: This flag is not needed and the final version will probably have this removed.

lavarou:
> What version(s) of composer will this work with?

So far it's been tested with 2.2 and 2.6. It should work with all versions >= 2.2.

agent/lib_composer.c
lavarou: Thanks for pointing it out. Memory leak is fixed in 060303ef42268a0f5d4949bbd3df42d800dd650b. I need to take a closer look at `retval`.

lavarou: https://github.com/newrelic/newrelic-php-agent/pull/962/commits/aab8068fb61792807fce5eac658bb126ee314223 changes how the agent interacts with Composer's runtime API - there's no longer need to allocate memory for the PHP code to be executed, because the code string is static. This patch is currently under test, but if it passes it will become the final version.

As far as `retval` goes `zend_eval_string` leaves it in good state as long as it returns `SUCCESS`. Once that happens, `zval_dtor` can be called. https://github.com/newrelic/newrelic-php-agent/pull/962/commits/aab8068fb61792807fce5eac658bb126ee314223 incorporates this approach.

agent/lib_composer.c
```diff
@@ -135,9 +136,13 @@ static void nr_execute_handle_autoload_composer_get_packages_information(
         = ""
         "(function() {"
         "  try {"
+        "    $root_package = \\Composer\\InstalledVersions::getRootPackage();"
         "    $packages = array();"
         "    foreach (\\Composer\\InstalledVersions::getAllRawData() as $installed) { "
         "      foreach ($installed['versions'] as $packageName => $packageData) {"
+        "        if ($packageName == @$root_package['name']) {"
+        "          continue;"
+        "        }"
```
lavarou, line 145: Make this more robust, like:
```
if (is_array($installed) && array_key_exists('root', $installed)) {
    $root_package = $installed['root'];

    if (is_array($root_package) && array_key_exists('name', $root_package) && 'foobar' == $root_package['name']) {
        echo "root package is foobar!" . PHP_EOL;
    } else {
        echo "not foobar!" . PHP_EOL;
    }
}
```

lavarou: Addressed with 2713cd3cb574e8c2e980166a0ed5a1e7d0a58e30.

agent/fw_drupal8.c
```diff
@@ -687,10 +688,10 @@ void nr_drupal_version() {
   if (nr_php_is_zval_valid_string(zval_version)) {
     char* version = Z_STRVAL_P(zval_version);
     if (NRINI(vulnerability_management_package_detection_enabled)) {
-      nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
+      p = nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME, version);
     }
-    nr_fw_support_add_package_supportability_metric(NRPRG(txn), PHP_PACKAGE_NAME,
-                                                    version);
+    nr_fw_support_add_package_supportability_metric(
+        NRPRG(txn), PHP_PACKAGE_NAME, version, p);
```
lavarou, line 694: This change (and similar changes in other `fw_*.c` modules) is required so that if composer api is enabled, the major version value, used in the metric name, will be from package info collected using composer api rather than from the package itself (from static VERSION constant or calling some version of `get_version` method in the package). The former method yields more useful and accurate results.

axiom/nr_php_packages.h
```diff
@@ -10,20 +10,43 @@
...
 
+typedef enum {
+  NR_PHP_PACKAGE_SOURCE_LEGACY,
+  NR_PHP_PACKAGE_SOURCE_COMPOSER
+} nr_php_package_source_priority_t;
+
 typedef struct _nr_php_package_t {
   char* package_name;
   char* package_version;
+  nr_php_package_source_priority_t source_priority;
```
lavarou, line 25: `source_priority` decides if `package_version` gets overwritten when a package with the same `package_name` is added to `nr_php_packages_t.data` again - see implementation of `nr_php_packages_add_package`.

axiom/tests/test_php_packages.c
```diff
@@ -203,4 +261,5 @@ void test_main(void* p NRUNUSED) {
   test_php_packages_to_json();
   test_php_package_exists_in_hashmap();
```
zsistla, line 262: Would be good to have a composer package added followed by composer package added for completeness.
`test_php_package_exists_in_hashmap` only tests legacy package followed by legacy package and `test_php_package_priority` only tests mixed packages.

agent/fw_drupal.c
```diff
@@ -20,6 +20,8 @@
 #include "util_memory.h"
 #include "util_strings.h"
 
+#define PHP_PACKAGE_NAME "drupal/drupal"
```
bduranleau-nr, line 23: [fw_drupal8.c](https://github.com/newrelic/newrelic-php-agent/blob/956976b451cd72a433a414d25eca6b9e5f2b06c4/agent/fw_drupal8.c#L21) defines this value as `"drupal/core"`. Is the difference here to use `"drupal/drupal"` intentional?

agent/fw_support.c
```diff
@@ -58,23 +58,39 @@ void nr_fw_support_add_logging_supportability_metric(nrtxn_t* txn,
...
+    return;
+  }
 
   /* The below for loop checks if the major version of the package is more than
    * one digit and keeps looping until a '.' is encountered or one of the
    * conditions is met.
    */
-  for (int i = 0; package_version[i] && i < MAJOR_VERSION_LENGTH - 1; i++) {
-    if ('.' == package_version[i]) {
+  for (int i = 0; version[i] && i < MAJOR_VERSION_LENGTH - 1; i++) {
```
bduranleau-nr, line 89: I'm a little confused by this loop logic- why the inclusion of `version[i]` here? Is there some guarantee that when this `const char*` is assigned that we aren't reading outside the bounds of the array? It doesn't really look like the code checks this currently and instead just seems to be the result of a `Z_STRVAL_P` operation

bduranleau-nr: I guess we're using it to check for the `'\0'` character which is functionally equivalent to `0` in C and break the loop in that case, assuming that every potential variation of the source of the `version` value is null-terminated.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  return NR_SUCCESS;
+}
+
+static void nr_execute_handle_autoload_composer_get_packages_information(
+    const char* vendor_path) {
+  zval retval;  // This is used as a return value for zend_eval_string.
+                // It will only be set if the result of the eval is SUCCESS.
+  int result = -1;
+
+  // nurunlikely because this should alredy be ensured by the caller
```
bduranleau-nr, line 77:
```suggestion
  // nrunlikely because this should alredy be ensured by the caller
```

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  if (nrunlikely(!NRINI(vulnerability_management_package_detection_enabled))) {
+    // do nothing when collecting package information for vulnerability
+    // management is disabled
+    return;
+  }
+
+  // nurunlikely because this should alredy be ensured by the caller
```
bduranleau-nr, line 84:
```suggestion
  // nrunlikely because this should alredy be ensured by the caller
```

bduranleau-nr: Actually I think you can remove both of these entirely- it doesn't seem possible to reach this function without having already checked for VM/ComposerAPI to be enabled in php_execute.c where `nr_composer_handle_autoload` is called. This method is also static, so unless you envision it being called by some other function that may not check for these values, I think it makes sense to remove both `nrunlikely` calls and just document the expectation in the function definition comment.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  if (nrunlikely(!NRINI(vulnerability_management_composer_api_enabled))) {
+    // do nothing when use of composer to collect package info is disabled
+    return;
+  }
+
+  // clang-format off
+  char* getallrawdata
```
bduranleau-nr, line 91: Should this be a `const char*` since it doesn't seem to get modified after definition?
```suggestion
  const char* getallrawdata
```

agent/tests/test_fw_support.c
```diff
@@ -11,7 +11,12 @@
 tlib_parallel_info_t parallel_info
     = {.suggested_nthreads = -1, .state_size = 0};
 
-static void test_fw_supportability_metrics(void) {
+// When package detection for vulnerability management is disbled,
```
hahuja2, line 14:
```suggestion
// When package detection for vulnerability management is disabled,
```

agent/php_newrelic.h
```diff
@@ -593,6 +593,8 @@ nrinibool_t
 nrinibool_t
     vulnerability_management_package_detection_enabled; /* newrelic.vulnerability_management.package_detection.enabled
                                                          */
+nrinibool_t
+    vulnerability_management_composer_api_enabled; /* newrelic.vulnerability_management.composer_api.enabled */
```
zsistla, line 597: I couldn't see any unit/integration tests where `vulnerability_management_package_detection_enabled;` is disabled but `newrelic.vulnerability_management.composer_api.enabled` is enabled.  Would be good to have a couple of tests.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+static int nr_execute_handle_autoload_composer_init(const char* vendor_path) {
+  char* code = NULL;
+  zval retval;
+  int result = -1;
```
zsistla, line 38:
```suggestion
  int result = FAILURE;
```

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  return NR_SUCCESS;
+}
+
+static void nr_execute_handle_autoload_composer_get_packages_information(
+    const char* vendor_path) {
+  zval retval;  // This is used as a return value for zend_eval_string.
+                // It will only be set if the result of the eval is SUCCESS.
+  int result = -1;
```
zsistla, line 75:
```suggestion
  int result = FAILURE;
```

You are comparing it to a [zend_result](https://github.com/php/php-src/blob/cf0a44a38a512e238db490282adf80682200d296/Zend/zend_types.h#L59) later, so we should ensure we stay in sync with the zend_result values.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+                   __func__);
+
+  result
+      = zend_eval_string(getallrawdata, &retval, "composer_getallrawdata.php");
+  if (SUCCESS != result) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s - composer_getallrawdata.php failed",
+                     __func__);
+    return;
+  }
+  if (IS_ARRAY == Z_TYPE(retval)) {
```
zsistla, line 132: could also consider checking for IS_NULL so we know if an exception occurred.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+        "        if (is_array($root_package) && array_key_exists('name', $root_package) && $packageName == $root_package['name']) {"
+        "          continue;"
+        "        }"
+        "        if (isset($packageData['pretty_version'])) {"
+        "          $packages[$packageName] = ltrim($packageData['pretty_version'], 'v');"
+        "        }"
+        "      }"
+        "    }"
+        "    return $packages;"
+        "  } catch (Exception $e) {"
```
zsistla, line 108: Is there a test case where calling this causes an exception?

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+    ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL(retval), package_name,
+                                  package_version) {
+      if (NULL == package_name || NULL == package_version) {
+        continue;
+      }
+      if (nr_php_is_zval_non_empty_string(package_version)) {
+        nrl_verbosedebug(NRL_INSTRUMENT, "package %s, version %s",
+                         NRSAFESTR(ZSTR_VAL(package_name)),
+                         NRSAFESTR(Z_STRVAL_P(package_version)));
+        nr_txn_add_php_package_from_source(
```
zsistla, line 144: Doesn't NRSAFESTR return the string `"<NULL>"` when it encounters a null?
So if using NRSAFESTR with the call to nr_txn_add_php_package_from_source, a null package_name/package_version will actually send valid strings (that say NULL) and won't hit the check for a null package_name [here](https://github.com/newrelic/newrelic-php-agent/pull/962/files#diff-aff431af9f447c5431ea6466c518a37c44b2491884a829f6c9e4675082ca8186R3509) or [here](https://github.com/newrelic/newrelic-php-agent/pull/962/files#diff-bffd93c552e94ddfafc8445c041ff9e4727959267c975fa3c62f67c185561df7R40)
And it'll miss the check for a null version [here](https://github.com/newrelic/newrelic-php-agent/pull/962/files#diff-bffd93c552e94ddfafc8445c041ff9e4727959267c975fa3c62f67c185561df7R50)

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  char* vendor_path = NULL;  // result of dirname(filename)
+  char* cp = NULL;
+
+  // vendor_path = dirname(filename):
+  // 1. copy filename to vendor_path
+  vendor_path = nr_strdup(filename);
+  // 2. // find last occurence of '/' in vendor_path
+  cp = nr_strrchr(vendor_path, '/');
+  // 3. replace '/' with '\0' to get the directory path
+  *cp = '\0';
```
zsistla, line 169: Need to check if cp is NULL before assigning a value.
nr_strrchr will return NULL if it doesn't find `'/'`

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+}
+
+static char* nr_execute_handle_autoload_composer_get_vendor_path(
+    const char* filename) {
+  char* vendor_path = NULL;  // result of dirname(filename)
+  char* cp = NULL;
+
+  // vendor_path = dirname(filename):
+  // 1. copy filename to vendor_path
+  vendor_path = nr_strdup(filename);
```
zsistla, line 165: is there a check if filename is NULL?  Possibly not needed since strdup makes an empty string if filename is null.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  return vendor_path;
+}
+
+static bool nr_execute_handle_autoload_composer_file_exists(
+    const char* vendor_path,
+    const char* filename) {
+  char* composer_magic_file = NULL;  // vendor_path + filename
+  bool file_exists = false;
+
+  composer_magic_file = nr_formatf("%s/%s", vendor_path, filename);
```
zsistla, line 180: Is there a check for filename != NULL? otherwise, this will generate an error.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+#define COMPOSER_MAGIC_FILE_1_LEN (sizeof(COMPOSER_MAGIC_FILE_1) - 1)
+// Composer runtime API files:
+#define COMPOSER_MAGIC_FILE_2 "composer/InstalledVersions.php"
+#define COMPOSER_MAGIC_FILE_2_LEN (sizeof(COMPOSER_MAGIC_FILE_2) - 1)
+#define COMPOSER_MAGIC_FILE_3 "composer/installed.php"
+#define COMPOSER_MAGIC_FILE_3_LEN (sizeof(COMPOSER_MAGIC_FILE_3) - 1)
+  char* vendor_path = NULL;  // result of dirname(filename)
+
+  vendor_path = nr_execute_handle_autoload_composer_get_vendor_path(filename);
+  if (NULL == vendor_path) {
```
zsistla, line 200: will vendor_path ever be null since nrstrdup [here](https://github.com/newrelic/newrelic-php-agent/pull/962/files#diff-d5f4e22aae64a7055a773470d7d61e625e9d9a782162e0a2b3d0c0638ce12682R165) will always return an empty string if given a null.

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+    nrl_verbosedebug(NRL_FRAMEWORK, "unable to get vendor path from '%s'",
+                     filename);
+    return;
+  }
+
+  if (!nr_execute_handle_autoload_composer_file_exists(vendor_path,
+                                                       COMPOSER_MAGIC_FILE_1)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "'%s' not found in '%s'",
+                     COMPOSER_MAGIC_FILE_1, vendor_path);
+    return;
```
zsistla, line 210: vendor_path needs to be freed

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  if (!nr_execute_handle_autoload_composer_file_exists(vendor_path,
+                                                       COMPOSER_MAGIC_FILE_2)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "'%s' not found in '%s'",
+                     COMPOSER_MAGIC_FILE_2, vendor_path);
+    return;
```
zsistla, line 217: vendor_path needs to be freed

agent/lib_composer.c
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  if (!nr_execute_handle_autoload_composer_file_exists(vendor_path,
+                                                       COMPOSER_MAGIC_FILE_3)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "'%s' not found in '%s'",
+                     COMPOSER_MAGIC_FILE_3, vendor_path);
+    return;
```
zsistla, line 224: vendor_path needs to be freed

tests/integration/autoloader/autoload-with-broken-composer-01/vendor/composer/installed.php
```diff
@@ -0,0 +1,31 @@
...
+*/
+  return array(
+    // Mocked data: root package
+    'root' => array(
+        'pretty_version' => 'v1.0.0',
+        'version' => '1.0.0.0',
+        'type' => 'project'
+    ),
+    // Mocked data: installed packages and their versions
+    'versions' =>  array(
```
zsistla, line 19: What about adding some tests where the package name is null version is valid, version is null name is valid, both version/package are null.

axiom/tests/test_php_packages.c
lavarou: Does 0ab342b51ed1be5f93331e182caa8669b81af2c1 address your concern?

agent/lib_composer.c
lavarou: @zsistla Would it be enough for a mocked `\\Composer\\InstalledVersions::getRootPackage()` or `\\Composer\\InstalledVersions::getAllRawData()` throw an `\Exception`? Or you have something else in mind?

agent/lib_composer.c
lavarou: Checking for `NULL` out of habit. I'm going to leave it in just in case `nr_execute_handle_autoload_composer_get_vendor_path` ever changes in a way that will result in `NULL` being a possible return value.

agent/lib_composer.c
lavarou: Fixed in fd3f639501a84d36e23fd773ee8ad166176b8330.

agent/lib_composer.c
lavarou: Fixed in fd3f639501a84d36e23fd773ee8ad166176b8330.

agent/lib_composer.c
lavarou: Fixed in fd3f639501a84d36e23fd773ee8ad166176b8330.

agent/lib_composer.c
lavarou: @zsistla Will printing the returned value with `nr_format_zval_for_debug` in `else` clause [here](https://github.com/newrelic/newrelic-php-agent/blob/feat/vm/user-composer-for-package-info/agent/lib_composer.c#L151-L154) be good enough?

zsistla: Good idea, that would be great.

agent/lib_composer.c
zsistla: That would be sufficient.

ZNeumann (approved):

agent/lib_composer.c
lavarou: `filename` is guaranteed to be not `NULL` here by [this](https://github.com/newrelic/newrelic-php-agent/blob/feat/vm/user-composer-for-package-info/agent/php_execute.c#L1026-L1032). However, I'll add `NULL` checks for `filename`.

agent/lib_composer.c
lavarou: `filename` is guaranteed to be not `NULL` here by [this](https://github.com/newrelic/newrelic-php-agent/blob/feat/vm/user-composer-for-package-info/agent/php_execute.c#L1026-L1032). However, I'll add `NULL` checks for `filename`.

lavarou: Added in cb83165d367509f3bb3439ad14b0aa61a21225c3.

agent/lib_composer.c
lavarou: Added in cb83165d367509f3bb3439ad14b0aa61a21225c3.

agent/lib_composer.c
lavarou: Implemented in af1170a5c377631bbb238bb48b527a7310935503.

agent/lib_composer.c
lavarou: @zsistla `package_name` and `package_version` are guaranteed to not be `NULL` [here](https://github.com/newrelic/newrelic-php-agent/blob/cb83165d367509f3bb3439ad14b0aa61a21225c3/agent/lib_composer.c#L137). Additionally, `package_version` is guaranteed to be a non-empty string [here](https://github.com/newrelic/newrelic-php-agent/blob/cb83165d367509f3bb3439ad14b0aa61a21225c3/agent/lib_composer.c#L140). I don't see a way for either `package_name` or `package_version` ending up as `"<NULL>"` generated by use of `NRSAFESTR`. To be frank, because of those earlier `NULL` checks, `NRSAFESTR` could be dropped altogether.

agent/lib_composer.c
lavarou: Good point! Addressed in 388c5d085d80f0d369de5e8e94209378f39587b6.

agent/lib_composer.c
zsistla: Sounds appropriate to drop NRSAFESTR when passed to nr_txn_add_php_package_from_source, it didn't seem the right way to use it there, but if you remove it from nr_txn_add_php_package_from_source, I'm good with it.

lavarou: Dropped `NRSAFESTR` in 3a356ea74f4fdc5cb74c09589fd98248b4d8bc2f.

agent/lib_composer.c
lavarou: This actually had to be reverted because `zend_eval_string` expects `char *` not `const char *`, so making `code` a `const char *` does not make much sense because it would have to be casted back to `char *`.

agent/lib_composer.c
zsistla: NRSAFESTR was okay (and safer to leave) when passing to nrl_*.
I only objected to it when passing to nr_txn_add_php_package_from_source

lavarou: The [same argumentation](https://github.com/newrelic/newrelic-php-agent/pull/962#discussion_r1773696404) applies when `NRSAFESTR` is used in `nrl_verbosedebug`. Let's be consistent.

agent/lib_composer.c
lavarou: Added in a73bbda1233929e99c92b475a8171142ec8924fd.

agent/php_newrelic.h
lavarou: Added in f6b438aa1a1ffcb7b693d27cf0aed1f7a56b895d.

tests/integration/autoloader/autoload-with-broken-composer-01/vendor/composer/installed.php
lavarou: Added in 5a73606d8399bcf6199bbdef4efb58bed536af6d.

agent/lib_composer.c
zsistla: I don't believe the same argument applies.  nrl_verbosedebug does not handle NULLs so needs to be passed a safestr where NULL has been converted to  the string "<NULL>" to display and would segfault if passed an actual NULL.
whereas nr_txn_add_php_package_from_source does handle NULLs fine and would in fact behave incorrectly if the string "<NULL>" were passed to it instead of an actual NULL.  
If we dealing with anything to do with zvals (which are technically shared space values that PHP or another entity could muck with), it's better to be careful (the reason why we usually strdup anything quickly that we've gotten from a zval).

lavarou: `package_name` and `package_version` are guaranteed to not be NULL [here](https://github.com/newrelic/newrelic-php-agent/blob/cb83165d367509f3bb3439ad14b0aa61a21225c3/agent/lib_composer.c#L137). Additionally, `package_version` is **guaranteed to be a non-empty string** [here](https://github.com/newrelic/newrelic-php-agent/blob/cb83165d367509f3bb3439ad14b0aa61a21225c3/agent/lib_composer.c#L140) and therefore use of `NRSAFESTR` could be dropped no matter where the value of at least `package_version` is accessed. However, as requested I restored use of `NRSAFESTR` when `package_name` and `package_version` are accessed for debug log message in 51f6d43a7dd9a41b4466ea6e9dd3796c6af9c656. Maybe a better approach would be to guarantee that `package_name` is a non-empty string with `nr_php_is_zval_non_empty_string` to strengthen [this check](https://github.com/newrelic/newrelic-php-agent/blob/cb83165d367509f3bb3439ad14b0aa61a21225c3/agent/lib_composer.c#L137).

zsistla (dismissed):
Thanks for all the hard work that went into this!  LGTM.

hahuja2 (dismissed):

mfulb (approved):
Very nice work Michal!

hahuja2 (approved):

agent/lib_composer.c
bduranleau-nr: Are you sure? According to [php-src](https://github.com/php/php-src/blob/master/Zend/zend_execute_API.c#L1383) `zend_eval_string` expects a `const char *` for it's first argument, not `char *`

lavarou: Yes, I'm sure - https://github.com/newrelic/newrelic-php-agent/actions/runs/11016894669/job/30593507308#step:7:197:
```
/usr/local/src/newrelic-php-agent/agent/lib_composer.c: In function 'nr_execute_handle_autoload_composer_get_packages_information':
/usr/local/src/newrelic-php-agent/agent/lib_composer.c:126:26: error: passing argument 1 of 'zend_eval_string' discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
       = zend_eval_string(getallrawdata, &retval, "composer_getallrawdata.php");
                          ^~~~~~~~~~~~~
In file included from /usr/local/include/php/Zend/zend_API.h:31,
                 from /usr/local/src/newrelic-php-agent/agent/php_includes.h:18,
                 from /usr/local/src/newrelic-php-agent/agent/php_agent.h:38,
                 from /usr/local/src/newrelic-php-agent/agent/lib_composer.c:6:
/usr/local/include/php/Zend/zend_execute.h:49:37: note: expected 'char *' but argument is of type 'const char *'
 ZEND_API int zend_eval_string(char *str, zval *retval_ptr, char *string_name);
                               ~~~~~~^~~
```

bduranleau-nr: For future reference: PHP 7.2 - 7.4 do not have the `const` modifier. `const` was added to this function arg in PHP 8.0.

PR #965, opened 2024-09-25
feat(agent): improve package major version metric creation

mfulb:
Change how the package major number metrics are created. This now occurs
in R_SHUTDOWN.  This allows the use of package versions from ALL sources
including the Composer API.  Whenever instrumentation for a package detects
a package, it can create a package suggestion.  Initially if a version is not
known it is fine to use PHP_PACKAGE_VERSION_UNKNOWN.  If later the
instrumentation determines a version from a class constant, etc, then
the suggestion can be updated with the version.

At the end of the transaction the suggestions are iterated over and
the actual package data (which could include Composer API data) is
referenced and a major number supportability metric is created with the
best version available.

### DIFF
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -713,10 +713,10 @@ void nr_php_txn_create_agent_php_version_metrics(nrtxn_t* txn) {
   nr_php_txn_create_php_version_metric(txn, version);
 }
 
-static void nr_php_txn_php_package_create_major_metric(void* value,
-                                                       const char* key,
-                                                       size_t key_len,
-                                                       void* user_data) {
+void nr_php_txn_php_package_create_major_metric(void* value,
+                                                const char* key,
+                                                size_t key_len,
+                                                void* user_data) {
   nrtxn_t* txn = (nrtxn_t*)user_data;
   nr_php_package_t* suggested = value;
   nr_php_package_t* actual = NULL;
@@ -738,12 +738,12 @@ static void nr_php_txn_php_package_create_major_metric(void* value,
   actual
       = nr_php_packages_get_package(txn->php_packages, suggested->package_name);
 
-  nrl_verbosedebug(NRL_INSTRUMENT,
-                   "Creating PHP Package Supportability Metric for package "
-                   "'%s', suggested version '%s', actual version '%s'",
-                   NRSAFESTR(suggested->package_name),
-                   NRSAFESTR(suggested->package_version),
-                   NRSAFESTR(NULL != actual ? actual->package_version : ""));
+  nrl_verbosedebug(
+      NRL_INSTRUMENT,
+      "Creating PHP Package Supportability Metric for package "
+      "'%s', suggested version '%s', actual version '%s'",
+      NRSAFESTR(suggested->package_name), NRSAFESTR(suggested->package_version),
+      NRSAFESTR(NULL != actual ? actual->package_version : "NULL"));
   nr_fw_support_add_package_supportability_metric(
       txn, suggested->package_name, suggested->package_version, actual);
 }
@@ -753,7 +753,7 @@ void nr_php_txn_create_packages_major_metrics(nrtxn_t* txn) {
     return;
   }
 
-  nr_php_packages_iterate(txn->php_package_suggestions,
+  nr_php_packages_iterate(txn->php_package_major_version_metrics_suggestions,
                           nr_php_txn_php_package_create_major_metric, txn);
 }
 
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -19,10 +19,12 @@
 #include "nr_agent.h"
 #include "nr_commands.h"
 #include "nr_header.h"
+#include "nr_php_packages.h"
 #include "nr_rum.h"
 #include "nr_segment_children.h"
 #include "nr_txn.h"
 #include "nr_version.h"
+#include "fw_support.h"
 #include "util_labels.h"
 #include "util_logging.h"
 #include "util_memory.h"
@@ -711,6 +713,40 @@ void nr_php_txn_create_agent_php_version_metrics(nrtxn_t* txn) {
   nr_php_txn_create_php_version_metric(txn, version);
 }
 
+static void nr_php_txn_php_package_create_major_metric(void* value,
+                                                       const char* key,
+                                                       size_t key_len,
+                                                       void* user_data) {
+  nrtxn_t* txn = (nrtxn_t*)user_data;
+  nr_php_package_t* package = value;
+
+  (void)key;
+  (void)key_len;
+
+  if (NULL == package) {
+    return;
+  }
+
+  if (package->options & NR_PHP_PACKAGE_OPTION_MAJOR_METRIC) {
+    nrl_verbosedebug(NRL_INSTRUMENT,
+                     "Creating PHP Package Supportability Metric for package "
+                     "'%s', version '%s'",
+                     NRSAFESTR(package->package_name),
+                     NRSAFESTR(package->package_version));
+    nr_fw_support_add_package_supportability_metric(
+        txn, package->package_name, package->package_version, NULL);
+  }
+}
+
+void nr_php_txn_create_packages_major_metrics(nrtxn_t* txn) {
+  if (NULL == txn) {
+    return;
+  }
+
+  nr_php_packages_iterate(txn->php_packages,
+                          nr_php_txn_php_package_create_major_metric, txn);
+}
+
 nr_status_t nr_php_txn_begin(const char* appnames,
                              const char* license TSRMLS_DC) {
   nrtxnopt_t opts;
@@ -1164,6 +1200,9 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
 
     /* Agent and PHP version metrics*/
     nr_php_txn_create_agent_php_version_metrics(txn);
+
+    /* PHP packages major number metrics */
+    nr_php_txn_create_packages_major_metrics(txn);
 
     /* Add CPU and memory metrics */
     nr_php_resource_usage_sampler_end(TSRMLS_C);
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -724,6 +724,10 @@ static void nr_php_txn_php_package_create_major_metric(void* value,
   (void)key;
   (void)key_len;
 
+  if (NULL == txn) {
+    return;
+  }
+
   if (NULL == suggested) {
     return;
   }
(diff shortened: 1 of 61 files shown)


axiom/nr_php_packages.h
```diff
@@ -43,7 +49,10 @@ typedef struct _nr_php_packages_t {
  *           nr_php_packages_add_package() is not called, then it must be freed
  *           by nr_php_package_destroy()
  */
-extern nr_php_package_t* nr_php_package_create_with_source(char* name, char* version, const nr_php_package_source_priority_t source_priority);
+extern nr_php_package_t* nr_php_package_create_with_source(
+    const char* name,
+    const char* version,
+    const nr_php_package_source_priority_t source_priority);
```
lavarou, line 55: Formatting only, right?

axiom/nr_php_packages.h
```diff
@@ -58,7 +67,8 @@ extern nr_php_package_t* nr_php_package_create_with_source(char* name, char* ver
  *           nr_php_packages_add_package() is not called, then it must be freed
  *           by nr_php_package_destroy()
  */
-extern nr_php_package_t* nr_php_package_create(char* name, char* version);
+extern nr_php_package_t* nr_php_package_create(const char* name,
+                                               const char* version);
```
lavarou, line 71: Formatting only, right?

axiom/nr_php_packages.h
```diff
@@ -149,8 +159,9 @@ static inline int nr_php_packages_has_package(nr_php_packages_t* h,
  *
  * Returns : Returns pointer to php package if the package exists or NULL
  */
-static inline nr_php_package_t* nr_php_packages_get_package(nr_php_packages_t* php_packages,
-                                              const char* package_name) {
+static inline nr_php_package_t* nr_php_packages_get_package(
+    nr_php_packages_t* php_packages,
+    const char* package_name) {
```
lavarou, line 164: Formatting only, right?

agent/php_txn.c
```diff
@@ -711,6 +713,46 @@ void nr_php_txn_create_agent_php_version_metrics(nrtxn_t* txn) {
...
+                   "Creating PHP Package Supportability Metric for package "
+                   "'%s', suggested version '%s', actual version '%s'",
+                   NRSAFESTR(suggested->package_name),
+                   NRSAFESTR(suggested->package_version),
+                   NRSAFESTR(actual ? actual->package_version : ""));
+  nr_fw_support_add_package_supportability_metric(
+      txn, suggested->package_name, suggested->package_version, actual);
+}
+
+void nr_php_txn_create_packages_major_metrics(nrtxn_t* txn) {
```
lavarou, line 747: I think this is a candidate for unit tests.

axiom/nr_php_packages.h
mfulb: Yes I ran it through `git-clang-format`.

mfulb: No I changed the pointers to be `const` as I saw no reason for them to not be.
Then `git-clang-format` did the rest.

axiom/nr_php_packages.h
mfulb: No I changed the pointers to be `const` as I saw no reason for them to not be.
Then `git-clang-format` did the rest.

axiom/nr_php_packages.h
mfulb: Yes this change is solely from `git-clang-format`.

axiom/nr_txn.h
```diff
@@ -1196,8 +1198,26 @@ nr_php_package_t* nr_txn_add_php_package_from_source(
  *
  * Returns : pointer to added package on success or NULL otherwise.
  */
-nr_php_package_t* nr_txn_add_php_package(nrtxn_t* txn,
-                                         char* package_name,
-                                         char* package_version);
+extern nr_php_package_t* nr_txn_add_php_package(nrtxn_t* txn,
+                                                char* package_name,
+                                                char* package_version);
```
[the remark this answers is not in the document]
mfulb: I was just making all the prototypes consistent is using `extern`.

agent/php_txn.c
```diff
@@ -711,6 +713,46 @@ void nr_php_txn_create_agent_php_version_metrics(nrtxn_t* txn) {
...
+  nr_php_package_t* suggested = value;
+  nr_php_package_t* actual = NULL;
+
+  (void)key;
+  (void)key_len;
+
+  if (NULL == suggested) {
+    return;
+  }
+
```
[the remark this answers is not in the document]
mfulb: Thanks I've added this check.

axiom/nr_txn.c
```diff
@@ -3519,3 +3521,24 @@ nr_php_package_t* nr_txn_add_php_package(nrtxn_t* txn,
...
+    return;
+  }
+
+  nrl_verbosedebug(NRL_TXN, "Suggesting package %s %s", NRSAFESTR(package_name),
+                   NRSAFESTR(package_version));
+
+  p = nr_php_package_create_with_source(package_name, package_version,
+                                        NR_PHP_PACKAGE_SOURCE_SUGGESTION);
+  nr_php_packages_add_package(txn->php_package_suggestions, p);
+}
```
lavarou, line 3544: What should this function do when `package_version` is `NULL`?

lavarou: FYI: `agent/tests/test_lib_aws_sdk_php.c` is testing current behavior [here](https://github.com/newrelic/newrelic-php-agent/blob/4ea22b6bc5d62439e690a5e2885e318a0b9b1a84/agent/tests/test_lib_aws_sdk_php.c#L124-L147).

agent/php_txn.c
```diff
@@ -1165,6 +1211,9 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
     /* Agent and PHP version metrics*/
     nr_php_txn_create_agent_php_version_metrics(txn);
 
+    /* PHP packages major number metrics */
```
lavarou, line 1214:
```suggestion
    /* PHP packages major version metrics */
```

axiom/nr_php_packages.c
```diff
@@ -116,6 +121,16 @@ nr_php_package_t* nr_php_packages_add_package(nr_php_packages_t* h,
   return p;
 }
 
+void nr_php_packages_iterate(nr_php_packages_t* packages,
+                             nr_php_packages_iter_t callback,
+                             void* userdata) {
+  if (NULL == packages || NULL == callback) {
```
lavarou, line 127: Eventual check for `NULL == userdata` happens in `callback`, yes?

lavarou: Ok, I see it happens [here](https://github.com/newrelic/newrelic-php-agent/blob/4ea22b6bc5d62439e690a5e2885e318a0b9b1a84/agent/php_txn.c#L727-L729).

axiom/nr_txn.h
```diff
@@ -286,6 +286,8 @@ typedef struct _nrtxn_t {
       custom_events;               /* Custom events created through the API. */
   nr_log_events_t* log_events;     /* Log events pool */
   nr_php_packages_t* php_packages; /* Detected php packages */
+  nr_php_packages_t* php_package_suggestions; /* Suggested packages for major
```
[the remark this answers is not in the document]
mfulb: I changed it in [9feead2](https://github.com/newrelic/newrelic-php-agent/pull/965/commits/9feead2a917772c7e2a5f4705ac184fe6087537d).

agent/tests/test_php_txn.c
```diff
@@ -8,9 +8,9 @@
 #include "php_agent.h"
 #include "php_txn_private.h"
 
-#define LIBRARY_NAME "php/php-package"
-#define PACKAGE_VERSION "1.2.3"
-#define PACKAGE_MAJOR_VERSION "1"
+#define LIBRARY_NAME "vendor_name/package_name"
+#define LIBRARY_VERSION "1.2.3"
+#define LIBRARY_MAJOR_VERSION "1"
 #define COMPOSER_PACKAGE_VERSION "2.1.3"
```
zsistla, line 14: since we know composer version can give versions with letters in them, can we have a case with that?
something like `4.0.x-dev`

axiom/tests/test_php_packages.c
```diff
@@ -301,6 +302,74 @@ static void test_php_package_priority(void) {
   nr_php_packages_destroy(&hm);
 }
 
+static void nr_php_packages_itereate_callback(void* value,
+                                                const char* key,
+                                                size_t key_len,
+                                                void* user_data) {
+  nr_php_package_t* package = value;
```
zsistla, line 309: check if package is null?

axiom/nr_txn.c
zsistla: > What should this function do when `package_version` is `NULL`? It currently creates a package with `PHP_PACKAGE_VERSION_UNKNOWN` as suggested version.

That seems appropriate.

axiom/tests/test_php_packages.c
mfulb: In this test it shouldn't ever be `NULL` so maybe failing would be a good thing?  What is primarily being tested here is that the callback is called with each package in the packages list.  I put a `NULL` guard for `buf` because I wanted to test that `NULL` is passed as `userdata` properly.

agent/tests/test_php_txn.c
mfulb: Since it is presented as a string I don't know anything cares what ASCII values the string is composed of?

axiom/nr_txn.c
mfulb: My intention was to have `NULL` mapped to `PHP_PACKAGE_VERSION_UNKNOWN`.  This makes its use easier because many places in the instrumentation we get `NULL` for the version and not having to transform it in each place seems a better mechanism.  This way the version can just be passed into this function and it is known it will be handled properly.

agent/tests/test_php_txn.c
zsistla: do we not extract the major version from it? isn't there some string manipulation/scraping at that point?

zsistla (approved):

axiom/tests/test_php_packages.c
```diff
@@ -7,6 +7,7 @@
 #include "tlib_main.h"
 #include "util_memory.h"
 
+#include "stdio.h"
```
mfulb, line 10:
```suggestion
```

agent/tests/test_php_txn.c
lavarou, line 1: Nice tests! 👍 ➕

lavarou (approved):
Really nicely done!

PR #970, opened 2024-10-07
refactor(agent): improve magic file recognition performance

lavarou:
Speed up speed package detection by performing a suffix match on the 'magic' file pattern with case insensitive string comparison instead of a substring search within a lowercased filename. This is possible because all of the 'magic' file search patterns patterns are right anchored.

Fixup of e11b992c with changes from 24c1c656.


zsistla (approved):
nice optimization

Issue #399, opened 2022-03-30
Update install script to check for newrelic.ini in "mods-available"


ZNeumann:
Currently, the use of a "mods-available" dir to house the newrelic.ini is not checked ([even though it is a perfectly valid system configuration](https://discuss.newrelic.com/t/relic-solution-why-do-i-have-a-20-newrelic-ini/62574)). This causes issues during agent upgrades, as the install script is unable to find any newrelic.ini files and creates a new/additional ini.

ZNeumann:
Installing from tar using the `newrelic-install.sh` script will not properly find a mods-available directory. Installing from `apt` WILL find the mods-available directory IF it is located in `/etc/php/VERSION/`. A custom PHP installation location will not have its mods-available directory found.

PR #971, opened 2024-10-09
fix(installer): tarball installer checks for ini in mods-available

ZNeumann:
Debian uses a directory called "mods-available" to house ini files. It's `phpenmod` and `phpdismod` allow command-line management of extensions in directory. It also creates a symlink ini file to this "mods-available" directory in the normal conf.d directory. The issue is that this symlink is prefixed with a number (for example 20-newrelic.ini). This prefix prevents our tarball installer from recognizing that a newrelic.ini already exists, resulting in 2 ini files: a newrelic.ini and a 20-newrelic.ini.

Our .deb package installer scouts out the "mods-available" directory and installs therein when possible. That means that if a customer installs via package but upgrades via tarball (not recommended), then their system gets the duplicates.

This change makes our tarball installer scout for a potential previous installation in mods-available.

Fixes https://github.com/newrelic/newrelic-php-agent/issues/399

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -1386,7 +1386,7 @@ EOF
 
       if [ -d "${cfg_pfx}/mods-available" -a -f "${cfg_pfx}/mods-available/newrelic.ini" ]; then
         pi_inidir_cli="${cfg_pfx}/mods-available"
-        if [ -n ${pi_inidir_dso} ]; then
+        if [ -n "${pi_inidir_dso}" ]; then
           pi_inidir_dso="${cfg_pfx}/mods-available"
         fi
       fi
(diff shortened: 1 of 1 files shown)


agent/newrelic-install.sh
```diff
@@ -1371,6 +1371,25 @@ EOF
...
+      # While this install script will not install into the mods-available
+      # directory, our .deb installer can. Therefore, we want to detect if
+      # newrelic has previously been installed in the mods-available directory
+      # so that we do not create an additional ini file -- which would result in
+      # the conf.d directory having both newrelic.ini and 20-newrelic.ini.
+      #
+
+      if [ -d "${cfg_pfx}/mods-available" -a -f "${cfg_pfx}/mods-available/newrelic.ini" ]; then
+        pi_inidir_cli="${cfg_pfx}/mods-available"
+        if [ -n ${pi_inidir_dso} ]; then
```
lavarou, line 1389: Double quote reference to `${pi_inidir_dso}` to prevent globbing and word splitting:
```suggestion
        if [ -n "${pi_inidir_dso}" ]; then
```

lavarou (approved):

zsistla (approved):

PR #976, opened 2024-10-14
fix(agent): don't skip arguments when calling `mysqli::real_connect`

lavarou:


agent/php_mysqli.c
```diff
@@ -430,7 +434,13 @@ static nr_status_t nr_php_mysqli_link_real_connect(
     ADD_IF_STR_SET(argv, argc, metadata->database);
     ADD_IF_INT_SET(argv, argc, metadata->port);
     ADD_IF_STR_SET(argv, argc, metadata->socket);
-    ADD_IF_INT_SET(argv, argc, metadata->flags);
+
+    /*
+     * Avoid defaulting the final $flags argument to null.
+     */
+    if (metadata->flags) {
```
zsistla, line 441: won't this miss the case where the flags are actually set to the integer 0?

zsistla: It's okay to not add the flag param if it is zero, since if it's not added, it's zero anyway.

```
mysqli_real_connect(
    mysqli $mysql,
    ?string $hostname = null,
    ?string $username = null,
    #[\SensitiveParameter] ?string $password = null,
    ?string $database = null,
    ?int $port = null,
    ?string $socket = null,
    int $flags = 0
): bool
```

agent/php_mysqli.c
```diff
@@ -398,39 +398,54 @@ static nr_status_t nr_php_mysqli_link_real_connect(
...
-    ADD_IF_STR_SET(argv, argc, metadata->socket);
-    ADD_IF_INT_SET(argv, argc, metadata->flags);
+    ADD_IF_STR_SET(false, argv, argc, metadata->database);
+    ADD_IF_INT_SET(false, argv, argc, metadata->port);
+    ADD_IF_STR_SET(false, argv, argc, metadata->socket);
+
+    /*
+     * Avoid defaulting the final $flags argument to null.
+     */
+    if (metadata->flags) {
```
kovshenin, line 446: Do we still need this check if we can use is_set/default_null to accomplish the same check?

agent/php_mysqli.c
```diff
@@ -398,39 +398,54 @@ static nr_status_t nr_php_mysqli_link_real_connect(
...
+    ZVAL_NULL(args[argc]);                        \
+    argc++;                                       \
+  }
+
+#define ADD_IF_STR_SET(is_set, args, argc, value) \
+  if (value) {                                    \
+    args[argc] = nr_php_zval_alloc();             \
+    nr_php_zval_str(args[argc], value);           \
+    argc++;                                       \
+  } else if (false == is_set) {                   \
```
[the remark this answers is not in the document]
zsistla: Renamed variable to `null_ok` to indicate null is acceptable or not.
Please see: [b1c7ab5](https://github.com/newrelic/newrelic-php-agent/pull/976/commits/b1c7ab503d97f9c0b67dba9d5a92ed36a8c35902)

agent/php_mysqli.c
zsistla: Good point, removed the check here: [b1c7ab5](https://github.com/newrelic/newrelic-php-agent/pull/976/commits/b1c7ab503d97f9c0b67dba9d5a92ed36a8c35902)

kovshenin: :+1:

ZNeumann (approved):

bduranleau-nr (approved):

PR #983, opened 2024-10-30
feat(agent): add support for CakePHP 4.x and 5.x

hahuja2:
This PR does the following:

- Modifies how the action name is retrieved, so that it is compatible with CakePHP 4.x and 5.x
- Adds correct detection file for CakePHP 4.x and 5.x
- Adds call to `nr_txn_suggest_package_supportability_metric` in `nr_cakephp_enable`

### DIFF
--- a/agent/fw_cakephp.c
+++ b/agent/fw_cakephp.c
@@ -321,9 +321,11 @@ void nr_cakephp_enable_2(TSRMLS_D) {
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
   nr_php_wrap_user_function_before_after_clean(
-      NR_PSTR("CakeException::__construct"), nr_cakephp_problem_2, NULL, NULL);
+      NR_PSTR("Cake\\Core\\Exception\\CakeException::__construct"),
+      nr_cakephp_problem_2, NULL, NULL);
 #else
-  nr_php_wrap_user_function(NR_PSTR("CakeException::__construct"),
-                            nr_cakephp_problem_2 TSRMLS_CC);
+  nr_php_wrap_user_function(
+      NR_PSTR("Cake\\Core\\Exception\\CakeException::__construct"),
+      nr_cakephp_problem_2 TSRMLS_CC);
 #endif
 }
--- a/agent/fw_cakephp.c
+++ b/agent/fw_cakephp.c
@@ -154,7 +154,7 @@ NR_PHP_WRAPPER(nr_cakephp_error_handler_wrapper) {
   }
 
   if (NULL != request_uri) {
-    nr_txn_set_path("CakePHP", NRPRG(txn), request_uri, NR_PATH_TYPE_URI,
+    nr_txn_set_path("CakePHP Exception", NRPRG(txn), request_uri, NR_PATH_TYPE_URI,
                     NR_OK_TO_OVERWRITE);
   } else {
     nrl_verbosedebug(NRL_FRAMEWORK, "%s: request uri is NULL", __func__);
--- a/agent/fw_cakephp.c
+++ b/agent/fw_cakephp.c
@@ -269,30 +269,38 @@ NR_PHP_WRAPPER_END
 /*
  * CakePHP 4.0+
  *
- * If the action or controller is not found during the dispatch process, the
- * appropriate Exception will be created and thrown. We wrap the CakeException
- * constructor instead of the Exception handler, since CakePHP allows for the
- * handler to be completely replaced.
- *
- * txn naming scheme:
- * In this case, `nr_txn_set_path` is called before `NR_PHP_WRAPPER_CALL` with
- * `NR_NOT_OK_TO_OVERWRITE` and as this corresponds to calling the wrapped
- * function in func_begin it needs to be explicitly set as a before_callback to
- * ensure OAPI compatibility. This entails that the first wrapped call gets to
- * name the txn.
+ * Report errors and exceptions caught by CakePHP's error handler.
  */
-NR_PHP_WRAPPER(nr_cakephp_problem_4) {
-  const char* name = "Exception";
+NR_PHP_WRAPPER(nr_cakephp_error_handler_wrapper) {
+  zval* exception = NULL;
+  char* request_uri
+      = nr_strdup(nr_php_get_server_global("REQUEST_URI" TSRMLS_CC));
 
-  (void)wraprec;
   NR_UNUSED_SPECIALFN;
+  (void)wraprec;
 
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_CAKEPHP);
 
-  nr_txn_set_path("CakePHP", NRPRG(txn), name, NR_PATH_TYPE_ACTION,
+  exception = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
+  if (!nr_php_is_zval_valid_object(exception)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "%s: exception is NULL or not an object",
+                     __func__);
+    goto end;
+  }
+
+  if (NR_SUCCESS
+      != nr_php_error_record_exception(
+          NRPRG(txn), exception, nr_php_error_get_priority(E_ERROR), true,
+          "Uncaught exception ", &NRPRG(exception_filters) TSRMLS_CC)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "%s: unable to record exception", __func__);
+  }
+
+  nr_txn_set_path("CakePHP", NRPRG(txn), request_uri, NR_PATH_TYPE_ACTION,
                   NR_NOT_OK_TO_OVERWRITE);
 
-  NR_PHP_WRAPPER_CALL;
+end:
+  nr_php_arg_release(&exception);
+  nr_free(request_uri);
 }
 NR_PHP_WRAPPER_END
 
@@ -313,16 +321,10 @@ void nr_cakephp_enable(TSRMLS_D) {
   nr_php_wrap_user_function(
       NR_PSTR("Cake\\Controller\\Controller::invokeAction"),
       nr_cakephp_name_the_wt_4 TSRMLS_CC);
+  nr_php_wrap_user_function(
+      NR_PSTR(
+          "Cake\\Error\\Middleware\\ErrorHandlerMiddleware::handleException"),
+      nr_cakephp_error_handler_wrapper TSRMLS_CC);
   nr_txn_suggest_package_supportability_metric(NRPRG(txn), PHP_PACKAGE_NAME,
                                                PHP_PACKAGE_VERSION_UNKNOWN);
-#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
-    && !defined OVERWRITE_ZEND_EXECUTE_DATA
-  nr_php_wrap_user_function_before_after_clean(
-      NR_PSTR("Cake\\Core\\Exception\\CakeException::__construct"),
-      nr_cakephp_problem_4, NULL, NULL);
-#else
-  nr_php_wrap_user_function(
-      NR_PSTR("Cake\\Core\\Exception\\CakeException::__construct"),
-      nr_cakephp_problem_4 TSRMLS_CC);
-#endif
 }
--- a/agent/fw_cakephp.c
+++ b/agent/fw_cakephp.c
@@ -148,7 +148,6 @@ NR_PHP_WRAPPER_END
  *
  */
 NR_PHP_WRAPPER(nr_cakephp_name_the_wt_2) {
-  zval* arg1 = 0;
   zval* this_var = 0;
   zval* czval = 0;
   char* controller = 0;
@@ -156,8 +155,9 @@ NR_PHP_WRAPPER(nr_cakephp_name_the_wt_2) {
   int clen = 0;
   int alen = 0;
   char* name = 0;
-  zval* params;
-  zval* azval;
+  zval* params = NULL;
+  zval* request = NULL;
+  zval action_param;
 
   (void)wraprec;
   NR_UNUSED_SPECIALFN;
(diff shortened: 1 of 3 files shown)


agent/fw_cakephp.c
```diff
@@ -324,17 +307,22 @@ void nr_cakephp_enable_1(TSRMLS_D) {
...
+      NR_PSTR("Cake\\Controller\\Controller::invokeAction"),
+      nr_cakephp_name_the_wt_4 TSRMLS_CC);
+  nr_txn_suggest_package_supportability_metric(NRPRG(txn), PHP_PACKAGE_NAME,
+                                               PHP_PACKAGE_VERSION_UNKNOWN);
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
   nr_php_wrap_user_function_before_after_clean(
-      NR_PSTR("CakeException::__construct"), nr_cakephp_problem_2, NULL, NULL);
+      NR_PSTR("Cake\\Core\\Exception\\CakeException::__construct"),
+      nr_cakephp_problem_4, NULL, NULL);
```
lavarou, line 322: I'm not sure this is needed. When Observer API is used to hook into Zend Engine, the agent should do the right thing and record the error on the transaction even though framework installs its own exception handler. See Yii's exception instrumentation [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.3/agent/fw_yii.c#L197-L224) for reference.

hahuja2: I have experimented with the above and noticed that the error is not recorded as a traced error. Instead, I have added code that retrieves the exception caught by CakePHP's error handler and reports it as a traced error: [86108b5](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/86108b52af6eaade06b1bbd578f703edf78c56e4)

agent/fw_hooks.h
```diff
@@ -13,7 +13,7 @@
 #include "php_execute.h"
 
 extern void nr_cakephp_enable_1(TSRMLS_D);
```
ZNeumann, line 15: Do we want to remove this old code because it is seemingly no longer called?

agent/php_execute.c
```diff
@@ -331,15 +331,8 @@ typedef struct _nr_framework_table_t {
...
-     * Watch out:
-     *   cake1.2 and cake1.3 use a subdirectory named 'cake' (lower case)
-     *   cake2.0 and on use a subdirectory named 'Cake' (upper case file name)
-     */
-    {"CakePHP", "cakephp", NR_PSTR("cake/libs/object.php"), nr_cakephp_special_1,
-     nr_cakephp_enable_1, NR_FW_CAKEPHP},
-    {"CakePHP", "cakephp", NR_PSTR("cake/core/app.php"), nr_cakephp_special_2,
-     nr_cakephp_enable_2, NR_FW_CAKEPHP},
+    {"CakePHP", "cakephp", NR_PSTR("cakephp/src/core/functions.php"), 0,
+     nr_cakephp_enable, NR_FW_CAKEPHP},
```
ZNeumann, line 335: Does this file only exist on cakePHP 4+? There are many names/comments referencing CakePHP 4+, but as far as I can tell, this is the only place en/disabling our instrumentation. Since we are presumably not testing on version 3 and below, we probably want to ensure that our instrumentation is disabled for that.

agent/fw_cakephp.c
```diff
@@ -281,35 +268,40 @@ NR_PHP_WRAPPER(nr_cakephp_problem_1) {
...
- * ensure OAPI compatibility. This entails that the first wrapped call gets to
- * name the txn.
+ * Report errors and exceptions caught by CakePHP's error handler.
  */
-NR_PHP_WRAPPER(nr_cakephp_problem_2) {
-  const char* name = "Exception";
+NR_PHP_WRAPPER(nr_cakephp_error_handler_wrapper) {
+  zval* exception = NULL;
+  char* request_uri
+      = nr_strdup(nr_php_get_server_global("REQUEST_URI" TSRMLS_CC));
```
ZNeumann, line 278: I don't see the need to `strdup` and then `free` for this string. Can't we just directly pass `nr_php_get_server_global("REQUEST_URI")` wherever this string is needed?

ZNeumann, line 278:
```suggestion
      = nr_strdup(nr_php_get_server_global("REQUEST_URI"));
```

agent/fw_cakephp.c
lavarou: Leaving information for posterity:
- [CakePHP 4.x Error & Exception Handling](https://book.cakephp.org/4/en/development/errors.html)
- [CakePHP 5.x Error & Exception Handling](https://book.cakephp.org/5/en/development/errors.html)

The agent correctly detects that CakePHP registers its default exception handler `Cake\Error\ExceptionTrap::handleException` ([CakePHP 4.x](https://github.com/cakephp/cakephp/blob/4.x/src/Error/ExceptionTrap.php#L186), [CakePHP 5.x](https://github.com/cakephp/cakephp/blob/5.x/src/Error/ExceptionTrap.php#L173)). However, that exception handler is used only for uncaught exceptions. In a CakePHP application, exceptions thrown in controller actions are caught in `Cake\Error\Middleware\ErrorHandlerMiddleware::process` method ([CakePHP 4.x](https://github.com/cakephp/cakephp/blob/5.x/src/Error/Middleware/ErrorHandlerMiddleware.php#L105-L121), [CakePHP 5.x](https://github.com/cakephp/cakephp/blob/5.x/src/Error/Middleware/ErrorHandlerMiddleware.php#L105-L121)). `Cake\Error\Middleware\ErrorHandlerMiddleware` is part of CakePHP application's default middleware stack ([CakePHP 4.x](https://github.com/cakephp/app/blob/4.x/src/Application.php#L80), [CakePHP 5.x](https://github.com/cakephp/app/blob/5.x/src/Application.php#L70)) so wrapping `Cake\Error\Middleware\ErrorHandlerMiddleware::handleException` is the right approach for the agent to record error events for exceptions thrown by controller actions in standard CakePHP applications.

agent/fw_hooks.h
lavarou: Also `nr_cakephp_special_1` and `nr_cakephp_special_1` don't seem to be used any more.

agent/fw_cakephp.c
```diff
@@ -324,17 +316,16 @@ void nr_cakephp_enable_1(TSRMLS_D) {
...
-                            nr_cakephp_problem_2 TSRMLS_CC);
-#endif
+void nr_cakephp_enable(TSRMLS_D) {
+  nr_php_wrap_user_function(
+      NR_PSTR("Cake\\Controller\\Controller::invokeAction"),
+      nr_cakephp_name_the_wt_4 TSRMLS_CC);
+  nr_php_wrap_user_function(
+      NR_PSTR(
+          "Cake\\Error\\Middleware\\ErrorHandlerMiddleware::handleException"),
+      nr_cakephp_error_handler_wrapper TSRMLS_CC);
```
lavarou, line 328:
```suggestion
  nr_php_wrap_user_function(
      NR_PSTR("Cake\\Controller\\Controller::invokeAction"),
      nr_cakephp_name_the_wt_4);
  nr_php_wrap_user_function(
      NR_PSTR(
          "Cake\\Error\\Middleware\\ErrorHandlerMiddleware::handleException"),
      nr_cakephp_error_handler_wrapper);
```

agent/fw_cakephp.c
```diff
@@ -281,35 +268,40 @@ NR_PHP_WRAPPER(nr_cakephp_problem_1) {
...
+      = nr_strdup(nr_php_get_server_global("REQUEST_URI" TSRMLS_CC));
 
-  (void)wraprec;
   NR_UNUSED_SPECIALFN;
+  (void)wraprec;
 
   NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_CAKEPHP);
 
-  nr_txn_set_path("CakePHP", NRPRG(txn), name, NR_PATH_TYPE_ACTION,
+  exception = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
lavarou, line 285:
```suggestion
  exception = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
```

agent/fw_cakephp.c
```diff
@@ -281,35 +268,40 @@ NR_PHP_WRAPPER(nr_cakephp_problem_1) {
...
+  }
+
+  if (NR_SUCCESS
+      != nr_php_error_record_exception(
+          NRPRG(txn), exception, nr_php_error_get_priority(E_ERROR), true,
+          "Uncaught exception ", &NRPRG(exception_filters) TSRMLS_CC)) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "%s: unable to record exception", __func__);
+  }
+
+  nr_txn_set_path("CakePHP", NRPRG(txn), request_uri, NR_PATH_TYPE_ACTION,
```
lavarou, line 299: Should fourth argument be `NR_PATH_TYPE_ACTION` or `NR_PATH_TYPE_URI`? Judging by third argument, it should be `NR_PATH_TYPE_URI`.

agent/fw_cakephp.c
hahuja2: Good point, fixed: [a1dbe2e](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/a1dbe2ef44c83262b6a1313d98dafe814d4d9916)

hahuja2: This line has been removed in this commit: [a1dbe2e](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/a1dbe2ef44c83262b6a1313d98dafe814d4d9916)

agent/fw_cakephp.c
```diff
@@ -281,35 +268,38 @@ NR_PHP_WRAPPER(nr_cakephp_problem_1) {
...
+
+  if (NR_SUCCESS
+      != nr_php_error_record_exception(
+          NRPRG(txn), exception, nr_php_error_get_priority(E_ERROR), true,
+          "Uncaught exception ", &NRPRG(exception_filters))) {
+    nrl_verbosedebug(NRL_FRAMEWORK, "%s: unable to record exception", __func__);
+  }
+
+  nr_txn_set_path("CakePHP", NRPRG(txn),
+                  nr_php_get_server_global("REQUEST_URI"), NR_PATH_TYPE_ACTION,
```
lavarou, line 298: This will result in memory leak: `nr_php_get_server_global` calls `nr_strndup` and result of `nr_strndup` is returned to the caller and the caller is responsible for freeing the memory. Look [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.3/agent/php_txn.c#L46-L52) for an example of a proper use of `nr_php_get_server_global`.

ZNeumann: ah, so previously (https://github.com/newrelic/newrelic-php-agent/pull/983#discussion_r1838710212) it was being dup'ed twice

lavarou: > ah, so previously ([#983 (comment)](https://github.com/newrelic/newrelic-php-agent/pull/983#discussion_r1838710212)) it was being dup'ed twice

Yes.

agent/fw_cakephp.c
hahuja2: I believe the fourth argument should stay as `NR_PATH_TYPE_ACTION` because if it is changed to `NR_PATH_TYPE_URI`, then the `WebTransaction` metric will not include the controller name or the action name.

agent/fw_cakephp.c
hahuja2: Good catch, fixed: [2a8975e](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/2a8975e2df808c464a50ae6ffa3266e9cebbbdba)

agent/php_execute.c
hahuja2: Tests for CakePHP 3.x have been added in our testing repo.

agent/fw_cakephp.c
lavarou: But isn't the transaction named after request uri not after `Controller/method`?

agent/fw_cakephp.c
lavarou: [2a8975e](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/2a8975e2df808c464a50ae6ffa3266e9cebbbdba) is not fixing the memory leak - the return value from `nr_php_get_server_global` needs to be freed by the caller. Look [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.3/agent/php_txn.c#L46-L52) for an example of a proper use of nr_php_get_server_global.

hahuja2: Thanks for catching that, I didn't need to do another `nr_strdup`. Fixed: [fd5ef92](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/fd5ef92a51c39cd4f1d6bfc4fb9b718df237e292)

agent/php_execute.c
ZNeumann: What about 2.x and 1.x? Does this enable on those too? Does our instrumentation work/break for those? Or are they old enough (or PHP version incompatible) that we don't need to consider it?

agent/fw_cakephp.c
hahuja2: After discussion, this has been changed to `NR_PATH_TYPE_URI`. I also have added a `if` check to ensure the request uri is not `NULL`: [fa779f5](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/fa779f5a01cf2ace614bfced4e68afd8dbccb1ec)

agent/php_execute.c
hahuja2: It does not enable for CakePHP 1.x/2.x because the detection file that we are using does not exist in CakePHP 1.x/2.x

agent/fw_hooks.h
hahuja2: Good point, I have removed all of the old unused code for CakePHP: [d0a6282](https://github.com/newrelic/newrelic-php-agent/pull/983/commits/d0a6282fda9ccd2ef8ae002e7c49eeef4e042155)

ZNeumann (dismissed):

agent/fw_cakephp.c
```diff
@@ -193,37 +76,24 @@ NR_PHP_WRAPPER(nr_cakephp_name_the_wt_2) {
...
-    NR_PHP_WRAPPER_CALL;
-    goto end;
-  }
-
   NR_PHP_WRAPPER_CALL;
 
-  params = nr_php_get_zval_object_property(arg1, "params" TSRMLS_CC);
-  if (0 == params) {
-    nrl_verbosedebug(NRL_FRAMEWORK, "CakePHP: no params found in request");
+  request = nr_php_call(this_var, "getRequest");
```
lavarou, line 81: `\Cake\Controller\Controller` class has `$request` property. Why not avoid calling PHP code via `nr_php_call` and use the same method that was used to obtain the controller name: i.e. `nr_php_get_zval_object_property(this_var, "request");`?

agent/fw_cakephp.c
```diff
@@ -193,37 +76,24 @@ NR_PHP_WRAPPER(nr_cakephp_name_the_wt_2) {
...
+    nrl_verbosedebug(NRL_FRAMEWORK, "CakePHP: no request found in controller");
     goto end;
   }
 
-  if (IS_ARRAY != Z_TYPE_P(params)) {
-    nrl_verbosedebug(NRL_FRAMEWORK, "CakePHP: request params is not an array");
+  nr_php_zval_str(&action_param, "action");
+  action_zval = nr_php_call(request, "getParam", &action_param);
+  zval_dtor(&action_param);
+  if (!nr_php_is_zval_valid_string(action_zval)) {
```
lavarou, line 90: Wouldn't `nr_php_is_zval_non_empty_string` be a better check here?
```suggestion
  if (!nr_php_is_zval_non_empty_string(action_zval)) {
```

agent/fw_cakephp.c
hahuja2: Using `nr_php_get_zval_object_property(this_var, "request");` does not work for CakePHP 3.x because the `$request` property is deprecated in version `3.6.0`: https://github.com/cakephp/cakephp/blob/571fb867995ccfa4034ec5b66577f0ab2acd4429/src/Controller/Controller.php#L126. It also mentions to use `getRequest` instead.

lavarou: `$name` property is also protected in [3.x](https://github.com/cakephp/cakephp/blob/3.x/src/Controller/Controller.php#L104), [4.x](https://github.com/cakephp/cakephp/blob/4.x/src/Controller/Controller.php#L118) and [5.x](https://github.com/cakephp/cakephp/blob/5.x/src/Controller/Controller.php#L114) and using `nr_php_get_zval_object_property(this_var, "name");` works so why it doesn't work for `$request`? The deprecation comment is about it not being `public` any more so PHP code will no longer be able to access `$request` property. However the agent has access to `private` and `protected` properties of `this`.

lavarou: I'm fine leaving it as is.

lavarou (approved):

ZNeumann (approved):

PR #985, opened 2024-11-14
ci: add security scan with trivy

lavarou:
Security scan with trivy will be run on push to main and dev, pull request, and daily. trivy will skip scanning vendor subdirectory because the agent is not using code from that directory.


.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,48 @@
+name: Security scan
+on:
+  push:
+    branches:
+      - main
+      - dev
+  pull_request:
+  schedule:
+    - cron: '0 9 * * *' # Same time as CI Cron
```
bduranleau-nr, line 9: Do we want/need this to run daily or would a weekly run suffice?

bduranleau-nr (dismissed):

zsistla (approved):

mfulb (approved):

PR #991, opened 2024-12-10
feat(axiom): Add message segment functionality

zsistla:
1) Added message segment.
Agent will now handle generic, external, datastore, and message segments.

2) Added unit tests for the changes

3) New functionality involves adding message specific attributes to segments/spans, generating message specific metrics, and honoring the message_tracer_parameters_enabled INI setting to disable/enable the attributes.

4)Things to keep in mind while reviewing:
* the commits describe a lot of what is going on
* Functionality-wise, message segment is a mix of external and datastore
* nr_segment_message is the new file


tests/integration/synthetics/test_happy_path_with_dt.php
```diff
@@ -62,6 +62,7 @@
         "nr.syntheticsJobId": "jjjjjjj-jjjj-1234-jjjj-jjjjjjjjjjjj",
         "nr.syntheticsMonitorId": "mmmmmmm-mmmm-1234-mmmm-mmmmmmmmmmmm",
         "externalDuration": "??",
+        "externalCallCount": 1,
```
lavarou, line 65: Where does this extra attribute come from?

zsistla: It was an external segment bug.
The call count should have been calculated in 2 different places, but it was only updated in one place.
See commit and commit message for info: 
https://github.com/newrelic/newrelic-php-agent/pull/991/commits/6932595b740cf04e4c277e103b6f4748ef932de9

axiom/nr_segment.c
```diff
@@ -313,6 +313,26 @@ static void nr_populate_http_spans(nr_span_event_t* span_event,
                                     segment->typed_attributes->external.status);
 }
 
+static void nr_populate_message_spans(nr_span_event_t* span_event,
+                                      const nr_segment_t* segment) {
+  nr_span_event_set_category(span_event, NR_SPAN_MESSAGE);
```
mfulb, line 318: Should this happen before the NULL check below it?

tests/integration/synthetics/test_happy_path_with_dt.php
lavarou: Could the fix for `externalCallCount` attribute in transaction event be separated out from this PR? It is not related to message segment feature and it also deserves updates to [integration/external](https://github.com/newrelic/newrelic-php-agent/tree/main/tests/integration/external) tests - I would expect at least some of them enhanced with `EXPECT_ANALYTICS_EVENTS` where the value of `externalCallCount` attribute is verified.

axiom/nr_segment_message.c
```diff
@@ -0,0 +1,231 @@
...
+/*
+ * Purpose : Set all the typed message attributes on the segment.
+ */
+static void nr_segment_message_set_attrs(
+    nr_segment_t* segment,
+    const nr_segment_message_params_t* params,
+    nrtxnopt_t options) {
+  nr_segment_message_t message_attributes = {0};
+
+  message_attributes.message_action = params->message_action;
```
mfulb, line 25: Does this function need a `NULL` check for `params`?

axiom/nr_segment_message.c
```diff
@@ -0,0 +1,231 @@
...
+
+  if (options.message_tracer_segment_parameters_enabled) {
+    message_attributes.destination_name = nr_strempty(params->destination_name)
```
mfulb, line 28: Could we have `nr_segment_set_message()` handle the `nr_strempty()` check (since it is already doing a `NULL` check) and have this code just pass the pointer through w/o additional conditionals here.

axiom/nr_segment_message.c
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/*
+ * Purpose : Create metrics for a completed message call and set the segment
+ *           name.
+ *
+ * Metrics created during this call
+ * ----------------------------------------------------------------------------------
+ * MessageBroker/all                                         Unscoped Always
+ * MessageBroker/{library}/all                               coped   Always
```
mfulb, line 48:
```suggestion
 * MessageBroker/{library}/all                               Scoped   Always
```

axiom/nr_segment_message.c
```diff
@@ -0,0 +1,231 @@
...
+   */
+
+  nrm_force_add(segment->txn->unscoped_metrics, "MessageBroker/all", duration);
+
+  if (nr_strempty(message_params->library)) {
+    library_string = "<unknown>";
+  } else {
+    library_string = message_params->library;
+  }
+  rollup_metric = nr_formatf("MessageBroker/%s/all", library_string);
```
mfulb, line 117: For a simple string like this would it be worth a performance gain to build it by concatenating strings?

axiom/nr_segment_message.c
```diff
@@ -0,0 +1,231 @@
...
+      destination_type_string = "Exchange";
+      break;
+    default:
+      destination_type_string = "<unknown>";
+      break;
+  }
+  /*
+   * Create the scoped metric
+   * MessageBroker/{Library}/{DestinationType}/{Action}/Named/{DestinationName}
+   * non-temp MessageBroker/{Library}/{DestinationType}/{Action}/Temp
```
mfulb, line 155: This comment is confusing - is "non-temp" supposed to be at the end of the previous line?

axiom/nr_segment_message.h
```diff
@@ -0,0 +1,53 @@
...
+typedef enum _nr_segment_message_destination_type_t {
+  NR_MESSAGE_DESTINATION_TYPE_QUEUE,
+  NR_MESSAGE_DESTINATION_TYPE_TOPIC,
+  NR_MESSAGE_DESTINATION_TYPE_TEMP_QUEUE,
+  NR_MESSAGE_DESTINATION_TYPE_TEMP_TOPIC,
+  NR_MESSAGE_DESTINATION_TYPE_EXCHANGE
+} nr_segment_message_destination_type_t;
+
+typedef struct {
+  /* All strings are null-terminated. When unset, the strings are ingored. */
```
mfulb, line 26: Does "unset" mean set to `NULL` or an empty string or both?

tests/integration/synthetics/test_happy_path_with_dt.php
zsistla: Please see PR https://github.com/newrelic/newrelic-php-agent/pull/1006

axiom/nr_segment.c
zsistla: The category is being set on the span event, so separate from segment logic at that point (and set_category checks the event for null). Spans can have a type even if no attributes are available (see datastore/external for similar behavior).
This is core logic, so the fewer messages the better, but in this case, no log message is needed because if we are setting span category with no attributes, the absence of attributes would show up both in the UI as well as in the verbosedebug daemon logs.

axiom/nr_segment_message.c
zsistla: Technically yes, but it is a private function to this file and the calling function already checked params for NULL so it seemed not needed.
But if you feel strongly about this, happy to put it in.

axiom/nr_segment_message.c
zsistla: Not sure how much performance difference it would be between one nr_formatf vs 2 nr_strcat calls since the nr_strcat will do multiple null checks and also strlen calls.

nr_strcat:
```
char* nr_strcat(char* dest, const char* src) {
  int dl;
  int sl;

  if (nrlikely((0 != dest) && (0 != src))) {
    dl = (int)(strlen)(dest);
    sl = (int)(strlen)(src);

    nr_memcpy(dest + dl, src, (size_t)(sl + 1));
    return dest + dl + sl;
  }

  if (nrunlikely((0 == dest))) {
    return 0;
  }

  dl = (int)(strlen)(dest);
  return dest + dl;
}
```

nr_formatf
```
char* nr_formatf(const char* fmt, ...) {
  int rv;
  va_list ap;
  char* ret = NULL;

  if (NULL == fmt) {
    return NULL;
  }

  va_start(ap, fmt);
  rv = vasprintf(&ret, fmt, ap);
  va_end(ap);

  if (rv < 0) {
    nr_free(ret);
    return NULL;
  }

  return ret;
}
```

I could try something similar to how we did this metric here: https://github.com/newrelic/newrelic-php-agent/blob/feat/sqs_instrumentation/agent/lib_aws_sdk_php.c#L482

axiom/tests/test_segment_private.c
```diff
@@ -328,6 +370,19 @@ static void test_destroy_typed_attributes(void) {
...
+  /*
+   * Test : Clean up typed attributes for a message segment
+   */
+  s.type = NR_SEGMENT_MESSAGE;
+  s.typed_attributes = nr_zalloc(sizeof(nr_segment_typed_attributes_t));
+  s.typed_attributes->message.destination_name = nr_strdup("queue_name");
+  s.typed_attributes->message.messaging_system = nr_strdup("aws_sqs");
+  s.typed_attributes->message.server_address = nr_strdup("localhost");
+
+  nr_segment_destroy_typed_attributes(NR_SEGMENT_MESSAGE, &s.typed_attributes);
```
mfulb, line 385: Does this need a test to verify the typed attributes were cleaned up?

axiom/nr_segment_message.c
zsistla: Not sure that would help here much, since nr_strlcpy would also check strlen each time it was called.

axiom/tests/test_span_event.c
```diff
@@ -434,6 +475,58 @@ static void test_span_events_extern_get_and_set(void) {
...
+      "should be the component we set 1", "chicken",
+      nr_span_event_get_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS));
+  nr_span_event_set_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS, "oracle");
+  tlib_pass_if_str_equal(
+      "should be the component we set 2", "oracle",
+      nr_span_event_get_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS));
+
+  nr_span_event_destroy(&event);
+}
+
```
mfulb, line 529: Are there any tests that verify destroying fields will actually set them to the expected value (NULL?)?

axiom/nr_txn.c
```diff
@@ -2582,10 +2588,16 @@ nrobj_t* nr_txn_event_intrinsics(const nrtxn_t* txn) {
       params, txn->unscoped_metrics, "WebFrontend/QueueTime", "queueDuration");
   nr_txn_add_metric_total_as_attribute(params, txn->unscoped_metrics,
                                        "External/all", "externalDuration");
+  nr_txn_add_metric_count_as_attribute(params, txn->unscoped_metrics,
```
mfulb, line 2591: Does this affect any integration/multiverse tests or do they all wildcard the External/all count so this has never been tested as a fixed value?

zsistla: This is superceded by the https://github.com/newrelic/newrelic-php-agent/pull/1006 PR; once this is rebased on that, the externalCallCount changes will go away.  But this doesn't require any changes in multiverse (I ran both this messagesegment branch and the sqs instrumentation branch against multiverse).

axiom/nr_segment_message.h
zsistla: unset means set to NULL, for message segment we further don't allow empty strings, so if we detect NULL or empty, we leave the attribute unset.

axiom/tests/test_span_event.c
zsistla: As with other tests, valgrind is used to verify proper mem cleanup.
Comment updates to indicate that.
https://github.com/newrelic/newrelic-php-agent/pull/991/commits/fde6d510b065a30bc0b242411b5a39a31b86da6a

axiom/tests/test_segment_private.c
zsistla: As with other tests, valgrind is used to verify proper mem cleanup.
Comment updates to indicate that.
https://github.com/newrelic/newrelic-php-agent/commit/fde6d510b065a30bc0b242411b5a39a31b86da6a

axiom/nr_segment_message.c
zsistla: Improved formatting, please see: [12af96f](https://github.com/newrelic/newrelic-php-agent/pull/991/commits/12af96f1c587a461066885a918d0d8cd36d3182c)

axiom/nr_segment_message.c
zsistla: Not quite sure what optimization you are suggesting.  Can you please clarify?

axiom/nr_segment_message.c
zsistla: Good idea!
Please see: [6aec35d](https://github.com/newrelic/newrelic-php-agent/pull/991/commits/6aec35d7650ec9941a75c8f8198a2f14765de0f3)

axiom/tests/test_segment_private.c
mfulb: I'm thinking more about the state of the pointers being set to a known value like `NULL`.  I am not sure valgrind would evaluate if this is happening.

axiom/nr_segment_message.h
mfulb: So would it be accurate for the comment to say "When NULL, the strings are ignored"?  Or is it "NULL or empty"?

axiom/nr_segment_message.c
mfulb: This is a recurring topic that comes up and I don't have a feel for how we've decided to handle it.  Maybe having a naming convention for functions to indicate they are not robustly checking arguments would be a long term solution.

axiom/nr_span_event.h
```diff
@@ -115,6 +145,8 @@ extern void nr_span_event_set_transaction_name(nr_span_event_t* event,
                                                const char* transaction_name);
 extern void nr_span_event_set_category(nr_span_event_t* event,
                                        nr_span_category_t category);
+extern void nr_span_event_set_spankind(nr_span_event_t* event,
+                                       nr_span_spankind_t category);
```
lavarou, line 149: Not that it matters much here, but just to avoid confusion in the future and help in searching through the code:
```suggestion
                                       nr_span_spankind_t spankind);
```

axiom/nr_txn.c
```diff
@@ -60,10 +60,8 @@ struct _nr_txn_attribute_t {
 #define NR_TXN_ATTRIBUTE_TRACE_ERROR \
   (NR_ATTRIBUTE_DESTINATION_TXN_TRACE | NR_ATTRIBUTE_DESTINATION_ERROR)
 
-#define NR_TXN_ATTR(X, NAME, DESTS)                     \
-  const nr_txn_attribute_t* X = &(nr_txn_attribute_t) { \
-    (NAME), (DESTS)                                     \
-  }
+#define NR_TXN_ATTR(X, NAME, DESTS) \
+  const nr_txn_attribute_t* X = &(nr_txn_attribute_t) { (NAME), (DESTS) }
```
lavarou, line 64: Why this change? `clang-format` wants it the way it is was before this change.

axiom/nr_span_event.c
lavarou, line 1: Avoid using hardcoded strings that are used in multiple places. E.g. `"messaging.destination.name"` is used in `nr_span_event_set_message` and `nr_span_event_get_message` and if it is mistyped in either of those places, the code will not work. A less error prone strategy is to define and use a macro. This way string is consistent between uses in `nr_span_event_set_message` and `nr_span_event_get_message`. Yes, tests should catch spelling mistakes in hardcoded strings and yes, there can be a spelling mistake in the macro itself. However with macro, the spelling error will be caught as early as at compile time rather than later in the testing process, which tightens build-test-correct loop. Please make this update for at least the strings used in `nr_span_event_set_message` and `nr_span_event_get_message`.

axiom/nr_span_event.h
```diff
@@ -21,9 +21,30 @@ typedef struct _nr_span_event_t nr_span_event_t;
...
+ * which states that span.kind is
+ * a) producer when the operation type is create or send(if the context is
+ * create) b) client when the operation type is create or send(if the context is
+ * NOT create) c) consumer when the operation type is process
+ */
+typedef enum {
+  NR_SPAN_PRODUCER,
+  NR_SPAN_CLIENT,
+  NR_SPAN_CONSUMER,
+  NR_SPAN_NO_SPANKIND
```
lavarou, line 45: With this naming convention it's hard to tell if `NR_SPAN_PRODUCER` is span category or span kind. Consider using `NR_SPAN_KIND_` prefix.

axiom/nr_span_event.h
```diff
@@ -21,9 +21,30 @@ typedef struct _nr_span_event_t nr_span_event_t;
...
+} nr_span_spankind_t;
```
lavarou, line 46: Why double the word `span`? Why not simply `nr_span_kind_t`?

axiom/tests/test_span_event.c
```diff
@@ -256,6 +256,47 @@ static void test_span_event_category(void) {
...
+                         "message", nr_span_event_get_category(event));
+
+  nr_span_event_destroy(&event);
+}
+
+static void test_span_event_spankind(void) {
+  nr_span_event_t* event = nr_span_event_create();
+
+  // Test : the default is NULL (spankind must be explicitly set)
+  tlib_pass_if_str_equal("The default category", NULL,
```
lavarou, line 270:
```suggestion
  tlib_pass_if_str_equal("empty kind is NULL", NULL,
```

axiom/tests/test_span_event.c
```diff
@@ -256,6 +256,47 @@ static void test_span_event_category(void) {
...
+                         nr_span_event_get_spankind(event));
+
+  // Test : A null event returns NULL
+  tlib_pass_if_null("The default category", nr_span_event_get_spankind(NULL));
```
lavarou, line 274:
```suggestion
  tlib_pass_if_null("get_spankind(NULL) does not crash", nr_span_event_get_spankind(NULL));
```

axiom/tests/test_span_event.c
```diff
@@ -256,6 +256,47 @@ static void test_span_event_category(void) {
...
+}
+
+static void test_span_event_spankind(void) {
+  nr_span_event_t* event = nr_span_event_create();
+
+  // Test : the default is NULL (spankind must be explicitly set)
+  tlib_pass_if_str_equal("The default category", NULL,
+                         nr_span_event_get_spankind(event));
+
+  // Test : A null event returns NULL
```
lavarou, line 273: Instead of comments, let's make them test messages!

axiom/nr_segment_message.c
mfulb: After discussion we decided this kind of optimization is better addressed by designed a new function and changing the metric name creation throughout the agent to use this function.  In this case the expected performance impact is minimal and doesn't warrant a specialized section of code for it.

axiom/tests/test_span_event.c
```diff
@@ -256,6 +256,47 @@ static void test_span_event_category(void) {
...
+  // Test : setting the spankind back and forth
+  nr_span_event_set_spankind(event, NR_SPAN_CLIENT);
+  tlib_pass_if_str_equal("Spankind should be the one we set - client", "client",
+                         nr_span_event_get_spankind(event));
+
+  // Test : setting the spankind back and forth
+  nr_span_event_set_spankind(event, NR_SPAN_CONSUMER);
+  tlib_pass_if_str_equal("Spankind should be the one we set - consumer",
+                         "consumer", nr_span_event_get_spankind(event));
+
```
lavarou, line 299: Do we care to test with invalid input, i.e. `spankind` argument passed is out of valid  range like this: `nr_span_event_set_spankind(event, 255);`?

lavarou: Something like this:
```suggestion
  nr_span_event_set_spankind(event, 255);
  tlib_pass_if_str_equal("Invalid spankind sets spankind to none (NULL)",
                         NULL, nr_span_event_get_spankind(event));

```

axiom/nr_segment_message.c
zsistla: Ticket created to create a optimized function for general use with metrics.

axiom/nr_segment_message.c
zsistla: Updated comment here: [97218d6](https://github.com/newrelic/newrelic-php-agent/pull/991/commits/97218d6120f6039daa904dcff59bcc18d54c45e1)

axiom/tests/test_span_event.c
```diff
@@ -434,6 +475,58 @@ static void test_span_events_extern_get_and_set(void) {
...
+  // Test : setting the server address back and forth behaves as expected
+  nr_span_event_set_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS, "chicken");
+  tlib_pass_if_str_equal(
+      "should be the component we set 1", "chicken",
+      nr_span_event_get_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS));
+  nr_span_event_set_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS, "oracle");
+  tlib_pass_if_str_equal(
+      "should be the component we set 2", "oracle",
+      nr_span_event_get_message(event, NR_SPAN_MESSAGE_SERVER_ADDRESS));
+
```
lavarou, line 526: Do we care to test with invalid input, i.e. member argument passed is out of valid range like this: nr_span_event_get_message(event, 255);?

axiom/tests/test_segment_private.c
zsistla: All of these use nr_free which sets to null, purpose of this isn't to test nr_free, just verify it's been called.

axiom/nr_segment_message.c
```diff
@@ -0,0 +1,238 @@
...
+ * attributes or not Returns: true on success.
+ *
+ * Note: This is a function private to this file and assumes the calling
+ * function has already checked the input parameters for NULL prior to calling
+ * this function.
+ */
+static void nr_segment_message_set_attrs(
+    nr_segment_t* segment,
+    const nr_segment_message_params_t* params,
+    nrtxnopt_t options) {
```
lavarou, line 33: Why are `options` passed by value (`nrtxnopt_t`) here instead of reference (`nrtxnopt_t *`)? Passing `options` by reference is more performant than passing them by reference because it avoids copying the whole `nrtxnopt_t` structure onto stack.

axiom/Makefile
```diff
@@ -17,7 +17,6 @@
 # Useful variables:
 #
 # AR:     The archiver used to build a static library (default: ar).
-# CC:     The C compiler to use (default: gcc).
```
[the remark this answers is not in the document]
zsistla: Not sure how this was deleted, good catch.
Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/nr_span_event.h
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/nr_txn.c
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/nr_span_event.h
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/tests/test_span_event.c
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/tests/test_span_event.c
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/tests/test_span_event.c
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/tests/test_span_event.c
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/tests/test_span_event.c
zsistla: Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/nr_segment_message.c
zsistla: Great point.
This can be improved further now that the function expectations were clarified to have already checked certain values for NULL.

Please see: https://github.com/newrelic/newrelic-php-agent/pull/991/commits/862604b081029206c2072275588076691d30560a

axiom/nr_segment_message.h
zsistla: Yes, NULL or empty.
please see: [925b94c](https://github.com/newrelic/newrelic-php-agent/pull/991/commits/925b94caf9fb5ce2999efcf57af0572f9a7938a0)

axiom/nr_span_event.c
lavarou: It would be easier to re-review the feedback if the feedback would be addressed atomically rather than lumped into a single commit. It takes longer to dissect the single commit to verify if the feedback has actually been addressed.

zsistla: I understand your suggestion and will keep it in mind.

For comment suggestions, I usually always use the commit from GitHub's web UI.  However, code changes sometimes don't compile cleanly, so I always like to try them locally before committing.

In the meantime, is the original feedback given in this thread addressed adequately?

axiom/tests/test_span_event.c
lavarou: https://github.com/newrelic/newrelic-php-agent/commit/862604b081029206c2072275588076691d30560a update is still not precise enough for my taste. But better than before.

axiom/nr_span_event.c
lavarou: I always test my suggestions and run clang-format on them to make it easier for the author.

axiom/nr_segment_message.c
lavarou: The approach from https://github.com/newrelic/newrelic-php-agent/commit/862604b081029206c2072275588076691d30560a is definitely more optimal than my suggestion, however it now hides the dependency of `nr_segment_message_set_attrs` on transaction's options. While reading `nr_segment_message_end` it's not obvious that `nr_segment_message_set_attrs` checks and acts differently depending on transaction's options. I think the best compromise between performance and code readability would be to either pass the value of `message_tracer_segment_parameters_enabled` option like this:
```
nr_segment_message_set_attrs(segment, message_params, segment->txn->options.message_tracer_segment_parameters_enabled);
```
or pointer to options like I suggested earlier. When value is passed by reference, the pointer is usually passed in register rather than put on stack (same for bool value) so this is still fairly performant.

axiom/tests/test_span_event.c
zsistla: The initial suggestion of "empty" was imprecise, what does empty mean, how do you empty it?
The more precise wording would be to say that the default, in lieu of explicitly setting it, is NULL.
https://github.com/newrelic/newrelic-php-agent/pull/991/commits/4f91a63f1a1220ef144733f0abc4ed0dad9328c4

axiom/nr_span_event.c
lavarou: > In the meantime, is the original feedback given in this thread addressed adequately?

I think it has.

axiom/nr_segment_message.c
zsistla: `options` is a part of segment, so anytime it is passed, a function can use all pieces of it.  Other areas of code don't explicitly pass options as an argument, they just use it as part of segment when needing to make decisions based on options.  It is duplicate to pass the same information twice.

I'd rather call it out in the function header. Change:
```
 * Purpose : Set all the typed message attributes on the segment.
```
to:
```
 * Purpose : Set all the typed message attributes on the segment.  The message_tracer_segment_parameters_enabled setting is checked in the txn options to determine if the attributes should be set. 
```

axiom/tests/test_segment_private.c
mfulb: My thought is the test is to verify `nr_segment_destroy_typed_attributes()` does these things properly - we shouldn't assume what it is using internally but treat it as an unknown and verify the state after calling it meets our expectations.  Maybe you can infer it called `nr_free()` if valgrind doesn't complain about leaks.  My preference is that the unit test has coverage of all these things and valgrind is just a backup for memory management issues.

zsistla: Please see: [0670292](https://github.com/newrelic/newrelic-php-agent/pull/991/commits/0670292ebb5d1562c4e7f43518bdeaf59e8a8186)

axiom/nr_segment_message.c
lavarou: > options is a part of segment, so anytime it is passed, a function can use all pieces of it. Other areas of code don't explicitly pass options as an argument, they just use it as part of segment when needing to make decisions based on options. It is duplicate to pass the same information twice.

Just because it's been done in the past doesn't mean it should be done going forward, especially if it makes the code less maintainable and clear. Updating the functions doc string is not enough because its behavior can be deduced by reading the code. But if the dependency on `message_tracer_segment_parameters_enabled` would be injected to the function (this technique is called [Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection)) it would make reading the code of `nr_segment_message_end` and understanding how `nr_segment_message_set_attrs` works more obvious without the need to switch the context too much.

zsistla: I disagree on this one.
Options are an inherent part of the txn and segment so they don't have to be explicitly passed to each and every function that will need the options.  That is the whole reason they were added to the segment->txn so that they wouldn't have to all be explicitly passed and so that it could be ensured that the options that were used remained consistent.  By breaking it out an passing explicitly, there is the chance of creating inconsistencies where the options passed are not the options that have actually been set for the segment and txn.

lavarou: > That is the whole reason they were added to the segment->txn so that they wouldn't have to all be explicitly passed and so that it could be ensured that the options that were used remained consistent.

`txn` pointer is a global in the agent, so there was no need to add this to the segment. However, this global cannot be accessed from axiom and adding `txn` pointer to the segment is was a workaround and that's how txn's options are also available from the segment. There are pros and cons to this approach. I will not argue this because this PR is not a place. The crux of my feedback was not to pass options by value and this has been addressed.

axiom/nr_segment_traces.c
```diff
@@ -578,3 +587,65 @@ void nr_segment_traces_create_data(
...
+/*
+ * Purpose : If available, add cloud attributes to segment.
+ *
+ * Params  : 1. segment to create and add agent attributes to
+ *           2. nr_segment_cloud_attrs_t* that contains the attributes
+ *
+ * Returns : void
+ *
+ */
+extern void nr_segment_traces_add_cloud_attributes(
```
lavarou, line 600: Why is this function defined in this file? It has nothing to do with segment traces.

lavarou (approved):

PR #993, opened 2024-12-10
fix PDO instrumentation in PHP 8.4

lavarou:
PHP 8.4 includes implementation of PDO driver specific sub-classes RFC (see https://wiki.php.net/rfc/pdo_driver_specific_subclasses). This means that when database access code uses new factory method `PDO::connect` to create PDO object, an instance of driver specific sub-class of `PDO` is returned instead of an instance of generic `PDO` class. This means that instrumentation of generic `PDO` class is not enough to provide instrumentation of datastores. Add wrappers for driver specific subclasses of `PDO` supported by the agent: `Pdo\Firebird`, `Pdo\Mysql`, `Pdo\Odbc`, `Pdo\Pgsql`, `Pdo\Sqlite`.


ZNeumann (approved):

tests/integration/pdo/mysql/base-class/test_prepared_stmt_1.php
```diff
@@ -0,0 +1,110 @@
...
+<?php require(realpath (dirname ( __FILE__ )) . '/../../skipif_mysql.inc');
+*/
+
+/*ENVIRONMENT
+DATASTORE_PRODUCT=MySQL
+DATASTORE_COLLECTION=tables
+*/
+
+/*INI
+;comment=Set explain_threshold to 0 to ensure that the slow query is recorded.
```
lavarou, line 27: I wanted the comments explaining these settings to be as close to the settings as possible. The reason for `;comment=...` construct is because `integration_runner` expects `key=value` format for anything (including comments) in `INI` directive.

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
```diff
@@ -0,0 +1,157 @@
...
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+When PDO base class constructor is used to create connection object
+to a database on a remote host via a TCP port,
+and database instance reporting is enabled, the agent should
+ - not generate errors
+ - record a datastore metrics
```
zsistla, line 12: Copy/paste comment occurs 27 times.
Suggestion to change:
`record a datastore metrics`
to be:
`record datastore metrics`

tests/integration/pdo/mysql/base-class/test_prepared_stmt_1.php
```diff
@@ -0,0 +1,110 @@
...
+newrelic.transaction_tracer.explain_threshold = 0
+;comment=Disable instance and database name reporting (it is tested separately).
```
zsistla, line 29: What's the advantage of disabling instance and database name reporting here?
It probably makes sense to have at least one test with instance and database name reporting disabled to ensure everything else still works, but since it's on by default, wouldn't it make sense to leave it on to ensure interoperability even if we aren't explicitly testing or looking for metrics for instance and database name reporting?

tests/integration/pdo/test_prepared_stmt_1.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
zsistla, line 6: Nice idea using the `.inc`s for code reusability!  Could you please add comments to the files to convey the context they expect and the general gist of what they do?

tests/integration/pdo/mysql/constructor/test_instance_reporting_2.php
```diff
@@ -0,0 +1,161 @@
+<?php
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
```
zsistla, line 7: This whole test seems duplicate of the same named [one in the base-class directory](https://github.com/newrelic/newrelic-php-agent/pull/993/files#diff-a28747bcacf6e28f41e0a8e6b95933d73810cfabc008fde280cf55ed9d815b31R8).

Shouldn't it be the `PDO's specialized subclass constructor is used to create connection object` version?

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
```diff
@@ -0,0 +1,157 @@
...
+      "db.statement": "DROP TABLE ENV[DATASTORE_COLLECTION];"
+    }
+  ]
+]
+*/
+
+require_once(realpath (dirname ( __FILE__ )) . '/../../test_instance_reporting.inc');
+require_once(realpath (dirname ( __FILE__ )) . '/../../../../include/config.php');
+
+test_instance_reporting(new PDO($PDO_MYSQL_DSN, $MYSQL_USER, $MYSQL_PASSWD), 0);
```
zsistla, line 157: The descriptions between this test and the test_instance_reporting_2.php are the same, but the code to connect is different here than for [reporting_2](https://github.com/newrelic/newrelic-php-agent/pull/993/files#diff-a28747bcacf6e28f41e0a8e6b95933d73810cfabc008fde280cf55ed9d815b31R156-R160).

Could the description be updated to explain the difference? And why the different way is needed?

Same comment to all reporting_1 vs reporting_2 test cases.
I see you've broken out the directories by base-class, constructor, factory, etc, but the reporting_2 cases for pdo/mysql all still reference base class and have different connection blocks than the rest of the pdo/mysql tests in their respective directories.
https://github.com/newrelic/newrelic-php-agent/pull/993/files#diff-ee3c25291b81a5eb33e3e6e95f4d3bb0c1e9c817001a1bc88af0ab8acc31453cR8
https://github.com/newrelic/newrelic-php-agent/pull/993/files#diff-9465b3b1aacc6abb291baf0a70e282de761c1fa60199e26771cc11e9fa6338a9R8

zsistla:
Overall LGTM, I just had a few comments (mostly style and comment questions).
It's great the fix is low risk and uses currently used code paths.
Awesome job increasing the test coverage and making it more easily extensible.

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
lavarou: The code to connect is the same and depends on the location of the test: base-class uses PDO base class constructor (`new PDO(...)`), factory uses static factory method from `PDO` class (`PDO::Connect(...)`), and constructor uses driver specific subclass of PDO base class (`new Pdo\{dbdriver}(...)`). What is different between test_instance_reporting_1 and test_instance_reporting_2 is DSN used in each of those methods: test_instance_reporting_1 uses remote host and TCP port vs test_instance_reporting_2 uses unix socket on local host and this difference is already reflected in tests' descriptions here: [1](https://github.com/newrelic/newrelic-php-agent/blob/feat/php-8.4/fix-pdo-instrumentation/tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php#L9) vs [2](https://github.com/newrelic/newrelic-php-agent/blob/feat/php-8.4/fix-pdo-instrumentation/tests/integration/pdo/mysql/base-class/test_instance_reporting_2.php#L9).

test_instance_reporting_1 and test_instance_reporting_2 differ in test expectation: see value of `"peer.hostname"` span's attribute in [1](https://github.com/newrelic/newrelic-php-agent/blob/feat/php-8.4/fix-pdo-instrumentation/tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php#L146) vs [2](https://github.com/newrelic/newrelic-php-agent/blob/feat/php-8.4/fix-pdo-instrumentation/tests/integration/pdo/mysql/base-class/test_instance_reporting_2.php#L145) as well as value of `"port_path_or_id"` in slow SQL trace [1](https://github.com/newrelic/newrelic-php-agent/blob/feat/php-8.4/fix-pdo-instrumentation/tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php#L93) vs [2](https://github.com/newrelic/newrelic-php-agent/blob/feat/php-8.4/fix-pdo-instrumentation/tests/integration/pdo/mysql/base-class/test_instance_reporting_2.php#L92).

tests/integration/pdo/mysql/base-class/test_prepared_stmt_1.php
lavarou: > What's the advantage of disabling instance and database name reporting here?

It makes maintaining these tests easier - these tests worry only about metrics, slow sql and span events, not so much about values of database instance attributes in the slow sql and span events. These tests also test functionality of disabling instance reporting - datastore instance attributes are not present in neither slow sql trace nor in span event.

> It probably makes sense to have at least one test with instance and database name reporting disabled to ensure everything else still works, but since it's on by default, wouldn't it make sense to leave it on to ensure interoperability even if we aren't explicitly testing or looking for metrics for instance and database name reporting?

There are tests dedicated to testing database instance reporting specifically (test_instance_reporting) and those tests test all possible and supported scenarios of the values of those attributes.

Testing too many things at once makes the tests harder to understand and harder to maintain. Slow sql functionality is different from datastore instance reporting functionality and that is why I decided to test them separately.

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
lavarou: Fixed in 9bb85e4fe01202286e4323acb21a212601e82b6e

tests/integration/pdo/test_prepared_stmt_1.inc
lavarou: Is bb4d7fba645f103122382e7a4943e4bb39f11797 what you had in mind?

tests/integration/pdo/mysql/base-class/test_query_3.php
```diff
@@ -0,0 +1,49 @@
...
+Datastore/operation/MySQL/select, 1
+Datastore/statement/MySQL/test/select, 1
+Datastore/operation/MySQL/drop, 1
+Datastore/statement/MySQL/test/drop, 1
+*/
+
+require_once(realpath (dirname ( __FILE__ )) . '/../../test_query_3.inc');
+require_once(realpath (dirname ( __FILE__ )) . '/../../../../include/config.php');
+
+test_pdo_query(new PDO($PDO_MYSQL_DSN, $MYSQL_USER, $MYSQL_PASSWD), 0);
```
mfulb, line 49: Are there any concerns about interactions of test runs when a test fails?  I'm thinking of each test creating and destroying a table - is it possible a table drop doesn't occur due to a failure, leading to an unclean test environment for the next test?

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
bduranleau-nr: Instead of `1` and `2`, can the test names be updated to reflect the core difference between them? Something like `_local` & `_remote` OR `_port` & `_socket` would go a long way in clarifying the nature of the test vs `_1` and `_2`.

tests/integration/events/test_database_duration.php
```diff
@@ -49,7 +49,7 @@ function test_slow_sql() {
   global $PDO_MYSQL_DSN, $MYSQL_USER, $MYSQL_PASSWD;
 
   $conn = new PDO($PDO_MYSQL_DSN, $MYSQL_USER, $MYSQL_PASSWD);
-  $result = $conn->query('select * from tables limit 1;');
+  $result = $conn->query('select * from information_schema.tables limit 1;');
```
mfulb, line 52: The changes above to `config.php` make it possible to override `MYSQL_DB` but it seems like the tests all hard code it to `information_schema` - what purpose does the override mechanism serve in this case?

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
```diff
@@ -0,0 +1,157 @@
...
+*/
+
+/*SKIPIF
+<?php require(realpath (dirname ( __FILE__ )) . '/../../skipif_mysql.inc');
+*/
+
+/*ENVIRONMENT
+DATASTORE_PRODUCT=MySQL
+DATASTORE_COLLECTION=test
+MYSQL_PORT=3306
```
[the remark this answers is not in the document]
mfulb: After discussion I understand this is necessary for as currently the `docker-compose.yml` does not define the `MYSQL_PORT` env var and it is used as part of the test expecations.

tests/integration/events/test_database_duration.php
mfulb: After discussion the hard coding of `information_schema` is OK as it is a standard reference in all relation databases we care about, and the override is used for other tests where the value comes from the env var.

tests/integration/pdo/mysql/base-class/test_instance_reporting_1.php
lavarou: @zsistla @bduranleau-nr Does ae4891bcf8222d98363a40c562df2daf2da3fcc2 work?

tests/integration/pdo/mysql/base-class/test_query_1.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
bduranleau-nr, line 1: Similar to the `port` vs `socket` naming, can we name the `query` tests to reflect the differences in the query being tested? For example, `_arg`, `_fetch_column`, `_fetch_class`, `_fetch_into`?

tests/integration/pdo/mysql/base-class/test_query_3.php
mfulb: That doesn't really address my concern - but I suppose if there is a failure of this nature it will be a bit trickier to troubleshoot but should be fairly obvious the cause was the previous test left the environment in an unclean state.

mfulb (approved):
Very thorough and encompassing improvement in our test coverage - nice Michal!

tests/integration/pdo/mysql/base-class/test_query_1.php
lavarou: @bduranleau-nr ce4c0fa51733fc493cbf2a1ee6e14826c7ecca6f?

bduranleau-nr (approved):
Very clean & detailed addition 👍

tests/integration/pdo/test_prepared_stmt_1.inc
zsistla: Much better, thank you!
Could you please also move the last line in the .incs up to the top so they aren't lost after the block of text of different.
So instead of this:
```
/*
  This file contains code that runs the PDO::query test using given connection object.
  It is common for all database driver specific tests that prepare connection objects
  using different methods:
  - ./mysql:
    - ./base-class
    - ./constructor
    - ./factory
  - ./pgsql:
    - ./base-class
    - ./constructor
    - ./factory
  - ./sqlite:
    - ./base-class
    - ./constructor
    - ./factory
  It tests one argument form of PDO::query().
*/
```

It's like this:
```
/*
  This file contains code that runs the PDO::query test using given connection object.  It tests one argument form of PDO::query().
  It is common for all database driver specific tests that prepare connection objects
  using different methods:
  - ./mysql:
    - ./base-class
    - ./constructor
    - ./factory
  - ./pgsql:
    - ./base-class
    - ./constructor
    - ./factory
  - ./sqlite:
    - ./base-class
    - ./constructor
    - ./factory
*/
```

lavarou: @zsistla bd5f70438b44a672e36e1c95d91df81e524707bd?

zsistla: Thanks!

zsistla (approved):

PR #995, opened 2024-12-11
fix(agent): fix Drupal error and exception handling

bduranleau-nr:


agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
+  if (!nr_php_is_zval_valid_object(exception)) {
+    nrl_verbosedebug(NRL_TXN, "Drupal: getException() returned a non-object");
+    goto end;
+  }
+
+  if (NR_SUCCESS
+      != nr_php_error_record_exception(NRPRG(txn), exception, priority, true,
+                                       NULL,
+                                       &NRPRG(exception_filters) TSRMLS_CC)) {
+    nrl_verbosedebug(NRL_TXN, "Drupal: unable to record exception");
```
bduranleau-nr, line 69: Should this be higher priority than `verbosedebug`?

agent/fw_drupal8.c
```diff
@@ -730,6 +785,32 @@ void nr_drupal8_enable(TSRMLS_D) {
                                     "er::getControllerFromDefinition"),
                             nr_drupal8_name_the_wt TSRMLS_CC);
 
+  /*
+   * ExceptionSubscribers handle Drupal errors and exceptions before
+   * the agent has the opportunity to capture them. Instrument several
+   * of these ExceptionSubscriber function `onException` methods in order
+   * to capture Exceptions and Errors in Drupal 9.x+
```
lavarou, line 792: Why these particular subscribers were selected?

agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
 
+NR_PHP_WRAPPER(nr_drupal_exception) {
+  int priority = nr_php_error_get_priority(E_ERROR);
+  zval* event = NULL;
+  zval* exception = NULL;
+
+  /* Warning avoidance */
+  (void)wraprec;
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL8);
```
mfulb, line 32: Should this use a variation of the `NR_PHP_WRAPPER_REQUIRE_FRAMEWORK` macro (which doesnt yet exist) to make sure a compatible version of Drupal is present?  This works with 9+ correct?

agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
+                     "have an `event` parameter");
+    goto end;
+  }
+
+  /*
+   * Get the exception from the event.
+   */
+  exception = nr_php_call(event, "getThrowable");
+  if (!nr_php_is_zval_valid_object(exception)) {
+    exception = nr_php_call(event, "getException");
```
lavarou, line 57: Can `exception` variable be repurposed like that? Doesn't it have to be freed first?

agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
+
+  if (!nr_php_is_zval_valid_object(exception)) {
+    nrl_verbosedebug(NRL_TXN, "Drupal: getException() returned a non-object");
+    goto end;
+  }
+
+  if (NR_SUCCESS
+      != nr_php_error_record_exception(NRPRG(txn), exception, priority, true,
+                                       NULL,
+                                       &NRPRG(exception_filters) TSRMLS_CC)) {
```
lavarou, line 68:
```suggestion
                                       &NRPRG(exception_filters))) {
```

agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
+
+  NR_PHP_WRAPPER_REQUIRE_FRAMEWORK(NR_FW_DRUPAL8);
+
+  if (NR_SUCCESS != nr_txn_record_error_worthy(NRPRG(txn), priority)) {
+    NR_PHP_WRAPPER_CALL;
+    goto end;
+  }
+
+  /* Get the event that was given. */
+  event = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS TSRMLS_CC);
```
lavarou, line 40:
```suggestion
  event = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
```

agent/fw_drupal8.c
bduranleau-nr: This fixes exceptions for 8 and exceptions _and_ errors for 9+. For Drupal 8 errors, this code has no impact (positive or negative) as the problem with error reporting originates from this [block](https://github.com/newrelic/newrelic-php-agent/blob/R11.4/agent/php_execute.c#L1389) unique to pre-oapi instrumentation, and D8 is only compatible with PHP up to 7.4.

agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
+  if (0 == nr_php_is_zval_valid_object(event)) {
+    nrl_verbosedebug(NRL_TXN,
+                     "Drupal: ExceptionSubscriber::onException() does not "
+                     "have an `event` parameter");
+    goto end;
+  }
+
+  /*
+   * Get the exception from the event.
+   */
```
mfulb, line 54: This code seems to rely on `nr_php_call()` gracefully handling if these methods did not exist.  Have you verified this is how it works?
I am also curious if these calls could return a valid zval that is not an exception object and perhaps it would be best to verify it is what we expect?
Or does `nr_php_error_record_exception()` handle this inspection and verification?

agent/fw_drupal8.c
```diff
@@ -20,6 +21,60 @@
...
+  exception = nr_php_call(event, "getThrowable");
+  if (!nr_php_is_zval_valid_object(exception)) {
+    exception = nr_php_call(event, "getException");
+  }
+
+  if (!nr_php_is_zval_valid_object(exception)) {
+    nrl_verbosedebug(NRL_TXN, "Drupal: getException() returned a non-object");
+    goto end;
+  }
```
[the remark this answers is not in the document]
bduranleau-nr: It's there to avoid having to call `nr_php_error_record_exception`, which avoids having to call `nr_php_error_zval_is_exception`, which avoids having to call `nr_php_object_instaceof_class`, then finally call `nr_php_is_zval_valid_object` if it's a non-object entirely and instead just release the `event` and `exception` directly.
Depending on how often this will trigger, it's either more efficient to keep this check (if we expect it to fail often enough) or more efficient to remove it (if we expect it to pass almost all of the time). I'm not positive which case is more likely, but if the consensus is that this check should pass in the overwhelming amount of cases, I'll remove it.

agent/fw_drupal8.c
bduranleau-nr: > This code seems to rely on nr_php_call() gracefully handling if these methods did not exist. Have you verified this is how it works?

I believe this is outside the scope of this PR. We use `nr_php_call` in this same manner throughout the agent, but for what it's worth, I do believe it handles the case that it does not exist gracefully.

> I am also curious if these calls could return a valid zval that is not an exception object and perhaps it would be best to verify it is what we expect?

The very first call in `nr_php_error_record_exception` is to `nr_php_error_zval_is_exception`.

agent/fw_drupal8.c
bduranleau-nr: This is heavily borrowed from our [symfony instrumentation](https://github.com/newrelic/newrelic-php-agent/blob/R11.4/agent/fw_symfony4.c#L49). I believe the assumption is that a failed `nr_php_call` will return `NULL`, in which case `exception` doesn't have to be freed. I can add a `free` if you'd like.

agent/fw_drupal8.c
bduranleau-nr: Removed the HttpExceptionSubscriber wrapper in favor of just the Logging and Final Exception Subscribers. Logging Subscriber should be the highest priority, since it logs an exception, but doesn't handle it. Final Subscriber is the last line of defense in case an exception has escaped all other subscribers.

agent/fw_drupal8.c
bduranleau-nr: addressed here: 82ec0d73ea92f604463b1623b5e93f43499ef474

lavarou (approved):
LGTM

PR #1001, opened 2024-12-19
fix(agent): address the aws-sdk-php versioning change

zsistla:
Concerns about future/past proofing to the checking prioritized the non-eval implementation vs using the eval method for checking the version. However, aws-sdk-php internal functions frequently change without notice in the release notes. An extra eval is used to avoid the fatal error that would occur in the VERY unlikely event of Aws/Sdk.php not being found and which addresses the previous concerns while decoupling versioning from internal functions.

### DIFF
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -30,9 +30,8 @@
  * error whenever it cannot find a class which must be caught. Calling this
  * from nr_aws_sdk_php_enable would allow the sdk version value to be set only
  * once. To avoid the VERY unlikely but not impossible fatal error, we need to
- * take an extra step to wrap the call in a try/catch block. This means we need
- * to have an additionial zend_string_eval to get the result, but we avoid fatal
- * errors.
+ * wrap the call in a try/catch block and make it a lambda so that we avoid
+ * fatal errors.
  */
 void nr_lib_aws_sdk_php_handle_version() {
   char* version = NULL;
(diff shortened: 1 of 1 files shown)


agent/lib_aws_sdk_php.c
```diff
@@ -22,49 +22,43 @@
...
- *
- * Concerns about future/past proofing to the checking prioritized the following
- * implementation vs using the eval method.
+ * settings, if the class is not found, PHP 8.2+ will generate an
+ * error whenever it cannot find a class which must be caught. Calling this
+ * from nr_aws_sdk_php_enable would allow the sdk version value to be set only
+ * once. To avoid the VERY unlikely but not impossible fatal error, we need to
+ * take an extra step to wrap the call in a try/catch block. This means we need
+ * to have an additionial zend_string_eval to get the result, but we avoid fatal
+ * errors.
```
lavarou, line 35: It is possible to use a single zend_eval_string to get the sdk version with error handling. Take a look at https://github.com/newrelic/newrelic-php-agent/blob/R11.4/agent/lib_composer.c#L91. That code defines a lambda (anonymous function) and calls it in place. The same idea can be used for the AWS SDK version retrieval. The PHP code to be used in a single `zend_string_eval` could look like this:
```php
(function() {
     $nr_aws_sdk_version = '';
     try {
          $nr_aws_sdk_version = Aws\\Sdk::VERSION;
     } catch (Throwable $e) {
     }
     return $nr_aws_sdk_version;
})();
```

lavarou (approved):
LGTM. Thanks for fixing this!

PR #1003, opened 2024-12-30
feat(agent): Add aws-sdk-php SQS instrumentation

zsistla:
Uses message segments to create SQS instrumentation.

Added unit tests and multiverse tests.


agent/lib_aws_sdk_php.h
```diff
@@ -20,4 +30,71 @@ extern void nr_lib_aws_sdk_php_handle_version();
...
     const char* service_name);
 
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO /* PHP8.1+ */
+/* Aside from service class and version detection, instrumentation is only
+ * supported with PHP 8.1+ */
+
+/*
+ * Purpose : Parses the QueueUrl to extract cloud_region, cloud_account_id, and
+ * destination_name.  The extraction sets all or none since the values from the
+ * same string and if it is malformed, it cannot be used.
```
lavarou, line 40:
```suggestion
 * destination_name.  The extraction sets all or none since the values are from
 * the same string and if it is malformed, it cannot be used.
```

agent/lib_aws_sdk_php.h
```diff
@@ -20,4 +30,71 @@ extern void nr_lib_aws_sdk_php_handle_version();
...
+ */
+extern char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
+                                                      NR_EXECUTE_PROTO);
+
+/*
+ * Purpose : Return the first argument to the Aws/AwsClient::__call function
+ * which should be the name of the command.
+ *
+ * Params  : 1. NR_EXECUTE_PROTO (execute_data, func_return_value)
+ *           2. command_name: name of the command being called
```
lavarou, line 90: Function's declaration doesn't have a second argument:
```suggestion
```

agent/lib_aws_sdk_php.h
```diff
@@ -20,4 +30,71 @@ extern void nr_lib_aws_sdk_php_handle_version();
...
+    const char* sqs_queueurl,
+    nr_segment_message_params_t* message_params,
+    nr_segment_cloud_attrs_t* cloud_attrs);
+
+/*
+ * Purpose : Handle when an SqsClient initiates a command
+ *
+ * Params  : 1. NR_EXECUTE_ORIG_ARGS (execute_data, func_return_value)
+ *           2. segment : if we instrument the commandName, we'll need to end
+ * the segment as a message segment Returns :
```
lavarou, line 63:
```suggestion
 * the segment as a message segment
 *
 * Returns :
```

agent/lib_aws_sdk_php.h
lavarou, line 1: The order of functions' arguments in descriptions does not match the order of arguments in function declarations.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,358 @@
...
+  };
+  nr_segment_cloud_attrs_t cloud_attrs = {0};
+
+  if (NULL == segment) {
+    return;
+  }
+
+  command_name_string
+      = nr_lib_aws_sdk_php_get_command_name(NR_EXECUTE_ORIG_ARGS);
+
```
lavarou, line 123: Check if `command_name_string` is `NULL`, and return early if it is.

agent/lib_aws_sdk_php.h
zsistla: in this PR: https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5

agent/lib_aws_sdk_php.h
zsistla: Thanks for catching, descriptions needed updating after a refactor. Updated here:
https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5

agent/lib_aws_sdk_php.h
zsistla: Thanks for catching, descriptions needed updating. Updated here:
https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5

agent/lib_aws_sdk_php.h
zsistla: Thanks for catching, descriptions needed updating after a refactor. Updated here:
https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5

agent/lib_aws_sdk_php.c
zsistla: Location has changed, but we do check for null here:
https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,409 @@
...
+  char* real_class_and_command = NULL;
+  nr_segment_t* segment = NULL;
+
+  klass
+      = nr_php_class_entry_name(Z_OBJCE_P(nr_php_execute_scope(execute_data)));
+  command_name_string
+      = nr_lib_aws_sdk_php_get_command_name(NR_EXECUTE_ORIG_ARGS);
+
+  if (NULL != command_name_string && NULL != klass) {
+    if (nr_striendswith(klass, nr_strlen(klass), AWS_SDK_PHP_SQSCLIENT_CLASS,
```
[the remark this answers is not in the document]
zsistla: Please see: [1c3e374](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/1c3e3740439f9304c7fe75567267bec2d0577267)

agent/lib_aws_sdk_php.h
```diff
@@ -20,4 +33,73 @@ extern void nr_lib_aws_sdk_php_handle_version();
...
+ * Returns :
+ *
+ */
+extern void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
+                                          char* command_name_string,
+                                          NR_EXECUTE_PROTO);
+
+/*
+ * Purpose : The second argument to the Aws/AwsClient::__call function should be
+ * an array containing an array of argument name:value pairs. Given an argument
```
lavarou, line 77: Is it an array containing an array or is it just an array containing arguments to the command stored as name_of_an_argument => value?

agent/lib_aws_sdk_php.h
lavarou: Functions' doc strings now match reality. However this has not been fixed in c1f163f7c43f1b731e1b94a515b0ec47cf4383f5 commit.

agent/lib_aws_sdk_php.h
lavarou: Functions' doc strings now match reality. However this has not been fixed in https://github.com/newrelic/newrelic-php-agent/commit/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5 commit.

agent/lib_aws_sdk_php.h
lavarou: Functions' doc strings now match reality. However this has not been fixed in https://github.com/newrelic/newrelic-php-agent/commit/c1f163f7c43f1b731e1b94a515b0ec47cf4383f5 commit.

agent/lib_aws_sdk_php.c
lavarou: c1f163f commit has nothing to do with `NULL` check. However the code has changed and this comment no longer applies.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,411 @@
...
+  if (0 != nr_php_is_zval_non_empty_string(command_name)) {
+    command_name_string = nr_strdup(Z_STRVAL_P(command_name));
+  }
+
+  nr_php_arg_release(&command_name);
+  return command_name_string;
+}
+
+void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
+                                   char* command_name_string,
```
lavarou, line 107: It might actually be better for this subroutine to take `command_name_string` as `zend_string`. The benefits are because `zend_string` type has additional information that could be useful to speed up determination of the command should be instrumented, namely it has length and hash, both which could be used to make this determination more performant.

agent/lib_aws_sdk_php.h
```diff
@@ -20,4 +33,73 @@ extern void nr_lib_aws_sdk_php_handle_version();
...
+ * which should be the name of the command.
+ *
+ * Params  : 1. NR_EXECUTE_PROTO (execute_data, func_return_value)
+ *
+ * Returns : the value of the command_name; NULL if does not exist
+ *
+ * Note: The caller is responsible for freeing the returned string value
+ *
+ */
+extern char* nr_lib_aws_sdk_php_get_command_name(NR_EXECUTE_PROTO);
```
lavarou, line 102: Could this return `zend_string *` instead of raw `char *`? See [this comment](https://github.com/newrelic/newrelic-php-agent/pull/1003/files#r1909385990).

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,411 @@
...
+            'BinaryValue' => 'string',
+            'StringListValues' => array('string', ... ),
+            'BinaryListValues' => array('string', ... ),
+            // DataType is required
+            'DataType' => 'string',
+        ),
+        // ... repeated
+    ),
+));
+
```
mfulb, line 58:
```suggestion
```

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,411 @@
...
+  /* Determine if we instrument this command. */
+  if (nr_streq(command_name_string, "sendMessage")
+      || nr_streq(command_name_string, "sendMessageBatch")) {
+    message_params.message_action = NR_SPAN_PRODUCER;
+    instrumented = true;
+  } else if (nr_streq(command_name_string, "receiveMessage")) {
+    message_params.message_action = NR_SPAN_CONSUMER;
+    instrumented = true;
+  }
```
mfulb, line 130: Would it make sense to catch the `else` case and flag it or return if we hit it?

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,411 @@
...
+      nr_lib_aws_sdk_php_sqs_handle(auto_segment, command_name_string,
+                                    NR_EXECUTE_ORIG_ARGS);
+    }
+  }
+
+  if (NR_SEGMENT_CUSTOM == auto_segment->type) {
+    /*
+     * It wasn't set to any of the segments that would autoclose the segment so
+     * we need to end the segment that we started in the before wrapper to
+     * handle external,datastore,message segments
```
lavarou, line 395:
```suggestion
```suggestion
   * We need to end the segment that we started in the 'before' wrapper if it
   * wasn't handled and ended by the handling function. Handling function would
   * have changed the segment type from from default (`NR_SEGMENT_CUSTOM`) if it
   * ended it.
```

agent/lib_aws_sdk_php.h
zsistla: It's an array containing an array of name:value pairs.

agent/lib_aws_sdk_php.c
zsistla: Interesting idea, but after looking at the numbers, I can't see that it would be more performant over the short command name streq comparisons.

For the command name comparison, hashes didn't prove to be more performant for command names less than 30 characters (pretty typical).

Because[ hashes are computed lazily](https://www.phpinternalsbook.com/php7/internal_types/strings/zend_strings.html#the-zend-string-api), worst case we'd need to compute the hash every time we encounter a command_name zend_string, best case, the zend_string hash is already computed.

With the hash compare, I looked at 2 different cases, one where we had to compute the hash before comparing the hash against 2-3 other hashes that we care about, and the other where no hash was computed and we just did the comparison.
For instance, say we have two-three command names we care about:
1) We can precompute both of those for later use so that won't take additional time.
2) When we encounter the command_name string, because it isn't persistent and it's hash wasn't needed, the hash hasn't been computed yet, so we need to compute it so this was included in the duration.
3) we have 2 hashes we care about, so worst case, we'd compare the zend_string hash against our known hashes, this was included in the duration.

vs 
1) worst case, two-three streq comparisons that were included in the duration

Results showed:
1) worst case of computing the hash for every was much worse than streq for both cases of 2 and 3 command names that we cared about:

two command hash to streq
>>> (hash_duration)=38217 (streq_duration)=28238

three command hash to streq
>>> (hash_duration)=47204 (streq_duration)=30784

2) for the best hash case, where we didn't compute the hash at all, the duration numbers were basically the same:

three command don't compute hash to streq
>>> (hash_duration)=29898 (streq_duration)=28304

Given that best case is basically the same, but worst case is worse, it doesn't make sense for this change.

It might have helped with class name comparisons since class names could get longer, but we already changed the logic there to only check for class name (not full path so comparing SqsClient instead of comparing Aws\Sqs\SqsClient) and for class we're already utilizing the zend_string length field.

agent/lib_aws_sdk_php.h
zsistla: See comment: https://github.com/newrelic/newrelic-php-agent/pull/1003#discussion_r1909652756

agent/lib_aws_sdk_php.c
zsistla: Nice idea.
`instrumented` bool and functionality were part of an earlier logic, so removing these as unneeded artifacts from that refactor.
Please see:
[e27135a](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/e27135a7898690d106fdfb09c3a6c06ffefb439a)

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,409 @@
...
+   * If we are unable to match any part of this, the whole decode is suspect and
+   * all values are discarded.
+   *
+   * Due to the overhead involved in escaping the original buffer, creating a
+   * regex, matching a regex, destroying a regex, this method was chosen as a
+   * more performant option because it's a very limited pattern.
+   */
+
+  if (NULL == nr_strlcpy(queueurl, sqs_queueurl, AWS_QUEUEURL_LEN_MAX)) {
+    /* Malformed queueurl, we can't decode this. */
```
mfulb, line 192: Looking at the code for `strlcpy()` the only reason this fails is because `sqs_queueurl` is NULL or zero length.  The comment seems to indicate to me some type of check of the actual contents of the `queueurl` is happening - is there more to this?

agent/lib_aws_sdk_php.c
lavarou: Thanks for digging into this and running some benchmarks - having this type of data always ensures correct decision is made. I can understand that hash computation may be less performant than doing `strcmp` for short strings. However, I still think that we can take advantage of string length. `nr_lib_aws_sdk_php_sqs_handle` subroutine will be called for all methods of `Aws\Sqs\SqsClient` class. Currently this function cares only for 3 methods: `sendMessage`, `sendMessageBatch` and `receiveMessage`. We can cheaply compare if input command's length is not 11 or 16 or 14 and short circuit immediately without doing 3 unnecessary `strcmp`s. Additionally, it only makes sense to do `strcmp` for `sendMessage` or `sendMessageBatch` if input command's length is 11 or 16 (which could actually be split into its own `else if` case to speed things up even further). The same applies to `receiveMessage` - it only makes sense to do `strcmp` if input command's length is 14.

zsistla: incorporating Strlen check does give significant improvements, especially for nonmatching cases. Test cases all involved three lengths/strings we were interested in.

regular steq for three commands that only succeeds on last compare (this would also be equal to the nomatch case) =43165

check len/string for each command, match len on the last check so one streq occurs: 
16672 

check len/string for each command, no matches, so no streq:
1783

I'll refactor the command check and also the class check.

zsistla: See [2a74246](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/2a7424662dce50a181b4dc55a8396ccbe8ebd607) for changes.

Since research showed comparing strlen first significantly reduced lookup times, especially in non-matching cases.
The helper function was refactored out and the wrapper refactored to use the already computed strlen in the zval/zend_string from nr_php_arg_get.
The helper function originally took the zval/zend_string and strduped it onto a char*.
Since we weren't helping by converting to string anymore, keeping the helper function would have meant having to do the less optimal options of the calling function remember to nr_php_arg_release an arg it didn't directly get, copy the zval, add a ref to the zval that would need to be freed.
Simpler just to extract the functionality back to the main function.  One positive of helper functions is they are easier to test, but shouldn't be at the detriment of code readability or optimization. nr_php_arg_get was already directly giving us the zval/zend_string availability that we wanted.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,409 @@
...
+    segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
+    if (NULL != segment) {
+      real_class_and_command
+          = nr_formatf("Custom/%s::%s", klass, command_name_string);
+      nr_segment_set_name(segment, real_class_and_command);
+    }
+  }
+
+  nr_free(command_name_string);
+  nr_free(real_class_and_command);
```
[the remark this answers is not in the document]
zsistla: I refactored before reading this comment, but I think the refactor addressed this concern.
[2a74246](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/2a7424662dce50a181b4dc55a8396ccbe8ebd607)

agent/lib_aws_sdk_php.c
zsistla: class comparison that first checks string length here:
[b1e868f](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/b1e868fda7210187e4b01c3576f1cd5fb9253876)

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,420 @@
...
+    /* Malformed queueurl, we can't decode this. */
+    return;
+  }
+  queueurl_pointer = queueurl;
+
+  /*
+   * Find the pattern of the AWS queueurl that should immediately precede the
+   * region.
+   */
+  if (0 != strncmp(queueurl_pointer, "https://sqs.", 12)) {
```
mfulb, line 203: Based on our conversation - replace hard coded numbers with a `#define` and `sizeof()`.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,420 @@
...
+  cloud_attrs.aws_operation = command_name_string;
+
+  command_arg_value = nr_lib_aws_sdk_php_get_command_arg_value(
+      AWS_SDK_PHP_SQSCLIENT_QUEUEURL_ARG, NR_EXECUTE_ORIG_ARGS);
+
+  /*
+   *  nr_lib_aws_sdk_php_sqs_parse_queueurl checks for NULL so safe pass
+   * command_arg_value directly in.
+   */
+  nr_lib_aws_sdk_php_sqs_parse_queueurl(command_arg_value, &message_params,
```
mfulb, line 142: Consider an optimization where a copy of `command_arg_value` is passed in and then the parsing function can modify this buffer in place and return pointers to the fields within this buffer to avoid strdup'ing the 3 fields in the parsing function.  Also removes the need to free these later.  Just need to document the parse function REQUIRES a copy that can be modified.

agent/lib_aws_sdk_php.c
zsistla: Please see: [09f10e8](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/09f10e854f9818202179412645b59fc8039b6c27)

agent/php_execute.c
```diff
@@ -1260,7 +1260,10 @@ static inline void nr_php_execute_segment_add_metric(
   if (create_metric) {
     nr_segment_add_metric(segment, buf, true);
   }
-  nr_segment_set_name(segment, buf);
+  if (!segment->name) {
```
mfulb, line 1263: How is this related to SQS instrumentation?  This is a code path run for all instrumentation so I'm trying to understand why this general a change is required for SQS and if this is not something we've handled differently before.

agent/lib_aws_sdk_php.c
zsistla: Please see: [f48df3f](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/f48df3f365177c9a137f6d11f5a9261681469b80)

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -6,12 +6,284 @@
...
+  char* arg_in_array_not_array
+      = "array("
+        "    0 => '1'"
+        ")";
+  array_arg = tlib_php_request_eval_expr(arg_in_array_not_array);
+  expr = nr_php_call(NULL, "two_param", first_arg, array_arg);
+  tlib_pass_if_not_null("Expression should evaluate.", expr);
+  nr_php_zval_free(&expr);
+  nr_php_zval_free(&array_arg);
+
```
mfulb, line 149: What if the `QueueUrl` key has a value which is not the correct type, like an `array`?

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -6,12 +6,284 @@
...
+    char* cloud_account_id,
+    char* destination_name) {
+  tlib_pass_if_str_equal("cloud_region should match.", cloud_attrs.cloud_region,
+                         cloud_region);
+  tlib_pass_if_str_equal("cloud_account_id should match.",
+                         cloud_attrs.cloud_account_id, cloud_account_id);
+  tlib_pass_if_str_equal("destination_name should match.",
+                         message_params.destination_name, destination_name);
+
+  nr_free(cloud_attrs.cloud_region);
```
mfulb, line 192: If the parse function is changed to not strdup the values this function will need to be modified too.

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -6,12 +6,284 @@
...
+#define INVALID_QUEUE_URL_1 "https://us-east-2.amazonaws.com/123456789012/SQS_QUEUE_NAME"
+#define INVALID_QUEUE_URL_2 "https://sqs.us-east-2.amazonaws.com/123456789012/"
+#define INVALID_QUEUE_URL_3 "https://sqs.us-east-2.amazonaws.com/SQS_QUEUE_NAME"
+#define INVALID_QUEUE_URL_4 "https://random.com"
+#define INVALID_QUEUE_URL_5 "https://sqs.us-east-2.amazonaws.com/123456789012"
+#define INVALID_QUEUE_URL_6 "https://sqs.us-east-2.amazonaws.com/"
+#define INVALID_QUEUE_URL_7 "https://sqs.us-east-2.amazonaws.com"
+#define INVALID_QUEUE_URL_8 "https://sqs.us-east-2.random.com/123456789012/SQS_QUEUE_NAME"
+  // clang-format on
+
```
mfulb, line 228: Can you add tests where the 2nd and 3rd args are `NULL` instead of pointing to the message and cloud structs?

agent/php_execute.c
zsistla: Not required for just SQS but for all AWS instrumentation.
Because we instrument the Aws/AwsClient::__call and decode the calling command, we are able to give a more meaningful name [here](https://github.com/newrelic/newrelic-php-agent/pull/1003/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R398).

So, instead of naming the segment `Aws//AwsClient::__call` we can name it `Aws\\Sqs\\SqsClient::sendMessage`.

However, when the segment was getting set in php_execute, it wasn't checking if a more meaningful name had already been assigned.  This just checks if a name has already been applied to the segment, and if so don't overwrite it; otherwise, continue as normal.

axiom/nr_segment_message.c
```diff
@@ -206,12 +208,15 @@ bool nr_segment_message_end(nr_segment_t** segment_ptr,
...
    * Therefore, we delete all children of the message segment.
+   * By destroying the tree we are able to destroy all descendants vs just
+   * destroying the child which then reparents all it's children to the segment.
    */
   if (segment) {
     for (size_t i = 0; i < nr_segment_children_size(&segment->children); i++) {
-      nr_segment_t* child = nr_segment_children_get(&segment->children, i);
-      nr_segment_discard(&child);
+      child = nr_segment_children_get(&segment->children, i);
+      nr_segment_destroy_tree(child);
```
mfulb, line 217: The header docs for this function say " * WARNING : This should only be called during transaction destruction."  Do we know this is safe at this point?

agent/tests/test_lib_aws_sdk_php.c
zsistla: In [this case](https://github.com/newrelic/newrelic-php-agent/pull/1003/files#diff-5cbabeb350f68acfaa71e353ed2387f1d2aa71b7b03201d820a283c3810e3aa2R220), the third argument queue_name would not exist.

In [this case](https://github.com/newrelic/newrelic-php-agent/pull/1003/files#diff-5cbabeb350f68acfaa71e353ed2387f1d2aa71b7b03201d820a283c3810e3aa2R224), we wouldn't decode the 2nd argument (account id).

If [one of the three is NULL, all will be NULL](https://github.com/newrelic/newrelic-php-agent/pull/1003/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R175); otherwise, the queueUrl is malformed and cannot be trusted.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,409 @@
...
+  } else if (command_name_len == AWS_SQS_RECEIVE_MESSAGE_COMMAND_LEN
+             && 0
+                    == nr_strncmp(AWS_SQS_RECEIVE_MESSAGE_COMMAND,
+                                  command_name_string,
+                                  AWS_SQS_RECEIVE_MESSAGE_COMMAND_LEN)) {
+    message_params.message_action = NR_SPANKIND_CONSUMER;
+  } else {
+    /* Nothing to do here so exit. */
+    return;
+  }
```
lavarou, line 131: This code has so much indirection and requires jumping back and forth to check if the macros have correct values and what those values actually are. Wouldn't this be easier to read and maintain:
```suggestion
#define COMMAND_IS(CMD)                  \
  (command_name_len == (sizeof(CMD) - 1) \
   && 0 == nr_strcmp(CMD, command_name_string))

  /* Determine if we instrument this command. */
  if (COMMAND_IS("sendBatch")) {
    message_params.message_action = NR_SPANKIND_PRODUCER;
  } else if (COMMAND_IS("sendMessage")) {
    message_params.message_action = NR_SPANKIND_PRODUCER;
  } else if (COMMAND_IS("receiveMessage")) {
    message_params.message_action = NR_SPANKIND_CONSUMER;
  } else {
    /* Nothing to do here so exit. */
    return;
  }
#undef IS_COMMAND
```
Suggested change avoids the redirection, and also avoids defining and maintaining the macros for each command and its length.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,409 @@
...
+  if (NULL == command_name_string || 0 == command_name_len) {
+    return;
+  }
+
+  /* Determine if we instrument this command. */
+  if (command_name_len == AWS_SQS_SEND_MESSAGE_BATCH_COMMAND_LEN
+      && 0
+             == nr_strncmp(AWS_SQS_SEND_MESSAGE_BATCH_COMMAND,
+                           command_name_string,
+                           AWS_SQS_SEND_MESSAGE_BATCH_COMMAND_LEN)) {
```
lavarou, line 114: What is the advantage of using `nr_strncmp` over `nr_strncmp` or `nr_streq`?

axiom/nr_segment_message.c
zsistla: Yes, the warning is because it permanently walks the tree and destroys all children and their children.  In this case, where we always do not want message segments to have children, that is exactly what we want.  In most other cases, discard is more appropriate since it reparents a discarded child segment's children to the discarded child segment's parent.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,409 @@
...
+      if (NULL != segment) {
+        real_class_and_command
+            = nr_formatf("Custom/%s::%s", klass, command_name_string);
+        nr_segment_set_name(segment, real_class_and_command);
+        nr_free(real_class_and_command);
+      }
+    }
+    /* Release the command_name. */
+    nr_php_arg_release(&command_name);
+  }
```
lavarou, line 415: newrelic-php-agent's code base doesn't seem to have a standard around structuring the code. Golang has the following standard defined in [Effective Go](https://go.dev/doc/effective_go#if):
> [...] a common situation where code must guard against a sequence of error conditions. The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements, the resulting code needs no else statements.

The readability of the above could could definitely benefit from applying Effective Go standards. Consider:
```
  if (NULL == klass) {
    return;
  }
  ...
  if (!nr_php_is_zval_non_empty_string(command_name)) {
    nr_php_arg_release(&command_name);
    return;
  }
  
```

lavarou: This suggestion is something to consider but is not blocking approval.

agent/lib_aws_sdk_php.c
lavarou: This suggestion is something to consider but is not blocking approval.

agent/lib_aws_sdk_php.c
lavarou: `nr_lib_aws_sdk_php_sqs_parse_queueurl` follows the Effective Go standards and is easy to read.

agent/lib_aws_sdk_php.c
zsistla: Hold off on looking at this, it needs additional rework.

agent/tests/test_lib_aws_sdk_php.c
mfulb: I mean something like
```
nr_lib_aws_sdk_php_sqs_parse_queueurl(NULL, NULL, NULL);
```

agent/php_execute.c
zsistla: Correct.

agent/lib_aws_sdk_php.c
zsistla: Agreed, the macro makes the code more readable, but it means we've got to compute the size of a known string every single time we go through the handle_sqs code.

The defines were both to avoid hardcoding values as well as to avoid having to compute sizeof/strlen for those three values every time we go through the code.

For the case where we match the last string, the numbers for computing the size everytime vs having it already computed is:
use macro duration =979650 
precomputed sizes duration =957323

For the case where no strings match:
macro=62778 
precomputed=56180
----------

We could combine approaches to improve readability while maintaining the optimization.

The macro could be something similar to:
```

   #define COMMAND_IS(CMD) \
  (command_name_len == AWS_ ## CMD ## _COMMAND ## _LEN && (nr_streq(AWS_ ## CMD ## _COMMAND, command_name_string)))
```
then the code could look like:
```
  /* Determine if we instrument this command. */
  if (COMMAND_IS(SQS_SEND_MESSAGE_BATCH)) {
    message_params.message_action = NR_SPANKIND_PRODUCER;
  } else if (COMMAND_IS(SQS_SEND_MESSAGE)) {
    message_params.message_action = NR_SPANKIND_PRODUCER;
  } else if (COMMAND_IS(SQS_RECEIVE_MESSAGE)) {
    message_params.message_action = NR_SPANKIND_CONSUMER;
  } else {
    /* Nothing to do here so exit. */
    return;
  }
#undef IS_COMMAND
```

The above improves readability but as seen below this still keeps the performance improvements
macro comparison when len/string matches:
macro that computes lengths each time 924522
macro that uses precomputed lengths 866615

macro comparison when len/string matches:
macro that computes lengths each time 68946
macro that uses precomputed lengths 65715

agent/tests/test_lib_aws_sdk_php.c
zsistla: With the parse_queueurl refactor, tests need to be updated with modifiable strings and to not free the previously strduped strings.
[0b194c2](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/0b194c2382d09799451eee9274fe9ddd02531586)

agent/lib_aws_sdk_php.c
zsistla: Tests needed to be updated for this refactor to pass modifiable strings not literals.
[0b194c2](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/0b194c2382d09799451eee9274fe9ddd02531586)

agent/lib_aws_sdk_php.c
zsistla: strncmp was needed since the defined strings had an extra char on the end that we could account for if we used the precomputed length instead of the length computed in nr_streq.
This could be made more readable with a [macro](https://github.com/newrelic/newrelic-php-agent/pull/1003#discussion_r1915814849).

agent/lib_aws_sdk_php.c
zsistla: Agree, but it's not a simple return that can occur since it needs a return value.
I can however, make a `goto end` so that we'll skip to the bottom then all the [logic here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L274) can execute for the actual return value.

agent/tests/test_lib_aws_sdk_php.c
zsistla: Added here: [b63f74f](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/b63f74fb558b260d03c91d29c90f623fe0918fb3)

agent/tests/test_lib_aws_sdk_php.c
zsistla: Please see: [0230262](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/023026234bf80e79ef7d5f555087d290c4758ab8)

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -211,6 +207,7 @@ static void test_nr_lib_aws_sdk_php_sqs_parse_queueurl() {
    */
   nr_segment_message_params_t message_params = {0};
   nr_segment_cloud_attrs_t cloud_attrs = {0};
+  char modifiable_string[512];
```
mfulb, line 232: Can you use `AWS_QUEUEURL_LEN_MAX` instead of "512"?

agent/lib_aws_sdk_php.c
```diff
@@ -150,24 +149,15 @@ void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
...
   nr_free(command_arg_value);
-
-  /*
-   * These are the message_params params that were strduped in
-   * nr_lib_aws_sdk_php_sqs_parse_queueurl
-   */
-  nr_free(cloud_attrs.cloud_region);
-  nr_free(message_params.destination_name);
-  nr_free(cloud_attrs.cloud_account_id);
 }
```
mfulb, line 153: I think `nr_lib_aws_sdk_php_sqs_parse_queueurl` should set the fields in `message_params` and `cloud_attrs` to `NULL` at the start as the function can exit w/o actually storing these values.  This could lead to issues when the caller uses these uninitialized values.

zsistla: The values are already being initialized [here](https://github.com/newrelic/newrelic-php-agent/pull/1003/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R94-R99).
All cloud_attrs start as NULL, all message_params not explicitly set are set to NULL during that initialization.

agent/tests/test_lib_aws_sdk_php.c
zsistla: Please see: [43fa2b8](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/43fa2b888d6583004d8159325270f1de6c0e16e9)

agent/tests/test_lib_aws_sdk_php.c
zsistla: Please see: [43fa2b8](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/43fa2b888d6583004d8159325270f1de6c0e16e9)

agent/lib_aws_sdk_php.c
mfulb: I guess it is a question of expectations - my initial expectation was that this function to `NULL` those fields out if the `queryurl` was unparsable to reflect no data was found to fill them in properly.

agent/php_execute.c
mfulb: I'm curious if this line of code was doing something similar for external segments and how it my interplay with this change?

https://github.com/newrelic/newrelic-php-agent/blob/94d15754d6602dbd798be6badfa92ff12f0c687f/axiom/nr_segment_external.c#L106

agent/lib_aws_sdk_php.c
zsistla: Those 3 fields are already set to NULL by the caller who NULLs all the unknown fields prior to finding values for them.  Caller is responsible for initializing the structs, not parseQ.
parseQ should only modify the fields if it has something meaninful to add; otherwise, it should leave them untouched.  At some point in future, we might give any of these values "unknown" or something as default then pass it to parseQ, we wouldn't want parseQ to ovewrite the default with NULL.  parseQ is only reponsible setting the values if they are known; otherwise, caller is responsible for setting struct values.

agent/php_execute.c
zsistla: External naming will be unaffected by this change (as also seen by the passing multiverse/integration tests).

That line in external is doing the same thing message is doing [here](https://github.com/newrelic/newrelic-php-agent/pull/991/files#diff-2cb229cedb60157f4b25c632c7fa42b3874431be0de9ad70f03900d6f98b45e7R231).
Datastore, external, and message all manually start a segment, add all the typed attributes and category names, then manually end the segment with the appropriate name so those manually started/stopped segments will never be closed in the way that will try and give it another name.

This will affect generic segments that are closed in func_end that we just grab the function name as a way of naming.

In this particular case in this PR, it's naming the segment that has a manually started/stopped message segment as its child.

agent/lib_aws_sdk_php.c
lavarou: Doesn't the `AWS_ ## CMD ## _COMMAND ## _LEN ` expand to `sizeof (AWS_ ## CMD ## _COMMAND)`? I don't see how this macro:
```
 #define COMMAND_IS(CMD) \
  (command_name_len == AWS_ ## CMD ## _COMMAND ## _LEN && (nr_streq(AWS_ ## CMD ## _COMMAND, command_name_string)))
```
differs from this macro:
```
#define COMMAND_IS(CMD) \
  (command_name_len == (sizeof(CMD) - 1) && nr_streq(CMD, command_name_string))
```
other than it causes preprocessor to expand one more macro at compile time. I don't see how this could have any impact on runtime performance. I really don't see the point in having macro for each command name and for its length if the command string is used in a single place.

agent/lib_aws_sdk_php.c
lavarou: What is the extra char on the end in the defined string that needs to be accounted for?

agent/lib_aws_sdk_php.c
zsistla: I ran numerous timing tests on this.

When comparing the condition macro and comparing original format(macro-ized str/len) and when using streq in both cases, the best performer was the original version using name/len precomputed macros.
The conditional macros were pretty close(but the 2nd macro used strncmp).

Maybe it's in the difference of using a macro that is precomputed at runtime and can be used as a constant vs a macro that can't be computed at runtime because it still has unknowns (command_name_len and command_name_string).

The original version using only macro-ized strings and macro-ized lengths is noticeably more performant when using streq.
I've run the benchmarks against this multiple times, and the original code using the len macro is always more performant.

However, when using nrstrcmp (as needed to pass the proper strlength for comparison when using macroized strings), the performance gain is lost.  Given this and the added readability of the macro, after slight modification, I'll put this in and remove the macro-ized strings/lengths.

I'll do a similar thing for class compare.

agent/php_execute.c
lavarou: When `nr_segment_set_name(segment, buf)` will not execute because segment name has already been set, but `nr_segment_add_metric(segment, buf, true);` on line 1261 will create a metric with a different name, will it cause any issues in the backend that the segment name does not match the metric name?

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,402 @@
...
+  zval* command_name = NULL;
+  const char* klass = NULL;
+  char* command_name_string = NULL;
+  char* real_class_and_command = NULL;
+  nr_segment_t* segment = NULL;
+  zend_class_entry* class_entry = NULL;
+  int klass_len = 0;
+
+  class_entry = Z_OBJCE_P(nr_php_execute_scope(execute_data));
+  klass = nr_php_class_entry_name(class_entry);
```
lavarou, line 361: Add a check if `NULL != class_entry`:
```suggestion
  if (NULL == class_entry) {
    return;
  }
  klass = nr_php_class_entry_name(class_entry);
```

agent/lib_aws_sdk_php.c
zsistla: See the link in the previous comment to see how the return value is computed on wrappers.

agent/lib_aws_sdk_php.c
zsistla: With the macro added, nrstrcmp is no longer needed.

defined strings and lengths were also removed.
See the changes:
[fa419e3](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/fa419e3526c46177e60f761426cfe74e5f799b80)
[7d9c4c0](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/7d9c4c03123dad8403b1268d85a8af004da51129)

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,405 @@
...
+  /*
+   * nr_lib_aws_sdk_php_sqs_parse_queueurl requires a modifiable string to
+   * populate message_params and cloud_attrs.
+   */
+  nr_lib_aws_sdk_php_sqs_parse_queueurl(command_arg_value, &message_params,
+                                        &cloud_attrs);
+
+  /* Add cloud attributes, if available. */
+
+  nr_segment_traces_add_cloud_attributes(segment, &cloud_attrs);
```
lavarou, line 139: Just for posterity - no `char *` pointers in `could_attrs` need to be freed because they're either pointers into `command_arg_value` which is freed before returning from this function or it is a pointer to `command_name_string` which is also freed.

agent/lib_aws_sdk_php.c
zsistla: This is what is needed to "exit" early from a wrapper: [1523be8](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/1523be8657cb7c18c98f51cad3e65250005a2063)

agent/lib_aws_sdk_php.c
zsistla: Cannot do a simple `return;` inside a wrapper.
See this [comment](https://github.com/newrelic/newrelic-php-agent/pull/1003#discussion_r1915836283) for the link to understand what logic goes into the return value.

`goto end;` is the way to exit early from a wrapper.
[1523be8](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/1523be8657cb7c18c98f51cad3e65250005a2063)

agent/php_execute.c
zsistla: Good point.
If segment is already named (rare case), use the name to create the metric; otherwise, business as usual.
[024ed47](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/024ed4785088f7e6206273e2d541c2514ed4720f)

agent/lib_aws_sdk_php.c
lavarou: I see. Also in case there's no custom logic is needed (like freeing memory), instead of a simple `return`, [`NR_PHP_WRAPPER_LEAVE`](https://github.com/newrelic/newrelic-php-agent/blob/R11.5/agent/php_wrapper.h#L305C9-L305C29) can also be used, which will `goto` [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L275) and end the wrapper correctly.

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,413 @@
...
+void nr_lib_aws_sdk_php_sqs_parse_queueurl(
+    char* sqs_queueurl,
+    nr_segment_message_params_t* message_params,
+    nr_segment_cloud_attrs_t* cloud_attrs) {
+  char* region = NULL;
+  char* queue_name = NULL;
+  char* account_id = NULL;
+  char* queueurl_pointer = NULL;
+
+  if (NULL == sqs_queueurl || NULL == message_params) {
```
lavarou, line 156:
```suggestion
  if (NULL == sqs_queueurl || NULL == message_params || NULL == cloud_attrs) {
```

lavarou (approved):

mfulb (approved):
Nice work Amber!

axiom/nr_segment_message.c
```diff
@@ -10,6 +10,7 @@
 #include "nr_header.h"
 #include "nr_segment_message.h"
 #include "nr_segment_private.h"
+#include "util_logging.h"
```
lavarou, line 13: Is this include needed? None of the code changes use `nrl_*` functions:
```suggestion
```

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,413 @@
...
+    message_params.message_action = NR_SPANKIND_PRODUCER;
+  } else if (AWS_COMMAND_IS("sendMessage")) {
+    message_params.message_action = NR_SPANKIND_PRODUCER;
+  } else if (AWS_COMMAND_IS("receiveMessage")) {
+    message_params.message_action = NR_SPANKIND_CONSUMER;
+  } else {
+    /* Nothing to do here so exit. */
+    return;
+  }
+#undef IS_AWS_COMMAND
```
lavarou, line 123:
```suggestion
#undef AWS_COMMAND_IS
```

agent/lib_aws_sdk_php.c
```diff
@@ -14,10 +14,413 @@
...
+    goto end;
+  }
+  command_name_string = Z_STRVAL_P(command_name);
+  klass_len = nr_php_class_entry_name_length(class_entry);
+
+#define AWS_CLASS_IS(KLASS, SHORT_KLASS) \
+  (klass_len == (sizeof(KLASS) - 1)      \
+   && nr_striendswith(klass, klass_len, SHORT_KLASS, sizeof(SHORT_KLASS) - 1))
+
+  if AWS_CLASS_IS ("Aws\\Sqs\\SqsClient", "SqsClient") {
```
lavarou, line 383:
```suggestion
  if (AWS_CLASS_IS("Aws\\Sqs\\SqsClient", "SqsClient")) {
```

agent/lib_aws_sdk_php.h
```diff
@@ -7,6 +7,21 @@
 #ifndef LIB_AWS_SDK_PHP_HDR
 #define LIB_AWS_SDK_PHP_HDR
 
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO /* PHP8.1+ */
+/* Service instrumentation only supported above PHP 8.1+*/
+
+/* SQS */
+#define SQS_LIBRARY_NAME "SQS"
```
lavarou, line 14: This is not used anywhere.

agent/lib_aws_sdk_php.h
```diff
@@ -7,6 +7,21 @@
...
+#define AWS_SQS_MESSAGING_SERVICE "aws_sqs"
```
lavarou, line 15: This is not used anywhere.

lavarou: Could be used [here](https://github.com/newrelic/newrelic-php-agent/blob/feat/sqs_instrumentation/agent/lib_aws_sdk_php.c#L95) and [here](https://github.com/newrelic/newrelic-php-agent/blob/feat/sqs_instrumentation/agent/lib_aws_sdk_php.c#L97).

zsistla: Thanks for catching, that was the intention.
Please see: [8a45142](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/8a451426aa9dc8ada6afa1ece55afee6b7d6f564)

agent/lib_aws_sdk_php.h
```diff
@@ -7,6 +7,21 @@
...
+#define AWS_SDK_PHP_SQSCLIENT_QUEUEURL_ARG "QueueUrl"
+#define AWS_QUEUEURL_LEN_MAX 512
```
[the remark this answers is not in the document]
zsistla: Removed the artifact of a refactor.
[62e66da](https://github.com/newrelic/newrelic-php-agent/pull/1003/commits/62e66da1d9c0bbbbc12677c18c2d5372c800f83e)

lavarou (approved):

mfulb (approved):

PR #1009, opened 2025-01-21
feat(agent): Add RabbitMQ instrumentation using the php-amqplib library

zsistla:
Most of the PR is basic instrumentation, retrieval of values and setting attributes or creating metrics and is very similar to patterns we've established in other instrumentation.
The DT header insertion logic is the trickiest bit since we are modifying the headers in flight. (look to drupal and laravel for similar logic).
Note: the DT header insertion logic, while trickiest, is something that can be turned off anytime by the user by setting  newrelic.distributed_tracing_exclude_newrelic_header to true.

Initial commit does the following:
* Detect library via magic file
* Detect package and version information.
* Basic unit tests

Subsequent commits:
* Add attributes needed for rabbitMQ to message segment
* Instrument basic_publish and basic_get
* add unit tests, multiverse tests
* added support for PHP 7.x
* DT header insertion/retrieval

Remaining:
* Add more multiverse tests, especially around the DT functionality


agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,82 @@
...
+      "     try {"
+      "          $nr_php_amqplib_version = PhpAmqpLib\\Package::VERSION;"
+      "     } catch (Throwable $e) {"
+      "     }"
+      "     return $nr_php_amqplib_version;"
+      "})();",
+      &retval, "Get nr_php_amqplib_version");
+
+  /* See if we got a non-empty/non-null string for version. */
+  if (SUCCESS == result) {
```
mfulb, line 45: What is the condition to detect when the `catch` was triggered in the lambda function?  What will be the value of `result` and of `retval` in that situation?

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,82 @@
...
+  if (NRINI(vulnerability_management_package_detection_enabled)) {
+    nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME,
+                           PHP_PACKAGE_VERSION_UNKNOWN);
+  }
+
+  /* Extract the version for aws-sdk 3+ */
+  nr_php_amqplib_handle_version();
+
+  /* Called when initializing all Clients */
+  // nr_php_wrap_user_function(NR_PSTR("Aws\\AwsClient::parseClass"),
```
mfulb, line 80: Is this left here intentionally?

agent/lib_php_amqplib.c
zsistla: result will be success.

agent/lib_php_amqplib.c
zsistla: AWS references removed
[feba725](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/feba725d6e9ba383d7a0d3f9352abe9fed7a6b50)

agent/tests/test_lib_php_amqplib.c
```diff
@@ -0,0 +1,110 @@
...
+      p);
+  tlib_pass_if_str_equal(
+      "nr_lib_php_amqplib_handle_version when PhpAmqpLib\\Package class is not "
+      "defined - "
+      "suggested version set to PHP_PACKAGE_VERSION_UNKNOWN",
+      PHP_PACKAGE_VERSION_UNKNOWN, p->package_version);
+
+  tlib_php_request_end();
+}
+
```
mfulb, line 102: Can you add a a case where the PhpAmqpLib/Package class exists BUT the class constant is not defined?  I'm concerned about future releases or a damaged installation where this constant isn't present.  The should exercise the `try` in the detection lambda function.

zsistla: Good idea [761de36](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/761de36198b4abbd2d6cfa7e61155fa000638a74)

agent/lib_php_amqplib.c
mfulb: Could we change it so we return empty string explicitly so it is not relying on a current behavior of PHP and so it is more clear how this code interacts with the calling C code?

mfulb: I understand the code now - the initialized value (empty string) is used as the `catch` doesn't actually need to modify its value.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+ */
+void nr_php_amqplib_handle_version() {
+  char* version = NULL;
+  zval retval;
+  int result = FAILURE;
+
+  result = zend_eval_string(
+      "(function() {"
+      "     $nr_php_amqplib_version = '';"
+      "     try {"
```
mfulb, line 84: I believe PHP provides tools to make sure this constant exists so it would avoid triggering an exception in the first place.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+          = nr_php_zend_hash_index_find(Z_ARRVAL_P(connect_constructor_params),
+                                        AMQP_CONSTRUCT_PARAMS_SERVER_INDEX);
+      if (nr_php_is_zval_non_empty_string(amqp_server)) {
+        message_params->server_address
+            = ENSURE_PERSISTENCE(Z_STRVAL_P(amqp_server));
+      }
+      amqp_port
+          = nr_php_zend_hash_index_find(Z_ARRVAL_P(connect_constructor_params),
+                                        AMQP_CONSTRUCT_PARAMS_PORT_INDEX);
+      if (nr_php_is_zval_valid_scalar(amqp_port)) {
```
mfulb, line 135: This checks if it is a bool, a valid string, long or double.  Are these all valid values for the port?

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+        message_params->server_port = Z_LVAL_P(amqp_port);
```
mfulb, line 136: The check above does not guarantee the value is a long.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+
+  amqp_exchange = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
+  if (nr_php_is_zval_non_empty_string(amqp_exchange)) {
+    /*
+     * In PHP 7.x, the following will create a strdup in
+     * message_params.destination_name that needs to be freed.
+     */
+    message_params.destination_name
+        = ENSURE_PERSISTENCE(Z_STRVAL_P(amqp_exchange));
+  } else {
```
mfulb, line 180: What if `ampq_exchange` is not a string at all?

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+#else
+  /*
+   * Because we had to strdup values to persist them beyond NR_PHP_WRAPPER_CALL,
+   * now we destroy them. There isn't a separate function to destroy all since
+   * some of the params are string literals and we don't want to strdup
+   * everything if we don't have to. RabbitMQ basic_publish PHP 7.x will only
+   * strdup server_address, destination_name, and
+   * messaging_destination_routing_key.
+   */
+  nr_free(message_params.server_address);
```
mfulb, line 246: I think it could be cleaner to have a matching macro to `ENSURE_PERSISTENCE` that could be used to un-persist these values in the appropriate way for the PHP is use.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+    /*
+     *  Get the exchange and routing key from the AMQPMessage
+     */
+    amqp_exchange = nr_php_get_zval_object_property(*retval_ptr, "exchange");
+    if (nr_php_is_zval_non_empty_string(amqp_exchange)) {
+      /* Used with consumer only; his is exchange name.  Exchange name is
+       * Default in case of empty string. */
+      message_params.messaging_destination_publish_name
+          = Z_STRVAL_P(amqp_exchange);
+    } else {
```
mfulb, line 321: What if `ampq_exchange` is not a string? (This applies to this pattern in any wrappers in this PR).

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,401 @@
...
+   * nr_php_amqplib_handle_version.
+   */
+  if (NRINI(vulnerability_management_package_detection_enabled)) {
+    nr_txn_add_php_package(NRPRG(txn), PHP_PACKAGE_NAME,
+                           PHP_PACKAGE_VERSION_UNKNOWN);
+  }
+
+  /* Extract the version for aws-sdk 3+ */
+  nr_php_amqplib_handle_version();
+  nr_php_amqplib_ensure_class();
```
mfulb, line 382: I am concerned on how this all works - if the file used to trigger this enable function is, for example, in a sequence of PHP calls that was in response to autoloading the class we need, and inside this sequence we start another sequence to load this class - is composer setup to handle this?

Is there not a file we can use in the situations this class is not triggered (the AWS context I guess?).

agent/lib_php_amqplib.c
zsistla: Good point, only allow long:
[095f82d](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/095f82dd24439b86044da7e882094468003b91df)

agent/lib_php_amqplib.c
zsistla: verify long: [095f82d](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/095f82dd24439b86044da7e882094468003b91df)

agent/lib_php_amqplib.c
zsistla: That would be doing an PHP extra call when this call already verifies and retrieves the value and is future proofed against any future exceptions.  Additionally, it will automatically load the class if it's not already loaded.

agent/lib_php_amqplib.c
zsistla: The [parameter](https://github.com/newrelic/newrelic-php-agent/pull/1009/files#diff-c7ef4a9c5f82fb59784f43ed3beebc548f983ae03675acbfa5b6b602e6d98e8dR164) list for the function call dictates that exchange is a string.

Additionally, the check [here](https://github.com/newrelic/newrelic-php-agent/pull/1009/files#diff-c7ef4a9c5f82fb59784f43ed3beebc548f983ae03675acbfa5b6b602e6d98e8dR190) guarantees it is a valid non-empty string before we set it (and strdup for php 7.x) here.

agent/lib_php_amqplib.c
zsistla: Great idea!
[71e5e02](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/71e5e02e7565d93c37274ae1a45b7e77d0f6779a)

agent/lib_php_amqplib.c
zsistla: The function param list and/or class definition dictates strings or not.  
nr_php_is_zval_non_empty_string verifies that the value is not null, is not empty, and is a zval string.  Only after that do we do anything with it.

agent/lib_php_amqplib.c
zsistla: No, none of the autoload files can guarantee this class being loaded in time for the AWS setup which is either do to the slight variance in an SSL connection or to the older rabbitmq engine that is running.  None of the files that loaded when running AWS tests worked to allow our instrumentation to be wrapped (different than the non-ssl, non-aws rabbitmq server which actually loaded the file as well as the class).

All this is doing is calling `class_exists` so pretty innocuous.  This is a php function which other functions can call in the middle of their and have easily handled in the middle of other function calls as well.  It's analogous to function calls that use other classes.  The class doesn't get loaded until it's used.  Same thing here.

agent/lib_php_amqplib.c
mfulb: Under the hood this is using `zend_read_property()` which can accept many data types.

agent/lib_php_amqplib.c
mfulb: Is it possible for this value to be a string?   Does the AWS SDK define the allowable types?

agent/lib_php_amqplib.c
zsistla: The check nr_php_is_zval_non_empty_string verifies the property is a string before we do anything with it.  If it's not a string, we ignore it.

agent/lib_php_amqplib.c
zsistla: php-amqplib only allows integers > 0.

agent/lib_php_amqplib.c
zsistla: Only create default if it is an empty string:
[ead28c2](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/ead28c21a438e64e0da8dc8dc53a1926246e73df)

agent/lib_php_amqplib.c
zsistla: Only set default if it is a valid but empty string
[ead28c2](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/ead28c21a438e64e0da8dc8dc53a1926246e73df)

agent/lib_php_amqplib.c
zsistla: The Drupal case is a lot more complicated. 
This case is only referencing a const in a class.
There are only a limited amount of errors that could occur:
1) the class doesn't exist
2)the const doesn't exist.

There are unit tests that verify both of these cases are handled.  
The defined() is unneeded in this case.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,760 @@
...
+ *
+ * Params  : 1. PhpAmqpLib\Connection family of connections that inherit from
+ * AbstractConnection
+ *           2. nr_segment_message_params_t* message_params that will be
+ * modified with port and host info, if available
+ *
+ * Returns : None
+ */
+static void nr_php_amqplib_ensure_class() {
+  zval retval_dtor;
```
lavarou, line 84: Why the `_dtor` suffix in the variable name? `_dtor` suggests a destructor function and this variable is supposed to store a return value of [`class_exist`](https://www.php.net/manual/en/function.class-exists.php) which is a scalar of type `bool`.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,760 @@
...
+ * Purpose : Retrieves host and port from an AMQP Connection and sets the
+ * host/port values in the message_params.
+ *
+ * Params  : 1. PhpAmqpLib\Connection family of connections that inherit from
+ * AbstractConnection
+ *           2. nr_segment_message_params_t* message_params that will be
+ * modified with port and host info, if available
+ *
+ * Returns : None
+ */
```
lavarou, line 82: This doc string does not apply to `nr_php_amqplib_ensure_class` defined right below it.

zsistla: Good catch copy/paste error.
[0009a56](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/0009a56d2412b3060104d30d04603aa28e005b9d)

agent/lib_php_amqplib.c
zsistla: It was to remind which variables needed to be freed and how.
There are two types that need freed, I suffixed them with _zvf (for zend val free with nr_php_zval_free) and _dtor (for free with zval_ptr_dtor).  If dtor suffix is confusing, what about zpd suffix (for zval ptr dtor)?

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,756 @@
...
+  int result = FAILURE;
+
+  result = zend_eval_string("class_exists('PhpAmqpLib\\Channel\\AMQPChannel');",
+                            &retval_dtor, "Get nr_php_amqplib_class_exists");
+  /*
+   * We don't need to check anything else at this point. If this fails, there's
+   * nothing else we can do anyway.
+   */
+
+  zval_dtor(&retval_dtor);
```
lavarou, line 90: If return value from `class_exists('PhpAmqpLib\Channel\AMQPChannel');` is not verified, why is it passed to `zend_eval_string`? Also I really don't understand why neither `result` nor return value from `class_exists('PhpAmqpLib\Channel\AMQPChannel');` are not verified. What is the purpose of `nr_php_amqplib_ensure_class`?

zsistla: It loads the class if it is loadable.  The call itself is sufficient to load the class.  We could check but there's nothing we would do differently either way.

This [big block](https://github.com/newrelic/newrelic-php-agent/pull/1009/files#diff-c7ef4a9c5f82fb59784f43ed3beebc548f983ae03675acbfa5b6b602e6d98e8dR57-R68) of text explains the purpose.

agent/lib_php_amqplib.c
zsistla: I understand the concern about the dtor suffix.

Yes, there is no convention, but the function that uses that style had a mix of variables some of which did not need to be freed, some of which needed nr_php_zval_free and some of which needed zval_dtor. Now it's very clear when looking at a variable like `retval_set_table_zvf` that it needs to be freed (or the calling function needs to be aware they are freeing it) and how and if it should be freed.  It's easy to check that everything that needed to be freed was accounted for.

It adds clarity to a function that has variables like:
```
  zval* amqp_properties_array = NULL;
  zval* dt_headers_zvf = NULL;
  zval* amqp_headers_table = NULL;
  zval* retval_set_property_zvf = NULL;
  zval* retval_set_table_zvf = NULL;
  zval application_headers_zpd;
  zval key_zval_zpd;
  zval amqp_table_retval_zpd;
  zval* key_exists = NULL;
  zval* amqp_table_data = NULL;
  zend_ulong key_num = 0;
  nr_php_string_hash_key_t* key_str = NULL;
  zval* val = NULL;
  int retval = FAILURE;
```
to see at a glance what will need to be freed and how during the course of the function.

see [88c904d](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/88c904dc4a857774c78c998af3b579f3ef3652a7) for the name change.

agent/lib_php_amqplib.c
lavarou: Why not use `nr_php_get_class_constant` like is done for Drupal [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.5/agent/fw_drupal8.c#L736)?

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,758 @@
...
+  /*
+   * We don't need to check anything else at this point. If this fails, there's
+   * nothing else we can do anyway.
+   */
+
+  zval_dtor(&retval_zpd);
+}
+
+/*
+ * Version detection will be called directly from PhpAmqpLib\\Package::VERSION
```
lavarou, line 96: What does it mean to be called from a class constant? `PhpAmqpLib\\Package::VERSION` is a class constant: https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Package.php#L8 so how can it call anything.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,758 @@
...
+ * nr_php_amqplib_handle_version will automatically load the class if it isn't
+ * loaded yet and then evaluate the string. To avoid the VERY unlikely but not
```
lavarou, line 98: Can you explain how this lambda construct automatically loads the class if it doesn't exist?

agent/lib_php_amqplib.c
zsistla: [4aa2522](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/4aa25223a683824afbcfabe17905fd6bdd24314d)

agent/lib_php_amqplib.c
zsistla: The lambda will not load the class if it doesn't exist, it will only load it if it is loadable.
clarified comment: [b5c7def](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/b5c7def9055f80ca86695ac6f3fb546d22bec5ba)

agent/lib_php_amqplib.c
zsistla: A couple of reasons.

1) That would be [two calls](https://github.com/newrelic/newrelic-php-agent/blob/R11.5/agent/fw_drupal8.c#L730-L741) vs one that is occuring now.
2) if the class is not loaded, the [nr_php_find_class](https://github.com/newrelic/newrelic-php-agent/blob/R11.5/agent/fw_drupal8.c#L730) call will fail since the class isn't loaded.  The eval_string forces it to load.

agent/lib_php_amqplib.c
lavarou: [This part](https://github.com/newrelic/newrelic-php-agent/blob/feat/rabbitmq_instrumentation/agent/lib_php_amqplib.c#L59-L63):

> [...] there are AWS MQ_BROKER specific but valid scenarios where the PhpAmqpLib\\Channel\\AMQPChannel class file does not explicitly load or does not load in time, and the instrumented functions are NEVER wrapped regardless of how many times they are called in one txn [...]

is very vague. Could you give an example?

[This part](https://github.com/newrelic/newrelic-php-agent/blob/feat/rabbitmq_instrumentation/agent/lib_php_amqplib.c#L63-L67) is also not very clear to me:

> [...] Specifically, this centered around the very slight but impactful differences when using managing the AWS MQ Broker connect vs causes an explicit load of the AMQPChannel class/file and PhpAmqpLib\Connection\AMQPSSLConnection which does NOT cause an explicit load of the AMQPChannelclass/file [...]

It feels like there some extra/missing verbs?

agent/php_execute.c
```diff
@@ -491,6 +491,10 @@ static nr_library_table_t libraries[] = {
 
     {"MongoDB", NR_PSTR("mongodb/src/client.php"), nr_mongodb_enable},
 
+    /* php-amqplib RabbitMQ >= 3.7 */
```
lavarou, line 494: I don't think this comment is accurate - `phpamqplib/connection/abstractconnection.php` file also exists in version 2.x (starting with version [v2.0.2](https://github.com/php-amqplib/php-amqplib/blob/v2.0.2/PhpAmqpLib/Connection/AbstractConnection.php)) of php-amqplib.

agent/php_execute.c
```diff
@@ -491,6 +491,10 @@ static nr_library_table_t libraries[] = {
...
+    {"php-amqplib", NR_PSTR("phpamqplib/connection/abstractconnection.php"),
+     nr_php_amqplib_enable},
```
lavarou, line 496: Since `phpamqplib/connection/abstractconnection.php` file also exists in version 2.x  of php-amqplib (starting with version [v2.0.2](https://github.com/php-amqplib/php-amqplib/blob/v2.0.2/PhpAmqpLib/Connection/AbstractConnection.php)), this code will enable agent's instrumentation of php-amqplib in projects that use version 2 of php-amqplib. Is this safe?

agent/lib_php_amqplib.c
zsistla: Comparing the two methods like that is a bit simplistic because it fails to account for different overheads involved with Drupal's version detection.

Drupal only works because it wraps a file here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L549
To call into the detect version function.

Which means drupal version is only detected when vulnerability_management_package_detection_enabled is enabled and it also means it adds additional overhead of one more file lookup every time any file is loaded 
https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L972

Whereas the current rabbitmq version detection only incurs overhead once on package detection and not every single time a file is loaded for ANY framework.

Furthermore, rabbitmq doesn't have a file that is loaded that it can hinge on.
These are the loaded files when connecting to a regular rabbitmq server:

```
2025-02-05 15:06:30.271 +0000 (4496 4496) verbosedebug: execute:            file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/Constants091.php'}
2025-02-05 15:06:30.272 +0000 (4496 4496) verbosedebug: execute:            file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AMQPStreamConnection.php'}
2025-02-05 15:06:30.274 +0000 (4496 4496) verbosedebug: execute:                  file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php'}
2025-02-05 15:06:30.276 +0000 (4496 4496) verbosedebug: execute:                          file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php'}
2025-02-05 15:06:30.277 +0000 (4496 4496) verbosedebug: execute:            file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/Constants080.php'}
2025-02-05 15:06:30.278 +0000 (4496 4496) verbosedebug: execute:              file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php'}
2025-02-05 15:06:30.279 +0000 (4496 4496) verbosedebug: execute:                file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php'}
2025-02-05 15:06:30.280 +0000 (4496 4496) verbosedebug: execute:                  file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPAbstractCollection.php'}
2025-02-05 15:06:30.284 +0000 (4496 4496) verbosedebug: execute:                    file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPBufferReader.php'}
2025-02-05 15:06:30.285 +0000 (4496 4496) verbosedebug: execute:                          file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPReader.php'}
2025-02-05 15:06:30.287 +0000 (4496 4496) verbosedebug: execute:                  file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPIOReader.php'}
2025-02-05 15:06:30.306 +0000 (4496 4496) verbosedebug: execute:                file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPTable.php'}
```

PhpAmqpLib\\Package.php is never explicitly loaded as a file, so we couldn't even link to it if we wanted.  We are forced to load the class as currently implemented.

agent/lib_php_amqplib.c
zsistla: Basically, the way AWS support rabbitmq in their MQ_BROKER differs slightly from official rabbitmq servers.

So this added function is needed only to support AWS's MQ_BROKER.

when connecting via SSL with rabbitmq's official server, these are the php-amqplib files that show up:
```
2025-02-05 15:28:29.672 +0000 (4543 4543) verbosedebug: execute:          file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/Constants091.php'}
2025-02-05 15:28:29.673 +0000 (4543 4543) verbosedebug: execute:            file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AMQPStreamConnection.php'}
2025-02-05 15:28:29.675 +0000 (4543 4543) verbosedebug: execute:                  file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php'}
2025-02-05 15:28:29.680 +0000 (4543 4543) verbosedebug: execute:                          file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php'}
2025-02-05 15:28:29.680 +0000 (4543 4543) verbosedebug: execute:            file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/Constants080.php'}
2025-02-05 15:28:29.682 +0000 (4543 4543) verbosedebug: execute:              file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php'}
2025-02-05 15:28:29.684 +0000 (4543 4543) verbosedebug: execute:                file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php'}
2025-02-05 15:28:29.686 +0000 (4543 4543) verbosedebug: execute:                  file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPAbstractCollection.php'}
2025-02-05 15:28:29.703 +0000 (4543 4543) verbosedebug: execute:                    file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPBufferReader.php'}
2025-02-05 15:28:29.704 +0000 (4543 4543) verbosedebug: execute:                          file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPReader.php'}
2025-02-05 15:28:29.706 +0000 (4543 4543) verbosedebug: execute:                  file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPIOReader.php'}
2025-02-05 15:28:29.731 +0000 (4543 4543) verbosedebug: execute:                file={'/multiverse/tests/php-amqplib/3.7.x/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPTable.php'}
```

When using THE SAME file, with only changes in the server name for the connection, these are the files that show up:
```
2025-01-17 18:44:08.370 +0000 (2411 2411) verbosedebug: execute:          file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/Constants091.php'}
2025-01-17 18:44:08.371 +0000 (2411 2411) verbosedebug: execute:            file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AMQPStreamConnection.php'}
2025-01-17 18:44:08.371 +0000 (2411 2411) verbosedebug: execute:                  file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php'}
2025-01-17 18:44:08.373 +0000 (2411 2411) verbosedebug: execute:              file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php'}
2025-01-17 18:44:08.373 +0000 (2411 2411) verbosedebug: execute:                file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php'}
2025-01-17 18:44:08.375 +0000 (2411 2411) verbosedebug: execute:                  file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPAbstractCollection.php'}
2025-01-17 18:44:08.546 +0000 (2411 2411) verbosedebug: execute:                    file={'/home/ubuntu/phptest/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPBufferReader.php'}
202
```

You can see the missing, but very key `PhpAmqpLib/Channel/AMQPChannel.php` file never gets explicitly loaded in the AWS version of events.  The class is not automatically loaded, but is available and can be resolved if called from within PHP.  Because of this the functions we instrument NEVER get wrapped when connecting to the MQ_BROKER and therefore our instrumentation is never triggered.  The explicit loading of the class is therefore needed to work with MQ_BROKER.

zsistla: Updated comment to clarify: [0e6ec90](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/0e6ec907de79c045fd659b95f846bdd9da26fd8b)

agent/php_execute.c
zsistla: That's a good point.  The should say something like `PHP Agent Supports RabbitMQ >= 3.7`

They aren't maintaining 2.x and the highest version works with PHP < 8.

Preliminary testing shows it basically works ok and creates the message segments.

I updated the comment here: [127e62a](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/127e62afa3d31e4861d051567c0746357c21f5cd).

agent/php_execute.c
zsistla: Preliminary testing shows it basically works fine and creates the message segments, but we aren't fully testing it.

Would your vote be to allow it or to force a check and not instrument if it's a version below 3.7.

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,770 @@
...
+     * Must be checked after PHP_WRAPPER_CALL to ensure txn didn't end during
+     * the call.
+     */
+    goto end;
+  }
+  /*
+   *The retval should be an AMQPMessage. nr_php_is_zval_* ops do NULL checks
+   * as well.
+   */
+  if (nr_php_is_zval_valid_object(*retval_ptr)) {
```
lavarou, line 678: Check if `retval_ptr` is not `NULL` before dereferencing it:
```suggestion
  if (NULL != retval_ptr && nr_php_is_zval_valid_object(*retval_ptr)) {
```

lavarou: Since `nr_rabbitmq_basic_get` is used as both `special_instrumentation` and `special_instrumentation_clean` callbacks, `retval_ptr` will be NULL in the latter case (which is when `PhpAmqpLib\Channel\AMQPChannel::basic_get` throws an exception).

lavarou: `NULL` checking in `nr_php_is_zval_valid_object` is done after `retval_ptr` is dereferenced (`*retval_ptr`) which will yield undefined behavior (`SIGSEGV`) if `retval_ptr`'s value is `NULL`.

agent/php_execute.c
lavarou: Are the signatures of `PhpAmqpLib\Channel\AMQPChannel::basic_publish` and `PhpAmqpLib\Channel\AMQPChannel::basic_get` methods the same between 2.x and 3.x?

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,770 @@
...
+      = nr_php_zend_hash_index_find(Z_ARRVAL_P(connect_constructor_params),
+                                    AMQP_CONSTRUCT_PARAMS_SERVER_INDEX);
+  if (nr_php_is_zval_non_empty_string(amqp_server)) {
+    message_params->server_address
+        = ENSURE_PERSISTENCE(Z_STRVAL_P(amqp_server));
+  }
+
+  amqp_port = nr_php_zend_hash_index_find(
+      Z_ARRVAL_P(connect_constructor_params), AMQP_CONSTRUCT_PARAMS_PORT_INDEX);
+  if (IS_LONG != Z_TYPE_P((amqp_port))) {
```
lavarou, line 197: `nr_php_is_zval_valid_integer` is what is needed here (it also checks for NULL):
```suggestion
  if (nr_php_is_zval_valid_integer(amqp_port)) {
```

agent/lib_php_amqplib.c
zsistla: But also correct, the retval is not needed since we aren't using it.
[23a8a34](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/23a8a34bc492bae2e63976440addb80ac7318286)

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,766 @@
...
+    }
+
+    amqp_routing_key
+        = nr_php_get_zval_object_property(*retval_ptr, "routingKey");
+    if (nr_php_is_zval_non_empty_string(amqp_routing_key)) {
+      message_params.messaging_destination_routing_key
+          = Z_STRVAL_P(amqp_routing_key);
+    }
+
+    nr_php_amqplib_retrieve_dt_headers(*retval_ptr);
```
lavarou, line 701: This function has a side effect - it sets DT headers in the current transaction (it eventually calls `nr_txn_accept_distributed_trace_payload`). This is not very obvious when reading the name of this function, so leaving this comment for future reviewers and maintainers.

agent/php_execute.c
zsistla: Yes

agent/lib_php_amqplib.c
zsistla: It said outdated suggestions cannot be applied.
suggested applied here:
[90da3b7](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/90da3b7c75ffb66a1d1548f89231d44f36edf022)

ZNeumann (approved):

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,807 @@
...
+
+  amqp_port = nr_php_zend_hash_index_find(
+      Z_ARRVAL_P(connect_constructor_params), AMQP_CONSTRUCT_PARAMS_PORT_INDEX);
+  if (nr_php_is_zval_valid_integer(amqp_port)) {
+    message_params->server_port = Z_LVAL_P(amqp_port);
+  }
+}
+
+/*
+ * Purpose : Applies DT headers to an inbound AMQPMessage.
```
lavarou, line 199: Isn't publishing (producing) an outbound action? The headers are added to the outgoing message that is pushed to the queue:
```suggestion
 * Purpose : Applies DT headers to an outbound AMQPMessage.
```

agent/lib_php_amqplib.c
lavarou: I'm ok with leaving this piece of code as is except for maybe changing the default return value of the lambda function to either `null` or `' '` (space). Having `''` (empty string) causes https://github.com/newrelic/newrelic-php-agent/blob/feat/rabbitmq_instrumentation/agent/lib_php_amqplib.c#L131 not to execute and remain `null`.

agent/php_execute.c
lavarou: If the method signatures are the same, then I would leave things as they are.

agent/lib_php_amqplib.c
zsistla: Good point.
I changed it to return null (which is a very clear differentiator between any string they might give and what response we might give) and added additional check for the null before calling the other functions.
[87f641e](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/87f641ecb0b768d06a513090ecb76a3bf3b57cc3)

agent/lib_php_amqplib.c
mfulb: Consider using zend_eval_stringl(NR_PSTR(<str)...)

agent/lib_php_amqplib.c
mfulb: This looks good - one final request - change the final arg to `zend_eval_string` to a function name descriptive of what it is doing.

agent/lib_php_amqplib.c
zsistla: zend_eval_stringl here: [55383f0](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/55383f0a2136c5ac33d3ca68a006654961d3732a)

zsistla: use nr_php_find_class first here: [73b47b2](https://github.com/newrelic/newrelic-php-agent/pull/1009/commits/73b47b2b29c14f239c855b25148549dbee0b7511)

agent/lib_php_amqplib.c
```diff
@@ -0,0 +1,833 @@
...
+        "nr_php_amqplib_class_exists_channel_amqpchannel");
+  }
+  /*
+   * We don't need to check anything else at this point. If this fails, there's
+   * nothing else we can do anyway.
+   */
+}
+
+/*
+ * Version detection will be called pulled from PhpAmqpLib\\Package::VERSION
```
lavarou, line 108:
```suggestion
 * Version information will be pulled from PhpAmqpLib\\Package::VERSION
```

lavarou (approved):

mfulb (dismissed):

lavarou (approved):

mfulb (approved):

PR #1010, opened 2025-01-21
fix(agent): properly name drupal cached pages

bduranleau-nr:
Previously, Drupal cached pages were aggregated under the `page_cache` name and metric. This PR injects logic that allows the agent to retrieve the `_controller` information on a request to meaningfully name a txn involving a cached page.

Adds new INI setting `newrelic.framework.drupal.page_cache_naming.enabled` to toggle behavior on/off.


agent/fw_drupal8.c
```diff
@@ -66,8 +66,7 @@ NR_PHP_WRAPPER(nr_drupal_exception) {
 
   if (NR_SUCCESS
       != nr_php_error_record_exception(NRPRG(txn), exception, priority, true,
-                                       NULL,
-                                       &NRPRG(exception_filters))) {
+                                       NULL, &NRPRG(exception_filters))) {
```
bduranleau-nr, line 69: Formatting

agent/fw_drupal8.c
```diff
@@ -134,12 +133,12 @@ static void nr_drupal8_add_method_callback(const zend_class_entry* ce,
...
-                                           size_t method_len,
-                                           nrspecialfn_t before_callback,
-                                           nrspecialfn_t after_callback,
-                                           nrspecialfn_t clean_callback) {
+    const zend_class_entry* ce,
+    const char* method,
+    size_t method_len,
+    nrspecialfn_t before_callback,
+    nrspecialfn_t after_callback,
+    nrspecialfn_t clean_callback) {
```
bduranleau-nr, line 141: Formatting

agent/fw_drupal8.c
```diff
@@ -164,13 +163,13 @@ static void nr_drupal8_add_method_callback_before_after_clean(
...
-                              class_method, nr_strlen(class_method),
-                              before_callback, after_callback, clean_callback);
+        class_method, nr_strlen(class_method), before_callback, after_callback,
+        clean_callback);
 
     nr_free(class_method);
   }
 }
-#endif // OAPI
+#endif  // OAPI
```
bduranleau-nr, line 172: Formatting

agent/fw_drupal8.c
```diff
@@ -501,8 +544,7 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_callback) {
 
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
-  zval* curr_hook
-      = (zval*)nr_stack_get_top(&NRPRG(drupal_invoke_all_hooks));
+  zval* curr_hook = (zval*)nr_stack_get_top(&NRPRG(drupal_invoke_all_hooks));
```
bduranleau-nr, line 547: Formatting

agent/fw_drupal8.c
```diff
@@ -515,7 +557,7 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_callback) {
   nr_drupal_hook_instrument(Z_STRVAL_P(module), Z_STRLEN_P(module),
                             NRPRG(drupal_invoke_all_hook),
                             NRPRG(drupal_invoke_all_hook_len) TSRMLS_CC);
-#endif // OAPI
+#endif  // OAPI
```
bduranleau-nr, line 560: Formatting

agent/fw_drupal8.c
```diff
@@ -536,7 +578,7 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with) {
     || defined OVERWRITE_ZEND_EXECUTE_DATA
   char* prev_hook = NULL;
   int prev_hook_len;
-#endif // not OAPI
+#endif  // not OAPI
```
bduranleau-nr, line 581: Formatting

agent/fw_drupal8.c
```diff
@@ -547,7 +589,7 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with) {
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
     nr_php_arg_release(&hook);
-#endif // OAPI
+#endif  // OAPI
```
bduranleau-nr, line 592: Formatting

agent/fw_drupal8.c
```diff
@@ -561,7 +603,7 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with) {
       = nr_strndup(Z_STRVAL_P(hook), Z_STRLEN_P(hook));
   NRPRG(drupal_invoke_all_hook_len) = Z_STRLEN_P(hook);
   NRPRG(check_cufa) = true;
-#endif // OAPI
+#endif  // OAPI
```
bduranleau-nr, line 606: Formatting

agent/fw_drupal8.c
```diff
@@ -581,14 +623,14 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with) {
...
+#endif  // not OAPI
 
-leave: ;
+leave:;
 #if ZEND_MODULE_API_NO < ZEND_8_0_X_API_NO \
     || defined OVERWRITE_ZEND_EXECUTE_DATA
   /* for OAPI, the _after callback handles this free */
   nr_php_arg_release(&hook);
-#endif // not OAPI
+#endif  // not OAPI
```
bduranleau-nr, line 633: Formatting

agent/fw_drupal8.c
```diff
@@ -605,7 +647,7 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
   nr_drupal_invoke_all_hook_stacks_pop();
 }
 NR_PHP_WRAPPER_END
-#endif // OAPI
+#endif  // OAPI
```
bduranleau-nr, line 650: Formatting

agent/fw_drupal8.c
```diff
@@ -642,10 +684,8 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
   nr_drupal8_add_method_callback_before_after_clean(
-                                 ce, NR_PSTR("invokeallwith"),
-                                 nr_drupal94_invoke_all_with,
-                                 nr_drupal94_invoke_all_with_after,
-                                 nr_drupal94_invoke_all_with_clean);
+      ce, NR_PSTR("invokeallwith"), nr_drupal94_invoke_all_with,
+      nr_drupal94_invoke_all_with_after, nr_drupal94_invoke_all_with_clean);
```
bduranleau-nr, line 688: Formatting

bduranleau-nr:
Do we want to add some kind of indication to the txn name or metric that this was the result of a cached page hit? For example, prepending the controller name/route with `cached`? Current implementation has no differentiation between a request that hits a cached page vs a non-cached page in the metrics the agent will report to the collector- maybe users would find this information useful to determine how often their cache is getting hit, or if there are problems specifically with cached pages vs non-cached.

agent/fw_drupal8.c
```diff
@@ -322,12 +321,57 @@ NR_PHP_WRAPPER_END
...
+  request = nr_php_arg_get(1, NR_EXECUTE_ORIG_ARGS);
+
+  if (0 == nr_php_is_zval_valid_object(request)) {
+    nrl_verbosedebug(
+        NRL_TXN, "Drupal 8 PageCache::get does not have a request parameter");
+
+    goto end;
+  }
+
+  if (false
```
mfulb, line 345: Small nit - but both `nr_php_is_zval_valid_object` and `nr_php_object_instanceof_class` return a value of "0" (not `false`) so can both of these checks use the same value for consistency?

agent/fw_drupal8.c
```diff
@@ -322,12 +321,57 @@ NR_PHP_WRAPPER_END
...
+  if (!NRINI(drupal_page_cache_naming)) {
+    name = nr_strdup("page_cache");
+    goto end;
+  }
+
+  nrfn = nr_php_find_function("newrelic_name_cached");
+  if (NULL == nrfn) {
+    nrl_warning(NRL_FRAMEWORK,
+                "Failed to retrieve NR cached page naming function");
+    goto end;
```
mfulb, line 362: This could probably set `name` to "page_cache" since this is what we just did above when naming wasn't requested.  Maybe change the warning to include a notice that "page_cache" was being used as the name because of the failure.

agent/fw_drupal8.c
```diff
@@ -752,6 +797,66 @@ void nr_drupal_version() {
...
+    if (inject) {
+      symfony_ce
+          = nr_php_find_class("Symfony\\Component\\HttpFoundation\\Request");
+      if (NULL == symfony_ce) {
+        inject = false;
+        nrl_warning(NRL_FRAMEWORK, "Missing Symfony Request Class");
+      }
+    }
+
+    if (inject) {
```
mfulb, line 822: Why two `if (inject)`?

agent/fw_drupal8.c
```diff
@@ -752,6 +797,66 @@ void nr_drupal_version() {
...
+
+  if (NRINI(drupal_page_cache_naming)) {
+    drupal_ce = nr_php_find_class("Drupal\\Core\\Routing\\RouteMatch");
+
+    if (NULL == drupal_ce) {
+      inject = false;
+      nrl_warning(NRL_FRAMEWORK, "Missing Drupal RouteMatch Class");
+    }
+
+    if (inject) {
```
mfulb, line 813: Can this be an `else` for the conditional on line 808?

agent/fw_drupal8.c
bduranleau-nr: da14399002566dc6741a84065b8f96b407f9064b

agent/fw_drupal8.c
bduranleau-nr: Maybe a better solution to multiple `name = nr_strdup("page_cache")` checks is to just check after `NR_PHP_WRAPPER_CALL` if `name == NULL` and set it to `page_cache` that way?

agent/fw_drupal8.c
bduranleau-nr: aebfeebf39718d891d7e03abb7e8672c4ca05efd

agent/fw_drupal8.c
mfulb: My concern was that it appeared we had a case where we missed naming - if your solution handles these it should work fine.

bduranleau-nr: I think 04cbaf6d702fdb4aa2ff99624c3861a2870b0d25 is a slight improvement on the fallback naming

bduranleau-nr: On testing, that seems to have broken something... I'll have to look into this a little more

PR #1011, opened 2025-01-22
refactor(agent): reduce number of segments created

zsistla:
Instead of a blanket segment_start in the "before" function that needs to be discarded in the majority of the cases, just create the segment when it is needed, add the attributes/metrics, then close the segment.

DONE:Multiverse SQS exception tests need to be updated.
DONE:Needs additional testing against backend.

### DIFF
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -91,7 +91,7 @@ void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* auto_segment,
                                    size_t command_name_len,
                                    NR_EXECUTE_PROTO) {
   char* command_arg_value = NULL;
-  nr_segment_t* segment = NULL;
+  nr_segment_t* message_segment = NULL;
 
   nr_segment_message_params_t message_params = {
       .library = SQS_LIBRARY_NAME,
@@ -129,12 +129,12 @@ void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* auto_segment,
    * only create the segment now, grab the parent segment start time, add our
    * special segment attributes/metrics then close the newly created segment.
    */
-  segment = nr_segment_start(NRPRG(txn), auto_segment, NULL);
-  if (NULL == segment) {
+  message_segment = nr_segment_start(NRPRG(txn), auto_segment, NULL);
+  if (NULL == message_segment) {
     return;
   }
   /* re-use start time from auto_segment started in func_begin */
-  segment->start_time = auto_segment->start_time;
+  message_segment->start_time = auto_segment->start_time;
   cloud_attrs.aws_operation = command_name_string;
 
   command_arg_value = nr_lib_aws_sdk_php_get_command_arg_value(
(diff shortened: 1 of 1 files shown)


zsistla:
Looks good:
image

agent/lib_aws_sdk_php.c
```diff
@@ -122,6 +124,16 @@ void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
   }
 #undef AWS_COMMAND_IS
 
+  /*
+   * By this point, it's been determined that this call will be instrumented so
+   * only create the segment now, grab the parent segment start time, add our
+   * special segment attributes/metrics then close the newly created segment.
+   */
+  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
```
lavarou, line 132: This segment should be parented to `auto_segment` (segment created for `Aws\AwsClient::__call`):
```suggestion
  segment = nr_segment_start(NRPRG(txn), auto_segment, NULL);
```

zsistla: It is not necessary in this case as it is automatically parented to the current segment.
See https://github.com/newrelic/newrelic-php-agent/pull/1011#issuecomment-2610429004 for the parenting.
Multiverse tests also all showed correct parenting.
We can be explicit about the parenting, but it doesn't change functionality.

agent/lib_aws_sdk_php.c
```diff
@@ -86,19 +86,21 @@
 
 //clang-format on
 */
-void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
+void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* auto_segment,
                                    char* command_name_string,
                                    size_t command_name_len,
                                    NR_EXECUTE_PROTO) {
   char* command_arg_value = NULL;
+  nr_segment_t* segment = NULL;
```
lavarou, line 94: The code readability would increase if this local variable would be named `message_segment`.

agent/lib_aws_sdk_php.c
```diff
@@ -122,6 +124,16 @@ void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
...
 
+  /*
+   * By this point, it's been determined that this call will be instrumented so
+   * only create the segment now, grab the parent segment start time, add our
+   * special segment attributes/metrics then close the newly created segment.
+   */
+  segment = nr_segment_start(NRPRG(txn), auto_segment, NULL);
+  if (NULL == segment) {
+    return;
+  }
```
lavarou, line 135:
```suggestion
  }
  /* re-use start time from auto_segment started in func_begin */
```

lavarou (approved):

lavarou (dismissed):

lavarou (approved):

agent/lib_aws_sdk_php.c
```diff
@@ -122,6 +124,17 @@ void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
...
+   * By this point, it's been determined that this call will be instrumented so
+   * only create the segment now, grab the parent segment start time, add our
+   * special segment attributes/metrics then close the newly created segment.
+   */
+  message_segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  if (NULL == message_segment) {
+    return;
+  }
+  /* re-use start time from auto_segment started in func_begin */
+  message_segment->start_time = auto_segment->start_time;
```
lavarou, line 137: Shouldn't `auto_segment->error` be duplicated here as well?

lavarou: After further consideration, this is not needed at the moment.

mfulb (approved):

PR #1012, opened 2025-01-27
chore: remove vendor subdir

lavarou:


Makefile
```diff
@@ -441,21 +439,14 @@ lasp-test-all:
...
 #
-# Vendored libraries
+# Check for protobuf-c (HAVE_PROTOBUF_C): 
+#  - axiom build needs protoc-c (protobuf-c-compiler)
+#  - agent build needs protobuf-c static library (protobuf-c-devel)
 #
-export GIT
-
-.PHONY: vendor vendor-clean
+.PHONY: protobuf-c
```
mfulb, line 446: Why was the `export GIT` line removed?

lavarou: `export GIT` was removed because it is not used anywhere. It used to be used in `vendor/Makefile` [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/vendor/Makefile#L37) and [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/vendor/Makefile#L57) and `vendor/Makefile` is removed in this PR. However, even `vendor/Makefile` used to set the default value for `GIT` if it was not set - see [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/vendor/Makefile#L6).

zsistla (approved):

PR #1023, opened 2025-02-13
feat(agent): Add AWS Lambda Relationship

ZNeumann:
Instrument AWS Lambda `invoke` and reconstruct the ARN to facilitate a complete service map.

### DIFF
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -367,7 +367,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
     }
     if (nr_php_is_zval_valid_array(metadata)) {
       zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
-      if (nr_php_is_zval_valid_string(uri)) {
+      if (nr_php_is_zval_valid_string(uri) && !nr_strempty(Z_STRVAL_P(uri))) {
         external_params.uri = Z_STRVAL_P(uri);
       }
     }
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -392,7 +392,7 @@ void nr_aws_sdk_mshutdown(void) {
 
 void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs) {
   zval* call_args = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
-  zval* this_obj = getThis();//NR_PHP_USER_FN_THIS();
+  zval* this_obj = NR_PHP_USER_FN_THIS();
   char* arn = NULL;
   char* function_name = NULL;
   char* region = NULL;
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -429,7 +429,7 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
   region = nr_regex_substrings_get_named(matches, "region");
   qualifier = nr_regex_substrings_get_named(matches, "qualifier");
 
-  /* suppliment missing information with API calls */
+  /* supplement missing information with API calls */
   if (nr_strempty(function_name)) {
     /*
      * Cannot get the needed data. Function name is required in the
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -15,9 +15,16 @@
 #include "fw_support.h"
 #include "util_logging.h"
 #include "nr_segment_message.h"
+#include "nr_segment_external.h"
 #include "lib_aws_sdk_php.h"
 
 #define PHP_PACKAGE_NAME "aws/aws-sdk-php"
+#define AWS_ARN_REGEX "(arn:(aws[a-zA-Z-]*)?:lambda:)?" \
+            "((?<region>[a-z]{2}((-gov)|(-iso([a-z]?)))?-[a-z]+-\\d{1}):)?" \
+            "((?<accountId>\\d{12}):)?" \
+            "(function:)?" \
+            "(?<functionName>[a-zA-Z0-9-\\.]+)" \
+            "(:(?<qualifier>\\$LATEST|[a-zA-Z0-9-]+))?"
 
 #if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO /* PHP8.1+ */
 /* Service instrumentation only supported above PHP 8.1+*/
@@ -295,6 +302,167 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
   cloud_attrs->cloud_region = region;
 }
 
+void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
+                                      char* command_name_string,
+                                      size_t command_name_len,
+                                      NR_EXECUTE_PROTO) {
+  nr_segment_t* external_segment = NULL;
+  zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
+
+  nr_segment_cloud_attrs_t cloud_attrs = {
+      .cloud_platform = "aws_lambda"
+  };
+
+  if (NULL == auto_segment) {
+    return;
+  }
+
+  if (NULL == command_name_string || 0 == command_name_len) {
+    return;
+  }
+
+  if (NULL == retval_ptr) {
+    /* Do not instrument when an exception has happened */
+    return;
+  }
+
+#define AWS_COMMAND_IS(CMD) \
+  (command_name_len == (sizeof(CMD) - 1) && nr_streq(CMD, command_name_string))
+
+  /* Determine if we instrument this command. */
+  if (AWS_COMMAND_IS("invoke")) {
+    // Command can be saved here if in the future
+    // we instrument more than 1 lambda command
+  } else {
+    return;
+  }
+#undef AWS_COMMAND_IS
+
+  /* reconstruct the ARN */
+  nr_aws_lambda_invoke(NR_EXECUTE_ORIG_ARGS, &cloud_attrs);
+  if (!cloud_attrs.cloud_resource_id) {
+    /* we do not want to instrument if we cannot reconstruct the ARN */
+    return;
+  }
+
+  /*
+   * By this point, it's been determined that this call will be instrumented so
+   * only create the segment now, grab the parent segment start time, add our
+   * special segment attributes/metrics then close the newly created segment.
+   */
+  external_segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  if (NULL == external_segment) {
+    return;
+  }
+  /* re-use start time from auto_segment started in func_begin */
+  external_segment->start_time = auto_segment->start_time;
+  cloud_attrs.aws_operation = command_name_string;
+
+  /* end the segment */
+  nr_segment_traces_add_cloud_attributes(external_segment, &cloud_attrs);
+  nr_segment_external_params_t external_params = {.library = "aws_sdk"};
+  zval* data = nr_php_get_zval_object_property(*retval_ptr, "data");
+  if (NULL != data && IS_ARRAY == Z_TYPE_P(data)) {
+    zval* status_code = nr_php_zend_hash_find(Z_ARRVAL_P(data), "StatusCode");
+    if (NULL != status_code && IS_LONG == Z_TYPE_P(status_code)) {
+      external_params.status = Z_LVAL_P(status_code);
+    }
+    zval* metadata = nr_php_zend_hash_find(Z_ARRVAL_P(data), "@metadata");
+    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)
+        && IS_ARRAY == Z_TYPE_P(Z_REFVAL_P(metadata))) {
+      zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(Z_REFVAL_P(metadata)), "effectiveUri");
+      if (NULL != uri && IS_STRING == Z_TYPE_P(uri)) {
+        external_params.uri = Z_STRVAL_P(uri);
+      }
+    }
+  }
+  nr_segment_external_end(&auto_segment, &external_params);
+}
+
+void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs) {
+  zval* call_args = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
+  zval* this_obj = NR_PHP_USER_FN_THIS();
+  char* arn = NULL;
+  char* function_name = NULL;
+  char* region = NULL;
+  zval* region_zval = NULL;
+  char* qualifier = NULL;
+  zval* qualifier_zval = NULL;
+  char* accountID = NULL;
+
+  /* verify arguments */
+  if (NULL == call_args || IS_ARRAY != Z_TYPE_P(call_args)) {
+    return;
+  }
+  zval* lambda_args = nr_php_zend_hash_index_find(Z_ARRVAL_P(call_args), 0);
+  if (NULL == lambda_args || IS_ARRAY != Z_TYPE_P(lambda_args)) {
+    return;
+  }
+  zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
+  if (NULL == lambda_name || IS_STRING != Z_TYPE_P(lambda_name)) {
+    return;
+  }
+
+  /* Compile the regex */
+  if (NULL == NRPRG(aws_arn_regex)) {
+    NRPRG(aws_arn_regex) = nr_regex_create(AWS_ARN_REGEX, 0, 0);
+  }
+
+  /* Extract all information possible from the passed lambda name via regex */
+  nr_regex_substrings_t* matches =
+      nr_regex_match_capture(NRPRG(aws_arn_regex),
+                             Z_STRVAL_P(lambda_name),
+                             Z_STRLEN_P(lambda_name));
+  function_name = nr_regex_substrings_get_named(matches, "functionName");
+  accountID = nr_regex_substrings_get_named(matches, "accountId");
+  region = nr_regex_substrings_get_named(matches, "region");
+  qualifier = nr_regex_substrings_get_named(matches, "qualifier");
+
+  /* suppliment missing information with API calls */
+  if (nr_strempty(function_name)) {
+    /*
+     * Cannot get the needed data. Function name is required in the
+     * argument, so this won't happen in normal operation
+     */
+    nr_regex_substrings_destroy(&matches);
+    return;
+  }
+  if (nr_strempty(accountID)) {
+    accountID = NRINI(aws_account_id);
+    accountID = "012345";
+  }
+  if (nr_strempty(region)) {
+    region_zval = nr_php_call(this_obj, "getRegion");
+    if (nr_php_is_zval_valid_string(region_zval)) {
+      region = Z_STRVAL_P(region_zval);
+    }
+  }
+  if (nr_strempty(qualifier)) {
+    qualifier_zval = NULL;//nr_php_call(this_obj, "getQualifier");
+    if (nr_php_is_zval_valid_string(qualifier_zval)) {
+      qualifier = Z_STRVAL_P(qualifier_zval);
+    }
+  }
+
+  if (!nr_strempty(accountID) && !nr_strempty(region)) {
+    // construct the ARN
+    if (qualifier) {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s:%s",
+                       region, accountID, function_name, qualifier);
+    } else {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s",
+                       region, accountID, function_name);
+    }
+
+    // Attatch the ARN
+    cloud_attrs->cloud_resource_id = arn;
+  }
+
+  nr_regex_substrings_destroy(&matches);
+  nr_php_zval_free(&region_zval);
+  nr_php_zval_free(&qualifier_zval);
+}
+
 char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
                                                NR_EXECUTE_PROTO) {
   zval* param_array = NULL;
@@ -381,6 +549,10 @@ NR_PHP_WRAPPER(nr_aws_client_call) {
 
   if (AWS_CLASS_IS("Aws\\Sqs\\SqsClient", "SqsClient")) {
     nr_lib_aws_sdk_php_sqs_handle(auto_segment, command_name_string,
+                                  Z_STRLEN_P(command_name),
+                                  NR_EXECUTE_ORIG_ARGS);
+  } else if (AWS_CLASS_IS("Aws\\Lambda\\LambdaClient", "LambdaClient")) {
+    nr_lib_aws_sdk_php_lambda_handle(auto_segment, command_name_string,
                                   Z_STRLEN_P(command_name),
                                   NR_EXECUTE_ORIG_ARGS);
   }
@@ -566,5 +738,6 @@ void nr_aws_sdk_php_enable() {
   nr_php_wrap_user_function_before_after_clean(
       NR_PSTR("Aws\\AwsClient::__call"), NULL, nr_aws_client_call,
       nr_aws_client_call);
+
 #endif
 }
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -367,7 +367,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
     }
     if (nr_php_is_zval_valid_array(metadata)) {
       zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
-      if (nr_php_is_zval_valid_string(uri) && !nr_strempty(Z_STRVAL_P(uri))) {
+      if (nr_php_is_zval_non_empty_string(uri)) {
         external_params.uri = Z_STRVAL_P(uri);
       }
     }
@@ -410,7 +410,7 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
     return;
   }
   zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
-  if (!nr_php_is_zval_valid_string(lambda_name)) {
+  if (!nr_php_is_zval_non_empty_string(lambda_name)) {
     return;
   }
 
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -474,7 +474,7 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
                        region, accountID, function_name);
     }
 
-    /* Attatch the ARN */
+    /* Attach the ARN */
     cloud_attrs->cloud_resource_id = arn;
   }
 
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -321,7 +321,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
     return;
   }
 
-  if (NULL == retval_ptr) {
+  if (NULL == *retval_ptr) {
     /* Do not instrument when an exception has happened */
     return;
   }
@@ -362,21 +362,36 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
   nr_segment_traces_add_cloud_attributes(external_segment, &cloud_attrs);
   nr_segment_external_params_t external_params = {.library = "aws_sdk"};
   zval* data = nr_php_get_zval_object_property(*retval_ptr, "data");
-  if (NULL != data && IS_ARRAY == Z_TYPE_P(data)) {
+  if (nr_php_is_zval_valid_array(data)) {
     zval* status_code = nr_php_zend_hash_find(Z_ARRVAL_P(data), "StatusCode");
-    if (NULL != status_code && IS_LONG == Z_TYPE_P(status_code)) {
+    if (nr_php_is_zval_valid_integer(status_code)) {
       external_params.status = Z_LVAL_P(status_code);
     }
     zval* metadata = nr_php_zend_hash_find(Z_ARRVAL_P(data), "@metadata");
-    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)
-        && IS_ARRAY == Z_TYPE_P(Z_REFVAL_P(metadata))) {
-      zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(Z_REFVAL_P(metadata)), "effectiveUri");
-      if (NULL != uri && IS_STRING == Z_TYPE_P(uri)) {
-        external_params.uri = Z_STRVAL_P(uri);
+    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)) {
+      metadata = Z_REFVAL_P(metadata);
+      if (IS_ARRAY == Z_TYPE_P(metadata)) {
+        zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
+        if (nr_php_is_zval_valid_string(uri)) {
+          external_params.uri = Z_STRVAL_P(uri);
+        }
       }
     }
   }
   nr_segment_external_end(&auto_segment, &external_params);
+}
+
+/* This stores the compiled regex to parse AWS ARNs. The compilation happens when
+ * it is first needed and is destroyed in mshutdown
+ */
+static nr_regex_t* aws_arn_regex;
+
+static void nr_aws_sdk_compile_regex(void) {
+  aws_arn_regex = nr_regex_create(AWS_ARN_REGEX, 0, 0);
+}
+
+void nr_aws_sdk_mshutdown(void) {
+  nr_regex_destroy(&aws_arn_regex);
 }
 
 void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs) {
@@ -391,26 +406,26 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
   char* accountID = NULL;
 
   /* verify arguments */
-  if (NULL == call_args || IS_ARRAY != Z_TYPE_P(call_args)) {
+  if (!nr_php_is_zval_valid_array(call_args)) {
     return;
   }
   zval* lambda_args = nr_php_zend_hash_index_find(Z_ARRVAL_P(call_args), 0);
-  if (NULL == lambda_args || IS_ARRAY != Z_TYPE_P(lambda_args)) {
+  if (!nr_php_is_zval_valid_array(lambda_args)) {
     return;
   }
   zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
-  if (NULL == lambda_name || IS_STRING != Z_TYPE_P(lambda_name)) {
+  if (!nr_php_is_zval_valid_string(lambda_name)) {
     return;
   }
 
-  /* Compile the regex */
-  if (NULL == NRPRG(aws_arn_regex)) {
-    NRPRG(aws_arn_regex) = nr_regex_create(AWS_ARN_REGEX, 0, 0);
+  /* Ensure regex exists */
+  if (NULL == aws_arn_regex) {
+    nr_aws_sdk_compile_regex();
   }
 
   /* Extract all information possible from the passed lambda name via regex */
   nr_regex_substrings_t* matches =
-      nr_regex_match_capture(NRPRG(aws_arn_regex),
+      nr_regex_match_capture(aws_arn_regex,
                              Z_STRVAL_P(lambda_name),
                              Z_STRLEN_P(lambda_name));
   function_name = nr_regex_substrings_get_named(matches, "functionName");
@@ -429,10 +444,10 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
   }
   if (nr_strempty(accountID)) {
     accountID = NRINI(aws_account_id);
-    accountID = "012345";
   }
   if (nr_strempty(region)) {
-    region_zval = nr_php_call(this_obj, "getRegion");
+    region_zval
+      = nr_php_get_zval_object_property(this_obj, "region");
     if (nr_php_is_zval_valid_string(region_zval)) {
       region = Z_STRVAL_P(region_zval);
     }
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -340,10 +340,6 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
 
   /* reconstruct the ARN */
   nr_aws_lambda_invoke(NR_EXECUTE_ORIG_ARGS, &cloud_attrs);
-  if (NULL == cloud_attrs.cloud_resource_id) {
-    /* we do not want to instrument if we cannot reconstruct the ARN */
-    return;
-  }
 
   /*
    * By this point, it's been determined that this call will be instrumented so
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -404,7 +404,6 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
   char* region = NULL;
   zval* region_zval = NULL;
   char* qualifier = NULL;
-  zval* qualifier_zval = NULL;
   char* accountID = NULL;
   bool using_account_id_ini = false;
 
@@ -461,16 +460,10 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
       region = Z_STRVAL_P(region_zval);
     }
   }
-  if (nr_strempty(qualifier)) {
-    qualifier_zval = NULL;//nr_php_call(this_obj, "getQualifier");
-    if (nr_php_is_zval_valid_string(qualifier_zval)) {
-      qualifier = Z_STRVAL_P(qualifier_zval);
-    }
-  }
 
   if (!nr_strempty(accountID) && !nr_strempty(region)) {
     // construct the ARN
-    if (qualifier) {
+    if (!nr_strempty(qualifier)) {
       arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s:%s",
                        region, accountID, function_name, qualifier);
     } else {
@@ -490,7 +483,6 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
   nr_free(region);
   nr_free(qualifier);
   nr_php_zval_free(&region_zval);
-  nr_php_zval_free(&qualifier_zval);
 }
 
 char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
--- a/agent/lib_aws_sdk_php.c
+++ b/agent/lib_aws_sdk_php.c
@@ -340,7 +340,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
 
   /* reconstruct the ARN */
   nr_aws_lambda_invoke(NR_EXECUTE_ORIG_ARGS, &cloud_attrs);
-  if (!cloud_attrs.cloud_resource_id) {
+  if (NULL == cloud_attrs.cloud_resource_id) {
     /* we do not want to instrument if we cannot reconstruct the ARN */
     return;
   }
@@ -352,6 +352,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
    */
   external_segment = nr_segment_start(NRPRG(txn), NULL, NULL);
   if (NULL == external_segment) {
+    nr_free(cloud_attrs.cloud_resource_id);
     return;
   }
   /* re-use start time from auto_segment started in func_begin */
@@ -379,6 +380,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
     }
   }
   nr_segment_external_end(&auto_segment, &external_params);
+  nr_free(cloud_attrs.cloud_resource_id);
 }
 
 /* This stores the compiled regex to parse AWS ARNs. The compilation happens when
@@ -404,6 +406,7 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
   char* qualifier = NULL;
   zval* qualifier_zval = NULL;
   char* accountID = NULL;
+  bool using_account_id_ini = false;
 
   /* verify arguments */
   if (!nr_php_is_zval_valid_array(call_args)) {
@@ -439,11 +442,17 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
      * Cannot get the needed data. Function name is required in the
      * argument, so this won't happen in normal operation
      */
+    nr_free(function_name);
+    nr_free(accountID);
+    nr_free(region);
+    nr_free(qualifier);
     nr_regex_substrings_destroy(&matches);
     return;
   }
   if (nr_strempty(accountID)) {
+    nr_free(accountID);
     accountID = NRINI(aws_account_id);
+    using_account_id_ini = true;
   }
   if (nr_strempty(region)) {
     region_zval
@@ -474,6 +483,12 @@ void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attr
   }
 
   nr_regex_substrings_destroy(&matches);
+  nr_free(function_name);
+  if (!using_account_id_ini) {
+    nr_free(accountID);
+  }
+  nr_free(region);
+  nr_free(qualifier);
   nr_php_zval_free(&region_zval);
   nr_php_zval_free(&qualifier_zval);
 }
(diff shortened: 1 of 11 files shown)


agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,166 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    return;
+  }
+  zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
+  if (NULL == lambda_name || IS_STRING != Z_TYPE_P(lambda_name)) {
+    return;
+  }
+
+  /* Compile the regex */
+  if (NULL == NRPRG(aws_arn_regex)) {
+    NRPRG(aws_arn_regex) = nr_regex_create(AWS_ARN_REGEX, 0, 0);
```
zsistla, line 408: Where/when does this get destroyed?

zsistla: Can we just precompile this in minit like wordpress does?
https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_wordpress.c#L878-L886

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,166 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    return;
+  }
+  zval* lambda_args = nr_php_zend_hash_index_find(Z_ARRVAL_P(call_args), 0);
+  if (NULL == lambda_args || IS_ARRAY != Z_TYPE_P(lambda_args)) {
+    return;
+  }
+  zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
+  if (NULL == lambda_name || IS_STRING != Z_TYPE_P(lambda_name)) {
+    return;
+  }
```
zsistla, line 404: This function will already do all the checking for you and if 'FunctionName' has a valid string value, it will return it to you .
https://github.com/newrelic/newrelic-php-agent/blob/main/agent/lib_aws_sdk_php.h#L78

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,166 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+     * argument, so this won't happen in normal operation
+     */
+    nr_regex_substrings_destroy(&matches);
+    return;
+  }
+  if (nr_strempty(accountID)) {
+    accountID = NRINI(aws_account_id);
+  }
+  if (nr_strempty(region)) {
+    region_zval = nr_php_call(this_obj, "getRegion");
```
zsistla, line 434: instead of doing an nr_php_call here (which has higher overhead), you could use something like:
```
  region_zval
      = nr_php_get_zval_object_property(this_obj, "region");
  if (!nr_php_is_zval_non_empty_string(region_zval)) {
 ... 
}
```

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,166 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  } else {
+    return;
+  }
+#undef AWS_COMMAND_IS
+
+  /* reconstruct the ARN */
+  nr_aws_lambda_invoke(NR_EXECUTE_ORIG_ARGS, &cloud_attrs);
+  if (!cloud_attrs.cloud_resource_id) {
+    /* we do not want to instrument if we cannot reconstruct the ARN */
+    return;
```
zsistla, line 345: I think it's still fine to instrument in this case.
It won't create the relationship, but the external segment can still be created.  The cloud attributes are extra attributes, not core external segment attributes.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,166 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  nr_segment_traces_add_cloud_attributes(external_segment, &cloud_attrs);
+  nr_segment_external_params_t external_params = {.library = "aws_sdk"};
+  zval* data = nr_php_get_zval_object_property(*retval_ptr, "data");
+  if (NULL != data && IS_ARRAY == Z_TYPE_P(data)) {
+    zval* status_code = nr_php_zend_hash_find(Z_ARRVAL_P(data), "StatusCode");
+    if (NULL != status_code && IS_LONG == Z_TYPE_P(status_code)) {
+      external_params.status = Z_LVAL_P(status_code);
+    }
+    zval* metadata = nr_php_zend_hash_find(Z_ARRVAL_P(data), "@metadata");
+    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)
```
zsistla, line 371: Could simplify this a bit by something like just adding this check:
```
  if (IS_REFERENCE == Z_TYPE_P(metadata)) {
    metadata = Z_REFVAL_P(metadata);
  }
```

Then the code could look something like:
```
if (NULL != metadata && IS_REFERENCE == Z_TYPE_P(metadata)) {
    metadata = Z_REFVAL_P(metadata);
}
if nr_php_is_zval_valid_array(metadata) {
      zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
      if (nr_php_is_zval_non_empty_string(uri)) {
        external_params.uri = Z_STRVAL_P(uri);
      }
    }
```

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,166 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  char* arn = NULL;
+  char* function_name = NULL;
+  char* region = NULL;
+  zval* region_zval = NULL;
+  char* qualifier = NULL;
+  zval* qualifier_zval = NULL;
+  char* accountID = NULL;
+
+  /* verify arguments */
+  if (NULL == call_args || IS_ARRAY != Z_TYPE_P(call_args)) {
```
zsistla, line 394: General comment, there are a lot of helper functions for zval here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_zval.h
 of the form `nr_php_is_zval_*` that will already do the NULL and type checking for you for most zval types.

agent/php_nrini.c
```diff
@@ -3100,6 +3100,18 @@ STD_PHP_INI_ENTRY_EX("newrelic.vulnerability_management.composer_api.enabled",
                      newrelic_globals,
                      nr_enabled_disabled_dh)
 
+/*
+ * Cloud relationship settings
+ */
+STD_PHP_INI_ENTRY_EX("newrelic.cloud.aws.account_id",
+                     "1",
+                     NR_PHP_REQUEST,
+                     nr_string_mh,
```
zsistla, line 3109: Can this call a function that verifies the string is always 12 chars long (according to the spec).

agent/scripts/newrelic.ini.template
```diff
@@ -1352,3 +1352,12 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
 ;
 ;newrelic.message_tracer.segment_parameters.enabled = true
+
+; Setting: newrelic.cloud.aws.account_id
+; Type   : string
+; Scope  : per-directory
+; Default: none
+; Info   : AWS account id that is used to map entities called
+;          via the AWS SDK.
+;
```
zsistla, line 1362: Java has wording around this: https://docs.newrelic.com/docs/apm/agents/java-agent/configuration/java-agent-configuration-config-file/#cloud-configuration

agent/lib_aws_sdk_php.c
ZNeumann: This has been moved to be compiled on first use and destroyed on mshutdown.

agent/lib_aws_sdk_php.c
ZNeumann: That function performs a strdup that I don't think is necessary. The only other place it is used, the caller frees the duplicated string in scope. Can I remove the strdup from it?

agent/lib_aws_sdk_php.c
ZNeumann: changed

agent/lib_aws_sdk_php.c
ZNeumann: changed in all places

agent/lib_aws_sdk_php.h
```diff
@@ -74,6 +74,23 @@ extern void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
...
+ * Purpose : Handle when a LambdaClient::invoke command happens
+ *
+ * Params  : 1. NR_EXECUTE_ORIG_ARGS (execute_data, func_return_value)
+ *           2. cloud_attrs : the cloud attributes pointer to be
+ *              populated with the ARN
+ *
+ * Returns :
+ *
+ */
+void nr_aws_lambda_invoke(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs);
```
lavarou, line 87: The function name suggests that calling this function will invoke aws lambda where in fact this is a helper function that parses arguments passed to `Aws\Lambda\LambdaClient::invoke`. The comment right above where this function is used, `/* reconstruct the ARN */`, tells more what this function does than its name. Please adjust the function name to better match its purpose. Maybe `nr_aws_sdk_lambda_client_invoke_parse_args`?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  if (AWS_COMMAND_IS("invoke")) {
+    // Command can be saved here if in the future
+    // we instrument more than 1 lambda command
+  } else {
+    return;
+  }
+#undef AWS_COMMAND_IS
+
+  /* reconstruct the ARN */
+  nr_aws_lambda_invoke(NR_EXECUTE_ORIG_ARGS, &cloud_attrs);
```
lavarou, line 342: This function name suggests that it will invoke aws lambda, but that's not what will happen. This is just a helper function that parses arguments passed to `Aws\Lambda\LambdaClient::invoke` and populates `cloud_attrs`.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+      metadata = Z_REFVAL_P(metadata);
+      if (IS_ARRAY == Z_TYPE_P(metadata)) {
+        zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
+        if (nr_php_is_zval_valid_string(uri)) {
+          external_params.uri = Z_STRVAL_P(uri);
+        }
+      }
+    }
+  }
+  nr_segment_external_end(&auto_segment, &external_params);
```
lavarou, line 378: Should the first argument be `&auto_segment` or `&external_segment`?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  zval* data = nr_php_get_zval_object_property(*retval_ptr, "data");
+  if (nr_php_is_zval_valid_array(data)) {
+    zval* status_code = nr_php_zend_hash_find(Z_ARRVAL_P(data), "StatusCode");
+    if (nr_php_is_zval_valid_integer(status_code)) {
+      external_params.status = Z_LVAL_P(status_code);
+    }
+    zval* metadata = nr_php_zend_hash_find(Z_ARRVAL_P(data), "@metadata");
+    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)) {
+      metadata = Z_REFVAL_P(metadata);
+      if (IS_ARRAY == Z_TYPE_P(metadata)) {
```
lavarou, line 370: Use `nr_php_is_zval_valid_array`, which does `NULL` checks:
```suggestion
      if (nr_php_is_zval_valid_array(metadata)) {
```

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    nr_free(function_name);
+    nr_free(accountID);
+    nr_free(region);
+    nr_free(qualifier);
+    nr_regex_substrings_destroy(&matches);
+    return;
+  }
+  if (nr_strempty(accountID)) {
+    nr_free(accountID);
+    accountID = NRINI(aws_account_id);
```
lavarou, line 449: What is the value of `NRINI(aws_account_id)` if `newrelic.cloud.aws.account_id` is not set? Empty string (`""`) or `NULL`?

agent/lib_aws_sdk_php.c
zsistla: Instead of first use (which would make the first instrumented call take much longer than any other call, why not have it setup either in minit or in nr_aws_sdk_php_enable if you want to wait until at least the agent knows aws-sdk is involved.

agent/lib_aws_sdk_php.c
zsistla: The other place it's used needs the strdup because it uses it and modifies the string in place to extract the values it needs (instead of using regex).

agent/lib_aws_sdk_php.c
```diff
@@ -15,9 +15,16 @@
 #include "fw_support.h"
 #include "util_logging.h"
 #include "nr_segment_message.h"
+#include "nr_segment_external.h"
 #include "lib_aws_sdk_php.h"
 
 #define PHP_PACKAGE_NAME "aws/aws-sdk-php"
+#define AWS_ARN_REGEX "(arn:(aws[a-zA-Z-]*)?:lambda:)?" \
```
zsistla, line 22: Please change to AWS_ARN_LAMBDA_REGEX so we can differentiate when we get to other services that need ARNs as well.

agent/lib_aws_sdk_php.c
zsistla, line 370:
```
    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)) {
      metadata = Z_REFVAL_P(metadata);
      if (IS_ARRAY == Z_TYPE_P(metadata)) {
```

This code only handles the case where it's a reference, but in order to handle the case where it's a reference to an array or when it's the actual array, it should be like the snippet here: https://github.com/newrelic/newrelic-php-agent/pull/1023#discussion_r1955391446

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+        }
+      }
+    }
+  }
+  nr_segment_external_end(&auto_segment, &external_params);
+  nr_free(cloud_attrs.cloud_resource_id);
+}
+
+/* This stores the compiled regex to parse AWS ARNs. The compilation happens when
+ * it is first needed and is destroyed in mshutdown
```
zsistla, line 383: https://github.com/newrelic/newrelic-php-agent/pull/1023#discussion_r1972515431

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    // construct the ARN
+    if (!nr_strempty(qualifier)) {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s:%s",
+                       region, accountID, function_name, qualifier);
+    } else {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s",
+                       region, accountID, function_name);
+    }
+
+    // Attatch the ARN
```
zsistla, line 470:
```suggestion
    /* Set the cloud_resource_id to the ARN */
```

Style guidelines when we open sourced directed to use the `/* */` style commenting.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    }
+  }
+
+  if (!nr_strempty(accountID) && !nr_strempty(region)) {
+    // construct the ARN
+    if (!nr_strempty(qualifier)) {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s:%s",
+                       region, accountID, function_name, qualifier);
+    } else {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s",
```
zsistla, line 466: Where does `arn` get freed?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  if (nr_strempty(region)) {
+    region_zval
+      = nr_php_get_zval_object_property(this_obj, "region");
+    if (nr_php_is_zval_valid_string(region_zval)) {
+      region = Z_STRVAL_P(region_zval);
+    }
+  }
+
+  if (!nr_strempty(accountID) && !nr_strempty(region)) {
+    // construct the ARN
```
zsistla, line 461:
```suggestion
    /* construct the ARN */
```

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  nr_regex_substrings_t* matches =
+      nr_regex_match_capture(aws_arn_regex,
+                             Z_STRVAL_P(lambda_name),
+                             Z_STRLEN_P(lambda_name));
+  function_name = nr_regex_substrings_get_named(matches, "functionName");
+  accountID = nr_regex_substrings_get_named(matches, "accountId");
+  region = nr_regex_substrings_get_named(matches, "region");
+  qualifier = nr_regex_substrings_get_named(matches, "qualifier");
+
+  /* suppliment missing information with API calls */
```
zsistla, line 434:
```suggestion
  /* supplement missing information with API calls */
```

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  if (nr_strempty(function_name)) {
```
zsistla, line 435: Since this is unlikely to ever happen, consider using `nrunlikely` here.

agent/php_nrini.c
```diff
@@ -3100,6 +3100,18 @@ STD_PHP_INI_ENTRY_EX("newrelic.vulnerability_management.composer_api.enabled",
                      newrelic_globals,
                      nr_enabled_disabled_dh)
 
+/*
+ * Cloud relationship settings
+ */
+STD_PHP_INI_ENTRY_EX("newrelic.cloud.aws.account_id",
+                     "1",
```
zsistla, line 3107:
```suggestion
                     "",
```

agent/lib_aws_sdk_php.c
zsistla: Default is set [here](https://github.com/newrelic/newrelic-php-agent/pull/1023/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R460), and should be set to empty string as default.
The code does check for empty/null here before trying to use it: https://github.com/newrelic/newrelic-php-agent/pull/1023/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R460

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  }
+
+  nr_regex_substrings_destroy(&matches);
+  nr_free(function_name);
+  if (!using_account_id_ini) {
+    nr_free(accountID);
+  }
+  nr_free(region);
+  nr_free(qualifier);
+  nr_php_zval_free(&region_zval);
```
zsistla, line 481: The nice thing about `nr_php_get_zval_object_property` is that `region_zval` no longer needs to be freed.

agent/lib_aws_sdk_php.c
zsistla: Ah, I found where it got freed.
Could you put a comment indicating that the calling function will be responsible for freeing?

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -458,6 +482,103 @@ static void test_nr_lib_aws_sdk_php_handle_version(void) {
   tlib_php_request_end();
 }
 
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
+static void test_nr_lib_aws_sdk_php_lambda_invoke() {
+  tlib_php_engine_create("newrelic.cloud.aws.account_id=\"test_account\"");
```
zsistla, line 487: This is one way to set an NRINI value.
You could also set it manually by just doing:
NRINI(aws_account_id) = NULL;
or NRINI(aws_account_id) ="123456789012";
so it's easier to do various test cases.

Can you please add a test case where `aws_account_id` is `NULL` and where it is `""`?

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -458,6 +482,103 @@ static void test_nr_lib_aws_sdk_php_handle_version(void) {
...
+  tlib_pass_if_not_null("Expression should evaluate.", expr);
+  nr_php_zval_free(&expr);
+  nr_php_zval_free(&expect_arg);
+  nr_php_zval_free(&array_arg);
+
+  /* Test invalid arg 1 */
+  args
+      = "array("
+        "    0 => array("
+        "        'FunctionName' => 123"
```
zsistla, line 545: Can you please add a test for where functionName is an empty string and where it is null?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+                       region, accountID, function_name, qualifier);
+    } else {
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s",
+                       region, accountID, function_name);
+    }
+
+    // Attatch the ARN
+    cloud_attrs->cloud_resource_id = arn;
+  }
+
```
zsistla, line 473: Can't we set the other cloud attributes here as well like
region and account_id?

agent/lib_aws_sdk_php.c
ZNeumann: In that case I would prefer not to do unnecessary strdups and wish to keep this as is. (or the other code can be modified to perform the strdup on the returned un-duped str)

agent/lib_aws_sdk_php.c
ZNeumann: I am willing to make the change either way, but my understanding of references was that you cannot have a reference to null

agent/lib_aws_sdk_php.c
zsistla: That particular function needs the strdrup since we are freeing the arg value that we extracted the string from before exiting the function, so the strdup cannot simply be removed.
It can be explored further, but that's probably beyond the scope of this PR.

agent/lib_aws_sdk_php.c
zsistla: The array check should be removed from the if block it is in; otherwise, we'll only extract what we need when metadata starts as a reference, not when it starts as a array.

Something like this:
```
if (NULL != metadata && IS_REFERENCE == Z_TYPE_P(metadata)) {
    metadata = Z_REFVAL_P(metadata);
}
if (nr_php_is_zval_valid_array(metadata)) {
      zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
      if (nr_php_is_zval_non_empty_string(uri)) {
        external_params.uri = Z_STRVAL_P(uri);
      }
    }
```

agent/lib_aws_sdk_php.h
lavarou: `nr_aws_lambda_invoke` is very similar in function to `nr_lib_aws_sdk_php_sqs_parse_queueurl` - both parse stuff (arguments in case of `nr_aws_lambda_invoke` and queue url in case of `nr_lib_aws_sdk_php_sqs_parse_queueurl`).

agent/lib_aws_sdk_php.c
ZNeumann: In an effort to minimize the number of times that this regex is unnecessarily compiled, I would like to keep the compilation where it is.

agent/lib_aws_sdk_php.c
ZNeumann: In general, I prefer to avoid using those constructs unless on critical performance code paths. Otherwise, outsmarting the branch predictor is rarely worth it.

agent/lib_aws_sdk_php.c
ZNeumann: region and account_id, if available, will be in the arn. It is not in the spec to add them and would just be duplicate data, adding to overhead and data ingest.

agent/tests/test_lib_aws_sdk_php.c
ZNeumann: added in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/tests/test_lib_aws_sdk_php.c
ZNeumann: aws_account_id ini should never be `NULL`, as it defauls to `""` and requires a valid 12-length string to set. Added test for "" in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: removed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/php_nrini.c
ZNeumann: fixed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: added in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: fixed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: See that thread.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,185 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    return;
+  }
+
+#define AWS_COMMAND_IS(CMD) \
+  (command_name_len == (sizeof(CMD) - 1) && nr_streq(CMD, command_name_string))
+
+  /* Determine if we instrument this command. */
+  if (AWS_COMMAND_IS("invoke")) {
+    // Command can be saved here if in the future
+    // we instrument more than 1 lambda command
```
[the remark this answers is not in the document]
ZNeumann: changed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: changed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: fixed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.c
ZNeumann: changed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/lib_aws_sdk_php.h
ZNeumann: changed in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/php_nrini.c
ZNeumann: added in [5c45a0e](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f)

agent/tests/test_lib_aws_sdk_php.c
ZNeumann: Turns out it will be null, rather than the default, when the 12-length requirement is violated. Added tests here [25475ef](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/25475efa60752f455c8ca6bf90d5ef8e15ed1cc0)

agent/lib_aws_sdk_php.c
zsistla: That's a failure in the spec to update that.  But I'm okay leaving as-is for now; it's easy to update later if it's needed.

agent/php_nrini.c
zsistla: Can you please add a test case to verify the functionality?

agent/lib_aws_sdk_php.c
zsistla: Resolving this since to keep discussion in the other thread.

agent/lib_aws_sdk_php.h
```diff
@@ -74,6 +74,24 @@ extern void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
...
+/*
+ * Purpose : Handle when a LambdaClient::invoke command happens
+ *
+ * Params  : 1. NR_EXECUTE_ORIG_ARGS (execute_data, func_return_value)
+ *           2. cloud_attrs : the cloud attributes pointer to be
+ *              populated with the ARN
+ *
+ * Returns :
+ *
+ * Note: The caller is responsible for freeing cloud_atters->cloud_resource_id
```
zsistla, line 86:
```suggestion
 * Note: The caller is responsible for freeing cloud_attrs->cloud_resource_id
```

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    if (nr_php_is_zval_valid_integer(status_code)) {
+      external_params.status = Z_LVAL_P(status_code);
+    }
+    zval* metadata = nr_php_zend_hash_find(Z_ARRVAL_P(data), "@metadata");
+    if (NULL != metadata  && IS_REFERENCE == Z_TYPE_P(metadata)) {
+      metadata = Z_REFVAL_P(metadata);
+    }
+    if (nr_php_is_zval_valid_array(metadata)) {
+      zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
+      if (nr_php_is_zval_valid_string(uri)) {
```
zsistla, line 370: Is an empty string acceptable, or should this be checking for a valid, nonempty string instead with `nr_php_is_zval_non_empty_string`?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  /* verify arguments */
+  if (!nr_php_is_zval_valid_array(call_args)) {
+    return;
+  }
+  zval* lambda_args = nr_php_zend_hash_index_find(Z_ARRVAL_P(call_args), 0);
+  if (!nr_php_is_zval_valid_array(lambda_args)) {
+    return;
+  }
+  zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
+  if (!nr_php_is_zval_valid_string(lambda_name)) {
```
zsistla, line 413: Is an empty string acceptable, or should this be checking for a valid, nonempty string instead with `nr_php_is_zval_non_empty_string`?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    using_account_id_ini = true;
+  }
+  if (nr_strempty(region)) {
+    zend_class_entry* base_class = NULL;
+    if (NULL != execute_data->func && NULL!= execute_data->func->common.scope) {
+       base_class = execute_data->func->common.scope;
+    }
+    region_zval
+      = nr_php_get_zval_object_property_with_class(this_obj, base_class, "region");
+    if (nr_php_is_zval_valid_string(region_zval)) {
```
zsistla, line 457: Is an empty string acceptable, or should this be checking for a valid, nonempty string instead with `nr_php_is_zval_non_empty_string`?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    cloud_attrs->cloud_resource_id = arn;
+  }
+
+  nr_regex_substrings_destroy(&matches);
+  nr_free(function_name);
+  if (!using_account_id_ini) {
+    nr_free(accountID);
+  }
+  /* if region_zval is a valid string, we have already freed region */
+  if (!nr_php_is_zval_valid_string(region_zval)) {
```
zsistla, line 487: Might need modification depending on this convo: https://github.com/newrelic/newrelic-php-agent/pull/1023/files#r1978181824

agent/lib_aws_sdk_php.c
ZNeumann: We check for it being non-empty later on https://github.com/newrelic/newrelic-php-agent/pull/1023/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R467

agent/lib_aws_sdk_php.c
ZNeumann: If it is empty, the regex will fail to capture anything and the function will return exit at that point.

agent/lib_aws_sdk_php.c
ZNeumann: If it is empty, there is no harm in attaching it to the external params, but also no gain. added a check in [b72ebc1](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/b72ebc1104e0bf871dca57a2a458e488e1962505)

agent/php_nrini.c
ZNeumann: tests were added in [25475ef](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/25475efa60752f455c8ca6bf90d5ef8e15ed1cc0)

agent/tests/test_lib_aws_sdk_php.c
zsistla: Why not just make it `""` if the length requirement is violated with an else statement [here](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f#diff-07c39860ed983532dfdc2a9d3093475beb7bd749067607c324a1e7718cffd92fR1237)?

ZNeumann: I followed the design of other functions that check for string length, like our [license key check](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/5c45a0ef346dbb2bf7e1b28a8ed0de4cbabcbb3f#diff-07c39860ed983532dfdc2a9d3093475beb7bd749067607c324a1e7718cffd92fL1163). If there is no reason to follow this design, then I can switch it.

agent/lib_aws_sdk_php.c
zsistla: instead of `if (nr_php_is_zval_valid_string(uri) && !nr_strempty(Z_STRVAL_P(uri)))`

Just use `nr_php_is_zval_non_empty_string` which does both of those things.

agent/lib_aws_sdk_php.c
zsistla: True, but looks like we [get to at least here](https://github.com/newrelic/newrelic-php-agent/blob/main/axiom/util_regex.c#L148), which could be skipped if we used nr_php_is_zval_non_empty_string to check for both valid and not empty string instead of just valid string.

agent/tests/test_lib_aws_sdk_php.c
zsistla: Hmm, I don't have a strong opinion one way or the other on this.
If it's consistent with other string behavior, I'm okay with it.

agent/php_nrini.c
zsistla: Nice!

agent/lib_aws_sdk_php.c
ZNeumann: changed in [50aaa3c](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/50aaa3c55da96aaa0b34449d87ec3623b33c908a)

agent/lib_aws_sdk_php.c
ZNeumann: changed in [50aaa3c](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/50aaa3c55da96aaa0b34449d87ec3623b33c908a)

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+
+  if (NULL == command_name_string || 0 == command_name_len) {
+    return;
+  }
+
+  if (NULL == *retval_ptr) {
+    /* Do not instrument when an exception has happened */
+    return;
+  }
+
```
lavarou, line 328: `Aws\Sqs\SqsClient` commands that end up throwing exceptions are still instrumented. Why a different approach was selected for `Aws\Lambda\LambdaClient`?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+#define AWS_COMMAND_IS(CMD) \
+  (command_name_len == (sizeof(CMD) - 1) && nr_streq(CMD, command_name_string))
+
+  /* Determine if we instrument this command. */
+  if (AWS_COMMAND_IS("invoke")) {
```
lavarou, line 333: Should [`invokeAsync`](https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-lambda-2015-03-31.html#invoke) be instrumented as well?

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  /* re-use start time from auto_segment started in func_begin */
+  external_segment->start_time = auto_segment->start_time;
+  cloud_attrs.aws_operation = command_name_string;
+
+  /* end the segment */
+  nr_segment_traces_add_cloud_attributes(external_segment, &cloud_attrs);
+  nr_segment_external_params_t external_params = {.library = "aws_sdk"};
+  zval* data = nr_php_get_zval_object_property(*retval_ptr, "data");
+  if (nr_php_is_zval_valid_array(data)) {
+    zval* status_code = nr_php_zend_hash_find(Z_ARRVAL_P(data), "StatusCode");
```
lavarou, line 360: Shouldn't `StatusCode` start with small `s`?

ZNeumann: Below is an example result. The status code is stored twice: once in the base result, and once in the metadata. The one in the base result which I am grabbing was the "S".

```
object(Aws\Result)#77 (1) {
 ["data":"Aws\Result":private]=>
 array(5) {
  ["Payload"]=>
  object(GuzzleHttp\Psr7\Stream)#74 (7) {
   ["stream":"GuzzleHttp\Psr7\Stream":private]=>
   resource(103) of type (stream)
   ["size":"GuzzleHttp\Psr7\Stream":private]=>
   NULL
   ["seekable":"GuzzleHttp\Psr7\Stream":private]=>
   bool(true)
   ["readable":"GuzzleHttp\Psr7\Stream":private]=>
   bool(true)
   ["writable":"GuzzleHttp\Psr7\Stream":private]=>
   bool(true)
   ["uri":"GuzzleHttp\Psr7\Stream":private]=>
   string(10) "php://temp"
   ["customMetadata":"GuzzleHttp\Psr7\Stream":private]=>
   array(0) {
   }
  }
  ["StatusCode"]=>
  int(200)
  ["FunctionError"]=>
  string(0) ""
  ["LogResult"]=>
  string(0) ""
  ["@metadata"]=>
  array(4) {
   ["statusCode"]=>
   int(200)
   ["effectiveUri"]=>
   string(84) "https://lambda.us-east-2.amazonaws.com/2015-03-31/functions/agent-lambda/invocations"
   ["headers"]=>
   array(8) {
    ["date"]=>
    string(29) "??"
    ["content-type"]=>
    string(16) "application/json"
    ["content-length"]=>
    string(2) "81"
    ["connection"]=>
    string(5) "close"
    ["x-amzn-requestid"]=>
    string(36) "??"
    ["x-amzn-remapped-content-length"]=>
    string(1) "0"
    ["x-amz-executed-version"]=>
    string(7) "$LATEST"
    ["x-amzn-trace-id"]=>
    string(95) "??"
   }
   ["transferStats"]=>
   array(1) {
    ["http"]=>
    array(1) {
     [0]=>
     array(0) {
     }
    }
   }
  }
 }
}
```

agent/lib_aws_sdk_php.c
ZNeumann: invokeasync is deprecated https://docs.aws.amazon.com/lambda/latest/api/API_InvokeAsync.html

agent/lib_aws_sdk_php.c
ZNeumann: It is a relatively arbitrary decision. Here is the reasoning for my decision:

This instrumentation heavily uses the return value, so lots of information will be missing. Further, an exception is likely to happen BEFORE the actual lambda call. If we create an external segment when no external call happened, that could mislead customer debugging efforts. And even if the exception happens after the lambda call, because ASW is using Guzzle under the hood, an external segment with the information will still be created. While this won't result in a mapped entity relationship due to lack of instrumentation, that is a tradeoff to correct for the first case.

Additionally, this instrumentation digs into the parameters and arguments lots. In the case of an exception, it seems likely to me that we won't be able to get all of the information we want to create the full ARN anyway.

agent/lib_aws_sdk_php.c
lavarou: > invokeasync is deprecated https://docs.aws.amazon.com/lambda/latest/api/API_InvokeAsync.html

This is not link to AWS for PHP SDK's documentation - it is a link to AWS's REST API. AWS for PHP SDK's documentation for [`invokeAsync`](https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-lambda-2015-03-31.html#invoke) doesn't say that `invokeAsync` is deprecated. Under the hook AWS for PHP SDK is not using https://docs.aws.amazon.com/lambda/latest/api/API_InvokeAsync.html for `invokeAsync`.

ZNeumann: From [that same page](https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-lambda-2015-03-31.html#invokeasync), it states to use invoke instead.

> For asynchronous function invocation, use Invoke.

However, I have tested using invokeAsync. 1) it still calls invoke under the hood and the arn is attached as expected. 2) It returns a promise which yields an object with the expected fields, rather than an array. This behavior still works with the current instrumentation, but missing everything from the return values.

Because 1) AWS recommends against invokeAsync 2) our instrumentation of invokeAsync still produces the requires information for a relationship to be formed and is just missing extraneous information and 3)  instrumentation of promises is very complicated and, as far as I know, has not been done in the agent, I see no reason to attempt to instrument this.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+      metadata = Z_REFVAL_P(metadata);
+    }
+    if (nr_php_is_zval_valid_array(metadata)) {
+      zval* uri = nr_php_zend_hash_find(Z_ARRVAL_P(metadata), "effectiveUri");
+      if (nr_php_is_zval_non_empty_string(uri)) {
+        external_params.uri = Z_STRVAL_P(uri);
+      }
+    }
+    
+  }
```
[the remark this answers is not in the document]
ZNeumann: This code was written in an effort to avoid calling php

agent/lib_aws_sdk_php.c
lavarou: > From that same page, it states to use invoke instead.
>
>> For asynchronous function invocation, use Invoke.

But it doesn't say it's deprecated so our agent should instrument this. Thanks for checking that under the hood `invokeAsync` calls `invoke` and agent behaves correctly when `invokeAsync` is used.

lavarou: > 3) instrumentation of promises is very complicated and, as far as I know, has not been done in the agent, I see no reason to attempt to instrument this.

It is done in the agent - see `guzzlehttp/guzzle`'s instrumentation (see [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/agent/lib_guzzle6.c)). Since you verified that under the hood `invokeAsync` calls `invoke`, and agent behaves correctly when `invokeAsync` is used, I consider this resolved.

agent/lib_aws_sdk_php.c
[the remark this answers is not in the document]
lavarou: As far as I know using `nr_php_call` does not incur additional overhead.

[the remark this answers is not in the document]
zsistla: Using nr_php_get_zval_object_property is also a pattern in the agent.  It is much more performant than nr_php_call.  There are tests that would show us if this structure ever changes, but the Data field, metadata, and status code are consistent across aws-sdk language versions.

All aws-sdk languages are thin wrappers around their API that communicates with their services.  The request/responses are detailed and consistent and as such all elements are very unlikely to change.

Look [here](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/getting-started_basic-usage.html) for more, but some relevant parts are:
```
Executing a successful operation returns an Aws\Result object. Instead of returning the raw XML or JSON data of a service (which is the SAME response returned to all language SDKs), the SDK coerces the response data into an associative array structure. It normalizes some aspects of the data based on its knowledge of the specific service and the underlying response structure.

The result structure of each API operation is documented in the API docs for each operation.
```

As such, [AwsResult](https://docs.aws.amazon.com/aws-sdk-php/v3/api/class-Aws.Result.html) is extremely unlikely to change (and if it does we have tests to catch it) and the implementation here is both performant and accurate.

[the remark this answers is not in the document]
zsistla: nr_php_call incurs much more overhead and the AwsResult structure will not change.

agent/lib_aws_sdk_php.c
zsistla: So there is some difference between message segments and external segments and also a difference between [SQS](https://docs.aws.amazon.com/aws-sdk-php/v3/api/class-Aws.Sqs.SqsClient.html) and [lambda](https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-lambda-2015-03-31.html#invoke).
SQS 

SQS message segments are able to extract all that is needed to create a basic message segment with the the initial parameters and doesn't rely on the retval for anything needed to create a basic message segment.

Lambda however does rely on the retval because the initial parameters are not guaranteed to have the URI.

agent/lib_aws_sdk_php.c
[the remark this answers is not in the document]
lavarou: > AwsResult structure will not change.

But underlying implementation might without any warning. I understand the strategy is to deal with it when it happens.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+    accountID = NRINI(aws_account_id);
+    using_account_id_ini = true;
+  }
+  if (nr_strempty(region)) {
+    zend_class_entry* base_class = NULL;
+    if (NULL != execute_data->func && NULL!= execute_data->func->common.scope) {
+       base_class = execute_data->func->common.scope;
+    }
+    region_zval
+      = nr_php_get_zval_object_property_with_class(this_obj, base_class, "region");
```
lavarou, line 456: Is this construct correct? Take a look at [this example](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/agent/lib_guzzle6.c#L411-L420). The first argument to `nr_php_get_zval_object_property_with_class` is the object to extract the property from, and the second argument is the class entry for the context from which the property should be extracted. What is this code trying to do?

ZNeumann: `LambdaClient` extends `AwsClient`. `region` exists on the `AwsClient`, not the `LambdaClient`. So if we wish to extract `region` from `this` (which is a `LambdaClient`), we must indicate the base class.

agent/lib_aws_sdk_php.c
```diff
@@ -295,6 +302,194 @@ void nr_lib_aws_sdk_php_sqs_parse_queueurl(
...
+  aws_arn_regex = nr_regex_create(AWS_LAMBDA_ARN_REGEX, 0, 0);
+}
+
+void nr_aws_sdk_mshutdown(void) {
+  nr_regex_destroy(&aws_arn_regex);
+}
+
+void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs) {
+  zval* call_args = nr_php_get_user_func_arg(2, NR_EXECUTE_ORIG_ARGS);
+  zval* this_obj = getThis();//NR_PHP_USER_FN_THIS();
```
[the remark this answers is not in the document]
ZNeumann: This was leftover from testing about https://github.com/newrelic/newrelic-php-agent/pull/1023#discussion_r1983857512. Using NR_PHP_USER_FN_THIS() in [fbf408f](https://github.com/newrelic/newrelic-php-agent/pull/1023/commits/fbf408fb81c7a6265e5978c6b0e51910fb1f79aa)

agent/lib_aws_sdk_php.c
lavarou: If the intent is to look for the `"region"` property either in `this` (which is `Aws\Lambda\LambdaClient`) and then fall back to looking for it in `this`'s base class (which is `Aws\AwsClient`), then wouldn't it be safer to get the `base_class` like this: `base_class = nr_php_find_class("aws\\awsclient");`?

lavarou: The best strategy to ensure that the agent will be able to retrieve the region is to call `getRegion` method on `this` like this: `region = nr_php_call(this, "getRegion");` It doesn't rely on any assumptions but rather on publicly available API of AWS SDK for PHP.

zsistla (approved):

PR #1027, opened 2025-02-20
feat(agent): Adds label forwarding to log events

mfulb:
This PR adds the ability for labels to be forwarded with any log messages forwarded by the PHP agent.


agent/php_txn.c
```diff
@@ -917,17 +1035,21 @@ nr_status_t nr_php_txn_begin(const char* appnames,
...
+  if (NULL == NRPRG(app)) {
     nrl_debug(NRL_INIT, "unable to begin transaction: app '%.128s' is unknown",
               appnames ? appnames : "");
     return NR_FAILURE;
   }
 
   attribute_config = nr_php_create_attribute_config(TSRMLS_C);
-  NRPRG(txn) = nr_txn_begin(NRPRG(app), &opts, attribute_config);
+  log_forwarding_labels
+      = nr_php_txn_get_log_forwarding_labels(NRPRG(app)->info.labels);
```
zsistla, line 1046: Why not just have this be part of `NRPRG(app)->info`? Why does it need to be passed separately? 

It could just be added as a var here: https://github.com/newrelic/newrelic-php-agent/pull/1027/files#diff-f5f8c5da8db458c336e8505b331bcb42a571175d89cdd0e65fbc39c1121a5960L881

mfulb: I considered that but it would be odd to put it in "appinfo" but send it to the daemon in the "transaction" payload.   Also I believe the information needed to make the filtered log labels list is not available when "appinfo" is initialized.

agent/scripts/newrelic.ini.template
```diff
@@ -1305,6 +1304,29 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
 ;newrelic.application_logging.forwarding.context_data.exclude = ""
 
+; Setting: newrelic.application_logging.forwarding.labels.enabled
+; Type   : boolean
+; Scope  : per-directory
+; Default: false
+; Info   : Toggles whether the agent adds labels as attributes to log records for 
+;          sending to New Relic.  The labels can be defined by the newrelic.labels
+;          configuration value, and filtered using the 
+;          newrelic.application_logging.forwawrding.labels.exclude configuration value.
```
lavarou, line 1314:
```suggestion
; Info   : Toggles whether the agent adds labels as attributes to log records which
;          are sent to New Relic.  The labels can be defined by the newrelic.labels
;          configuration value, and filtered using the
;          newrelic.application_logging.forwarding.labels.exclude configuration value.
```

agent/scripts/newrelic.ini.template
```diff
@@ -1305,6 +1304,29 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
+;
+;newrelic.application_logging.forwarding.labels.enabled = false
+
+; Setting: newrelic.application_logging.forwarding.labels.exclude
+; Type   : string
+; Scope  : per-directory
+; Default: ""
+; Info   : A list of labels to NOT add as attributes to logs which are forwarded
+;          to New Relic.  This list can be separated by commas.
```
lavarou, line 1323: Please update description to state that label matching is case insensitive and multiple list elements *MUST* be separated by commas.

tests/integration/logging/monolog2/test_monolog_labels_forwarding_exclude_05.php
```diff
@@ -0,0 +1,216 @@
...
+]
+*/
+
+
+/*EXPECT_LOG_EVENTS
+[
+    {
+      "common": {
+        "attributes": { }
+      },
```
lavarou, line 87: For other reviewers - sending `"common"` object with empty `"attributes"` is an existing functionality.

axiom/tests/.gitignore
```diff
@@ -95,7 +95,6 @@ test_segment
 test_segment_children
 test_segment_datastore
 test_segment_external
-test_segment_message
```
lavarou, line 98: I think this has been removed by a mistake during rebase.

agent/php_txn.c
zsistla: Why wouldn't the info needed to make the filtered list not be available when appinfo is initialized?
Isn't the info is already available from the INI?


Otherwise, [all of this logic is](https://github.com/newrelic/newrelic-php-agent/pull/1027/files#diff-f5f8c5da8db458c336e8505b331bcb42a571175d89cdd0e65fbc39c1121a5960R566-R649) being done and added to the overhead of the beginning of every txn when it's a known value that won't change and only needs to be done once even if you still send it via the txn payload.

You could compute it once here after the labels are added: https://github.com/newrelic/newrelic-php-agent/pull/1027/files#diff-f5f8c5da8db458c336e8505b331bcb42a571175d89cdd0e65fbc39c1121a5960L881.
Then you could keep your same logic to send it in the txn payload
by just doing 
```
log_forwarding_labels
      = NRPRG(app)->info.log_labels;
```


The above would keep all the sending labels in the txn payload logic, but would save the agent from having to do all the string manipulation and exclude comparisons needed to recompute the log labels (which never change since they are set in the INI) in every txn.

Alternatively, you could also just theoretically send it with the appinfo payload once, then it could be set somewhere on the daemon side where the daemon could consult the appinfo for any log labels before sending a logevent.

lavarou: > Alternatively, you could also just theoretically send it with the appinfo payload once, then it could be set somewhere on the daemon side where the daemon could consult the appinfo for any log labels before sending a logevent.

I also like this idea.

agent/scripts/newrelic.ini.template
zsistla: An example (like [regular labels provides](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/scripts/newrelic.ini.template#L863C18-L863C50)) could also be useful.

zsistla:
Thanks for adding the new tests_monolog_label* tests.
I did a first look through and overall they look good.
A few questions/requests:

1) Could you add additional tests where:
* Labels list is not set, but exclude log labels are
* Log forwarding disabled, but exclude log labels are set (I.e., not default)
* Exclude label list includes a label that matches multiple case sensitive labels; expected behavior is they are all removed.
* Exclude label list has same entry more than once
2) Did you run the new integration/multiverse label tests with valgrind?
3) What is expected behavior if a label has a comma?

agent/php_txn.c
```diff
@@ -666,6 +777,11 @@ static void nr_php_txn_send_metrics_once(nrtxn_t* txn TSRMLS_DC) {
   nrm_force_add(NRTXN(unscoped_metrics), metname, 0);
   nr_free(metname);
 
+  metname = nr_formatf("Supportability/Logging/Labels/PHP/%s",
```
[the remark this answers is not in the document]
mfulb: I will create a PR associated with the next release to add this.  I will also create a ticket as part of the epic to track this.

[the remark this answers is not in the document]
mfulb: Ticket has been added.

axiom/tests/.gitignore
mfulb: Actually I added it because it was missing then changed my mind and added it to this PR instead:
https://github.com/newrelic/newrelic-php-agent/pull/1026

mfulb: I had it in this PR, removed it, put it in https://github.com/newrelic/newrelic-php-agent/pull/1026 and then rebased.  Weird it didn't complain about a conflict - I'll look into it.

agent/php_txn.c
mfulb: It was discussed that the agent does not have a parsed list of labels until `nr_php_txn_begin()` executes at the start of a transaction.  Therefore the list of labels for forwarding with log events is not available until this time.  The current design is based on this limitation.  A future effort will be needed to first move label parsing into MINIT and then the log labels computation could be moved there as well.

agent/scripts/newrelic.ini.template
mfulb: Updated in commit https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/277d03449135311cb9dd9d0661d8eaa0982f69e4.

agent/php_txn.c
```diff
@@ -536,6 +536,117 @@ static nrobj_t* nr_php_txn_get_labels() {
...
+   */
+  log_labels = nro_new(NR_OBJECT_HASH);
+  for (int i = 0; i < nro_getsize(label_keys); i++) {
+    const char* key = nro_get_array_string(label_keys, i + 1, NULL);
+    const char* value = nro_get_hash_string(labels, key, NULL);
+    char* lower_key = nr_string_to_lowercase(key);
+    bool exclude = false;
+    nr_status_t rv;
+
+    exclude = nro_get_hash_boolean(exclude_labels_hash, lower_key, &rv);
```
zsistla, line 625: `exclude` is set as a bool, but `nro_get_hash_boolean` returns an `int`, not a bool.

```
int nro_get_hash_boolean(const nrobj_t* obj,
                         const char* key,
                         nr_status_t* errp)
```
It returns -1 on failure/error cases and 0/1 to indicate a setting of false/true.

Consider saving some steps/checks and removing the need for a rv by setting exclude to be an unsigned int.

It could then be simplified a bit, so that instead of:
```
   bool exclude = false;
    nr_status_t rv;

    exclude = nro_get_hash_boolean(exclude_labels_hash, lower_key, &rv);
if (NR_SUCCESS != rv) {
      exclude = false;
    }
    if (!exclude) {
      nro_set_hash_string(log_labels, key, value);
    } else {
      nrl_verbosedebug(NRL_TXN, "%s: Excluding label %s", __FUNCTION__, key);
    }
```

It could simplified to be:
```
if (nro_get_hash_boolean(exclude_labels_hash, lower_key, NULL) != 1) {
      nro_set_hash_string(log_labels, key, value);
    } else {
      nrl_verbosedebug(NRL_TXN, "%s: Excluding label %s", __FUNCTION__, key);
    }
```

agent/php_txn.c
```diff
@@ -536,6 +536,117 @@ static nrobj_t* nr_php_txn_get_labels() {
...
+    if (NR_SUCCESS != rv) {
+      exclude = false;
+    }
+    if (!exclude) {
+      nro_set_hash_string(log_labels, key, value);
+    } else {
+      nrl_verbosedebug(NRL_TXN, "%s: Excluding label %s", __FUNCTION__, key);
```
zsistla, line 632: Key was not checked for NULL, this could generate a segfault.  Check for NULL earlier, or wrap in NRSAFESTR.

agent/php_txn.c
```diff
@@ -536,6 +536,117 @@ static nrobj_t* nr_php_txn_get_labels() {
...
+  nro_iteratehash(labels, nr_php_txn_collect_label_keys_iter,
+                  (void*)label_keys);
+
+  /* filter by going over the list of label keys, seeing if it exists in the
+   * exclude hash, and if it does skip it otherwise copy key/value for label
+   * to the log labels
+   */
+  log_labels = nro_new(NR_OBJECT_HASH);
+  for (int i = 0; i < nro_getsize(label_keys); i++) {
+    const char* key = nro_get_array_string(label_keys, i + 1, NULL);
```
zsistla, line 619: A NULL check for the key could help skip this round of the loop.

agent/php_txn.c
```diff
@@ -536,6 +536,117 @@ static nrobj_t* nr_php_txn_get_labels() {
...
+    char* lower_key = nr_string_to_lowercase(key);
+    bool exclude = false;
+    nr_status_t rv;
+
+    exclude = nro_get_hash_boolean(exclude_labels_hash, lower_key, &rv);
+    if (NR_SUCCESS != rv) {
+      exclude = false;
+    }
+    if (!exclude) {
+      nro_set_hash_string(log_labels, key, value);
```
zsistla, line 630: Also would be a useless call if passing in a NULL key.

agent/php_txn.c
```diff
@@ -536,6 +536,117 @@ static nrobj_t* nr_php_txn_get_labels() {
...
+    return NULL;
+  }
+
+  /* if logging labels are disabled then nothing to do */
+  if (0 == NRINI(log_forwarding_labels_enabled)) {
+    nrl_verbosedebug(NRL_TXN, "%s: Log forwarding labels disabled",
+                     __FUNCTION__);
+    return NULL;
+  }
+
```
zsistla, line 583: A check if `log_forwarding_labels_exclude` is an empty string (most common case) would allow an early exit from the function.

daemon/internal/newrelic/log_events_test.go
```diff
@@ -0,0 +1,158 @@
...
+	if string(json) != expected {
+		t.Errorf("expected JSON %s, got %s", expected, string(json))
+	}
+}
+
+func TestSetLogForwardingLabels(t *testing.T) {
+	events := NewLogEvents(10)
+	id := AgentRunID(`12345`)
+	log_data := []byte(`{"message":"test log event"}`)
+	label_data := []byte(`[{"label_type":"type1","label_value":"value1"},{"label_type":"type2","label_value":"value2"}]`)
```
zsistla, line 54: Could you please add a test where label name has spaces?

protocol/flatbuffers/protocol.fbs
```diff
@@ -118,7 +118,8 @@ table Transaction {
   sampling_priority:      double;  // added in the 8.2 PHP agent release
   span_events:            [Event];
   log_events:             [Event]; // added in the 10.1 PHP agent release
-  php_packages:           Event;   // added in the ??? PHP agent release
+  php_packages:           Event;   // added in the 10.17 PHP agent release
+  log_forwarding_labels:  Event;   // added in the ??? PHP agent release
```
zsistla, line 122: I think you can go ahead and put 11.7 in there instead of ???

agent/php_txn.c
zsistla: It's still possible to send it with the appinfo payload once, then it could be set somewhere on the daemon side where the daemon could consult the appinfo for any log labels before sending a logevent which seems a more correct way of doing it then sending it with every txn.

Send log labels with info (so only send once when setting up the app).
Parse it here and assign it to a meaningful variable (this could be a standalone variable or a structure if we anticipate wanting to reuse other app info without having to send it in txn).
https://github.com/newrelic/newrelic-php-agent/blob/4bb2f2a74e5c2b1e505bb9a7ccd3ade46eb02b6c/daemon/internal/newrelic/commands.go#L293
Pass it as a variable here to EventHarvestConfig: https://github.com/newrelic/newrelic-php-agent/blob/4bb2f2a74e5c2b1e505bb9a7ccd3ade46eb02b6c/daemon/internal/newrelic/app.go#L225 and set it on EventConfigs
Now, when a harvest is being created for this app here: https://github.com/newrelic/newrelic-php-agent/blob/4bb2f2a74e5c2b1e505bb9a7ccd3ade46eb02b6c/daemon/internal/newrelic/harvest.go#L47
It will have access to the log labels that it can pass to NewLogEvents so can be set on the LogEvents struct: https://github.com/newrelic/newrelic-php-agent/blob/feat/apm_logging_labels/daemon/internal/newrelic/log_events.go#L19
It'll be available to add to all log events at that point.

protocol/flatbuffers/protocol.fbs
mfulb, line 122:
```suggestion
  log_forwarding_labels:  Event;   // added in the 11.7 PHP agent release
```

agent/php_txn.c
```diff
@@ -536,6 +536,117 @@ static nrobj_t* nr_php_txn_get_labels() {
...
+                  (void*)label_keys);
+
+  /* filter by going over the list of label keys, seeing if it exists in the
+   * exclude hash, and if it does skip it otherwise copy key/value for label
+   * to the log labels
+   */
+  log_labels = nro_new(NR_OBJECT_HASH);
+  for (int i = 0; i < nro_getsize(label_keys); i++) {
+    const char* key = nro_get_array_string(label_keys, i + 1, NULL);
+    const char* value = nro_get_hash_string(labels, key, NULL);
```
lavarou, line 620: The value is not needed until it is certain the label has not been filtered out. If it is filtered out, there's no need to retrieve it.

axiom/nr_txn.h
```diff
@@ -389,7 +393,8 @@ void nr_txn_enforce_security_settings(nrtxnopt_t* opts,
  */
 extern nrtxn_t* nr_txn_begin(nrapp_t* app,
                              const nrtxnopt_t* opts,
-                             const nr_attribute_config_t* attribute_config);
+                             const nr_attribute_config_t* attribute_config,
+                             const nrobj_t* log_forwarding_labels);
```
lavarou, line 397: These two arguments are not documented by doc string. I don't know if this is something we care about or not.

lavarou (approved):
Looks good considering all limitations of agent<->daemon/integration_runner design.

daemon/internal/newrelic/log_events_test.go
mfulb: We also have an integration test for this scenario:
tests/integration/logging/monolog2/test_monolog_labels_forwarding_basic_withspaces.php

agent/php_txn.c
mfulb: Due to the current way the integration test runner works I am leaving this implementation as is.

axiom/nr_txn.h
mfulb: Good point - I addressed this omission with https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/7a02270164facfc55fcdff042cd96e8d6bb5e164

agent/php_txn.c
mfulb: I've modified the iter callback to no store a `key` with a value of `NULL` so it should not be possible to  reach this section of code with a `NULL` value.
See https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/c13d6c02f56cab1a6e3c0a011ab2966220771fde

agent/php_txn.c
mfulb: I've added the NRSAFESTR macro just in case although with the change in https://github.com/newrelic/newrelic-php-agent/commit/c13d6c02f56cab1a6e3c0a011ab2966220771fde the value of `key` should not be `NULL` here.
See https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/a238c676c9e40e1361284eec0f890d2b8a6fe0e6

agent/php_txn.c
mfulb: If the exclude rule is empty we need to copy all the labels into the log labels structure so I don't see how we could exit early.

agent/php_txn.c
mfulb: Good observation - I have implemented this in https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/90955f063775140a7d031c1db2a118965d0ddc9b

agent/php_txn.c
mfulb: Thank you - I've made this change in https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/90955f063775140a7d031c1db2a118965d0ddc9b

agent/php_txn.c
mfulb: I've added a `NULL` check also for this loop which should protect against this situation:
https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/90955f063775140a7d031c1db2a118965d0ddc9b

protocol/flatbuffers/protocol.fbs
mfulb: Handled here https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/467c8a2f1bde998311d2633ba374edc5cddf58a8

mfulb:
> Thanks for adding the new tests_monolog_label* tests. I did a first look through and overall they look good. A few questions/requests:
> 
> 1. Could you add additional tests where:
> 
> * Labels list is not set, but exclude log labels are
> * Log forwarding disabled, but exclude log labels are set (I.e., not default)
> * Exclude label list includes a label that matches multiple case sensitive labels; expected behavior is they are all removed.
> * Exclude label list has same entry more than once
> 
> 2. Did you run the new integration/multiverse label tests with valgrind?
> 3. What is expected behavior if a label has a comma? Can you add a test case, so we can document the expected behavior that way?

I think commit https://github.com/newrelic/newrelic-php-agent/pull/1027/commits/65f9ba2949816c25dcda4b78163d2dc4e1ae0b21 will address #1 and #3.

I did run the integration tests (no multiverse tests for log labels (yet)) with valgrind.  A couple of CLM related tests passed valgrind but failed because an additional span ("name": "Custom\/Monolog\\Logger::pushHandler") was detected.  I'm suspecting this is because it ran slower with valgrind than w/o so it was included whereas it is not when valgrind is not used?

agent/php_txn.c
zsistla: Why do we need to copy at that point?
Can't we just return labels?

mfulb: Ok I misunderstood what you mean by "early exit".  Yes it would in theory be possible to reduce the work done in the case of no excludes.

zsistla: Multiple paths to a refactor being considered for future efforts.

zsistla (approved):

lavarou (approved):

zsistla (approved):
Thanks for the work on this!

mfulb:
Multiverse tests are failing because this PR adds a metric and any existing test with an `EXPECT_METRICS` will fail as they will not have this new metric as an expectation.
Running this PR against the corrected multiverse tests PASSED however.

PR #1029, opened 2025-02-24
style(agent): drupal code formatting

bduranleau-nr:


agent/fw_drupal8.c
lavarou, line 1: While you're at it, could you add `clang-format off/on` protection around these lines so that they don't break mid-word on auto-format:
https://github.com/newrelic/newrelic-php-agent/blob/1654b33d734cb911350972d86011bc3aeefecfc3/agent/fw_drupal8.c#L766-L767
https://github.com/newrelic/newrelic-php-agent/blob/1654b33d734cb911350972d86011bc3aeefecfc3/agent/fw_drupal8.c#L770-L771
https://github.com/newrelic/newrelic-php-agent/blob/1654b33d734cb911350972d86011bc3aeefecfc3/agent/fw_drupal8.c#L782-L783

agent/fw_drupal_common.h
```diff
@@ -147,23 +147,23 @@ nr_status_t module_invoke_all_parse_module_and_hook_from_strings(
...
- * Purpose: Before an invoke_all style call, adds the hook to that hook states stacks
+ * Purpose: Before an invoke_all style call, adds the hook to that hook states
+ * stacks
  *
- * Params  : 1. A zval holding a copy of the hook invoked, to be managed by the hook
- *              states stacks and freed by nr_drupal_invoke_all_hook_stacks_pop() after the
- *              invoke_all call completes
+ * Params  : 1. A zval holding a copy of the hook invoked, to be managed by the
+ * hook states stacks and freed by nr_drupal_invoke_all_hook_stacks_pop() after
+ * the invoke_all call completes
```
lavarou, line 158: This doesn't seem right. The same file shows `Purpose and `Params` docs strings, that span multiple lines, indented.

bduranleau-nr: I've manually fixed this here: fca6ace05bfc8294c8f017d6d77e743c6ed95ee4
For future reference, the reason for this is not due to a mis-configuration of clang-format. This is due to the line length limit (80 chars) and clang-format not being clever enough to match the indentation style in a docstring like this. These kinds of cases will require manual intervention.

lavarou (approved):

PR #1030, opened 2025-02-25
feat(agent): Drupal hook attribute instrumentation

bduranleau-nr:
Adds support for Drupal Attribute Hooks added in Drupal 11.1.

### DIFF
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -640,7 +640,6 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
           nrl_warning(NRL_FRAMEWORK,
                       "hookImplementationsMap[hook = %s]: invalid value",
                       NRSAFESTR(ZEND_STRING_VALUE(hook_key)));
-          nr_free(hookpath);
           return false;
         }
 
@@ -652,7 +651,6 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
             nrl_warning(NRL_FRAMEWORK,
                         "hookImplementationsMap[class = %s]: invalid value",
                         NRSAFESTR(ZEND_STRING_VALUE(class_key)));
-            nr_free(hookpath);
             return false;
           }
 
@@ -665,7 +663,6 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
               nrl_warning(NRL_FRAMEWORK,
                           "hookImplementationsMap[method = %s]: invalid value",
                           NRSAFESTR(ZEND_STRING_VALUE(method_key)));
-              nr_free(hookpath);
               return false;
             }
 
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -645,14 +645,15 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
   ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), hook_key,
                                 hook_val) {
     if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
-      nrl_warning(NRL_FRAMEWORK, "hookImplementationsMap[hook]: invalid value");
+      nrl_warning(NRL_FRAMEWORK,
+                  "hookImplementationsMap[hook]: invalid key or value");
       return false;
     }
 
     ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_val), class_key, class_val) {
       if ((NULL == class_key) || (0 == nr_php_is_zval_valid_array(class_val))) {
         nrl_warning(NRL_FRAMEWORK,
-                    "hookImplementationsMap[class]: invalid value");
+                    "hookImplementationsMap[class]: invalid key or value");
         return false;
       }
 
@@ -661,7 +662,7 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
         if ((NULL == method_key)
             || (0 == nr_php_is_zval_valid_string(module_val))) {
           nrl_warning(NRL_FRAMEWORK,
-                      "hookImplementationsMap[method]: invalid value");
+                      "hookImplementationsMap[method]: invalid key or value");
           return false;
         }
 
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -673,8 +673,9 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
         }
 
         if (0
-            == nr_stricmp(ZEND_STRING_VALUE(class_key),
-                          "Drupal\\Core\\Extension\\ProceduralCall")) {
+            == nr_striendswith(
+                ZEND_STRING_VALUE(class_key), ZEND_STRING_LEN(class_key),
+                NR_PSTR("Drupal\\Core\\Extension\\ProceduralCall"))) {
           hookpath = nr_formatf("%s", ZEND_STRING_VALUE(method_key));
         } else {
           hookpath = nr_formatf("%s::%s", ZEND_STRING_VALUE(class_key),
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -609,11 +609,122 @@ NR_PHP_WRAPPER_END
 #endif  // OAPI
 
 /*
+ * Purpose: Instrument Drupal Attribute Hooks for Drupal 11.1+
+ *
+ * Params: 1. A zval pointer to the moduleHandler instance in use by Drupal.
+ *
+ * Return: bool
+ *
+ */
+static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
+  zval* hook_implementation_map = NULL;
+
+  nr_php_string_hash_key_t* hook_key = NULL;
+  zval* hook_val = NULL;
+  nr_php_string_hash_key_t* class_key = NULL;
+  zval* class_val = NULL;
+  nr_php_string_hash_key_t* method_key = NULL;
+  zval* module_val = NULL;
+  zend_ulong key_num = 0;
+
+  char* hook_str = NULL;
+  char* class_str = NULL;
+  char* method_str = NULL;
+  char* module_str = NULL;
+  char* hookpath = NULL;
+
+  hook_implementation_map = nr_php_get_zval_object_property(
+      module_handler, "hookImplementationsMap");
+  if (hook_implementation_map) {
+    if (nr_php_is_zval_valid_array(hook_implementation_map)) {
+      ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), key_num,
+                                hook_key, hook_val) {
+        (void)key_num;
+        if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
+          nrl_warning(NRL_FRAMEWORK,
+                      "hookImplementationsMap[hook = %s]: invalid value",
+                      NRSAFESTR(ZEND_STRING_VALUE(hook_key)));
+          return false;
+        }
+
+        ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(hook_val), key_num, class_key,
+                                  class_val) {
+          (void)key_num;
+          if ((NULL == class_key)
+              || (0 == nr_php_is_zval_valid_array(class_val))) {
+            nrl_warning(NRL_FRAMEWORK,
+                        "hookImplementationsMap[class = %s]: invalid value",
+                        NRSAFESTR(ZEND_STRING_VALUE(class_key)));
+            return false;
+          }
+
+          ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(class_val), key_num, method_key,
+                                    module_val) {
+            (void)key_num;
+
+            nr_free(hook_str);
+            nr_free(class_str);
+            nr_free(method_str);
+            nr_free(module_str);
+            nr_free(hookpath);
+
+            if ((NULL == method_key)
+                || (0 == nr_php_is_zval_valid_string(module_val))) {
+              nrl_warning(NRL_FRAMEWORK,
+                          "hookImplementationsMap[method = %s]: invalid value",
+                          NRSAFESTR(ZEND_STRING_VALUE(method_key)));
+              return false;
+            }
+
+            hook_str = nr_strdup(ZEND_STRING_VALUE(hook_key));
+            class_str = nr_strdup(ZEND_STRING_VALUE(class_key));
+            method_str = nr_strdup(ZEND_STRING_VALUE(method_key));
+            module_str = nr_strdup(Z_STRVAL_P(module_val));
+
+            if (0
+                == nr_stricmp(ZEND_STRING_VALUE(class_key),
+                              "Drupal\\Core\\Extension\\ProceduralCall")) {
+              hookpath = nr_formatf("%s", method_str);
+            } else {
+              hookpath = nr_formatf("%s::%s", class_str, method_str);
+            }
+
+            nr_php_wrap_user_function_drupal(hookpath, nr_strlen(hookpath),
+                                             module_str, nr_strlen(module_str),
+                                             hook_str, nr_strlen(hook_str));
+          }
+          ZEND_HASH_FOREACH_END();
+        }
+        ZEND_HASH_FOREACH_END();
+      }
+      ZEND_HASH_FOREACH_END();
+
+    } else {
+      nrl_warning(NRL_FRAMEWORK,
+                  "hookImplementationsMap property not a valid array");
+      return false;
+    }
+  } else {
+    nrl_warning(NRL_FRAMEWORK, "NULL hookImplementationsMap object property");
+    return false;
+  }
+
+  nr_free(hook_str);
+  nr_free(class_str);
+  nr_free(method_str);
+  nr_free(module_str);
+  nr_free(hookpath);
+
+  return true;
+}
+
+/*
  * Purpose : Wrap the invoke() method of the module handler instance in use.
  */
 NR_PHP_WRAPPER(nr_drupal8_module_handler) {
   zend_class_entry* ce = NULL;
   zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
+  bool hook_attribute_instrumentation = false;
 
   NR_UNUSED_SPECIALFN;
   (void)wraprec;
@@ -635,20 +746,26 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
 
   ce = Z_OBJCE_P(*retval_ptr);
 
-  nr_drupal8_add_method_callback(ce, NR_PSTR("getimplementations"),
-                                 nr_drupal8_post_get_implementations TSRMLS_CC);
-  nr_drupal8_add_method_callback(ce, NR_PSTR("implementshook"),
-                                 nr_drupal8_post_implements_hook TSRMLS_CC);
-  /* Drupal 9.4 introduced a replacement method for getImplentations */
+  hook_attribute_instrumentation
+      = nr_drupal_hook_attribute_instrument(*retval_ptr);
+
+  if (!hook_attribute_instrumentation) {
+    nr_drupal8_add_method_callback(
+        ce, NR_PSTR("getimplementations"),
+        nr_drupal8_post_get_implementations TSRMLS_CC);
+    nr_drupal8_add_method_callback(ce, NR_PSTR("implementshook"),
+                                   nr_drupal8_post_implements_hook TSRMLS_CC);
+    /* Drupal 9.4 introduced a replacement method for getImplentations */
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
-  nr_drupal8_add_method_callback_before_after_clean(
-      ce, NR_PSTR("invokeAllWith"), nr_drupal94_invoke_all_with,
-      nr_drupal94_invoke_all_with_after, nr_drupal94_invoke_all_with_clean);
+    nr_drupal8_add_method_callback_before_after_clean(
+        ce, NR_PSTR("invokeallwith"), nr_drupal94_invoke_all_with,
+        nr_drupal94_invoke_all_with_after, nr_drupal94_invoke_all_with_clean);
 #else
-  nr_drupal8_add_method_callback(ce, NR_PSTR("invokeallwith"),
-                                 nr_drupal94_invoke_all_with TSRMLS_CC);
+    nr_drupal8_add_method_callback(ce, NR_PSTR("invokeallwith"),
+                                   nr_drupal94_invoke_all_with TSRMLS_CC);
 #endif
+  }
 }
 NR_PHP_WRAPPER_END
 
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -709,7 +709,6 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
 NR_PHP_WRAPPER(nr_drupal8_module_handler) {
   zend_class_entry* ce = NULL;
   zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
-  bool hook_attribute_instrumentation = false;
 
   NR_UNUSED_SPECIALFN;
   (void)wraprec;
@@ -731,26 +730,23 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
 
   ce = Z_OBJCE_P(*retval_ptr);
 
-  hook_attribute_instrumentation
-      = nr_drupal_hook_attribute_instrument(*retval_ptr);
-
-  if (!hook_attribute_instrumentation) {
-    nr_drupal8_add_method_callback(
-        ce, NR_PSTR("getimplementations"),
-        nr_drupal8_post_get_implementations TSRMLS_CC);
-    nr_drupal8_add_method_callback(ce, NR_PSTR("implementshook"),
-                                   nr_drupal8_post_implements_hook TSRMLS_CC);
-    /* Drupal 9.4 introduced a replacement method for getImplentations */
+  if (nr_drupal_hook_attribute_instrument(*retval_ptr)) {
+    NR_PHP_WRAPPER_LEAVE;
+  }
+  nr_drupal8_add_method_callback(ce, NR_PSTR("getimplementations"),
+                                 nr_drupal8_post_get_implementations TSRMLS_CC);
+  nr_drupal8_add_method_callback(ce, NR_PSTR("implementshook"),
+                                 nr_drupal8_post_implements_hook TSRMLS_CC);
+  /* Drupal 9.4 introduced a replacement method for getImplentations */
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
-    nr_drupal8_add_method_callback_before_after_clean(
-        ce, NR_PSTR("invokeallwith"), nr_drupal94_invoke_all_with,
-        nr_drupal94_invoke_all_with_after, nr_drupal94_invoke_all_with_clean);
+  nr_drupal8_add_method_callback_before_after_clean(
+      ce, NR_PSTR("invokeallwith"), nr_drupal94_invoke_all_with,
+      nr_drupal94_invoke_all_with_after, nr_drupal94_invoke_all_with_clean);
 #else
-    nr_drupal8_add_method_callback(ce, NR_PSTR("invokeallwith"),
-                                   nr_drupal94_invoke_all_with TSRMLS_CC);
+  nr_drupal8_add_method_callback(ce, NR_PSTR("invokeallwith"),
+                                 nr_drupal94_invoke_all_with TSRMLS_CC);
 #endif
-  }
 }
 NR_PHP_WRAPPER_END
 
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -608,7 +608,8 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
 NR_PHP_WRAPPER_END
 #endif  // OAPI
 
-static bool nr_validate_key_val_arr(nr_php_string_hash_key_t* key, zval* val) {
+static bool nr_is_invalid_key_val_arr(nr_php_string_hash_key_t* key,
+                                      zval* val) {
   if (NULL == key || 0 == ZEND_STRING_LEN(key)
       || 0 == nr_php_is_zval_valid_array(val)
       || 0 == zend_hash_num_elements(Z_ARRVAL_P(val))) {
@@ -649,14 +650,14 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
 
   ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), hook_key,
                                 hook_val) {
-    if (nr_validate_key_val_arr(hook_key, hook_val)) {
+    if (nr_is_invalid_key_val_arr(hook_key, hook_val)) {
       nrl_warning(NRL_FRAMEWORK,
                   "hookImplementationsMap[hook]: invalid key or value");
       return false;
     }
 
     ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_val), class_key, class_val) {
-      if (nr_validate_key_val_arr(class_key, class_val)) {
+      if (nr_is_invalid_key_val_arr(class_key, class_val)) {
         nrl_warning(NRL_FRAMEWORK,
                     "hookImplementationsMap[class]: invalid key or value");
         return false;
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -631,11 +631,6 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
   hook_implementation_map = nr_php_get_zval_object_property(
       module_handler, "hookImplementationsMap");
 
-  if (NULL == hook_implementation_map) {
-    nrl_warning(NRL_FRAMEWORK, "NULL hookImplementationsMap object property");
-    return false;
-  }
-
   if (!nr_php_is_zval_valid_array(hook_implementation_map)) {
     nrl_warning(NRL_FRAMEWORK,
                 "hookImplementationsMap property not a valid array");
--- a/agent/fw_drupal8.c
+++ b/agent/fw_drupal8.c
@@ -630,68 +630,66 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
 
   hook_implementation_map = nr_php_get_zval_object_property(
       module_handler, "hookImplementationsMap");
-  if (hook_implementation_map) {
-    if (nr_php_is_zval_valid_array(hook_implementation_map)) {
-      ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_implementation_map),
-                                    hook_key, hook_val) {
-        if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
-          nrl_warning(NRL_FRAMEWORK,
-                      "hookImplementationsMap[hook = %s]: invalid value",
-                      NRSAFESTR(ZEND_STRING_VALUE(hook_key)));
-          return false;
-        }
 
-        ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_val), class_key,
-                                      class_val) {
-          if ((NULL == class_key)
-              || (0 == nr_php_is_zval_valid_array(class_val))) {
-            nrl_warning(NRL_FRAMEWORK,
-                        "hookImplementationsMap[class = %s]: invalid value",
-                        NRSAFESTR(ZEND_STRING_VALUE(class_key)));
-            return false;
-          }
-
-          ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(class_val), method_key,
-                                        module_val) {
-            if ((NULL == method_key)
-                || (0 == nr_php_is_zval_valid_string(module_val))) {
-              nrl_warning(NRL_FRAMEWORK,
-                          "hookImplementationsMap[method = %s]: invalid value",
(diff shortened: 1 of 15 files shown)


agent/fw_drupal8.c
```diff
@@ -632,20 +748,26 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
...
+    nr_drupal8_add_method_callback_before_after_clean(
+        ce, NR_PSTR("invokeallwith"), nr_drupal94_invoke_all_with,
+        nr_drupal94_invoke_all_with_after, nr_drupal94_invoke_all_with_clean);
 #else
-  nr_drupal8_add_method_callback(ce, NR_PSTR("invokeallwith"),
-                                 nr_drupal94_invoke_all_with TSRMLS_CC);
+    nr_drupal8_add_method_callback(ce, NR_PSTR("invokeallwith"),
+                                   nr_drupal94_invoke_all_with TSRMLS_CC);
 #endif
+  }
```
lavarou, line 770: It would be easier on reviewers for the code to look like this:
```
  if (nr_drupal_hook_attribute_instrument(*retval_ptr)) {
    return;
  }
```
and leave lines 635-648 unchanged. Maybe the name `nr_drupal_hook_attribute_instrument` could be improved in this context.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+  char* hook_str = NULL;
+  char* class_str = NULL;
+  char* method_str = NULL;
+  char* module_str = NULL;
+  char* hookpath = NULL;
+
+  hook_implementation_map = nr_php_get_zval_object_property(
+      module_handler, "hookImplementationsMap");
+  if (hook_implementation_map) {
+    if (nr_php_is_zval_valid_array(hook_implementation_map)) {
```
lavarou, line 643: newrelic-php-agent's code base doesn't seem to have a standard around structuring the code. Golang has the following standard defined in [Effective Go](https://go.dev/doc/effective_go#if):

> [...] a common situation where code must guard against a sequence of error conditions. The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements, the resulting code needs no else statements.

The readability of the above could could definitely benefit from applying Effective Go standards. Consider:
```
  hooks_instrumented = false;
  hook_implementation_map = nr_php_get_zval_object_property(
      module_handler, "hookImplementationsMap");
  if (NULL == hook_implementation_map) {
    goto LEAVE;
  }
  if (!nr_php_is_zval_valid_array(hook_implementation_map)) {
    goto LEAVE;
  }
...
  // all checks passed, hooks have been instrumented:
  hooks_instrumented = true;
LEAVE:
  NR_FREE_HOOK_MEM
  return hooks_instrumented;
```

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+      ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), key_num,
+                                hook_key, hook_val) {
```
lavarou, line 645: If `key_num` is not needed, and `hook_key` is a string, consider using `ZEND_HASH_FOREACH_STR_KEY_VAL` iterator.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+        (void)key_num;
+        if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
+          nrl_warning(NRL_FRAMEWORK,
+                      "hookImplementationsMap[hook = %s]: invalid value",
+                      NRSAFESTR(ZEND_STRING_VALUE(hook_key)));
+          NR_FREE_HOOK_MEM
+          return false;
+        }
+
+        ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(hook_val), key_num, class_key,
```
lavarou, line 655: [This comment](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1972313208) applies here as well.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+          if ((NULL == class_key)
+              || (0 == nr_php_is_zval_valid_array(class_val))) {
+            nrl_warning(NRL_FRAMEWORK,
+                        "hookImplementationsMap[class = %s]: invalid value",
+                        NRSAFESTR(ZEND_STRING_VALUE(class_key)));
+            NR_FREE_HOOK_MEM
+            return false;
+          }
+
+          ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(class_val), key_num, method_key,
```
lavarou, line 667: [This comment](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1972313208) applies here as well.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+                                    module_val) {
+            (void)key_num;
+
+            NR_FREE_HOOK_MEM
```
lavarou, line 671: This is confusing. Why is it here and not after when it was used, i.e. after call to `nr_php_wrap_user_function_drupal`?

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+                          "hookImplementationsMap[method = %s]: invalid value",
+                          NRSAFESTR(ZEND_STRING_VALUE(method_key)));
+              NR_FREE_HOOK_MEM
+              return false;
+            }
+
+            hook_str = nr_strdup(ZEND_STRING_VALUE(hook_key));
+            class_str = nr_strdup(ZEND_STRING_VALUE(class_key));
+            method_str = nr_strdup(ZEND_STRING_VALUE(method_key));
+            module_str = nr_strdup(Z_STRVAL_P(module_val));
```
lavarou, line 685: There is no need for any of these `strdup`s:
- `ZEND_STRING_VALUE(class_key)` and `ZEND_STRING_VALUE(method_key)` can be simply passed as arguments to `nr_formatf` that construct `hookpath`
- `Z_STRVAL_P(module_val)` and `ZEND_STRING_VALUE(hook_key)` can be simply passed as arguments to `nr_php_wrap_user_function_drupal` which does `strdup`. As an added benefit, zend_strings have length associated with them so there's no need to call strlen for neither `module_val` nor `hook_key`.

lavarou: With the suggestion above, there's really no need for `NR_FREE_HOOK_MEM` macro to exist as well.

agent/fw_drupal8.c
zsistla: hook_implementation_map doesn't need to be checked for NULL here because nr_php_is_zval_valid_array (like all nr_php_is_zval_* funcs) also checks for NULL.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+  char* hookpath = NULL;
+
+  hook_implementation_map = nr_php_get_zval_object_property(
+      module_handler, "hookImplementationsMap");
+  if (hook_implementation_map) {
+    if (nr_php_is_zval_valid_array(hook_implementation_map)) {
+      ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), key_num,
+                                hook_key, hook_val) {
+        (void)key_num;
+        if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
```
zsistla, line 647: The extra NULL check isn't needed since nr_php_is_zval_valid_array will check for NULL already.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+          }
+
+          ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(class_val), key_num, method_key,
+                                    module_val) {
+            (void)key_num;
+
+            NR_FREE_HOOK_MEM
+
+            if ((NULL == method_key)
+                || (0 == nr_php_is_zval_valid_string(module_val))) {
```
zsistla, line 674: nr_php_is_zval_valid_string checks for NULL so extra check isn't needed.

agent/fw_drupal8.c
zsistla: Since it's a wrapper, it'd need to be:
```
  if (nr_drupal_hook_attribute_instrument(*retval_ptr)) {
    NR_PHP_WRAPPER_LEAVE;
  }
```

lavarou: I always forget that this code is in the context of `NR_PHP_WRAPPER`. Thanks for pointing it out!

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+                == nr_stricmp(ZEND_STRING_VALUE(class_key),
+                              "Drupal\\Core\\Extension\\ProceduralCall")) {
+              hookpath = nr_formatf("%s", method_str);
+            } else {
+              hookpath = nr_formatf("%s::%s", class_str, method_str);
+            }
+
+            nr_php_wrap_user_function_drupal(hookpath, nr_strlen(hookpath),
+                                             module_str, nr_strlen(module_str),
+                                             hook_str, nr_strlen(hook_str));
```
zsistla, line 697: Can't the memory be freed here right after this function is called?  So that is doesn't go through the loop again and strdup/nr_format values again that haven't had a chance to be freed?  If it's freed here right after it's been allocated, the NR_FREE_HOOK_MEM calls don't need to happen everywhere.

And if [this comment](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1972391782) is addressed, then NR_FREE_HOOK_MEM isn't needed, and the only thing that would be needed would be one call to nr_free(hookpath) after nr_php_wrap_user_function_drupal.

agent/fw_drupal8.c
bduranleau-nr: In order to guarantee the memory is free before trying to reassign. In a triple nested loop it's much safer to force free variables at the start than the end in my opinion.

agent/fw_drupal8.c
bduranleau-nr: I intentionally chose not to use `goto` here. In general I believe `goto` is a bad practice that interrupts the structure of code and adds complexity and possible side-effects.

agent/fw_drupal8.c
bduranleau-nr: I chose this over `ZEND_HASH_FOREACH_STR_KEY_VAL` because this is the better documented zend macro (outside of the agent code) and this paradigm is more common (`void key_num`) in the code. Anecdotally I also observed some strange errors when I first tried to use `STR_KEY_VAL` initially and just decided to go with what worked.

agent/fw_drupal8.c
bduranleau-nr: `nr_php_is_zval_valid_array` is checking the value of `hook_val`. The `NULL` check here is for `hook_key`. They are different variables and are not guaranteed to both be non-null.

agent/fw_drupal8.c
```diff
@@ -605,12 +605,128 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+                      NRSAFESTR(ZEND_STRING_VALUE(hook_key)));
+          NR_FREE_HOOK_MEM
+          return false;
+        }
+
+        ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(hook_val), key_num, class_key,
+                                  class_val) {
+          (void)key_num;
+          if ((NULL == class_key)
+              || (0 == nr_php_is_zval_valid_array(class_val))) {
```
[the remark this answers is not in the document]
bduranleau-nr: The checks are for different variables.

agent/fw_drupal8.c
bduranleau-nr: The checks are for different variables.

agent/fw_drupal8.c
bduranleau-nr: > So that is doesn't go through the loop again and strdup/nr_format values again that haven't had a chance to be freed?

How would that happen when the [first](https://github.com/newrelic/newrelic-php-agent/pull/1030/files?diff=unified&w=0#diff-6c636f6e3733aae919f4c4cf66ec5226065346587de8cbaf3918f483f9d0f523R671) operation in the loop is freeing the memory?

zsistla: Ah, true, but the first operation doesn't need to free the memory.
It can be simplified so that only hookpath needs to be freed, and that can be freed directly after using it.

If strdups are removed as suggested [here](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1972391782), then NR_FREE_HOOK_MEM isn't needed, and the only thing that would be needed would be one call to nr_free(hookpath) after nr_php_wrap_user_function_drupal.

agent/fw_drupal8.c
zsistla: If you remove the [strdups](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1972391782) and free the nr_formatted hookpath string right after it's used, these could be just `return`s to exit early because at that point, it won't need to call NR_FREE_HOOK_MEM everywhere.

agent/fw_drupal8.c
zsistla: Isn't it safer to free them immediately after t[hey are allocated and used and no longer needed after the call to `nr_php_wrap_user_function_drupal`](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#diff-6c636f6e3733aae919f4c4cf66ec5226065346587de8cbaf3918f483f9d0f523R682-R697)?  Why wait for it to loop through again and sprinkle calls to NR_FREE_HOOK_MEM throughout when you can just free [right after using](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1972498681)?

agent/fw_drupal8.c
lavarou: `goto` is a C language construct, like `switch` or `if`. And, like any construct, it can be used to improve code’s maintainability and readability but it can also be used to make the code a mess. I'm in favor of [Flattening Arrow Code](https://blog.codinghorror.com/flattening-arrow-code/). I'm not saying `goto` is always the way to go, but in this particular case, it would help.

agent/fw_drupal8.c
lavarou: > better documented

Better documented how? `ZEND_HASH_FOREACH_KEY_VAL` and `ZEND_HASH_FOREACH_STR_KEY_VAL` are defined in `Zend/zend_hash.h` for each PHP version newrelic-php-agent supports. This particular situation calls for  `ZEND_HASH_FOREACH_STR_KEY_VAL` not `ZEND_HASH_FOREACH_KEY_VAL` because key's index is not used. Both macros are used in newrelic-php-agent's code base - [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/agent/lib_monolog.c#L279) and [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/agent/php_api.c#L1377). However, the latter example has a [comment](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/agent/php_api.c#L1364-L1371) which explain why `ZEND_HASH_FOREACH_KEY_VAL` was used instead of `ZEND_HASH_FOREACH_STR_KEY_VAL`. The situation described in the comment is no longer applicable and `ZEND_HASH_FOREACH_STR_KEY_VAL` should be used going forward.

> Anecdotally I also observed some strange errors when I first tried to use STR_KEY_VAL initially and just decided to go with what worked.

What was the strange behavior you observed?

agent/fw_drupal8.c
```diff
@@ -632,20 +748,26 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
...
 
-  nr_drupal8_add_method_callback(ce, NR_PSTR("getimplementations"),
-                                 nr_drupal8_post_get_implementations TSRMLS_CC);
-  nr_drupal8_add_method_callback(ce, NR_PSTR("implementshook"),
-                                 nr_drupal8_post_implements_hook TSRMLS_CC);
-  /* Drupal 9.4 introduced a replacement method for getImplentations */
+  hook_attribute_instrumentation
+      = nr_drupal_hook_attribute_instrument(*retval_ptr);
+
+  if (!hook_attribute_instrumentation) {
```
lavarou, line 754: `nr_drupal_hook_attribute_instrument` will only return `true` if it's able to walk the entire `hookImplementationsMap`. What will happen if it fails mid way? Which hook instrumentation will be used? Does `nr_drupal_hook_attribute_instrument` cleanup its partial work if it fails to walk the entire `hookImplementationsMap`? Could you add a test for this case?

agent/fw_drupal8.c
bduranleau-nr: Changed here 3ded06d732c7cd7b4ee87b62f4d57a3dbbc60ef8

agent/fw_drupal8.c
bduranleau-nr: Addressed here 3ded06d732c7cd7b4ee87b62f4d57a3dbbc60ef8

agent/fw_drupal8.c
lavarou: > See if [3ded06d](https://github.com/newrelic/newrelic-php-agent/commit/3ded06d732c7cd7b4ee87b62f4d57a3dbbc60ef8) addresses your concern

See https://github.com/newrelic/newrelic-php-agent/pull/1030#discussion_r1973926151

agent/fw_drupal8.c
bduranleau-nr: If walking the map fails halfway through, we exit the function and return false. The wraprecs that are already created remain, and we revert back to the old method of wrapping hooks. 

> Which hook instrumentation will be used?

The answer would be "both". The attribute based instrumentation would persist for anything it managed to wrap, and the other methods act as a fallback to hopefully plug the gaps. I don't see anticipate any negative side-effects from this, do you?

> Could you add a test for this case?

This would be difficult and doesn't fit neatly into our established test paradigms, The dependency here is a lot of Drupal code, from `Drupal::moduleHandler` to the `hookImplementationsMap`. Is there a specific concern you have with this failing partway though walking the map?

agent/fw_drupal8.c
bduranleau-nr: 3c68c3d7063bf0c0915a551b9c68d289cbb78732

agent/fw_drupal8.c
bduranleau-nr: a68bd59f4757076ea81e722c16b5cb020eb8fd4c

agent/fw_drupal8.c
bduranleau-nr: Changed here 6b3d5b1536f40234b2de87bca522ed11a4637118
I can't reproduce the errors I was seeing the first time I tried using this, so I have to assume something else was at play when I was testing it.

agent/fw_drupal8.c
lavarou: Addressed with https://github.com/newrelic/newrelic-php-agent/commit/6b3d5b1536f40234b2de87bca522ed11a4637118

agent/fw_drupal8.c
lavarou: Addressed with https://github.com/newrelic/newrelic-php-agent/commit/6b3d5b1536f40234b2de87bca522ed11a4637118

agent/fw_drupal8.c
bduranleau-nr: Given the refactors, do you still want to see `goto` used to escape the loops, or is just `return`ing enough?

lavarou: Considering current code structure, it's ok to flatten/short-circuit nested `if`s with `return`.

bduranleau-nr: 42d5af5211cb76d57b49d057caf1cd6fc00b4546

agent/fw_drupal8.c
```diff
@@ -605,6 +605,92 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+    nrl_warning(NRL_FRAMEWORK,
+                "hookImplementationsMap property not a valid array");
+    return false;
+  }
+
+  ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), hook_key,
+                                hook_val) {
+    if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
+      nrl_warning(NRL_FRAMEWORK,
+                  "hookImplementationsMap[hook = %s]: invalid value",
```
lavarou, line 646: This can happen when `NULL == hook_key`, so more appropriate message would be:
```suggestion
                  "hookImplementationsMap[hook = %s]: invalid key or value",
```

agent/fw_drupal8.c
```diff
@@ -605,6 +605,92 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+                  NRSAFESTR(ZEND_STRING_VALUE(hook_key)));
```
lavarou, line 647: This will seg fault if `hook_key` is `NULL` - `ZEND_STRING_VALUE` dereferences pointer it is passed.

agent/fw_drupal8.c
lavarou: Given https://github.com/newrelic/newrelic-php-agent/pull/1030#discussion_r1974304375, I would feel more comfortable having a test with mocked implementation of  `Drupal\Core\Extension\ModuleHandlerInterface` that has corrupted `hookImplementationsMap` property in every possible way - invalid type of the property (not an array), first level has invalid key and value, second level has invalid key and value, third level has invalid key and value.

lavarou: Take a look [here](https://github.com/newrelic/newrelic-php-agent/blob/R11.6/tests/integration/frameworks/drupal/test_invoke_all_with.php#L109-L116) for inspiration how to mock implementation of `Drupal\Core\Extension\ModuleHandlerInterface` and use it in integration test.

agent/fw_drupal8.c
bduranleau-nr: The warnings have been fixed here d876e0e684483bae8ee442f7420476f0f4958d4a

agent/fw_drupal8.c
bduranleau-nr: `value` here was intended to be generic, as in "an invalid value was returned" and not the specifc map value for a key, but I've fixed it here ea0e4cfae60efa443919a75fd33a15d1aea8bc37

agent/fw_drupal8.c
```diff
@@ -605,6 +605,89 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+
+  nr_php_string_hash_key_t* hook_key = NULL;
+  zval* hook_val = NULL;
+  nr_php_string_hash_key_t* class_key = NULL;
+  zval* class_val = NULL;
+  nr_php_string_hash_key_t* method_key = NULL;
+  zval* module_val = NULL;
+
+  char* hookpath = NULL;
+
```
zsistla, line 627: Could check `module_handler` for NULL and exit early.

agent/fw_drupal8.c
```diff
@@ -605,6 +605,89 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+  hook_implementation_map = nr_php_get_zval_object_property(
+      module_handler, "hookImplementationsMap");
+
+  if (NULL == hook_implementation_map) {
+    nrl_warning(NRL_FRAMEWORK, "NULL hookImplementationsMap object property");
+    return false;
+  }
+
+  if (!nr_php_is_zval_valid_array(hook_implementation_map)) {
```
zsistla, line 636: `nr_php_is_zval_valid_array` will already check `hook_implementation_map` for NULL ; it doesn't need the extra check [right above](https://github.com/newrelic/newrelic-php-agent/pull/1030/files#diff-6c636f6e3733aae919f4c4cf66ec5226065346587de8cbaf3918f483f9d0f523R631).

agent/fw_drupal8.c
```diff
@@ -605,6 +605,89 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+                "hookImplementationsMap property not a valid array");
+    return false;
+  }
+
+  ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_implementation_map), hook_key,
+                                hook_val) {
+    if ((NULL == hook_key) || (0 == nr_php_is_zval_valid_array(hook_val))) {
+      nrl_warning(NRL_FRAMEWORK,
+                  "hookImplementationsMap[hook]: invalid key or value");
+      return false;
```
zsistla, line 647: Do we want to return here or just skip this round of the loop?
What if the first hook_key or hook_val is invalid, but the next one is good?

agent/fw_drupal8.c
```diff
@@ -605,6 +605,89 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+    }
+
+    ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hook_val), class_key, class_val) {
+      if ((NULL == class_key) || (0 == nr_php_is_zval_valid_array(class_val))) {
+        nrl_warning(NRL_FRAMEWORK,
+                    "hookImplementationsMap[class]: invalid key or value");
+        return false;
```
zsistla, line 654: Similar to https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1974532294, do we want to exit the whole function or just skip this round of the loop.

agent/fw_drupal8.c
```diff
@@ -605,6 +605,89 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+      }
+
+      ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(class_val), method_key,
+                                    module_val) {
+        if ((NULL == method_key)
+            || (0 == nr_php_is_zval_valid_string(module_val))) {
+          nrl_warning(NRL_FRAMEWORK,
+                      "hookImplementationsMap[method]: invalid key or value");
+          return false;
```
zsistla, line 663: https://github.com/newrelic/newrelic-php-agent/pull/1030/files#r1974532294
skip this round of the loop or completely exit?

agent/fw_drupal8.c
bduranleau-nr: I'd rather exit. If a key or value isn't valid, something is very wrong and there's no point continuing. The structure for this map is clearly and explicitly defined and managed by internal Drupal. If one is wrong, it's likely all of them are wrong.

agent/fw_drupal8.c
bduranleau-nr: `module_handler` is already checked for non-`NULL` in the calling function by `nr_php_object_instanceof_class`.

agent/fw_drupal8.c
bduranleau-nr: removed 2d515f1623b5ce74898dc2679b6ca4f0d16312af

agent/fw_drupal8.c
bduranleau-nr: Each example of a "bad" `hookImplementationsMap` would require it's own unique instance and definition of the `ModuleHandler` interface class, and all we'd really be testing is that the agent doesn't blow up. There isn't a return value we'd be able to check. I don't think it's appropriate to test this here- that's what our SOAK + valgrind and Multiverse tests are for.

lavarou: > Each example of a "bad" hookImplementationsMap would require it's own unique instance and definition of the ModuleHandler interface class, and all we'd really be testing is that the agent doesn't blow up.

That is exactly right - these are corner cases that need to be covered by the tests! None of the tests you mention cover any of these corner case scenarios.

bduranleau-nr: I disagree that those tests are necessary. Every value here is checked before it's used. If the `hook_implementation_map` isn't an array, the function will return false. If the `hook_key` is `NULL` or the `hook_val` is not a valid array, the function will return false. If `class_key` is `NULL` or `class_val` is not a valid array, the function will return false. If `method_key` is `NULL` or `module_val` is not a valid string, the function will return false. Every `nr_php_string_hash_key_t` variable is guaranteed by `ZEND_HASH_FOREACH_STR_KEY_VAL` to be  a `zend_string`. We've checked every single value we intend to use at every step in the process before we attempt to use it.

Mocking Drupal tests gives us a false view of the state of agent support. We already know that our mocked Drupal tests for `page_cache` are objectively wrong, but it doesn't trigger any errors. Even if done correctly, we open ourselves up to bitrot and the risks of using stale mocked framework code to verify our instrumentation.

If you would really like these tests incorporated, I'll add them. I'm just not seeing the value at this moment.

bduranleau-nr: tests: 5a902824771112b41a060f455ffd74847a10c608

lavarou: This looks good. Could you add one positive test case to show what the agent expects `hookImplementationsMap` to look like? It will be easier to look at this mocked example than try to reconstruct it based on the C code or digging through Drupal's codebase.

agent/fw_drupal8.c
zsistla: How does it look internally if hookkey exists but has no classes associated with it yet?  Will hookval be an empty array or NULL?

zsistla: Yes, but my question is, if a hook exists, but there are no classes associated with it, internally, will the hook value be NULL or will it be an empty array, or is hook never added to the map if there are no classes associated with it?

agent/fw_drupal8.c
bduranleau-nr: 0836487a42ccc5fae16f9d4e267f0651931965ec

tests/integration/frameworks/drupal/test_hook_implementations_map_valid.php
```diff
@@ -0,0 +1,35 @@
...
+newrelic.framework = drupal8
+*/
+
+/*EXPECT_TRACED_ERRORS null */
+
+/*EXPECT_ERROR_EVENTS null */
+
+/*EXPECT
+*/
+
```
zsistla, line 21: All of the EXPECT blocks for the valid/invalid tests are the same.

`nr_php_user_wraprec_create_named` creates a supportability metric like `Supportability/InstrumentedFunction/{ClassName}:{func_name}`
`nr_php_user_wraprec_create_named`  is called from `nr_php_add_custom_tracer_named`
`nr_php_add_custom_tracer_named` is called from `nr_php_wrap_user_function`
`nr_php_wrap_user_function` is called from `nr_php_wrap_user_function_drupal`

So, if we are calling nr_php_wrap_user_function_drupal here https://github.com/newrelic/newrelic-php-agent/pull/1030/files#diff-6c636f6e3733aae919f4c4cf66ec5226065346587de8cbaf3918f483f9d0f523R680
So if we are handling a valid hookImplementation map, shouldn't we have successfully wrapped some things and expect to see some Supportability/InstrumentedFunction metrics in an `EXPECT_METRICS_EXIST` block?

Similarly we can check for the existence/non-existence of the metrics in the other test cases.

bduranleau-nr: The `EXPECT` blocks are all the same because they are only there to ensure there isn't any unexpected output that would indicate an error has occurred. 

If we add the expectation for InstrumentedFunction, the only recognized metric returned is 
`"name": "Supportability/InstrumentedFunction/Drupal::moduleHandler"`. I believe this is because the Supportability metric isn't generated when the function is wrapped, it's generated when the wrapped function is called. `nr_txn_create_fn_supportability_metric` is only responsible for generating the `InstrumentedFunction` name, which then gets assigned to the `wraprec->supportability_metric` field for use when called. All of the hook/class/method names are mocked in these tests, and do not actually exist, so aren't callable. 

For `InstrumentedFunction` metrics, we're better off using Multiverse tests. These tests are not intended to perfectly replicate Drupal behavior- it's just to make sure we don't crash when parsing this map.

zsistla: Heading over to look at the multiverse tests now.

tests/integration/frameworks/drupal/mock_module_handler_empty_array.php
```diff
@@ -0,0 +1,34 @@
...
+/* Verify agent behavior when key value is an empty array */
+
+namespace Drupal\Core\Extension {
+  interface ModuleHandlerInterface
+  {
+    public function invokeAllWith($hook_str, $callback);
+  }
+  class ModuleHandler implements ModuleHandlerInterface
+  {
+    protected array $hookImplementationsMap = array(
```
zsistla, line 16: In the test run actions, the PHP 7.2 and 7.3 test runs are failing due to some PHP language change around this.  Since this is only mocking 11.1 structure anyway, those PHPs can be excluded in a skipif.

bduranleau-nr: typed properties weren't introduced until [7.4](https://php.watch/versions/7.4/typed-properties)

fixed here 2481b5be246166c69a9c80ac97899151ee9b39fb

bduranleau-nr: Technically we could set the skipif to 8.3+ since that's all that 11.1+ supports. I'm torn- I want to make sure we aren't impacting the prior Drupal hook instrumentation methodology, but this map won't exist for less than PHP 8.3

zsistla: No strong opinions on this one.

For the mocked versions, either way is fine.  It's okay to leave for php 7.4+ since it's just a mock and doesn't really run, as you mentioned it just verifies that the structure parsing works and handles various valid/invalid values.

And for multiverse, you'll already be excluding PHPs based on actual drupal code and the PHPs it supports, so it should be fine.

agent/fw_drupal8.c
```diff
@@ -605,6 +605,94 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
 NR_PHP_WRAPPER_END
 #endif  // OAPI
 
+static bool nr_validate_key_val_arr(nr_php_string_hash_key_t* key, zval* val) {
```
zsistla, line 608: The name made me think valid values would return true, and invalid values would return false.

lavarou: I agree with @zsistla's observation.

agent/fw_drupal8.c
```diff
@@ -605,6 +605,94 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+  zval* module_val = NULL;
+
+  char* hookpath = NULL;
+
+  hook_implementation_map = nr_php_get_zval_object_property(
+      module_handler, "hookImplementationsMap");
+
+  if (!nr_php_is_zval_valid_array(hook_implementation_map)) {
+    nrl_warning(NRL_FRAMEWORK,
+                "hookImplementationsMap property not a valid array");
```
lavarou, line 643: How frequently will this message be logged with agent's default log level when running Drupal < 11.1?

agent/fw_drupal8.c
```diff
@@ -605,6 +605,94 @@ NR_PHP_WRAPPER(nr_drupal94_invoke_all_with_clean) {
...
+            || 0 == nr_php_is_zval_valid_string(module_val)) {
+          nrl_warning(NRL_FRAMEWORK,
+                      "hookImplementationsMap[method]: invalid key or value");
+          return false;
+        }
+
+        if (0
+            == nr_stricmp(ZEND_STRING_VALUE(class_key),
+                          "Drupal\\Core\\Extension\\ProceduralCall")) {
+          hookpath = nr_formatf("%s", ZEND_STRING_VALUE(method_key));
```
lavarou, line 674: What is more likely:
- `class_key` starting with `Drupal\Core`
- `class_key` ending with `Drupal\Core\Extension\ProceduralCall`?

Or put it in a different way: how often will `nr_stricmp` have to scan through `Drupal\Core` only to fail on the next character? Is it possible to optimize this check so that it fails sooner than later?

lavarou: I wonder if using `nr_striendswith(ZEND_STRING_VALUE(class_key), ZEND_STRING_LEN(class_key), NR_PSTR("Drupal\\Core\\Extension\\ProceduralCall"))` wouldn’t be more optimal. Worst case, `nr_striendswith` will still have to scan the whole string. Best case, `nr_striendswith` will fail fast on length comparison.

zsistla: I agree with @lavarou 's suggestion; that's a good optimization.

agent/fw_drupal8.c
zsistla: Does this need to be warning vs something like verbosedebug?

agent/fw_drupal8.c
bduranleau-nr: 315ecbb1b8d7e0687d239eaa84b03eb8816ccbcf

agent/fw_drupal8.c
bduranleau-nr: fda25fe30b931b180adb4472eb8363aab24e47fb

agent/fw_drupal8.c
bduranleau-nr: fda25fe30b931b180adb4472eb8363aab24e47fb

lavarou: I think you meant 977e9efcdc1f42e42c346276778942b819b8523f, which looks good to me.

agent/fw_drupal8.c
```diff
@@ -643,7 +736,7 @@ NR_PHP_WRAPPER(nr_drupal8_module_handler) {
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
   nr_drupal8_add_method_callback_before_after_clean(
-      ce, NR_PSTR("invokeAllWith"), nr_drupal94_invoke_all_with,
+      ce, NR_PSTR("invokeallwith"), nr_drupal94_invoke_all_with,
```
lavarou, line 739: Rebase issue. Case of `invokeAllWith` must not be changed by this PR.

bduranleau-nr:
```suggestion
      ce, NR_PSTR("invokeAllWith"), nr_drupal94_invoke_all_with,
```

zsistla (approved):

lavarou (approved):

PR #1037, opened 2025-03-14
feat(agent): Add AWS DynamoDb Instrumentation

zsistla:
Added functionality to instrument
```
Call	                 collection(tablename)	operation	product

Create Table	TableName parameter	create_table	DynamoDB
Delete Item	TableName parameter	delete_item	DynamoDB
Delete Table	TableName parameter	delete_table	DynamoDB
Get Item	TableName parameter	get_item	DynamoDB
Put Item	TableName parameter	put_item	DynamoDB
Query	        TableName parameter	query	        DynamoDB
Scan	        TableName parameter	scan	        DynamoDB
Update Item	TableName parameter	update_item	DynamoDB
```

Unit tests added.
See also relevant multiverse PR.
See also relevant soak test PR.

Needed a modification to axiom to:
* add db.system
* modify datastore to allow for datastore that don't have concept of database name


agent/lib_aws_sdk_php.c
```diff
@@ -14,17 +14,16 @@
 #include "fw_hooks.h"
 #include "fw_support.h"
 #include "util_logging.h"
-#include "nr_segment_message.h"
-#include "nr_segment_external.h"
```
bduranleau-nr, line 18: Why the removal of these 2 includes?

zsistla: moved here: https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-25fbc57db8661cdcd450cbe94214a9df48b6bad7c768af9ae1cdf2e01c7af0beR10-R14

agent/lib_aws_sdk_php.c
```diff
@@ -519,6 +518,172 @@ char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
...
+  }
+
+  this_obj = NR_PHP_USER_FN_THIS();
+  func = nr_php_execute_function(NR_EXECUTE_ORIG_ARGS);
+
+  if (NULL == this_obj || NULL == func) {
+    return;
+  }
+
+  if (NULL != func->common.scope) {
```
bduranleau-nr, line 546: if `func->common.scope` is equal to `NULL`, shouldn't we exit early here? A `NULL` `base_class` arg fed to `nr_php_get_zval_object_property_with_class` will result in a `NULL` return value from that function each time it's called.

zsistla: No, we shouldn't exit early.  This value is just needed to get region and endpoint/port, but we can still create a message segment without those since region isn't required and we know the endpoint/port defaults and can use those.

bduranleau-nr: That creates a pretty complex codepath that's difficult to manually trace in the rest of this function in that case- for example, `region_zval` [here](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R550) will be `NULL`, which will result in `cloud_attrs->cloud_region` [here](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R553) to not be set, `endpoint_zval` [here](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R556) will be `NULL`, resulting in this whole `if` [block](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R558-R575) being skipped, etc. Can we separate the logic here more explicitly into a path where the `base_class` is `NULL` vs where it isn't to take the guesswork out? Looking briefly at the `test_lib_aws_sdk_php.c` tests, I can't find where this scenario is explicitly tested. Do we have unit tests for the case where `base_class` is `NULL`?

bduranleau-nr: Based on my understanding of the current code everything in this [block](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R550-R575) will be skipped if `func->common.scope` is `NULL`, so it probably makes sense to include all of that in the body of this `if` statement for clarity.

zsistla: I changed the the scope of the if statement to improve clarity of what is impacted by the value:
[29627f0](https://github.com/newrelic/newrelic-php-agent/pull/1037/commits/29627f067b432743506b53aeeb6c423f174de678)

However, there isn't a test for explicitly testing if `func->common.scope` is NULL.

We're instrumenting a function of a class which means PHP has stored the class entry in `func->common.scope`.  By the time we get here, we've already checked that we have a [valid object and func](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R542) so if it is valid, `func->common.scope`.  There is not a valid case where obj if valid and func is valid but common.scope is null.

To change this, but keep all other parts (like object and func) valid in order to hit this if statement, in the unit test I'd have to manually go into the zend_execute_data and change that value to NULL.  It doesn't seem meaningful, but if you feel strongly about the coverage for this case, I could add the manual step to break zend_execute_data so that `func->common.scope` is NULL.

bduranleau-nr: Let me think on it while I go through the rest of the changes and get back to you

agent/lib_aws_sdk_php.h
```diff
@@ -8,6 +8,11 @@
 #define LIB_AWS_SDK_PHP_HDR
 
 #include "nr_segment_message.h"
+#include "nr_segment_external.h"
+#include "nr_datastore_instance.h"
+#include "nr_segment_datastore.h"
+#include "nr_segment_datastore_private.h"
+
```
bduranleau-nr, line 15: my clang formatter removes this extra newline
```suggestion
```

agent/lib_aws_sdk_php.h
```diff
@@ -111,6 +136,40 @@ void nr_aws_sdk_mshutdown(void);
...
+             Note: A host/port other than the AWS DynamoDb default is very
+ uncommon and used only for development environments
+ *
+ * Params  : 1. datastore_params : datastore params to modify
+ *           2. cloud_attrs : cloud attributes to modify
+ *           3. NR_EXECUTE_ORIG_ARGS (execute_data, func_return_value)
+ * Returns :
+ * Note: caller is responsible for freeing the cloud_attrs->cloud_resource_id
+ and datastore_params->collection and datastore_params.instance.port_path_or_id
+ */
```
bduranleau-nr, line 167: It looks like the formatter did a number on this section. I tried to fix it on my machine but you may want to double check it yourself:
```suggestion
/*
 * Purpose : Populate DynamoDbClient datastore_params and cloud_attrs.  This
 *           will extract region and host/port and set the value in the
 *           appropriate struct.
 *
 *           Note: A host/port other than the AWS DynamoDb default is very
 *           uncommon and used only for development environments
 *
 * Params  : 1. datastore_params : datastore params to modify
 *           2. cloud_attrs : cloud attributes to modify
 *           3. NR_EXECUTE_ORIG_ARGS (execute_data, func_return_value)
 * Returns :
 * Note: caller is responsible for freeing the cloud_attrs->cloud_resource_id
 *       and datastore_params->collection and
 *       datastore_params.instance.port_path_or_id
 */

```

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -61,24 +61,83 @@ NR_PHP_WRAPPER(expect_arg_value_null) {
 }
 NR_PHP_WRAPPER_END
 
+NR_PHP_WRAPPER(aws_dynamodb_set_params_wrapper) {
+  // (void)wraprec;
```
bduranleau-nr, line 65: My editor complains about an unused `wraprec` parameter in the `NR_PHP_WRAPPER` call
```suggestion
  (void)wraprec;
```

agent/lib_aws_sdk_php.c
```diff
@@ -309,9 +308,7 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
   nr_segment_t* external_segment = NULL;
   zval** retval_ptr = NR_GET_RETURN_VALUE_PTR;
 
-  nr_segment_cloud_attrs_t cloud_attrs = {
-      .cloud_platform = "aws_lambda"
-  };
+  nr_segment_cloud_attrs_t cloud_attrs = {.cloud_platform = "aws_lambda"};
```
mfulb, line 311: Formatting change?

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -61,24 +61,83 @@ NR_PHP_WRAPPER(expect_arg_value_null) {
...
+  zval* test_name = NULL;
+  zval* expected = NULL;
+
+  datastore_params.instance = &instance;
+
+  /*
+   * argument 1 is is used to pass in the test name
+   */
+
+  test_name = nr_php_get_user_func_arg(3, NR_EXECUTE_ORIG_ARGS);
```
bduranleau-nr, line 80: Is this supposed to take `1` as an argument instead of `3`?

agent/lib_aws_sdk_php.c
```diff
@@ -332,7 +329,8 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
   /* Determine if we instrument this command. */
   if (AWS_COMMAND_IS("invoke")) {
     /* reconstruct the ARN */
-    nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_ORIG_ARGS, &cloud_attrs);
+    nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_ORIG_ARGS,
+                                               &cloud_attrs);
```
mfulb, line 333: Formatting change?

agent/lib_aws_sdk_php.c
```diff
@@ -390,7 +387,9 @@ void nr_aws_sdk_mshutdown(void) {
   nr_regex_destroy(&aws_arn_regex);
 }
 
-void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs) {
+void nr_aws_sdk_lambda_client_invoke_parse_args(
+    NR_EXECUTE_PROTO,
+    nr_segment_cloud_attrs_t* cloud_attrs) {
```
mfulb, line 392: Formatting change?

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -61,24 +61,83 @@ NR_PHP_WRAPPER(expect_arg_value_null) {
...
+      NRSAFESTR(datastore_params.collection));
+
+  if (nr_php_is_zval_valid_string(expected)) {
+    expected_string = Z_STRVAL_P(expected);
+  }
+
+  if (nr_php_is_zval_valid_string(test_name)) {
+    test_name_string = Z_STRVAL_P(test_name);
+  } else {
+    test_name_string = "Parmas should match expected.";
```
bduranleau-nr, line 112:
```suggestion
    test_name_string = "Params should match expected.";
```

agent/tests/test_lib_aws_sdk_php.c
```diff
@@ -483,20 +554,397 @@ static void test_nr_lib_aws_sdk_php_handle_version(void) {
...
+      "private ?endpoint_class $endpoint;"
+      "function base_func($command, $args, $expects) {return;}"
+      "function __construct(?string $region = null, ?int $port = null, ?string "
+      "$host = null) {"
+      "$this->region = $region;"
+      "$this->endpoint = new endpoint_class($port, $host);"
+      "}"
+      "}"
+      "class top_class extends base_class {"
+      "}");
```
bduranleau-nr, line 578: suggestion to make this construct a little more human readable:
```suggestion
  // clang-format off
  const char* endpoint_class = 
      "class endpoint_class {"
        "public ?string $host;"
        "public ?int $port;"
        "function __construct(?int $port = null, ?string $host = null) {"
          "$this->host = $host;"
          "$this->port = $port;"
        "}"
      "}"
      "class base_class {"
        "private ?string $region;"
        "private ?endpoint_class $endpoint;"
        "function base_func($command, $args, $expects) {return;}"
        "function __construct(?string $region = null, ?int $port = null, ?string $host = null) {"
          "$this->region = $region;"
          "$this->endpoint = new endpoint_class($port, $host);"
        "}"
      "}"
      "class top_class extends base_class {}";
  // clang-format on
  tlib_php_request_eval(endpoint_class);
```

agent/tests/test_lib_aws_sdk_php.c
zsistla: Good catch,
please see: [efd8799](https://github.com/newrelic/newrelic-php-agent/pull/1037/commits/efd87996476535e265a3d9b51053c23f518f4c20)

agent/lib_aws_sdk_php.c
```diff
@@ -519,6 +518,171 @@ char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
...
+    datastore_params->collection = table_name;
+  }
+  if (!nr_strempty(NRINI(aws_account_id))) {
+    account_id = NRINI(aws_account_id);
+  }
+
+  if (NULL != datastore_params->collection && NULL != account_id
+      && NULL != cloud_attrs->cloud_region) {
+    /* Must be freed by caller */
+    cloud_attrs->cloud_resource_id = nr_formatf(
```
mfulb, line 597: What would be the best behavior here if `cloud_attrs->cloud_resource_id` was non-NULL?

agent/lib_aws_sdk_php.c
```diff
@@ -519,6 +518,171 @@ char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
...
+        /* Only try to get a port if we have a valid host. */
+        port_zval = nr_php_get_zval_object_property(endpoint_zval, "port");
+        if (nr_php_is_zval_valid_integer(port_zval)) {
+          /* Must be freed by caller */
+          datastore_params->instance->port_path_or_id
+              = nr_formatf(NR_INT64_FMT, Z_LVAL_P(port_zval));
+        } else {
+          /* In case where host was found but port was not, spec says return
+           * unknown for port. */
+          datastore_params->instance->port_path_or_id = nr_strdup("unknown");
```
mfulb, line 571: Would this just be the default port if unset?

agent/lib_aws_sdk_php.c
zsistla, line 311: Format only

agent/lib_aws_sdk_php.c
bduranleau-nr: So the idea here is that we're only ever instrumenting `Aws\DynamoDb\DynamoDbClient` with this function so `func->common.scope` will always have a scope of `DynamoDbClient` and can therefore never be `NULL` by the time we've reached this codepath?

agent/lib_aws_sdk_php.c
zsistla: No because we were able to extract a host just not a port.  Once we've extracted a host, we cannot assume a port to go with it.  Default port only ever pairs with default host.

agent/lib_aws_sdk_php.c
zsistla: Best behavior?  if non-NULL it gets passed back to caller who uses it to end the segment before freeing it.

agent/lib_aws_sdk_php.c
```diff
@@ -519,6 +518,171 @@ char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
...
+  } else if (AWS_COMMAND_IS("putItem")) {
+    datastore_params.operation = AWS_SDK_PHP_DYNAMODBCLIENT_PUT_ITEM;
+  } else if (AWS_COMMAND_IS("query")) {
+    datastore_params.operation = AWS_SDK_PHP_DYNAMODBCLIENT_QUERY;
+  } else if (AWS_COMMAND_IS("scan")) {
+    datastore_params.operation = AWS_SDK_PHP_DYNAMODBCLIENT_SCAN;
+  } else if (AWS_COMMAND_IS("updateItem")) {
+    datastore_params.operation = AWS_SDK_PHP_DYNAMODBCLIENT_UPDATE_ITEM;
+  } else {
+    /* Nothing to do here so exit. */
```
bduranleau-nr, line 649: Is there a default value that `datastore_params.operation` should be set to here to indicate that it fell through the expected cases? Or should this field just be `NULL` in that case? Do we need to explicitly set to `NULL` if that's the case?

zsistla: No because in this case we directly exit and no datastore segment is started or ended, so datastore_params (which is a local var) is not used and ignored.

agent/lib_aws_sdk_php.c
```diff
@@ -371,14 +369,13 @@ void nr_lib_aws_sdk_php_lambda_handle(nr_segment_t* auto_segment,
...
     }
-    
   }
   nr_segment_external_end(&external_segment, &external_params);
   nr_free(cloud_attrs.cloud_resource_id);
 }
 
-/* This stores the compiled regex to parse AWS ARNs. The compilation happens when
- * it is first needed and is destroyed in mshutdown
+/* This stores the compiled regex to parse AWS ARNs. The compilation happens
```
zsistla, line 377: format only

agent/lib_aws_sdk_php.c
zsistla, line 392: format only

agent/lib_aws_sdk_php.c
```diff
@@ -409,7 +408,8 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
   if (!nr_php_is_zval_valid_array(lambda_args)) {
     return;
   }
-  zval* lambda_name = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
+  zval* lambda_name
+      = nr_php_zend_hash_find(Z_ARRVAL_P(lambda_args), "FunctionName");
```
zsistla, line 412: format only

agent/lib_aws_sdk_php.c
```diff
@@ -420,10 +420,8 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
   }
 
   /* Extract all information possible from the passed lambda name via regex */
-  nr_regex_substrings_t* matches =
-      nr_regex_match_capture(aws_arn_regex,
-                             Z_STRVAL_P(lambda_name),
-                             Z_STRLEN_P(lambda_name));
+  nr_regex_substrings_t* matches = nr_regex_match_capture(
+      aws_arn_regex, Z_STRVAL_P(lambda_name), Z_STRLEN_P(lambda_name));
```
zsistla, line 424: format only

agent/lib_aws_sdk_php.c
```diff
@@ -449,11 +447,12 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
   }
   if (nr_strempty(region)) {
     zend_class_entry* base_class = NULL;
-    if (NULL != execute_data->func && NULL!= execute_data->func->common.scope) {
-       base_class = execute_data->func->common.scope;
+    if (NULL != execute_data->func
+        && NULL != execute_data->func->common.scope) {
+      base_class = execute_data->func->common.scope;
```
zsistla, line 452: format only

agent/lib_aws_sdk_php.c
```diff
@@ -467,11 +466,11 @@ void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_clo
...
     if (!nr_strempty(qualifier)) {
-      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s:%s",
-                       region, accountID, function_name, qualifier);
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s:%s", region, accountID,
+                       function_name, qualifier);
     } else {
-      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s",
-                       region, accountID, function_name);
+      arn = nr_formatf("arn:aws:lambda:%s:%s:function:%s", region, accountID,
+                       function_name);
```
zsistla, line 473: format only

agent/lib_aws_sdk_php.c
zsistla: While that is true, it's more than that.  Any PHP obj with a function will ALWAYS have the scope in `func->common.scope`, so if we've already verified we have an obj and the obj has a func, `func->common.scope` will exist.  There's no class/func that we could create to test this case. The only way to test a NULL in common.scope that has already passed the non-null obj and non-null func test would be to take the zend_execute_data created when the function is invoked and break it by manually setting that field to NULL prior to passing it to `nr_lib_aws_sdk_php_dynamodb_set_params`

bduranleau-nr: I'm good with not adding a test for this case

bduranleau-nr (dismissed):

agent/lib_aws_sdk_php.c
mfulb: I am curious if this is an expected situation so it needs to be handled to not leak memory, or if it is not expected then what is the best course?

agent/lib_aws_sdk_php.c
```diff
@@ -519,6 +518,171 @@ char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
...
+                                         NR_EXECUTE_ORIG_ARGS);
+
+  /*
+   * By this point, the datastore params are decoded, grab the parent segment
+   * start time, add the special segment attributes/metrics then close the newly
+   * created segment.
+   */
+  datastore_segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  if (NULL == datastore_segment) {
+    return;
```
mfulb, line 670: Will the parameters mentioned in the comment just above going to be freed properly after this `return`?

agent/lib_aws_sdk_php.c
```diff
@@ -519,6 +518,171 @@ char* nr_lib_aws_sdk_php_get_command_arg_value(char* command_arg_name,
...
+
+  if (NULL == command_name_string || 0 == command_name_len) {
+    return;
+  }
+
+#define AWS_COMMAND_IS(CMD) \
+  (command_name_len == (sizeof(CMD) - 1) && nr_streq(CMD, command_name_string))
+
+  /* Determine if we instrument this command. */
+  if (AWS_COMMAND_IS("createTable")) {
```
mfulb, line 632: There are defines for all of these strings - can we use them instead of having the literal value duplicated in two places?

agent/lib_aws_sdk_php.h
```diff
@@ -87,7 +110,9 @@ extern void nr_lib_aws_sdk_php_sqs_handle(nr_segment_t* segment,
  *
  * Note: The caller is responsible for freeing cloud_attrs->cloud_resource_id
  */
-void nr_aws_sdk_lambda_client_invoke_parse_args(NR_EXECUTE_PROTO, nr_segment_cloud_attrs_t* cloud_attrs);
+void nr_aws_sdk_lambda_client_invoke_parse_args(
+    NR_EXECUTE_PROTO,
+    nr_segment_cloud_attrs_t* cloud_attrs);
```
mfulb, line 115: This appears to be a formatting only change.

agent/lib_aws_sdk_php.h
```diff
@@ -111,6 +136,44 @@ void nr_aws_sdk_mshutdown(void);
...
+ *           Note: A host/port other than the AWS DynamoDb default is very
+ *           uncommon and used only for development environments
+ *
+ * Params  : 1. datastore_params : datastore params to modify
+ *           2. cloud_attrs : cloud attributes to modify
+ *           3. NR_EXECUTE_ORIG_ARGS (execute_data, func_return_value)
+ * Returns :
+ * Note: caller is responsible for freeing the cloud_attrs->cloud_resource_id
+ *       and datastore_params->collection and
+ *       datastore_params.instance.port_path_or_id
```
mfulb, line 169:
```suggestion
 * Note: caller is responsible for freeing:
 *       - cloud_attrs->cloud_resource_id
 *       - datastore_params->collection
 *       - datastore_params.instance.port_path_or_id
```

axiom/nr_segment_datastore.c
```diff
@@ -208,9 +220,9 @@ bool nr_segment_datastore_end(nr_segment_t** segment_ptr,
    * The allWeb and allOther rollup metrics are created at the end of the
    * transaction since the background status may change.
    */
-  scoped_metric
-      = create_metrics(segment, duration, datastore_string, collection,
-                       operation, &datastore, params->instance);
+  scoped_metric = create_metrics(segment, duration, datastore_string,
+                                 params->datastore.type, collection, operation,
+                                 &datastore, params->instance);
```
mfulb, line 225: Formatting change only possibly?

axiom/nr_segment_datastore.h
```diff
@@ -22,6 +22,7 @@ typedef struct _nr_segment_datastore_params_t {
                        extracted from the SQL for SQL segments. */
   char* operation;  /* The null-terminated operation; if NULL, this will be
                        extracted from the SQL for SQL segments. */
+  char* db_system;  /* Database management system (DBMS) product being used.*/
```
mfulb, line 25: Is this only currently populated for DynamobDB?
Do we need to add support for existing DBMS as well in another effort?

agent/lib_aws_sdk_php.c
zsistla: Good catch, please see: [02c00c9](https://github.com/newrelic/newrelic-php-agent/pull/1037/commits/02c00c96d55cbddc1e8b71c2cfcc3b56f343bba7)

agent/lib_aws_sdk_php.c
zsistla: There are no defines for these strings.

You'll note that these [02c00c9](https://github.com/newrelic/newrelic-php-agent/pull/1037/commits/02c00c96d55cbddc1e8b71c2cfcc3b56f343bba7) are determined by spec so that all agents report the same string regardless of a particular language's syntax.
For PHP, the aws-sdk-php command is `createTable` but we need to report `create_table` which are two different strings.

axiom/nr_segment_datastore.c
zsistla: Yes.

axiom/nr_segment_datastore.h
zsistla: Yes this is only currently populated for DynamoDB to provide minimal impact to customers since it provides duplicate information to datastore.component attribute but was required to build the relationship.

agent/lib_aws_sdk_php.c
zsistla: It is expected to be non-null for dynamodb but regardless of null/non-null, the calling function [knows](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R656) and will [call nr_free](https://github.com/newrelic/newrelic-php-agent/pull/1037/files#diff-b5410abe33a951704c8600c7b0b214d3c431cee94dcb3089567ec0ec9f328fc1R681) which will either free or ignore if NULL.

agent/lib_aws_sdk_php.h
zsistla: yes.

mfulb (approved):

bduranleau-nr (approved):

PR #1042, opened 2025-04-01
refactor(agent): optimize user function instrumentation lookup for PHPs 8.0+

lavarou:
Reduce agent’s CPU overhead when [observer API](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_observer.h) is used to hook into Zend Engine. This is achieved by:
* Simplifying the checks the agent needs to perform to determine if special instrumentation for user function is provided. Before this change, on every function execution the agent used a hashmap to check if special instrumentation for user function is provided. However, when the observer API is used to hook into Zend Engine, this can be done only once per user function execution at the time when the user function is executed for the first time.
* Reducing the number of times the name of the file is compared with the list of ‘magic files’ to detect package and/or framework used. Before this change on every file execution the agent scanned the list of magic files twice. However, when the observer API is used to hook into Zend Engine, this can be done only once per file execution.


agent/php_user_instrument.c
```diff
@@ -114,7 +116,33 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
+
+static inline nruserfn_t* nr_php_wraprec_lookup_get(zend_function* zf) {
+  nruserfn_t *wraprec = NULL;
+  
+  if (NULL != RUN_TIME_CACHE(&zf->op_array)) {
+    wraprec = ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle));
+  }
+  if (NULL != wraprec && NRPRG(pid) != wraprec->pid) {
+    wraprec = NULL;
+  }
```
lavarou, line 142: I'm not certain this safeguard is needed at all. I need to investigate this a bit more.

agent/php_user_instrument.c
```diff
@@ -290,10 +317,15 @@ static void nr_php_wrap_user_function_internal(nruserfn_t* wraprec TSRMLS_DC) {
   nr_php_wrap_zend_function(orig_func, wraprec TSRMLS_CC);
 }
 
-static nruserfn_t* nr_php_user_wraprec_create(void) {
-  return (nruserfn_t*)nr_zalloc(sizeof(nruserfn_t));
+nruserfn_t* nr_php_user_wraprec_create(void) {
+  nruserfn_t* wr = (nruserfn_t*)nr_zalloc(sizeof(nruserfn_t));
+  /* NRPRG(pid) is set in rinit but it's sometimes needed before rinit,
+  *  i.e. when wraprec is added via INI */
+  wr->pid = (0 != NRPRG(pid) ? NRPRG(pid) : nr_getpid());
```
lavarou, line 324: See https://github.com/newrelic/newrelic-php-agent/pull/1042/files#r2025568617

agent/php_execute.c
```diff
@@ -93,6 +93,12 @@
  * conditions wherever possible.
  */
 
+static inline nruserfn_t* nr_php_get_wraprec_from_op_array_extension(const char* fn, zend_function* zf) {
+  nruserfn_t* wraprec = (nruserfn_t*)ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle));
+  nrl_verbosedebug(NRL_AGENT, "%s from %s, op_array_extension=%p, wraprec=%p, wraprec->pid=%d", __func__, fn, ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)), wraprec, wraprec->pid);
```
zsistla: This will segfault if `fn` is NULL.

Needs a check if `zf` is NULL before dereferencing it.

agent/php_execute.c
```diff
@@ -93,6 +93,12 @@
  * conditions wherever possible.
  */
 
+static inline nruserfn_t* nr_php_get_wraprec_from_op_array_extension(const char* fn, zend_function* zf) {
+  nruserfn_t* wraprec = (nruserfn_t*)ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle));
```
zsistla: The cache slots are not available if the func is a trampoline so it should be checked before invoking `ZEND_OP_ARRAY_EXTENSION`
Note the safety instructions here:
https://github.com/php/php-src/blob/master/Zend/zend_extensions.c#L276-L289

agent/php_minit.c
```diff
@@ -722,6 +722,11 @@ PHP_MINIT_FUNCTION(newrelic) {
   nrl_debug(NRL_INIT, "MINIT processing done");
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO /* PHP 7.4+ */
   NR_PHP_PROCESS_GLOBALS(zend_offset) = zend_get_resource_handle(dummy);
+#if ZEND_MODULE_API_NO >= ZEND_8_4_X_API_NO /* PHP 7.4+ */
```
zsistla: Comment says PHP 7.4+, but the statement is checking for PHP 8.4 or greater.

agent/php_minit.c
```diff
@@ -722,6 +722,11 @@ PHP_MINIT_FUNCTION(newrelic) {
...
+  NR_PHP_PROCESS_GLOBALS(op_array_extension_handle) = zend_get_internal_function_extension_handle("newrelic");
```
zsistla: Why is `op_array_extension_handle` being set to the internal function extension handle for PHP 8.4+?

agent/php_minit.c
```diff
@@ -722,6 +722,11 @@ PHP_MINIT_FUNCTION(newrelic) {
...
+#else
+  NR_PHP_PROCESS_GLOBALS(op_array_extension_handle) = zend_get_op_array_extension_handle("newrelic");
```
zsistla, line 738: Why are we only using this for 8.0-8.3?  Why not 8.4?

agent/php_observer.c
```diff
@@ -75,13 +75,27 @@
...
+    nrl_verbosedebug(NRL_AGENT, "Registering Observer API handlers for user function %s",
+            NRSAFESTR(nr_php_op_array_function_name(NR_OP_ARRAY)));
+  } else {
+    nrl_verbosedebug(NRL_AGENT, "Registering Observer API handlers for file %s",
+            NRSAFESTR(nr_php_op_array_file_name(NR_OP_ARRAY)));
+  }
+
+  wr = nr_php_get_wraprec(execute_data->func);
+  // store the wraprec in the op_array extension for the duration of the request for later lookup
+  ZEND_OP_ARRAY_EXTENSION(&execute_data->func->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)) = wr;
```
zsistla: check for [trampolining](https://github.com/php/php-src/blob/master/Zend/zend_extensions.c#L281-L289) before using.

agent/php_user_instrument.c
```diff
@@ -130,6 +130,10 @@ static nr_php_wraprec_hashmap_t* user_function_wrappers;
 static inline void nr_php_wraprec_lookup_set(nruserfn_t* wr,
                                              zend_function* zf) {
   nr_php_wraprec_hashmap_update(user_function_wrappers, zf, wr);
+  // store the wraprec in the op_array extension for the duration of the request for later lookup
+  // for situation when wraprec is added after first execution of the function
+  ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)) = wr;
```
zsistla: check for [trampolining](https://github.com/php/php-src/blob/master/Zend/zend_extensions.c#L281-L289) before using zend_op_array_extension.

Does `zf` need a NULL check before dereferencing?

agent/php_user_instrument.c
```diff
@@ -291,7 +295,9 @@ static void nr_php_wrap_user_function_internal(nruserfn_t* wraprec TSRMLS_DC) {
 }
 
 static nruserfn_t* nr_php_user_wraprec_create(void) {
-  return (nruserfn_t*)nr_zalloc(sizeof(nruserfn_t));
+  nruserfn_t* wr = (nruserfn_t*)nr_zalloc(sizeof(nruserfn_t));
+  wr->pid = nr_getpid();
```
zsistla: check `wr` for NULL.

agent/php_execute.c
lavarou: `nr_php_get_wraprec_from_op_array_extension` has been removed.

agent/php_execute.c
lavarou: nr_php_get_wraprec_from_op_array_extension has been removed.

agent/php_minit.c
lavarou: This code no longer exists.

agent/php_minit.c
lavarou: This code no longer exists.

agent/php_user_instrument.c
lavarou: Checking that `wr` is not `NULL` is redundant - it is guaranteed by `nr_zalloc` that its return value is not `NULL`. See https://github.com/newrelic/newrelic-php-agent/blob/R11.7/axiom/util_memory.c#L57-L61.

agent/php_user_instrument.c
```diff
@@ -114,7 +116,33 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
+  zend_init_func_run_time_cache(&zf->op_array);
+  if (NULL != RUN_TIME_CACHE(&zf->op_array)) {
+    ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)) = wr;
+  }
+}
+
+static inline nruserfn_t* nr_php_wraprec_lookup_get(zend_function* zf) {
+  nruserfn_t *wraprec = NULL;
+  
+  if (NULL != RUN_TIME_CACHE(&zf->op_array)) {
```
zsistla, line 137: is `zf` guaranteed to be non-null here?

agent/php_user_instrument.c
```diff
@@ -114,7 +116,33 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
+static inline void nr_php_wraprec_lookup_set(nruserfn_t* wr,
+                                             zend_function* zf) {
+  // for situation when wraprec is added after first execution of the function
+  // store the wraprec in the op_array extension for the duration of the request for later lookup
+  // The op_array extension slot for function may not be initialized yet because it is
+  // initialized only on the first call made to the function in that request. This would
+  // mean that run_time_cache is NULL and wraprec cannot be stored yet! It will be stored
+  // on the first call to the function when observer is registered for that function.
+  zend_init_func_run_time_cache(&zf->op_array);
```
zsistla, line 128: Is `zf` guaranteed to be non-null?

agent/php_user_instrument.c
zsistla: Doesn't this protect us in case another extension has written to the cache?

agent/php_user_instrument.c
lavarou: It is after 7ae06753f2aefabb807ec3d0c39d84055b389575.

agent/php_observer.c
```diff
@@ -75,13 +76,58 @@
...
+    zend_long fn_hash = zend_hash_func(ZSTR_VAL(func_name), ZSTR_LEN(func_name));
+    nrl_always("class_name=%s, class_name->h=%zu, cn_hash=%zu, function_name=%s, function_name->h=%zu, fn_hash=%zu",
+                ZSTR_VAL(scope_name), scope_name->h, cn_hash,
+                ZSTR_VAL(func_name), func_name->h, fn_hash);
+#endif
+  } else if (OP_ARRAY_IS_A_FUNCTION(NR_OP_ARRAY)){
+    func_name = execute_data->func->op_array.function_name;
+#if 0
+    zend_long fn_hash = zend_hash_func(ZSTR_VAL(func_name), ZSTR_LEN(func_name));
+    nrl_always("function_name=%s, function_name->h=%zu, fn_hash=%zu", ZSTR_VAL(func_name), func_name->h, fn_hash);
```
ZNeumann, line 115: Probably want these log levels to be verbosedebug

lavarou: I plan on removing the code between `#if 0`.

agent/php_observer.c
```diff
@@ -75,13 +76,58 @@
...
+     * Let's get the framework info.
+     */
+    nr_php_execute_file(NR_OP_ARRAY, execute_data, NULL TSRMLS_CC);
+    return handlers;
+  }
+
+  if (OP_ARRAY_IS_A_METHOD(NR_OP_ARRAY)) {
+    scope_name = execute_data->func->op_array.scope->name;
+    func_name = execute_data->func->op_array.function_name;
+#if 0
```
mfulb, line 104: Are these just for debugging?  Maybe if we want to leave some troubleshooting code around have a define like DEBUG_FCALL_HANDLERS at the top of the file which is false by default instead of `#if 0`?

agent/php_user_instrument_wraprec_hashmap.c
```diff
@@ -0,0 +1,485 @@
...
+
+  if (created) {
+    *created = true;
+  }
+  return nr_func_hashmap_add_internal(hashmap, hash, key);
+}
+
+static void nr_func_hashmap_destroy_bucket_internal(nr_func_bucket_t** bucket_ptr) {
+  nr_func_bucket_t* bucket = *bucket_ptr;
+  nr_free(bucket->key->name);
```
mfulb, line 149: Possibly good idea to NULL check these struct pointer usage?

agent/php_user_instrument_wraprec_hashmap.c
```diff
@@ -0,0 +1,485 @@
...
+  scope->name_hash = 0;
+
+  /* If scope::method, then break into two strings */
+  for (int i = 0; i < full_name_len; i++) {
+    if ((':' == full_name[i]) && (':' == full_name[i + 1])) {
+      func->is_method = true;
+      scope->name = full_name;
+      scope->name_len = i;
+      func->name = full_name + i + 2;
+      func->name_len = full_name_len - i - 2;
```
mfulb, line 373: What if '::' was last two chars in string and `func->name_len` was 0 how is this handled by rest of code?

agent/php_user_instrument.c
lavarou: Yes. `nr_php_wraprec_lookup_set` is only called [here](https://github.com/newrelic/newrelic-php-agent/blob/refactor/optimize-wraprec-lookup/agent/php_user_instrument.c#L263) and `nr_php_wrap_zend_function` is only called [here](https://github.com/newrelic/newrelic-php-agent/blob/refactor/optimize-wraprec-lookup/agent/php_user_instrument.c#L317) or [here](https://github.com/newrelic/newrelic-php-agent/blob/refactor/optimize-wraprec-lookup/agent/php_user_instrument.c#L495), in both situations the function `nr_php_wrap_zend_function` is only called after ensuring `zend_function*` is not NULL.

zsistla:
Please update the PR description.

agent/fw_drupal8.c
```diff
@@ -140,22 +140,26 @@ static void nr_drupal8_add_method_callback_before_after_clean(
...
 
   if (NULL == ce) {
     nrl_verbosedebug(NRL_FRAMEWORK, "Drupal 8: got NULL class entry in %s",
                      __func__);
     return;
   }
 
-  function = nr_php_find_class_method(ce, method);
+  methodLC = nr_string_to_lowercase(method);
+  function = nr_php_find_class_method(ce, methodLC);
```
mfulb, line 152: Can we free `methodLC` after its use here and remove the 2 frees below?

agent/php_user_instrument_wraprec_hashmap.c
lavarou: This code is almost a verbatim copy of [this code](https://github.com/newrelic/newrelic-php-agent/blob/refactor/optimize-wraprec-lookup/agent/php_user_instrument.c#L338-L358). Should this concern be addressed in both places?

agent/fw_drupal8.c
lavarou: See cf07d9ae69c122c463a810396fd0e666c916f878.

agent/php_user_instrument.c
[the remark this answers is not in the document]
lavarou: See c494a450fff8b31ed0313cc700e885f9cd9bf531.

agent/php_observer.c
```diff
@@ -75,13 +76,58 @@
...
+                ZSTR_VAL(func_name), func_name->h, fn_hash);
+#endif
+  } else if (OP_ARRAY_IS_A_FUNCTION(&execute_data->func->op_array)){
+    func_name = execute_data->func->op_array.function_name;
+#if 0
+    zend_long fn_hash = zend_hash_func(ZSTR_VAL(func_name), ZSTR_LEN(func_name));
+    nrl_always("function_name=%s, function_name->h=%zu, fn_hash=%zu", ZSTR_VAL(func_name), func_name->h, fn_hash);
+#endif
+  }
+
```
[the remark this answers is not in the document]
lavarou: See c494a450fff8b31ed0313cc700e885f9cd9bf531.

agent/php_user_instrument_wraprec_hashmap.c
lavarou: I'll add a check for last character being `:` to `nr_php_add_custom_tracer_named`.

agent/php_observer.c
lavarou: This code has been cleaned up and simplified in 09eca036c48b27d7d27a20137b268d9543a59676. There is no need for any of this debugging code.

agent/php_user_instrument_wraprec_hashmap.c
lavarou: Input validation added in adccd7f9c6f9de98970c52cf9d1a9c16d894cb73.

agent/php_minit.c
```diff
@@ -491,6 +492,18 @@ PHP_MINIT_FUNCTION(newrelic) {
...
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
+  /* 
+   * The user function wraprec hashmap must be initialized before INI processing
+   * because INI processing adds wraprecs:
+   *  - newrelic.webtransaction.name.functions
+   *  - newrelic.transaction_tracer.custom
+  */
+  if (NR_SUCCESS != nr_php_user_instrument_wraprec_hashmap_init()) {
+    nrl_error(NRL_AGENT, "Failed to initialize user function instrumentation");
```
ZNeumann, line 503: If this happens, do we want to `goto disbad`? I am unsure of the implications of skipping `nr_php_register_ini_entries`, as all other `goto disbad`'s are afterward.

ZNeumann (approved):

agent/php_user_instrument.c
```diff
@@ -303,13 +362,6 @@ static nruserfn_t* nr_php_user_wraprec_create_named(const char* full_name,
   int klass_len;
   nruserfn_t* wraprec;
 
-  if (0 == full_name) {
```
zsistla, line 306: Since it's no longer validating `full_name` here, can you please leave a comment that this function expects `full_name` to be a prevalidated non-null, non-zero length string that doesn't end in `:`?

agent/php_user_instrument_wraprec_hashmap.c
```diff
@@ -0,0 +1,485 @@
...
+  nr_func_hashmap_key_t func_key = {0};
+  nr_func_hashmap_t* funcs_ht = NULL;
+  bool is_new_wraprec = false;
+  nruserfn_t* wraprec = NULL;
+
+
+  if (NULL == scope_ht || NULL == global_funcs_ht) {
+    return NULL;
+  }
+
```
zsistla, line 407: Since it's no longer validating `namestr` here, can you please leave a comment that this function expects `namestr` to be a prevalidated non-null, non-zero length string that doesn't end in `:`?

agent/php_user_instrument_wraprec_hashmap.c
zsistla: The change looks good to me, just had request for two additional comments to clarify the func is no longer validating and expects a prevalidated value.
https://github.com/newrelic/newrelic-php-agent/pull/1042#discussion_r2067357743
https://github.com/newrelic/newrelic-php-agent/pull/1042#discussion_r2067364238

agent/php_observer.c
zsistla: Should we output the hash value in show executes?  Maybe just for PHP 8+?

agent/fw_drupal8.c
mfulb: Now I think about it is there a reason for a temporary variable at all (aside from code clarity)?

agent/tests/test_user_instrument.c
```diff
@@ -145,6 +145,28 @@ static void test_hashmap_wraprec() {
...
+  nruserfn_t* wr = NULL;
+  tlib_php_request_start();
+
+  wr = nr_php_add_custom_tracer_named(NULL, 10);
+  tlib_pass_if_null("add_custom_tracer_named with NULL name", wr);
+
+  wr = nr_php_add_custom_tracer_named("function_name", 0);
+  tlib_pass_if_null("add_custom_tracer_named with length == 0", wr);
+
+  wr = nr_php_add_custom_tracer_named("function_name", -1);
```
mfulb, line 158: Im curious what this is doing - the function called accepts `size_t` which I believe is unsigned, but -1 is being passed in - does the compiler flag this?  I would think the receiving function would receive a very large positive integer unless something is happening under the hood.

agent/php_user_instrument.c
```diff
@@ -428,27 +513,40 @@ nruserfn_t* nr_php_add_custom_tracer_callable(zend_function* func TSRMLS_DC) {
...
 
   nr_php_wrap_zend_function(func, wraprec TSRMLS_CC);
-#if ZEND_MODULE_API_NO < ZEND_7_4_X_API_NO
   nr_php_add_custom_tracer_common(wraprec);
-#endif
 
   return wraprec;
 }
 
+static inline bool nr_php_user_instrument_is_name_valid(const char* namestr,
```
mfulb, line 521: This name is a bit misleading - if I pass "$)O#()@::@)@)@@" it will return `true` won't it?  How strongly do we care about the name being "valid"?

mfulb: Or "a::b::c::d" will pass as well.

mfulb: Since there is an API to add these it would be probably good if somewhere these strings are robustly checked - perhaps best in the API code before this level is called?

agent/fw_drupal8.c
mfulb: This looks good.

agent/php_user_instrument_wraprec_hashmap.c
lavarou: See 575aa95bd52a225774fb2428fa7520d64fdea37f.

agent/php_user_instrument.c
lavarou: See 575aa95bd52a225774fb2428fa7520d64fdea37f.

agent/php_user_instrument_wraprec_hashmap.c
```diff
@@ -0,0 +1,482 @@
...
+} nr_scope_bucket_t;
+
+typedef struct _nr_scope_hashmap {
+  size_t log2_num_buckets;
+  nr_scope_bucket_t** buckets;
+  size_t elements;
+} nr_scope_hashmap_t;
+
+static nr_scope_hashmap_t* nr_scope_hashmap_create_internal(size_t log2_num_buckets) {
+  nr_scope_hashmap_t* hashmap;
```
zsistla, line 205:
```suggestion
  nr_scope_hashmap_t* hashmap = NULL;
```

General comment regarding the other unitialized pointers in the PR, it's good practice to initialize the pointers to NULL.

agent/php_user_instrument_wraprec_hashmap.c
```diff
@@ -0,0 +1,482 @@
...
+}
+
+nr_status_t nr_php_user_instrument_wraprec_hashmap_init(void) {
+  if (NULL == scope_ht) {
+    scope_ht = nr_scope_hashmap_create_internal(0);
+  }
+  if (NULL == global_funcs_ht) {
+    global_funcs_ht = nr_func_hashmap_create_internal(0);
+  }
+  return NR_SUCCESS;
```
zsistla, line 393: With only one return value, does this need to return anything?

agent/php_user_instrument_wraprec_hashmap.c
mfulb: This function is used as a helper and in that context `*bucket_ptr` is guaranteed to be non-NULL.  However, I believe a NULL check for `bucket->key` is warranted as that field is not as readily obvious to not possibly be NULL.

agent/php_user_instrument_wraprec_hashmap.c
lavarou: This was changed in https://github.com/newrelic/newrelic-php-agent/pull/1042/commits/64408b5fc9c7aecb0ba197263a540b0f53bc38ca.

agent/php_minit.c
```diff
@@ -491,6 +492,18 @@ PHP_MINIT_FUNCTION(newrelic) {
...
   nr_php_generate_internal_wrap_records();
 
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
+  /* 
+   * The user function wraprec hashmap must be initialized before INI processing
+   * because INI processing adds wraprecs:
+   *  - newrelic.webtransaction.name.functions
+   *  - newrelic.transaction_tracer.custom
+  */
+  if (NR_SUCCESS != nr_php_user_instrument_wraprec_hashmap_init()) {
```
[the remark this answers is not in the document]
lavarou: This was changed in https://github.com/newrelic/newrelic-php-agent/pull/1042/commits/64408b5fc9c7aecb0ba197263a540b0f53bc38ca.

agent/php_user_instrument_wraprec_hashmap.c
lavarou: Added in c9563c2c11298078e426c88b6c39380812e5091f.

agent/php_user_instrument_wraprec_hashmap.c
lavarou: See 55ea3c5a1fdf2bc65e4df93f9a2f65db0ed40747.

agent/php_user_instrument.c
lavarou: A method of using `NRPRG(pid)` to validate if memory pointed to by the value retrieved from `ZEND_OP_ARRAY_EXTENSION`, has been allocated by the agent, has been replaced by a unique memory marker in 3cc2311c5278b0d47de0d6f9b461a60a67343d42.

agent/php_minit.c
mfulb, line 502: Have you verified this condition works as expected?  I noticed the code coverage showed the body of the condition was not exercised.

agent/php_user_instrument.c
```diff
@@ -114,7 +116,66 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
-#if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
+static inline void nr_php_wraprec_lookup_set(nruserfn_t* wr,
+                                             zend_function* zf) {
+  // The function cache slots are not available if the function is a trampoline
+  if (zf->op_array.fn_flags & ZEND_ACC_CALL_VIA_TRAMPOLINE) {
+    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+      char* name = nr_php_function_debug_name(zf);
+      nrl_verbosedebug(NRL_INSTRUMENT, "%s - %s is a trampoline function",
+                       __func__, name);
```
mfulb, line 127: Can `name` be `NULL`?

agent/php_user_instrument.c
```diff
@@ -114,7 +116,66 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
+  }
+
+  if (NULL != RUN_TIME_CACHE(&zf->op_array)) {
+    wraprec = ZEND_OP_ARRAY_EXTENSION(&zf->op_array, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle));
+  }
+  if (NULL != wraprec && wraprec->magic != NR_USERFN_T_MAGIC) {
+    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+      char* name = nr_php_function_debug_name(zf);
+      nrl_verbosedebug(NRL_INSTRUMENT, "%s - wraprec for {%s} is invalid",
+                       __func__, name);
```
mfulb, line 171: Can `name` be `NULL`?

agent/fw_drupal8.c
```diff
@@ -140,14 +140,17 @@ static void nr_drupal8_add_method_callback_before_after_clean(
...
+  char* methodLC = NULL;
 
   if (NULL == ce) {
     nrl_verbosedebug(NRL_FRAMEWORK, "Drupal 8: got NULL class entry in %s",
                      __func__);
     return;
   }
 
-  function = nr_php_find_class_method(ce, method);
+  methodLC = nr_string_to_lowercase(method);
```
mfulb, line 151: Would a `NULL` check for `method` be good here?  Maybe we could short circuit out - and it would shield some log messages below from trying to pass `NULL` to a "`%s`" argument.

agent/php_user_instrument.c
```diff
@@ -114,7 +116,66 @@ int nr_zend_call_orig_execute_special(nruserfn_t* wraprec,
...
+static inline nruserfn_t* nr_php_wraprec_lookup_get(zend_function* zf) {
+  nruserfn_t *wraprec = NULL;
+
+  // The function cache slots are not available if the function is a trampoline
+  if (zf->op_array.fn_flags & ZEND_ACC_CALL_VIA_TRAMPOLINE) {
+    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+      char* name = nr_php_function_debug_name(zf);
+      nrl_verbosedebug(NRL_INSTRUMENT, "%s - %s is a trampoline function",
+                       __func__, name);
+      nr_free(name);
```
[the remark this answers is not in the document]
lavarou: It can. Good point. Addressed with 5383c94898dfc00492c05a874a70cc1c5766c8ca. Is this acceptable?

agent/php_user_instrument.c
lavarou: It can. Good point. Addressed with 5383c94898dfc00492c05a874a70cc1c5766c8ca. Is this acceptable?

agent/php_observer.c
```diff
@@ -82,6 +83,38 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
+    nr_php_execute_file(NR_OP_ARRAY, execute_data, NULL TSRMLS_CC);
+    return handlers;
+  }
+
+  // The function cache slots are not available if the function is a trampoline
+  if (execute_data->func->op_array.fn_flags & ZEND_ACC_CALL_VIA_TRAMPOLINE) {
+    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+      char* name = nr_php_function_debug_name(execute_data->func);
+      nrl_verbosedebug(NRL_INSTRUMENT, "%s - %s is a trampoline function",
+                       __func__, name);
```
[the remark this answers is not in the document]
lavarou: It can. Good point. Addressed with 5383c94898dfc00492c05a874a70cc1c5766c8ca. Is this acceptable?

agent/fw_drupal8.c
lavarou: I can add a `NULL` check, but it seems redundant - `method` is equal to `"invokeAllWith"` because [this](https://github.com/newrelic/newrelic-php-agent/blob/5383c94898dfc00492c05a874a70cc1c5766c8ca/agent/fw_drupal8.c#L645-L647) is the only place `nr_drupal8_add_method_callback_before_after_clean` is used.

agent/php_user_instrument_wraprec_hashmap.c
```diff
@@ -0,0 +1,489 @@
...
+  nr_scope_hashmap_key_t* scope,
+  const char* full_name,
+  int full_name_len) {
+
+  if (nrunlikely(NULL == full_name)) {
+    return;
+  }
+  if (nrunlikely(full_name_len <= 0)) {
+    return;
+  }
```
zsistla, line 357: Since this is only called from `nr_php_user_instrument_wraprec_hashmap_add` which already validates name and name_len, could we remove the extra checks and add the extra comment like `nr_php_user_instrument_wraprec_hashmap_add` has?

(here's the snippet since I can't link it directly because github keeps hiding the large diff)
```
/* This function expects namestr and namestrlen to be validated with
 * nr_php_user_instrument_is_name_valid() before being passed in.
 * Specifically, it requires that:
 * - namestrlen be greater than 0
 * - namestr must not be NULL and must not end with `:` (colon) . */

nruserfn_t* nr_php_user_instrument_wraprec_hashmap_add(const char* namestr, size_t namestrlen) {
```

agent/tests/test_user_instrument.c
```diff
@@ -145,6 +145,32 @@ static void test_hashmap_wraprec() {
...
+  // side effects will happen when this silent type conversion takes place.
+  wr = nr_php_add_custom_tracer_named("function_name", -1);
+  tlib_pass_if_null("add_custom_tracer_named with length < 0", wr);
+
+  wr = nr_php_add_custom_tracer_named(NR_PSTR("scope_name::"));
+  tlib_pass_if_null("add_custom_tracer_named with name ending with :", wr);
+
+  wr = nr_php_add_custom_tracer_named(NR_PSTR("scope_name::function_name"));
+  tlib_pass_if_not_null("add_custom_tracer_named with valid name", wr);
+
```
zsistla, line 170: Thanks for adding the tests for add_custom_tracer_named.
A test for the following would be good for completeness:
```
/* Valid func name, no scope */
  wr = nr_php_add_custom_tracer_named(NR_PSTR("function_name"));
  tlib_pass_if_not_null("add_custom_tracer_named with valid func name, no scope", wr);
```

tests/integration/lang/trampoline/test_trampoline_00.php
```diff
@@ -0,0 +1,39 @@
+<?php
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+Test that trampoline functions not blow up.
```
zsistla, line 8: "do not blow up"
It looks copy/pasted in other tests.

zsistla (approved):

agent/php_user_instrument_wraprec_hashmap.c
lavarou: See c6c50d97dc5cf67311a4b504a7173d02223b8486.

mfulb (approved):

agent/tests/test_user_instrument.c
lavarou: See 9748cf5649dc3f3e179ae74842ede3a1a03119d8.

zsistla (approved):

zsistla (approved):

mfulb (approved):

PR #1049, opened 2025-04-16
feat(agent): clm off by default

bduranleau-nr:


tests/integration/api/add_custom_parameter/test_span_event.php
```diff
@@ -8,13 +8,6 @@
 Test that custom transaction attributes are added to span events.
 */
```
hahuja2, line 10: Since this PR is turning CLM off by default, we will have many duplicate tests that set CLM to off already. I think those tests should be removed in this PR also. Here is an example: https://github.com/newrelic/newrelic-php-agent/blob/765888d9998b6af7897fd43e03c75acb3f5619b8/tests/integration/api/add_custom_parameter/test_span_event.php5.php#L17

bduranleau-nr: These php5/clm off tests have been pruned in this commit fae50f7468f22d6f6c4e9d71f179e1879c922d02
Are there any others you're aware of?

hahuja2: Looks good to me, I don't see any additional tests that need to be removed.

lavarou (approved):

PR #1055, opened 2025-04-21
feat(agent): composer api package detection on by default

bduranleau-nr:
Changes the default configuration value for composer-based package detection to `true`.

An additional configuration value, `newrelic.vulnerability_management.composer_api.per_process_detection`, has also been added to control the frequency at which composer sampling occurs. By default, this setting will be enabled, only sampling Composer API data once per php-fpm process. This improves baseline performance to make the enablement of Composer API package detection a negligible overhead impact.

### DIFF
--- a/agent/php_nrini.c
+++ b/agent/php_nrini.c
@@ -537,6 +537,28 @@ static PHP_INI_MH(nr_preload_framework_library_detection_mh) {
   return SUCCESS;
 }
 
+static PHP_INI_MH(nr_composer_per_process_detection_mh) {
+  int val;
+
+  (void)entry;
+  (void)NEW_VALUE_LEN;
+  (void)mh_arg1;
+  (void)mh_arg2;
+  (void)mh_arg3;
+  (void)stage;
+  NR_UNUSED_TSRMLS;
+
+  val = nr_bool_from_str(NEW_VALUE);
+
+  if (-1 == val) {
+    return FAILURE;
+  }
+
+  NR_PHP_PROCESS_GLOBALS(composer_api_per_process_detection) = val ? 1 : 0;
+
+  return SUCCESS;
+}
+
 static PHP_INI_MH(nr_loglevel_mh) {
   nr_status_t rv;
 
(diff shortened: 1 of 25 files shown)


agent/php_nrini.c
```diff
@@ -537,6 +537,28 @@ static PHP_INI_MH(nr_preload_framework_library_detection_mh) {
...
+  (void)mh_arg1;
+  (void)mh_arg2;
+  (void)mh_arg3;
+  (void)stage;
+  NR_UNUSED_TSRMLS;
+
+  val = nr_bool_from_str(NEW_VALUE);
+
+  if (-1 == val) {
+    return FAILURE;
```
mfulb, line 554: I think it would be useful for the user to log when an invalid string was given for an option like this.
Something like
`nrl_warning("The value %s is not valid for <insert INI value name>, using default value");`

agent/scripts/newrelic.ini.template
```diff
@@ -1363,11 +1363,21 @@ newrelic.daemon.logfile = "/var/log/newrelic/newrelic-daemon.log"
...
-;newrelic.vulnerability_management.composer_api.enabled = false
+;newrelic.vulnerability_management.composer_api.enabled = true
+
+; Setting: newrelic.vulnerability_management.composer_api.per_process_detection
+; Type   : boolean
+; Scope  : system
+; Default: true
+; Info   : Controls the frequency at which Composer API samples the runtime environment
+;          for package data. When set to `true`, sampling will only occur once per process.
+;          If false, Composer will sample the environment every request.
```
mfulb, line 1378: ### 
```suggestion
;          If false, Composer will sample the environment every request, increasing the frequency which this package detection is performed.
```

agent/php_nrini.c
mfulb: I think it would be better to print out the `NEW_VALUE` so they can see the invalid string they used in the INI file.  The current change would just print out that "-1" is an invalid value?

bduranleau-nr: c43e3c01aee8b369d914a1f2ccf25f38d4742175

agent/php_nrini.c
```diff
@@ -537,6 +537,33 @@ static PHP_INI_MH(nr_preload_framework_library_detection_mh) {
...
+  (void)mh_arg2;
+  (void)mh_arg3;
+  (void)stage;
+  NR_UNUSED_TSRMLS;
+
+  val = nr_bool_from_str(NEW_VALUE);
+
+  if (-1 == val) {
+    nrl_warning(NRL_INIT,
+                "The value %s is not valid for the "
```
mfulb, line 555:
```suggestion
                "The value \"%s\" is not valid for the "
```

mfulb (approved):

PR #1059, opened 2025-04-24
chore(agent): remove guzzle 3.x from agent codebase

hahuja2:
This PR removes guzzle `3.x` from the agent codebase.


agent/lib_guzzle_common.c
```diff
@@ -231,33 +231,6 @@ char* nr_guzzle_response_get_header(const char* header,
     if (Z_STRLEN_P(retval) > 0) {
       value = nr_strndup(Z_STRVAL_P(retval), Z_STRLEN_P(retval));
     }
-  } else if (nr_php_object_instanceof_class(
-                 retval, "Guzzle\\Http\\Message\\Header" TSRMLS_CC)) {
```
lavarou, line 235: Does this class exist only in Guzzle 3?

hahuja2: Yes, this class only exists in Guzzle 3.

lavarou (approved):

PR #1060, opened 2025-05-12
refactor(agent): improve logging when adding named tracers

lavarou:
Make log messages easier to understand when adding a named custom tracer by
logging only a single message. Instead of logging one message when the wraprec
is added, and two messages when wraprec is re-used, log a single message that
indicates if the wraprec was added or re-used.


ZNeumann (approved):

lavarou:
Superseded by #1061.

PR #1097, opened 2025-07-01
tests: fix integration_runner opcache module loading

mfulb:
With recent official PHP docker images the Zend OPCache extension is loaded by default.  This is causing some test failures as PHP will output a warning message when the `integration_runner` attempts to load this extension.  As a result any `EXPECT` statements in tests will fail due to this extra output.

The solution chosen was to use the “-m” option for the `php` and `php-cgi` binaries to probe if the Zend OPCache module is loaded by default in the PHP environment.  Both binaries are used in testing so the defaults of both are considered.  The value of the integration_runner command line option `opcache_off` is also noted.  When a test is going to be run this information is used to determine if the `zend_extension=opcache.so` is needed as well as setting the values of `opcache.enable` and `opcache.enable_cli`.  These INI values needed special handling because it is possible these are overwritten by values in the INI stanza of a given test case.  Also consideration is needed to if the `PHPMODULES` stanza exists in the test case and specifies loading the `opcache.so` module.

### DIFF
--- a/daemon/cmd/integration_runner/main.go
+++ b/daemon/cmd/integration_runner/main.go
@@ -453,8 +453,14 @@ func main() {
 	if numFailed > 0 {
 		os.Exit(1)
 	}
-	if *flagWarnIsFail && numWarned > 0 {
-		os.Exit(2)
+
+	if numWarned > 0 {
+		if *flagWarnIsFail {
+			fmt.Println("WARNING: some tests were warned, but are treated as failures because --warnisfail is true")
+			os.Exit(2)
+		} else {
+			fmt.Printf("WARNING: some tests were warned, but are not treated as failures because --warnisfail is false\n")
+		}
 	}
 }
 
@@ -541,6 +547,12 @@ func runTest(t *integration.Test) {
 		if skipRE.Match(body) {
 			reason := string(bytes.TrimSpace(head(body)))
 			t.Skip(reason)
+			return
+		}
+
+		if warnRE.Match(body) && *flagWarnIsFail {
+			reason := string(bytes.TrimSpace(head(body)))
+			t.Warn(reason)
 			return
 		}
 	}


daemon/internal/newrelic/integration/test.go
```diff
@@ -214,7 +214,28 @@ func (t *Test) MakeRun(ctx *Context) (Tx, error) {
...
+			(ctx.OPCacheModuleLoaded[ctx.PHP]) {
+			for k, v := range t.PhpModules {
+				if !strings.Contains(v, "opcache.so") {
+					phpModulesCopy[k] = v
+				}
+			}
+		}
+	} else {
+		fmt.Printf("ERROR - UNEXPECTED - Running C test: %s\n", t.Path)
+		os.Exit(1)
```
ZNeumann: I know that we currently do not seem to run any "C Tests", but is there a reason to `Exit(1)` here rather than leave the C code paths in tact?

ZNeumann (dismissed):

daemon/internal/newrelic/integration/transaction.go
```diff
@@ -46,9 +46,43 @@ func PhpTx(src Script, env, settings map[string]string, ctx *Context) (Tx, error
...
+
+		// see if settings affect opcache config
+		// if so then we will not set config below
+		if k == "opcache.enable" {
+			setOPCacheEnable = false
+		} else if k == "opcache.enable_cli" {
+			setOPCacheEnableCLI = false
+		}
+	}
+	if ctx.UseOPCache {
```
lavarou, line 69: `ctx.UseOPCache` is actually the negation of the value of the `--opcacheoff` flag passed to `integration_runner`, which is `false` by default. This means that this branch is the default behavior. This also means that that `--opcacheoff` does not affect presence of `opcache.so` module.

daemon/internal/newrelic/integration/test.go
```diff
@@ -214,15 +222,70 @@ func (t *Test) MakeRun(ctx *Context) (Tx, error) {
...
+		}
 	}
+
+	// Make a copy of t.PhpModules and remove any entries containing "opcache.so"
+	// if opcache.so is loaded by default
+	//
+	// three cases:
+	// 1. C test - no opcache.so needed as PHP is not going to be run
+	// 2. Web test and php-cgi has opcache.so loaded by default - remove any PHPMODULE spec for opcache.so
+	// 3. PHP test and php has opcache.so loaded by default - remove any PHPMODULE spec for opcache.so
```
lavarou: The "C test" case has been eliminated:
```suggestion
	// two cases:
	// 1. Web test and php-cgi has opcache.so loaded by default - remove any PHPMODULE spec for opcache.so
	// 2. PHP test and php has opcache.so loaded by default - remove any PHPMODULE spec for opcache.so
```

daemon/internal/newrelic/integration/transaction.go
mfulb: That seems correct.

daemon/internal/newrelic/integration/test.go
mfulb: I was thinking we'd run it this way and see if it is ever triggered.   As it doesn't seem to be I was thinking a followup action would be to remove the support code for C testing.

lavarou (approved):

ZNeumann (approved):

Issue #1099, opened 2025-07-10
Excessive logging of `hookImplementationsMap[hook]: invalid key or value`


richardgaunt:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

This started with the upgrade to `11.10.0.24`  for new relic php agent

We are getting many, many (1000s) of logs during deployment:

`hookImplementationsMap[hook]: invalid key or value`

This log was implemented as part of: https://github.com/newrelic/newrelic-php-agent/pull/1030

I have run a check within my code base validating the `hookImplementationsMap` (in `ModuleHandler`) but cannot see any issue with the `ModuleHandler::hookImplementationsMap`

See below (I may have got something wrong as the c code is not easily readable for a php developer - please let me know and I can reimplement in my code base):
```  
protected function isValidImplementation() {
    foreach ($this->hookImplementationsMap as $hook_key => $hook_value) {
      // Remimplementing
      //      if (NULL == key || 0 == ZEND_STRING_LEN(key)
      //        || 0 == nr_php_is_zval_valid_array(val)
      //        || 0 == zend_hash_num_elements(Z_ARRVAL_P(val))) {
      //        return true;
      //      } else {
      //        return false;
      //      }

      if ($hook_key === NULL || $hook_value === '' || !is_array($hook_value) || count($hook_value) === 0) {
        return $hook_key;
      }
    }
  }
```

## Steps to Reproduce
Deploy the server with the new relic 

## Expected Behavior

Should not be getting these logs.

## Relevant Logs / Console output

`warning: hookImplementationsMap[hook]: invalid key or value`

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )

We are utilising this image (note the new relic download and installation):

https://github.com/uselagoon/lagoon-images/blame/main/images/php-fpm/8.3.Dockerfile

richardgaunt:
Might be good for the new relic agent to specify which entry it found to be invalid

richardgaunt:
I can confirm that blackfire is installed but not activated on our setup

zsistla:
Thank you for your feedback.  This issue has been addressed in the latest New Relic PHP Agent Release [12.0.0.25](https://github.com/newrelic/newrelic-php-agent/releases/tag/v12.0.0.25).  Logging has been modified to reduce the number of logs and to provide more information on the error condition.

seanhamlin:
[...]

PR #1100, opened 2025-07-11
refactor(agent): Improve invalid hookImplementationsMap[hook] log messages.

richardgaunt:
Add information to the log message about what hook is invalid.
#1099


richardgaunt:
@bduranleau-nr Thank you that is great and your PR is an improvement.

PR #1101, opened 2025-07-11
refactor(agent): improve drupal hook map error logging

bduranleau-nr:


agent/fw_drupal8.c
```diff
@@ -605,10 +605,19 @@ NR_PHP_WRAPPER_END
...
-                                      zval* val) {
+                                      zval* val,
+                                      const char* key_ident) {
   if (NULL == key || 0 == ZEND_STRING_LEN(key)
       || 0 == nr_php_is_zval_valid_array(val)
       || 0 == zend_hash_num_elements(Z_ARRVAL_P(val))) {
+    if (NULL != key) {
+      nrl_warning(NRL_FRAMEWORK,
+                  "hookImplementationsMap[%s]: invalid value for key '%s'",
+                  key_ident, ZEND_STRING_VALUE(key));
```
zsistla: enclose both key_idents in `NRSAFESTR` just in case

agent/fw_drupal8.c
```diff
@@ -605,10 +605,19 @@ NR_PHP_WRAPPER_END
 #endif  // OAPI
 
 static bool nr_is_invalid_key_val_arr(nr_php_string_hash_key_t* key,
-                                      zval* val) {
+                                      zval* val,
+                                      const char* key_ident) {
   if (NULL == key || 0 == ZEND_STRING_LEN(key)
       || 0 == nr_php_is_zval_valid_array(val)
       || 0 == zend_hash_num_elements(Z_ARRVAL_P(val))) {
+    if (NULL != key) {
```
zsistla: if you break up the if, you don't have to do the NULL check again.

could be something like:
```
if (NULL == key || 0 == ZEND_STRING_LEN(key)
invalid key msg
return false
if 0 == nr_php_is_zval_valid_array(val)
      || 0 == zend_hash_num_elements(Z_ARRVAL_P(val)))
invalid value msg 
return false

File: agent/fw_drupal8.c (hunk 1 above)
Comment (bduranleau-nr) [reply agent/fw_drupal8.c]: I don't think `NRSAFESTR` is necessary here- `key_ident` is not derived, it is a manually defined and passed `const char *` passed explicitly by the developer

Comment (bduranleau-nr) [reply agent/fw_drupal8.c]: wrapping the `key` variable might make sense though

File: agent/fw_drupal8.c (hunk 2 above)
Comment (bduranleau-nr) [reply agent/fw_drupal8.c]: 726fb76db25e505ba800028cd4e04211c93b7369

File: agent/fw_drupal8.c (hunk 1 above)
Comment (bduranleau-nr) [reply agent/fw_drupal8.c]: 726fb76db25e505ba800028cd4e04211c93b7369

File: agent/fw_drupal8.c [hunk 3]
```diff
@@ -646,25 +655,28 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
...
         return false;
       }
 
       ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(class_val), method_key,
                                     module_val) {
         if (NULL == method_key
             || 0 == nr_php_is_zval_valid_string(module_val)) {
-          nrl_warning(NRL_FRAMEWORK,
-                      "hookImplementationsMap[method]: invalid key or value");
+          if (NULL != method_key) {
```
zsistla: Could similarly split this to check 
if (NULL == method_key
then output the invalid key msg and return false
then check 0 == nr_php_is_zval_valid_string(module_val))
and output the invalid val msg and return false

Would make it more streamlined and save us the extra null check.

agent/fw_drupal8.c
```diff
@@ -605,14 +605,23 @@ NR_PHP_WRAPPER_END
...
 
 static bool nr_is_invalid_key_val_arr(nr_php_string_hash_key_t* key,
-                                      zval* val) {
-  if (NULL == key || 0 == ZEND_STRING_LEN(key)
-      || 0 == nr_php_is_zval_valid_array(val)
+                                      zval* val,
+                                      const char* key_ident) {
+  if (NULL == key || 0 == ZEND_STRING_LEN(key)) {
+    nrl_warning(NRL_FRAMEWORK, "hookImplementationsMap[%s]: invalid key",
+                key_ident);
```
zsistla, line 612:
```suggestion
                NRSAFESTR(key_ident));
```

agent/fw_drupal8.c
bduranleau-nr: b3531f513dc2811e0bd42420e75683261bd175f2

ZNeumann (approved):

zsistla (approved):

PR #1103, opened 2025-07-15
fix(daemon): aggregate packages throughout harvest

bduranleau-nr:
Fixes a bug in the daemon behavior where only the last VM package data payload from the agent was harvested by the daemon. See this line to better understand the general behavior at play: https://github.com/newrelic/newrelic-php-agent/blob/765888d9998b6af7897fd43e03c75acb3f5619b8/daemon/internal/newrelic/php_packages.go#L57

This issue was exposed by SOAK testing the composer VM once-per-process package detection, where the composer package data was being overridden by the legacy VM data. The fix required a refactor to how the daemon stores and processes package data, while still observing the need to de-duplicate data sent to the `update_loaded_modules` endpoint for processing. 

The `filterPhpPackages` logic was removed from `App` and placed under the `PhpPackages` struct as the new `Filter` function. 

The former `data` `JSONString` field has been converted into a `PhpPackagesKey` `slice` to store all of the package data sent by the agent for a single harvest cycle. Data received from the agent is no longer stored as JSON, but unmarshalled and verified before the daemon accepts it and stores it as a `PhpPackagesKey` `slice` with the raw name and version. Invalid package data received from the agent will no longer be stored in `data` by default and is instead discarded. To store the filtered package data, a new `filteredPkgs` `PhpPackagesKey` `slice` field has been added to the `PhpPackages` struct.

The responsibility of marshalling the final json payload for the backend endpoint has been consolidated into the `CollectorJSON` function for simpler data handling, and better accessibility for testing. `SetPhpPackages` has been removed and logic consolidated into the parent caller, `AddPhpPackagesFromData`.

### DIFF
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -221,6 +221,7 @@ func (t *Test) HandlePHPModules(php_executable string, ctx *Context) map[string]
 }
 
 func (t *Test) MakeRun(ctx *Context) (Tx, error) {
+
 	// we don't support running C tests - assert this so we can
 	// troubleshoot if we try to run a C test
 	if t.IsC() {
@@ -834,20 +835,22 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 	}
 }
 
-var MetricScrubRegexps = []*regexp.Regexp{
-	regexp.MustCompile(`CPU/User Time`),
-	regexp.MustCompile(`CPU/User/Utilization`),
-	regexp.MustCompile(`Memory/Physical`),
-	regexp.MustCompile(`Supportability/execute/user/call_count`),
-	regexp.MustCompile(`Supportability/execute/allocated_segment_count`),
-	regexp.MustCompile(`Memory/RSS`),
-	regexp.MustCompile(`^Supportability\/Locale`),
-	regexp.MustCompile(`^Supportability\/InstrumentedFunction`),
-	regexp.MustCompile(`^Supportability\/TxnData\/.*`),
-	regexp.MustCompile(`^Supportability/C/NewrelicVersion/.*`),
-	regexp.MustCompile(`^Supportability/PHP/Version/.*`),
-	regexp.MustCompile(`^Supportability/PHP/AgentVersion/.*`),
-}
+var (
+	MetricScrubRegexps = []*regexp.Regexp{
+		regexp.MustCompile(`CPU/User Time`),
+		regexp.MustCompile(`CPU/User/Utilization`),
+		regexp.MustCompile(`Memory/Physical`),
+		regexp.MustCompile(`Supportability/execute/user/call_count`),
+		regexp.MustCompile(`Supportability/execute/allocated_segment_count`),
+		regexp.MustCompile(`Memory/RSS`),
+		regexp.MustCompile(`^Supportability\/Locale`),
+		regexp.MustCompile(`^Supportability\/InstrumentedFunction`),
+		regexp.MustCompile(`^Supportability\/TxnData\/.*`),
+		regexp.MustCompile(`^Supportability/C/NewrelicVersion/.*`),
+		regexp.MustCompile(`^Supportability/PHP/Version/.*`),
+		regexp.MustCompile(`^Supportability/PHP/AgentVersion/.*`),
+	}
+)
 
 func (t *Test) Compare(harvest *newrelic.Harvest) {
 	if nil != t.expect {
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -734,7 +734,7 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 
 	// we don't currently test multiple requests that could result in needing to track package history
 	// or filter packages. An empty map here will allow all packages to be reported unfiltered.
-	harvest.PhpPackages.FilterData(make(map[newrelic.PhpPackagesKey]struct{}))
+	harvest.PhpPackages.Filter(make(map[newrelic.PhpPackagesKey]struct{}))
 	audit, err := newrelic.IntegrationData(harvest.PhpPackages, newrelic.AgentRunID("?? agent run id"), time.Now())
 	if nil != err {
 		t.Fatal(err)
--- a/daemon/internal/newrelic/integration/test.go
+++ b/daemon/internal/newrelic/integration/test.go
@@ -221,7 +221,6 @@ func (t *Test) HandlePHPModules(php_executable string, ctx *Context) map[string]
 }
 
 func (t *Test) MakeRun(ctx *Context) (Tx, error) {
-
 	// we don't support running C tests - assert this so we can
 	// troubleshoot if we try to run a C test
 	if t.IsC() {
@@ -733,6 +732,9 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 		}
 	}
 
+	// we don't currently test multiple requests that could result in needing to track package history
+	// or filter packages. An empty map here will allow all packages to be reported unfiltered.
+	harvest.PhpPackages.FilterData(make(map[newrelic.PhpPackagesKey]struct{}))
 	audit, err := newrelic.IntegrationData(harvest.PhpPackages, newrelic.AgentRunID("?? agent run id"), time.Now())
 	if nil != err {
 		t.Fatal(err)
@@ -763,24 +765,14 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 				len(expectedPackages), len(actualPackages), expectedPackages, actualPackages))
 			return
 		}
-		for i, _ := range expectedPackages {
+		for i := range expectedPackages {
 			var matchingIdx int = -1
 			for j, pkg := range actualPackages {
-				//fmt.Printf("Comparing %s to %s\n", pkg.Name, expectedPackages[i].Name)
 				if pkg.Name == expectedPackages[i].Name {
-					//fmt.Printf("Match - index = %d\n", j)
 					matchingIdx = j
 					break
 				}
 			}
-
-			//fmt.Printf("MatchingIdx: %d\n", matchingIdx)
-			//fmt.Printf("expectedPatckages[%d]: %+v\n", i, expectedPackages[i])
-			// if -1 != matchingIdx {
-			// 	fmt.Printf("actualPackages[%d]: %+v\n", matchingIdx, actualPackages[matchingIdx])
-			// } else {
-			// 	fmt.Printf("no match in actualPackages!\n")
-			// }
 
 			if -1 != matchingIdx {
 				testPackageNameOnly := false
@@ -823,7 +815,7 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 		}
 
 		// create notes for all packages in the actual list not in the expected list
-		for ii, _ := range actualPackages {
+		for ii := range actualPackages {
 			var found bool = false
 			for _, pkg := range expectedPackages {
 				if pkg.Name == actualPackages[ii].Name {
@@ -842,22 +834,20 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 	}
 }
 
-var (
-	MetricScrubRegexps = []*regexp.Regexp{
-		regexp.MustCompile(`CPU/User Time`),
-		regexp.MustCompile(`CPU/User/Utilization`),
-		regexp.MustCompile(`Memory/Physical`),
-		regexp.MustCompile(`Supportability/execute/user/call_count`),
-		regexp.MustCompile(`Supportability/execute/allocated_segment_count`),
-		regexp.MustCompile(`Memory/RSS`),
-		regexp.MustCompile(`^Supportability\/Locale`),
-		regexp.MustCompile(`^Supportability\/InstrumentedFunction`),
-		regexp.MustCompile(`^Supportability\/TxnData\/.*`),
-		regexp.MustCompile(`^Supportability/C/NewrelicVersion/.*`),
-		regexp.MustCompile(`^Supportability/PHP/Version/.*`),
-		regexp.MustCompile(`^Supportability/PHP/AgentVersion/.*`),
-	}
-)
+var MetricScrubRegexps = []*regexp.Regexp{
+	regexp.MustCompile(`CPU/User Time`),
+	regexp.MustCompile(`CPU/User/Utilization`),
+	regexp.MustCompile(`Memory/Physical`),
+	regexp.MustCompile(`Supportability/execute/user/call_count`),
+	regexp.MustCompile(`Supportability/execute/allocated_segment_count`),
+	regexp.MustCompile(`Memory/RSS`),
+	regexp.MustCompile(`^Supportability\/Locale`),
+	regexp.MustCompile(`^Supportability\/InstrumentedFunction`),
+	regexp.MustCompile(`^Supportability\/TxnData\/.*`),
+	regexp.MustCompile(`^Supportability/C/NewrelicVersion/.*`),
+	regexp.MustCompile(`^Supportability/PHP/Version/.*`),
+	regexp.MustCompile(`^Supportability/PHP/AgentVersion/.*`),
+}
 
 func (t *Test) Compare(harvest *newrelic.Harvest) {
 	if nil != t.expect {
--- a/daemon/internal/newrelic/php_packages.go
+++ b/daemon/internal/newrelic/php_packages.go
@@ -10,8 +10,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"time"
-
-	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/log"
 )
 
 type PhpPackagesKey struct {
@@ -22,7 +20,7 @@ type PhpPackagesKey struct {
 // phpPackages represents all detected packages reported by an agent.
 type PhpPackages struct {
 	numSeen      int
-	data         []JSONString
+	data         []PhpPackagesKey
 	filteredPkgs []PhpPackagesKey
 }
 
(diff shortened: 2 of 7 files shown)


daemon/internal/newrelic/integration/test.go
```diff
@@ -763,25 +765,15 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 				len(expectedPackages), len(actualPackages), expectedPackages, actualPackages))
 			return
 		}
-		for i, _ := range expectedPackages {
```
zsistla, line 766: Why was the `_` removed?  Since it's value, it isn't explicitly needed to be omitted.  Is there a convention followed in our code?  Is there a compelling reason to remove it?  Does `go` have any standards around this?

Although, why not just do 
`i, ePkg := range expectedPackages {`

then use `ePkg.Name` directly [here](https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-c7f1af7665ab0865f15e79eeba7e73a7295fe66e0ff9205b9781c7b0d7c75faeR771)?

daemon/internal/newrelic/integration/test.go
```diff
@@ -763,25 +765,15 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
...
+		for i := range expectedPackages {
 			var matchingIdx int = -1
 			for j, pkg := range actualPackages {
-				//fmt.Printf("Comparing %s to %s\n", pkg.Name, expectedPackages[i].Name)
 				if pkg.Name == expectedPackages[i].Name {
-					//fmt.Printf("Match - index = %d\n", j)
```
zsistla, line 771: Were all of these comments left in previously for ease of future debugging efforts?

daemon/internal/newrelic/integration/test.go
```diff
@@ -823,7 +815,7 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 		}
 
 		// create notes for all packages in the actual list not in the expected list
-		for ii, _ := range actualPackages {
+		for ii := range actualPackages {
```
zsistla, line 819: why not just use 
`for ii, aPkg:= range actualPackages {`

Then use aPgk directly later [here](https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-c7f1af7665ab0865f15e79eeba7e73a7295fe66e0ff9205b9781c7b0d7c75faeR821) and [here](https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-c7f1af7665ab0865f15e79eeba7e73a7295fe66e0ff9205b9781c7b0d7c75faeR827)?

daemon/internal/newrelic/integration/test.go
bduranleau-nr: This is an auto-formatting change. Correct, you do not need the `_` for looping with an `int` value.

> Although, why not just do
i, ePkg := range expectedPackages {
then use ePkg.Name directly [here](https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-c7f1af7665ab0865f15e79eeba7e73a7295fe66e0ff9205b9781c7b0d7c75faeR771)?

I did not write the logic for this section of code and changing it beyond what the autoformatter caught and removing comments is outside the scope of this PR.

daemon/internal/newrelic/integration/test.go
```diff
@@ -842,22 +834,20 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
...
-		regexp.MustCompile(`Memory/RSS`),
-		regexp.MustCompile(`^Supportability\/Locale`),
-		regexp.MustCompile(`^Supportability\/InstrumentedFunction`),
-		regexp.MustCompile(`^Supportability\/TxnData\/.*`),
-		regexp.MustCompile(`^Supportability/C/NewrelicVersion/.*`),
-		regexp.MustCompile(`^Supportability/PHP/Version/.*`),
-		regexp.MustCompile(`^Supportability/PHP/AgentVersion/.*`),
-	}
-)
+var MetricScrubRegexps = []*regexp.Regexp{
```
zsistla: Is this just a formatting change?

daemon/internal/newrelic/integration/test.go
bduranleau-nr: I don't know the origin of these comments.

daemon/internal/newrelic/php_packages.go
```diff
@@ -20,8 +19,9 @@ type PhpPackagesKey struct {
 
 // phpPackages represents all detected packages reported by an agent.
```
zsistla: Please add to the comment explaining the updated PhpPackages struct.

daemon/internal/newrelic/integration/test.go
bduranleau-nr: Changing this logic is outside the scope of this PR. This was an auto-change generated by Go's formatter.

daemon/internal/newrelic/integration/test.go
bduranleau-nr: Yes.

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +34,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
+// the JSON format received from the agent is:
+//
+//	[["package_name","version",{}],...]
+//
+// for each entry, assign the package name and version to the `PhpPackagesKey`
+// struct and use the key to verify data does not exist in the map. If the
+// key does not exist, add it to the map and the slice of filteredPkgs to be
+// sent in the current harvest.
+func (packages *PhpPackages) Filter(pkgHistory map[PhpPackagesKey]struct{}) {
+	if packages.data == nil {
```
zsistla: Shouldn't we first check if `packages` is not null and output a message like we did [here](https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-5a6f19b43bcbc6817e348e335fabbac40594503c9cc8cea5669e5ae57ad7e9c6L47)?

bduranleau-nr: Yes- added here 0825866a5f9a29d777faa010c1ddd45d0bd4ac1e

daemon/internal/newrelic/php_packages.go
bduranleau-nr: Let me know if this makes sense 261196c931a8f8d514b40549b10cb6476e3d4a24

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +34,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
+		return
+	}
 
-	if nil == packages {
-		return fmt.Errorf("packages is nil!")
+	for _, pkgKey := range packages.data {
+		_, ok := pkgHistory[pkgKey]
+		if !ok {
+			pkgHistory[pkgKey] = struct{}{}
```
zsistla, line 72: Why an empty struct?

daemon/internal/newrelic/integration/test.go
zsistla: Could you mark what is autoformatter?

daemon/internal/newrelic/integration/test.go
zsistla: Did the autoformatter remove them?

daemon/internal/newrelic/php_packages.go
zsistla: Much better, thanks!

daemon/internal/newrelic/integration/test.go
```diff
@@ -842,22 +834,20 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
...
-		regexp.MustCompile(`Supportability/execute/allocated_segment_count`),
-		regexp.MustCompile(`Memory/RSS`),
-		regexp.MustCompile(`^Supportability\/Locale`),
-		regexp.MustCompile(`^Supportability\/InstrumentedFunction`),
-		regexp.MustCompile(`^Supportability\/TxnData\/.*`),
-		regexp.MustCompile(`^Supportability/C/NewrelicVersion/.*`),
-		regexp.MustCompile(`^Supportability/PHP/Version/.*`),
-		regexp.MustCompile(`^Supportability/PHP/AgentVersion/.*`),
-	}
-)
```
bduranleau-nr: Formatting

daemon/internal/newrelic/integration/test.go
```diff
@@ -823,7 +815,7 @@ func (t *Test) comparePhpPackages(harvest *newrelic.Harvest) {
 		}
 
 		// create notes for all packages in the actual list not in the expected list
-		for ii, _ := range actualPackages {
```
bduranleau-nr, line 826: Formatting

daemon/internal/newrelic/integration/test.go
bduranleau-nr, line 766: Formatting

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +37,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
 		return []byte(`["Jars",[]]`), nil
 	}
 
-	buf := &bytes.Buffer{}
+	var buf bytes.Buffer
 
 	estimate := 512
 	buf.Grow(estimate)
 	buf.WriteByte('[')
-	buf.WriteString("\"Jars\",")
```
zsistla, line 79: Is this an autoformatter change?

daemon/internal/newrelic/processor_test.go
```diff
@@ -19,9 +19,11 @@ import (
 	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/utilization"
```
bduranleau-nr: Formatting (for this whole file)- 
there was initially a change made to this file that was committed and reverted, but the autoformatter changes persisted

daemon/internal/newrelic/integration/test.go
bduranleau-nr: I did while I was debugging a problem with the tests

daemon/internal/newrelic/php_packages_test.go
```diff
@@ -23,68 +22,88 @@ func TestNewPhpPackages(t *testing.T) {
...
+	emptyStr := []byte(``)
+	emptyNameStr := []byte(`[["","1.2.3",{}]]`)
+	missingElem := []byte(`[["package_a", "1.2.3"]]`)
+	excessElem := []byte(`[["package_a", "1.2.3", "x.y.z", {}]]`)
 
-	err := nilpkg.SetPhpPackages(validData)
-	if !strings.Contains(err.Error(), "packages is nil") {
-		t.Fatalf("Expected error 'packages is nil!', got '%s'", err.Error())
+	err := nilpkg.AddPhpPackagesFromData(validPkgData)
+	if err == nil {
```
zsistla, line 39: General comment for all `err == nil` checks.
This is just checking that error is not nil, it's not checking for the actual error message like the [original test](https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-371705fffc11618f2e231e986a0609103a0a8fab90c6ff76aedfe2ec82e41042L32) so if an error is caused somewhere else, the test wouldn't catch it.  The check for the actual message would verify the correct error condition is triggered by the test.

daemon/internal/newrelic/php_packages.go
bduranleau-nr: We store an empty struct as the value to indicate that an index has been hit, since we don't care about the actual value but instead the existence of the package key in the map. This is a trick in Go to populate a map where you are only concerned with the existence of a key.

daemon/internal/newrelic/php_packages.go
bduranleau-nr: No, this is intentional

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +37,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
+	if packages == nil || packages.data == nil {
+		return
+	}
 
-	if nil == packages {
-		return fmt.Errorf("packages is nil!")
+	for _, pkgKey := range packages.data {
+		_, ok := pkgHistory[pkgKey]
+		if !ok {
+			pkgHistory[pkgKey] = struct{}{}
```
zsistla, line 72: What is the behavior if the harvest fails?

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +37,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
+// the map of packages should persist for the duration of the
+// current connection
+//
+// the JSON format received from the agent is:
+//
+//	[["package_name","version",{}],...]
+//
+// for each entry, assign the package name and version to the `PhpPackagesKey`
+// struct and use the key to verify data does not exist in the map. If the
+// key does not exist, add it to the map and the slice of filteredPkgs to be
```
zsistla, line 62: `add it to the map`
is this the `App` structure map of PHP Packages the reporting app has encountered?

daemon/internal/newrelic/php_packages_test.go
```diff
@@ -163,3 +184,108 @@ func TestPackagesEmpty(t *testing.T) {
...
+func TestFilterPackageData(t *testing.T) {
+	info := AppInfo{}
+	app := NewApp(&info)
+	pkg := NewPhpPackages()
+	expect_a := PhpPackagesKey{"package_a", "1.2.3"}
+	expect_b := PhpPackagesKey{"package_b", "1.2.3"}
+	expect_c := PhpPackagesKey{"package_c", "1.2.3"}
+
+	// Test nil package data
+	pkg.Filter(app.PhpPackages)
```
zsistla, line 231: Doesn't `Filter` also now modify the `App` structure map of PHP Packages the reporting app has encountered?  Shouldn't we also verify the correctness of `app.PhpPackages` as well in the test?

bduranleau-nr: We verify it here by feeding it duplicate data https://github.com/newrelic/newrelic-php-agent/pull/1103/files#diff-371705fffc11618f2e231e986a0609103a0a8fab90c6ff76aedfe2ec82e41042R291

daemon/internal/newrelic/php_packages_test.go
bduranleau-nr: b4afa3f723cae3bfaf0d32bf89e833c6b91c97e6

daemon/internal/newrelic/php_packages.go
bduranleau-nr: This is referencing the map structure stored under the `App`, yes.

daemon/internal/newrelic/php_packages.go
bduranleau-nr: A failed harvest will result in dropped package info. The map does not get reset.

daemon/internal/newrelic/php_packages_test.go
zsistla: Would explicitly checking size or that a key exists be beneficial?

daemon/internal/newrelic/php_packages.go
zsistla: So if we call `Filter` and then harvest ends up getting dropped, hasn't it already modified the app map even though we didn't end up sending the filtered list?

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +37,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
+
+		version, ok := pkg[1].(string)
+		if !ok || len(version) == 0 {
+			return fmt.Errorf("unable to parse package version")
+		}
+
+		packages.data = append(packages.data, PhpPackagesKey{name, version})
+	}
+
 	packages.numSeen = 1
```
mfulb, line 117: I'm not sure what this field does any more - but if we keep it should it be incremented by the number of packages created in the `for` loop above?

daemon/internal/newrelic/php_packages.go
bduranleau-nr: Copying discussion here- tracking this behavior in an internal ticket to be fixed in a separate PR

daemon/internal/newrelic/php_packages.go
```diff
@@ -34,51 +37,112 @@ func (packages *PhpPackages) NumSaved() float64 {
...
+			return fmt.Errorf("invalid php package json structure: %+v", pkg)
+		}
+
+		name, ok := pkg[0].(string)
+		if !ok || len(name) == 0 {
+			return fmt.Errorf("unable to parse package name")
+		}
+
+		version, ok := pkg[1].(string)
+		if !ok || len(version) == 0 {
```
zsistla, line 106: What would be behavior for legacy package that is sent without a version?

daemon/internal/newrelic/php_packages_test.go
bduranleau-nr: Adding a check for map size here: 54b3e249a733c859081dcb30b80cd0fdd2802122

daemon/internal/newrelic/php_packages.go
bduranleau-nr: legacy will send a blank space (`" "` instead of `""`), not a completely empty field, so this will not trigger the failure state.

mfulb (approved):
Thanks Brian!

daemon/internal/newrelic/integration/test.go
lavarou: These debug messages were added in https://github.com/newrelic/newrelic-php-agent/pull/962/commits/76334f27a0e16e706faeaff04e71830e4ae3ffbf and commented out in https://github.com/newrelic/newrelic-php-agent/pull/962/commits/b925fed7a95350525fc3e320fde2907aac39ae12.

daemon/internal/newrelic/php_packages.go
```diff
@@ -7,21 +7,24 @@ package newrelic
...
 
 type PhpPackagesKey struct {
 	Name    string
 	Version string
 }
 
-// phpPackages represents all detected packages reported by an agent.
+// PhpPackages represents all detected packages reported by an agent
+// for a harvest cycle, as well as the filtered list of packages not
+// yet seen by the daemon for the lifecycle of the current process to
```
lavarou: "the lifecycle of the current process" - could this comment be more specific? Is it lifecycle of the current php process, daemon process, or some other process?

bduranleau-nr: daemon process d94b0d099e19e8f081bd5311789347ad85ddcd03

daemon/internal/newrelic/php_packages.go
```diff
@@ -7,78 +7,142 @@ package newrelic
...
-	numSeen int
-	data    JSONString
+	numSeen      int
+	data         []PhpPackagesKey
+	filteredPkgs []PhpPackagesKey
 }
 
-// NumSeen returns the total number PHP packages payloads stored.
+// NumSaved returns the total number PHP packages payloads stored.
 // Should always be 0 or 1.  The agent reports all the PHP
```
lavarou: Is "Should always be 0 or 1" still valid? Now that daemon potentially holds on to multiple php packages payloads received from the agent, shouldn't this represent the number of times the daemon received php packages payload from the agent in a given harvest cycle?

daemon/internal/newrelic/php_packages.go
lavarou: I'm also interested in how this gets resolved. See related comment: https://github.com/newrelic/newrelic-php-agent/pull/1103/files#r2261468915

daemon/internal/newrelic/php_packages.go
lavarou: @zsistla Some more explanation what this change is about can be found [here](https://go.dev/ref/spec#String_literals):
> Raw string literals are character sequences between back quotes, as in \`foo\`. Within the quotes, any character may appear except back quote.

This means that double quote (`"`) does not have to be escaped with `\`.

lavarou (dismissed):

daemon/internal/newrelic/php_packages.go
bduranleau-nr: As far as I can tell this value isn't sent to the backend or used anywhere other than the conditional checks as an indicator that `packages` have been set for the current harvest. I don't believe it is used in any kind of metrics or instrumentation data, so maybe we should consider changing it entirely to a `bool`. I think this was added as an artifact of the other Event types that report seen vs saved, but it doesn't really fit cleanly into the packages VM logic.

zsistla (dismissed):

daemon/internal/newrelic/php_packages.go
lavarou: At least the comment should be updated because it is misleading.

bduranleau-nr: 906481a4c4b29d27ec46b24541dfa25bd029ff6e

lavarou (approved):

PR #1105, opened 2025-08-05
Cache composer api packages so these are considered each requst

mfulb:


agent/php_txn.c
```diff
@@ -1151,6 +1151,24 @@ nr_status_t nr_php_txn_begin(const char* appnames,
...
+      && (NULL != NR_PHP_PROCESS_GLOBALS(composer_php_packages))) {
+    nrl_verbosedebug(NRL_FRAMEWORK,
+                     "composer packages already detected, using cached values");
+#if 0
+    nr_php_packages_destroy(&NRPRG(txn)->php_packages);
+    NRPRG(txn)->php_packages
+        = nr_php_packages_clone(NR_PHP_PROCESS_GLOBALS(composer_php_packages));
+    nrl_verbosedebug(NRL_FRAMEWORK, "composer packages cloned from cache");
+#else
+    nrl_verbosedebug(NRL_INSTRUMENT, "passing known PHP packages = %p",
```
mfulb, line 1165: This code needs some wrapping conditional so it isn't used if agent is running Composer API every request.

Probably wrapping it with:
```
if (NR_PHP_PROCESS_GLOBALS(composer_api_per_process_detection)) {
...
}
```

mfulb:
This approach yielded reasonable performance, but the use of process global to store package info persistently is considered problematic.  A different approach was discussed and finally adopted.

PR #1107, opened 2025-08-07
fix(agent): fix legacy package detection issues

lavarou:
When Composer runtime API is used to get packages information, the
agent should not generate packages information using legacy methods.
This solves the problem of package data detected with the help of
Composer's runtime API not to be sent to the backend in environments
using “once-per-process” setting. The reason for is that the first 
request generates package information using Composer's runtime API
but every next request only generates package information using legacy
method. This in conjunction with the fact that daemon is only sending
the last package information received from the daemon leads to a
situation that package information returned by Composer runtime API
is lost and never makes it to the backend.

### DIFF
--- a/agent/lib_composer.c
+++ b/agent/lib_composer.c
@@ -69,23 +69,26 @@ static int nr_execute_handle_autoload_composer_init(const char* vendor_path) {
   return NR_SUCCESS;
 }
 
-static bool nr_execute_handle_autoload_composer_get_packages_information(
+static nr_composer_api_call_result_t
+nr_execute_handle_autoload_composer_get_packages_information(
     const char* vendor_path) {
   zval retval;  // This is used as a return value for zend_eval_string.
                 // It will only be set if the result of the eval is SUCCESS.
   int result = FAILURE;
+  nr_composer_api_call_result_t api_call_result
+      = NR_COMPOSER_API_CALL_RESULT_UNSET;
 
   // nrunlikely because this should alredy be ensured by the caller
   if (nrunlikely(!NRINI(vulnerability_management_package_detection_enabled))) {
     // do nothing when collecting package information for vulnerability
     // management is disabled
-    return false;
+    return NR_COMPOSER_API_CALL_RESULT_INVALID_USE;
   }
 
   // nrunlikely because this should alredy be ensured by the caller
   if (nrunlikely(!NRINI(vulnerability_management_composer_api_enabled))) {
     // do nothing when use of composer to collect package info is disabled
-    return false;
+    return NR_COMPOSER_API_CALL_RESULT_INVALID_USE;
   }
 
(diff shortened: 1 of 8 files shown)


agent/lib_composer.c
```diff
@@ -69,23 +69,25 @@ static int nr_execute_handle_autoload_composer_init(const char* vendor_path) {
...
-    return;
+    return NR_COMPOSER_API_STATUS_INVALID_USE;
   }
 
   // nrunlikely because this should alredy be ensured by the caller
   if (nrunlikely(!NRINI(vulnerability_management_composer_api_enabled))) {
     // do nothing when use of composer to collect package info is disabled
-    return;
+    return NR_COMPOSER_API_STATUS_INVALID_USE;
   }
```
bduranleau-nr, line 91: Since this method is static to this file and the caller (`nr_composer_handle_autoload`) is only called if [these checks](https://github.com/newrelic/newrelic-php-agent/blob/765888d9998b6af7897fd43e03c75acb3f5619b8/agent/php_execute.c#L884-L892) pass, I would suggest removing these conditional checks here entirely, which could also allow us to remove the `NR_COMPOSER_API_STATUS_INVALID_USE` `enum`

zsistla: This is a valid point.

I noticed this earlier when thinking about [this line](https://github.com/newrelic/newrelic-php-agent/pull/1107/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R880) and wondered what the behavior would be depending on the other ENUMs that this could generate. In particular I thought about the behavior with the invalid_use ENUM, but when I looked through the code above it, I saw this would never get set because it would exit somewhere around lines [855-864](https://github.com/newrelic/newrelic-php-agent/pull/1107/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625L855-L864) before even calling this function.

lavarou: > I noticed this earlier when thinking about [this line](https://github.com/newrelic/newrelic-php-agent/pull/1107/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R880) and wondered what the behavior would be depending on the other ENUMs that this could generate. In particular I thought about the behavior with the invalid_use ENUM [...]

[This line](https://github.com/newrelic/newrelic-php-agent/pull/1107/files#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R880) means that Composer's runtime API will not be used again regardless of the outcome of an attempt of calling Composer's runtime API as long as there was an attempt to call it. So the only interesting status is `NR_COMPOSER_API_STATUS_UNSET`.

PR #1122, opened 2025-09-20
fix(agent): Fix Laravel queue job exception handling

zsistla:
Multiverse and soak tests revealed a bug with naming Job transactions when the job throws an exception.

The issue is that we [key off](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_laravel_queue.c#L337) of raiseBeforeJobEvent and raiseAfterJobEvent, but in the case of failure due to exception(https://github.com/illuminate/queue/blob/12.x/Worker.php#L447 ), raiseAfterJobEvent doesn't get called, so we never do the final end txn to close things off [here](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_laravel_queue.c#L355) and when the next retry calls [raiseBeforeJobEvent](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/fw_laravel_queue.c#L300) again it ends the txn and discards anything that happened before (in this case the logs/exceptions/segment of the previous retry attempt).

Solution is to wrap the following so we can handle the exception case in the after/clean callbacks.
 {Illuminate\Queue\Worker} function={process} async
 https://github.com/illuminate/queue/blob/12.x/Worker.php#L425 

Illuminate\Queue\SyncQueue::executeJob/push for sync
https://github.com/illuminate/queue/blob/12.x/SyncQueue.php#L123


agent/fw_laravel_queue.c
```diff
@@ -610,8 +644,6 @@ NR_PHP_WRAPPER(nr_laravel_queue_worker_process) {
     nr_free(txn_name);
   }
 
-  NR_PHP_WRAPPER_CALL;
-
```
lavarou, line 614: Why was this removed? `nr_laravel_queue_worker_process` is only used for PHPs < 8.0 and `NR_PHP_WRAPPER_CALL` is needed here for the job to actually execute.

agent/fw_laravel_queue.c
```diff
@@ -650,13 +682,13 @@ NR_PHP_WRAPPER(nr_laravel_queue_worker_process) {
...
+   * End the useless transaction and then start a new transaction so our
    * instrumentation continues to fire, knowing that we'll ignore that
-   * transaction either when Worker::process() is called again or when
-   * WorkCommand::handle() exits.
+   * transaction if the worker continues when Worker::process() is called again.
    */
-  nr_php_txn_end(0, 0 TSRMLS_CC);
+  nr_php_txn_end(1, 0 TSRMLS_CC);
   nr_php_txn_begin(NULL, NULL TSRMLS_CC);
+  NR_PHP_WRAPPER_CALL;
```
lavarou: Why was this changed? `nr_laravel_queue_worker_process` is only used for PHPs < 8.0 and NR_PHP_WRAPPER_CALL must be called earlier for the job to actually execute and for the agent to capture job's execution as background transaction.

agent/fw_laravel_queue.c
```diff
@@ -234,124 +234,158 @@ static char* nr_laravel_queue_job_txn_name(zval* job TSRMLS_DC) {
...
     }
-
-    nr_laravel_queue_set_cat_txn(job TSRMLS_CC);
-
-    nr_txn_set_path("Laravel", NRPRG(txn), txn_name, NR_PATH_TYPE_CUSTOM,
-                    NR_OK_TO_OVERWRITE);
   }
-  nr_free(txn_name);
-  nr_php_arg_release(&job);
   NR_PHP_WRAPPER_CALL;
```
lavarou: `NR_PHP_WRAPPER_CALL` is a no-op and can be removed.

zsistla: Not exactly a noop: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L290
And it will get call [automatically](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L274) anyway if it isn't explicitly called so good to leave it here to remind that it is called.

agent/fw_laravel_queue.c
```diff
@@ -234,124 +234,158 @@ static char* nr_laravel_queue_job_txn_name(zval* job TSRMLS_DC) {
...
+      job exceptions handler and then thrown again AFTER we've already ended the
+      txn for the job
+      *
+      */
+      zval exception;
+      ZVAL_OBJ(&exception, EG(exception));
+      nr_status_t st;
+      st = nr_php_error_record_exception(
+          NRPRG(txn), &exception, 50, false /* add to segment */,
+          NULL /* use default prefix */, &NRPRG(exception_filters));
```
lavarou: Why `nr_php_error_record_exception` for PHPs >= 8.0 does not use the same arguments as for PHPs < 8.0 - see [here](https://github.com/newrelic/newrelic-php-agent/blob/R12.1/agent/fw_laravel_queue.c#L643-L646)? I.e. set `priority` to `NR_PHP_ERROR_PRIORITY_UNCAUGHT_EXCEPTION`, set `add_to_segment` to `true` and set `prefix` to `"Unhandled exception within Laravel Queue job: "`.

zsistla: We don't need to add it to the segment for oapi because it was already added here: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_execute.c#L1969

We just need to add it to the txn.  I noticed the priority and prefix as well when I was looking another comment of yours and already updated it here: [1ec2a7e](https://github.com/newrelic/newrelic-php-agent/pull/1122/commits/1ec2a7e5ed837df1fd64f6f1391d2bdf1a881a81)

agent/fw_laravel_queue.c
lavarou: This block of code is when `ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO` and `!defined OVERWRITE_ZEND_EXECUTE_DATA`, so https://github.com/newrelic/newrelic-php-agent/blob/R12.1/agent/php_wrapper.h#L292 is a no-op. [Remainder of the code](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L295-L302) in this macro is not important in this context. Using `NR_PHP_WRAPPER_CALL` in code for PHPs >= 8.0 (when agent is using Observer API to hook into Zend Engine) is confusing.

agent/fw_laravel_queue.c
zsistla: Great catch, thanks!!
[1ec2a7e](https://github.com/newrelic/newrelic-php-agent/pull/1122/commits/1ec2a7e5ed837df1fd64f6f1391d2bdf1a881a81)

agent/fw_laravel_queue.c
zsistla: As mentioned, NR_PHP_WRAPPER_CALL gets called in NR_PHP_WRAPPER_END anyway, and [this block of code](https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_wrapper.h#L295-L302) is definitely run in this particular case because we have started/ended a txn in the wrapper.
It's needed to clear auto_segment which would otherwise be pointing to a segment that's already been freed.

zsistla: It doesn't really matter, and I can remove it because it will get trigged in func end, but it is still going to get called and execute that code block for this particular case.
[69509f2](https://github.com/newrelic/newrelic-php-agent/pull/1122/commits/69509f2d4bcf6acdedeef62c52f27cd65bd371ba)

agent/fw_laravel_queue.c
zsistla: [c473d6c](https://github.com/newrelic/newrelic-php-agent/pull/1122/commits/c473d6c16c32b1d2aab593895fb745688adf32a8)

lavarou (dismissed):

agent/fw_laravel_queue.c
```diff
@@ -234,124 +234,158 @@ static char* nr_laravel_queue_job_txn_name(zval* job TSRMLS_DC) {
...
+ * Illuminate\\Queue\\SyncQueue::raiseBeforeJobEvent
+ * Raise the before queue job event.
+ *
+ * @param  \Illuminate\Contracts\Queue\Job  $job
+ * @return void
+ *  protected function raiseBeforeJobEvent(Job $job)
+ *
+ * The reason these functions are used for txn naming:
+ * 1. It has been a consistent API called directly before the Job across Laravel
+ * going back to at leaset 6
```
hahuja2:
```suggestion
 * going back to at least 6
```

agent/tests/test_fw_laravel_queue.c
```diff
@@ -0,0 +1,361 @@
...
+
+  tlib_pass_if_not_null("Txn should not be null at the start of the test.",
+                        NRPRG(txn));
+  txn_time = nr_txn_start_time(NRPRG(txn));
+
+  nr_txn_set_path("ToBeDiscarded", NRPRG(txn), "Farewell", NR_PATH_TYPE_CUSTOM,
+                  NR_OK_TO_OVERWRITE);
+  tlib_pass_if_str_equal("Path should exist", "Farewell", NRTXN(path));
+
+  /* Create the mocked Worker and call process*/
```
hahuja2:
```suggestion
  /* Create the mocked Worker and call trycatchExecuteJob, which calls executeJob*/
```

hahuja2 (dismissed):

PR #1124, opened 2025-09-29
feat(agent): Add nonce to RUM scripts

poespas:
We ran into the problem that our users can't have a CSP-secure website because of the injected inline RUM scripts without a nonce-attribute, this PR will be able to add the nonce when it's set through the php ini configuration:

Example:
```
newrelic.browser_monitoring.nonce_rum_header="abc123"
newrelic.browser_monitoring.nonce_rum_footer="abc456"
```


hahuja2:
Hi @poespas, thank you for taking the time to open this PR! The functionality to add `nonce` attribute to the returned JavaScript string can be achieved by utilizing existing API functions `newrelic_get_browser_timing_header` and `newrelic_get_browser_timing_footer` where you can pass `false` to both API functions so they return the raw JavaScript string without `<script>` tags. You can then build the tags yourself and add custom attributes such as `nonce` attribute. Here is documentation explaining how to do this: https://docs.newrelic.com/docs/apm/agents/php-agent/features/browser-monitoring-php-agent/#nonce-attribute.

PR #1132, opened 2025-11-12
feat(agent): add supportability metric for laravel horizon and queue

hahuja2:
This PR does the following:

- Wrap `Laravel\Horizon\Console\WorkCommand` to check if Laravel Horizon is being used and if it is, then set `is_horizon_used` global flag to `true`.
- Check value of `is_horizon_used` flag which will determine what supportability metric is sent. (`Supportability/library/Laravel/Horizon/used` or `Supportability/library/Laravel/Queue/used`)

### DIFF
--- a/agent/php_globals.h
+++ b/agent/php_globals.h
@@ -80,7 +80,7 @@ typedef struct _nrphpglobals_t {
                                               enabled */
   nr_composer_api_status_t composer_api_status; /* Composer API's status */
   char* docker_id; /* 64 byte hex docker ID parsed from /proc/self/mountinfo */
-  bool is_horizon_used; /* Set to true if Laravel Horizon is used */
+  bool laravel_horizon_worker_used; /* Set to true if Laravel Horizon is used */
 
   /* Original PHP callback pointer contents */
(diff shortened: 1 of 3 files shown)


agent/php_globals.h
```diff
@@ -80,6 +80,7 @@ typedef struct _nrphpglobals_t {
                                               enabled */
   nr_composer_api_status_t composer_api_status; /* Composer API's status */
   char* docker_id; /* 64 byte hex docker ID parsed from /proc/self/mountinfo */
+  bool is_horizon_used; /* Set to true if Laravel Horizon is used */
```
lavarou: I suggest adding more context to the name of this global: `laravel_horizon_worker_used`.

zsistla: I agree, more context is needed in the var name.

hahuja2: Good point, changed: [1604a0e](https://github.com/newrelic/newrelic-php-agent/pull/1132/commits/1604a0e718dff79290cecf3c523312c6c7d40702)

lavarou (approved):

zsistla (approved):

PR #1140, opened 2025-12-09
feat(axiom): Provide libc name+version and os distribution name and version

zsistla:
This is supplemental to the legacy “OS version” information so if we can’t resolve the libc or distribution information, it’s okay, legacy is still there. 

In addition to the "OS version" which is still being sent unchanged, this PR now also sends:
  "OS Distro Name": single name representing the linux distribution like `ubuntu`, `debian`, `centos`, `alpine`, `amzn`, etc.
  "OS Distro Version": single value representing the linux distribution version like `12`, `24.04`,  `3.23.0`, etc.
  "OS LIBC Version": libc name+version (version only for glibc) like `GLIBC 2.39` or `MUSL`

The info is also additionally displayed in the log like so:
2025-12-09 03:07:33.950 +0000 (26653 26653) info: New Relic 12.3.0.0 ("nephrite" - "a17d4aa0fea8") [daemon='@/tmp/newrelic-daemon-26638.sock'  php='8.4.15' zts=no sapi='cli'  pid=26653 ppid=26638 uid=82 euid=82 gid=82 egid=82 backtrace=no startup=init osdistro='alpine 3.22.2' oslibc='MUSL' os='Linux' rel='6.12.54-linuxkit' mach='aarch64' ver='#1 SMP Tue Nov  4 21' node='610d93b01e37']

Expanded the system info structure to hold the additional information for the linux distribution name and version and the libc name and version(for MUSL it is MUSL only no version).  Updated the nr_system_t destroy function to free additional values.
The linux distribution and version are extracted from `etc/os-release`.
The libc version for glibc is found via a system call provided by glibc.
This involved refactoring `nr_system_get_system_information` so that it it still initializes the structure but now calls the following helper functions to populate the fields:
  `nr_system_get_system_info_from_uname`
  `nr_system_get_system_info_from_osrelease`
  `nr_system_get_system_info_libc`

### DIFF
--- a/axiom/tests/test_system.c
+++ b/axiom/tests/test_system.c
@@ -81,6 +81,31 @@ static void test_get_system_info_from_osrelease(void) {
   /* Unexpected Cases */
 
   nr_system_get_system_info_from_osrelease(
+      sys, REFERENCE_DIR "/osrelease_twoversionid_ubuntu_24.04");
+  tlib_pass_if_str_equal(
+      "if valid filename but two versionIds should be last one, VERSION_ID "
+      "sys->distro_version_id should be set",
+      sys->distro_version_id, "24.04");
+  tlib_pass_if_str_equal(
+      "if valid filename  ID sys->distro_id should be "
+      "set",
+      sys->distro_id, "ubuntu");
+  FREE_SYS_DISTRO_VALUES;
+
+  nr_system_get_system_info_from_osrelease(
+      sys, REFERENCE_DIR "/osrelease_twoids_ubuntu_24.04");
+  tlib_pass_if_str_equal(
+      "if valid filename, VERSION_ID "
+      "sys->distro_version_id should be set",
+      sys->distro_version_id, "24.04");
+  tlib_pass_if_str_equal(
+      "if valid filename but 2 IDs, ID should be set to last one, ID "
+      "sys->distro_id should be "
+      "set",
+      sys->distro_id, "ubuntu");
+  FREE_SYS_DISTRO_VALUES;
+
+  nr_system_get_system_info_from_osrelease(
       sys, REFERENCE_DIR "/osrelease_leadingwhitespace_ubuntu_24.04");
   tlib_pass_if_str_equal(
       "if valid filename but leading whitespace, VERSION_ID "
--- a/axiom/util_system.c
+++ b/axiom/util_system.c
@@ -73,28 +73,14 @@ void nr_system_get_system_info_from_osrelease(nr_system_t* sys,
   int len = 0;
   char* value = NULL;
 
-#define REMOVE_LEADING_WHITESPACE                                        \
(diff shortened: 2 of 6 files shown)


axiom/util_system.c
```diff
@@ -39,21 +45,115 @@ char* nr_system_get_hostname(void) {
...
+  const char* libc_version = NULL;
+
+  if (NULL == sys) {
+    return;
+  }
+
+#if defined(__GLIBC__)
+  libc_version = gnu_get_libc_version();
+#else
+/* Currently unable to extract MUSL version. */
```
bduranleau-nr: I would omit the `#else` case here entirely. MUSL devs have made it aggressively clear they don't ever intend to add identifiers into their implementation like `__GLIBC__`

zsistla: removed here [a327cf5](https://github.com/newrelic/newrelic-php-agent/pull/1140/commits/a327cf5950fdadedce1d9afc4269c31944d68246)

axiom/tests/test_system.c
```diff
@@ -20,6 +24,133 @@ static void test_system_get_hostname(void) {
...
+                    sys->distro_id);
+  tlib_pass_if_null("if NULL filename sys->distro_version_id should be NULL",
+                    sys->distro_version_id);
+
+  /* Sending a valid filename without VERSION_ID or ID shouldn't modify sys*/
+  nr_system_get_system_info_from_osrelease(
+      sys, REFERENCE_DIR "/osrelease_missingid_missingversionid");
+  tlib_pass_if_null("if NULL filename sys->distro_id should be NULL",
+                    sys->distro_id);
+  tlib_pass_if_null("if NULL filename sys->distro_version_id should be NULL",
```
bduranleau-nr: These test failure descriptions seem incorrect, since the filename supplied is not `NULL`

zsistla: good catch, updated here: [f711b55](https://github.com/newrelic/newrelic-php-agent/pull/1140/commits/f711b55b4dcaee38051e033fa0affd69122057df)

axiom/util_system.h
```diff
@@ -9,23 +9,65 @@
 #ifndef UTIL_SYSTEM
 #define UTIL_SYSTEM
 
+#define VERSION_ID_STRING "VERSION_ID="
+#define ID_STRING "ID="
+#define VERSION_ID_STRING_LEN strlen(VERSION_ID_STRING)
+#define ID_STRING_LEN strlen(ID_STRING)
```
lavarou: `strlen` is a runtime operation, `sizeof` is calculated at compile time (no overhead at runtime):
```suggestion
#define VERSION_ID_STRING_LEN (sizeof(VERSION_ID_STRING)-1)
#define ID_STRING_LEN (sizeof(ID_STRING)-1)
```

PR #1149, opened 2025-12-22
feat(agent): add support for PHP 8.5

bduranleau-nr:


axiom/tests/test_memory.c
```diff
@@ -13,6 +13,13 @@
 
 #include "tlib_main.h"
 
+// GCC 14 introduced changes that throw a fatal error on allocations of size 0
+// For PHP versions built on GCC 14+, suppress these errors so we may continue
+// to test on older GCC versions.
```
mfulb: The suppression of the errors is so we can actually test on GCC 14+, not older versions.  Older versions will continue to compile this file w/o the #pragma needed for GCC 14+.

files/Dockerfile
```diff
@@ -111,11 +111,17 @@ RUN \
 # memcache
 # Pre 8.0 requires 4.0.5.2
 RUN \
-  php_cmp=$(php -r "echo version_compare(PHP_VERSION, '8.0.0', '>');"); \
+  php_cmp=$(php -r "echo version_compare(PHP_VERSION, '8.5.0', '>=');"); \
   if [ "$php_cmp" = 1 ]; then \
-    pecl install memcache && docker-php-ext-enable memcache; \
+  # pecl install memcache && docker-php-ext-enable memcache; \
+    echo "unsupported"; \
```
mfulb:
```suggestion
    echo "memcache unsupported for PHP 8.5+"; \
```

axiom/tests/test_memory.c
bduranleau-nr: [done](https://github.com/newrelic/newrelic-php-agent/pull/1149/commits/5adc8c27ff3f19350e4b59dd393ed978d39926fd)

Issue #1148, opened 2025-12-18
Stop calling Request::get on Symfony 7.4 / 8.0 as it is deprecated


ruudk:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

Since Symfony 7.4 (released a few weeks ago) calling `get` on Request is deprecated, see: https://symfony.com/blog/new-in-symfony-7-4-request-class-improvements#deprecated-the-get-method

But it seems that New Relic is calling this a lot, see:
https://github.com/newrelic/newrelic-php-agent/blob/f69754b6aafe44c818cb383c8f1875f3a0e7fa3d/agent/fw_symfony4.c#L175-L178

https://github.com/newrelic/newrelic-php-agent/blob/f69754b6aafe44c818cb383c8f1875f3a0e7fa3d/agent/fw_symfony4.c#L190-L191

Both these cases should be replaced with `$request->attributes->get`

PR #1151, opened 2026-01-07
fix(agent): do not call 'get' directly on Symfony requests

ZNeumann:
Calling `get` on a request object is not best practice, and has been deprecated. It is advised to instead call get on the proper member, in our case, `attributes`.

closes https://github.com/newrelic/newrelic-php-agent/issues/1148

### DIFF
--- a/agent/fw_symfony4.c
+++ b/agent/fw_symfony4.c
@@ -130,9 +130,9 @@ NR_PHP_WRAPPER_END
 NR_PHP_WRAPPER(nr_symfony4_name_the_wt) {
   zval* event = NULL;
   zval* request = NULL;
-  zval* request_attrs;
-  zval* route_rval;
-  zval* controller_rval;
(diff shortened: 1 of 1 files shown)


agent/fw_symfony4.c
```diff
@@ -130,6 +130,9 @@ NR_PHP_WRAPPER_END
 NR_PHP_WRAPPER(nr_symfony4_name_the_wt) {
   zval* event = NULL;
   zval* request = NULL;
+  zval* request_attrs;
+  zval* route_rval;
+  zval* controller_rval;
```
zsistla: set to NULL

ZNeumann: fixed in [2f83818](https://github.com/newrelic/newrelic-php-agent/pull/1151/commits/2f838184dc0995f9915493b9abe38e002b7d832f)

PR #1152, opened 2026-01-07
Update issue templates

cpearcey:
We are attempting to move more general requests and support the support forums or account reps.


cpearcey:
Updating templates

PR #1154, opened 2026-01-09
feat(agent): update instrumentation to be compatible with MongoDB 2.x

### LINKED ISSUE #1147: Lost APM instrumentation of mongodb with mongo driver 2.x (works with 1.21.x)
## Description

After upgrade to mongo driver 2.1 we lost the mongo instrumentation in the APM

## Steps to Reproduce

Upgrade mongo driver to 2.1 and instrumentation of mongodb is lost

Downgrade to 1.21.1 and works again

## Expected Behavior

APM mongodb instrumentation should work in 2.x

## Your Environment

PHP 8.4
Agent Version: 12.2.27

## Screenshot

Image

hahuja2:
This PR removes redundant `Executable` interface check which caused instrumentation to break for MongoDB 2.x because the `Executable` interface has been removed. This check is not needed because for all operations we implement, they will only be instrumented if `execute` method is called, so this makes current check for `Executable` interface redundant as this interface only defined `execute` method.

closes #1147

### DIFF
--- a/agent/lib_mongodb.c
+++ b/agent/lib_mongodb.c
@@ -190,7 +190,6 @@ NR_PHP_WRAPPER(nr_mongodb_operation_before) {
 NR_PHP_WRAPPER_END
 
 NR_PHP_WRAPPER(nr_mongodb_operation_after) {
-  const char* this_klass = "MongoDB\\Operation\\Executable";
   zval* collection = NULL;
   zval* database = NULL;
   zval* server = NULL;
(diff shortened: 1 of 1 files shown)


agent/lib_mongodb.c
lavarou: This is no longer used.

hahuja2: Good point, removed: [bcb941a](https://github.com/newrelic/newrelic-php-agent/pull/1154/commits/bcb941a105c5d821b1906d5f5cd63e1fb4e942e7)

agent/lib_mongodb.c
```diff
@@ -246,11 +234,7 @@ NR_PHP_WRAPPER(nr_mongodb_operation_after) {
                                           &instance.port_path_or_id);
 
 leave:
```
lavarou: This label is not used anymore.

hahuja2: Removed: [70b199e](https://github.com/newrelic/newrelic-php-agent/pull/1154/commits/70b199ef5369f98afbbf090c208f54e3ac4f0b83)

lavarou (approved):

zsistla (approved):

PR #1157, opened 2026-01-16
fix(agent): fix drupal hook implementation

bduranleau-nr:
Addresses the change introduced in Drupal 11.3 to how hooks are managed by Drupal's ModuleHandler. Required to maintain hook instrumentation in the Agent.


agent/fw_drupal8.c
```diff
@@ -625,36 +625,77 @@ static bool nr_is_invalid_key_val_arr(nr_php_string_hash_key_t* key,
...
+ *   ...
+ *
+ * To provide hook instrumentation, we loop through all the elements of the
+ * array and extract the hook key name, class key name, function key name and
+ * module name, and construct a wrapper using that data for each hook.
  */
-static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
-  zval* hook_implementation_map = NULL;
-
+static bool nr_drupal_parse_hookmap(zval* hookmap) {
```
zsistla, line 689: Both `nr_drupal_parse_hooklist` and `nr_drupal_parse_hookmap` don't check if the param is NULL or a valid zval array since it's checked in `nr_drupal_hook_attribute_instrument`.  Could you please add that requirement  (that `nr_drupal_parse_hooklist` and `nr_drupal_parse_hookmap` assume the parameter to be a valid non-null zval array) to the function description of each?

zsistla:
Can you please add some mocked integration tests, similar to what was done previously here: https://github.com/newrelic/newrelic-php-agent/tree/main/tests/integration/frameworks/drupal

agent/fw_drupal8.c
```diff
@@ -704,6 +745,137 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
...
+  bool uses_hooklist = false;
+
+  hook_implementation_map = nr_php_get_zval_object_property(
+      module_handler, "hookImplementationsMap");
+
+  if (!nr_php_is_zval_valid_array(hook_implementation_map)) {
+    hook_implementation_map
+        = nr_php_get_zval_object_property(module_handler, "hookLists");
+    if (!nr_php_is_zval_valid_array(hook_implementation_map)) {
+      nrl_verbosedebug(NRL_FRAMEWORK, "hookLists property not a valid array");
```
zsistla: The comment could be misleading, it could be on drupal 11.2 but hooks_implementation_map was somehow invalid so then we tried this case and then got that hooksList was invalid even though there is no hookslist in 11.2.

agent/fw_drupal8.c
bduranleau-nr: The current comment reads pretty clearly to me that null is not welcome in either function:
```
@param zval* hookmap: a pointer to the zval containing the ModuleHandler `hookImplementationsMap` class property.
```
Do you just want me to add `non-null` before `pointer`?

agent/fw_drupal8.c
bduranleau-nr: 81f580b8cc2bbfb9243ef4516773531211e5b7d4

agent/fw_drupal8.c
zsistla: how about `valid, non-null` before `pointer`?

bduranleau-nr:
> Can you please add some mocked integration tests, similar to what was done previously here: https://github.com/newrelic/newrelic-php-agent/tree/main/tests/integration/frameworks/drupal

New tests added along with a general reorganization of the Drupal integration tests by breaking change version: dfc1a61794fc2756ecb636b6ba16500b2baa352a

agent/fw_drupal8.c
bduranleau-nr: aa59a777c2c1d960cd804bc9819e550066fd5b37

tests/integration/frameworks/drupal/11.3/mock_module_handler_invalid_map.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* Verify agent behavior when hookImplemementationsMap is not an array */
```
zsistla: hooksList instead of hookImplemementationsMap?

bduranleau-nr: 278710e34057f9eef58d1127befe525ec1f75dd5

agent/fw_drupal8.c
```diff
@@ -704,6 +745,138 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
...
+ *   ...
+ *
+ * To provide hook instrumentation, we loop through all the elements of the
+ * array and extract the hook key name, function key name, and module name,
+ * and construct a wrapper using that data for each hook.
+ */
+static bool nr_drupal_parse_hooklist(zval* hooklist) {
+  nr_php_string_hash_key_t* hook_key = NULL;
+  zval* hook_val = NULL;
+  nr_php_string_hash_key_t* func_key = NULL;
```
mfulb, line 795: Is there a good reason to use indirection to define the type when it is just a `zend_string` and the zend calls below would not work unless it is?

Same for `hook_key` above.

agent/fw_drupal8.c
```diff
@@ -704,6 +745,138 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
...
+ *
+ * Return: bool
+ *
+ */
+static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
+  zval* hook_implementation_map = NULL;
+
+  bool uses_hooklist = false;
+
+  hook_implementation_map = nr_php_get_zval_object_property(
```
mfulb, line 857: I think it would improve the code readability if this code to determine if it is a "hookLists" was in a function like `nr_drupal_hook_is_hooklist()`.

Then you could combine the two ifs and get rid of the `uses_hooklist` bool.

```
if (nr_drupal_hook_is_hooklist(module_handler)) {
       nr_drupal_parse_hooklist(hook_implementation_map);
} else {
       nr_drupal_parse_hookmap(hook_implementation_map);
}
```

agent/fw_drupal8.c
bduranleau-nr: I don't understand the question- are you asking why we use `nr_php_string_hash_key_t` to store the `func_key` and `hook_key`? If so, it's because it's the datatype required to operate on a PHP array/list/map structure when using the `ZEND_HASH_FOREACH_STR_KEY_VAL` api. You can't just use `zend_string` directly. There's no guarantee of the datatype you'll receive from the key or value field.

agent/fw_drupal8.c
bduranleau-nr: I don't think that would improve readability. Where is the `hook_implementation_map` coming from in that example? Right now it's generated directly in the function it's used. A solution like the one suggested above would require passing `module_handler` to a different function, then doing the same logic to pull out the hook list or map, and then we'd still have to determine some way to determine which one we're using.
Unless you're suggesting using `nr_php_get_zval_object_property` to return `true` or `false` in one function, then again in another function to return the actual property zval itself, in which case I would argue that's certainly less performant (2 calls instead of one) and no more clear than the current code.

mfulb: Additionally I don't like that `hook_implementation_map` is overloaded with different values - one path actually uses a "hookList" so now we're trusting the value send to two different function calls on line 872 and 874 was setup correctly.

It seems like the `uses_hooklist` boolean isnt even needed if the conditional on line 860 was reversed and the call to `nr_drupal_parse_hookmap()` was there?

Just seems like this could be simplified.

agent/fw_drupal8.c
mfulb: Why use `nr_php_string_hash_key_t*` when it is really a `zend_string*`?  It is used in a PHP macro below so why use a NR typedef when we have the native type available?
See https://github.com/newrelic/newrelic-php-agent/blob/cbc3ffda2766989020ae5bc357d4002106ead26f/agent/php_compat.h#L16

bduranleau-nr: This is the construct used throughout the agent when operating within the `ZEND_HASH_FOREACH_KEY_VAL` macro loop. If you search the agent code, every instance of `ZEND_HASH_FOREACH_KEY_VAL` is used alongside an `nr_php_string_hash_key_t` variable. I don't think it makes much sense to break that pattern here.

mfulb: I understand now this is an established pattern.

agent/fw_drupal8.c
mfulb: After further discussion we determined this is a good way to handle the two cases of finding the hook mapping information.

mfulb (approved):

agent/fw_drupal8.c
```diff
@@ -704,6 +745,138 @@ static bool nr_drupal_hook_attribute_instrument(zval* module_handler) {
...
+ */
+static bool nr_drupal_parse_hooklist(zval* hooklist) {
+  nr_php_string_hash_key_t* hook_key = NULL;
+  zval* hook_val = NULL;
+  nr_php_string_hash_key_t* func_key = NULL;
+  zval* module_val = NULL;
+
+  ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(hooklist), hook_key, hook_val) {
+    if (nr_is_invalid_key_val_arr(hook_key, hook_val, "hook")) {
+      return false;
```
zsistla, line 800: NONBLOCKING: Does this need a verbosedebug log message?

zsistla (approved):

agent/fw_drupal8.c
bduranleau-nr: The function being called here (`nr_is_invaliid_key_val_arr`) will generate log messages if the key is `NULL`, the string is invalid, the array is invalid, or the number of elements in the current array is zero, so I opted not to include another message here to avoid overloading.

PR #1158, opened 2026-01-26
fix(agent): ensure consistent datastore instance naming

hahuja2:
This PR does the following:

- Fix issue where instance metric is being added before `nr_datastore_instance_set_host` and `nr_datastore_instance_set_port_path_or_id` are called. This caused issues with datastore instance naming when `host` and `port_path_or_id` are `NULL` because the metric was shown as  `(null)/(null)` which is incorrect behavior as it should be shown as `unknown/unknown`.
- Add unit tests to verify behavior.

### DIFF
--- a/axiom/tests/test_segment_datastore.c
+++ b/axiom/tests/test_segment_datastore.c
@@ -312,6 +312,91 @@ static void test_instance_info_empty(void) {
   nr_txn_destroy(&txn);
 }
 
+static void test_instance_info_null(void) {
+  nrtxn_t* txn = new_txn(0);
+  nrtime_t duration = 4 * NR_TIME_DIVISOR;
+  nr_segment_datastore_params_t params = sample_segment_datastore_params();
+  nr_segment_t* segment = NULL;
+  nr_datastore_instance_t instance = {
+      .host = NULL,
+      .port_path_or_id = NULL,
+      .database_name = NULL,
+  };
+  const char* tname = "instance info null";
+  params.instance = &instance;
+
+  // Test: Instance info all NULL
+  segment = nr_segment_start(txn, NULL, NULL);
+  segment->start_time = 1 * NR_TIME_DIVISOR;
+  segment->stop_time = 1 * NR_TIME_DIVISOR + duration;
+  test_segment_datastore_end_and_keep(&segment, &params);
+  test_segment_metric_created(tname, segment->metrics,
+                              "Datastore/instance/MongoDB/unknown/unknown",
+                              false);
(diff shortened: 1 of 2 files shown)


axiom/tests/test_segment_datastore.c
```diff
@@ -312,6 +312,91 @@ static void test_instance_info_empty(void) {
...
+  instance.port_path_or_id = "";
+  instance.database_name = "";
+  test_segment_datastore_end_and_keep(&segment, &params);
+  test_segment_metric_created(tname, segment->metrics,
+                              "Datastore/instance/MongoDB/unknown/unknown",
+                              false);
+  test_datastore_segment(&segment->typed_attributes->datastore, tname,
+                         "MongoDB", NULL, NULL, NULL, NULL, NULL, "unknown",
+                         "unknown", "unknown");
+  nr_txn_destroy(&txn);
```
lavarou, line 396: Isn't this already tested in `test_instance_info_empty` [here](https://github.com/newrelic/newrelic-php-agent/blob/5c1ee596fbff94a9596a3bc5f84e05f0cd89ee34/axiom/tests/test_segment_datastore.c#L303-L305)?

lavarou (approved):

axiom/tests/test_segment_datastore.c
hahuja2: Yeah, but in that [example](https://github.com/newrelic/newrelic-php-agent/blob/5c1ee596fbff94a9596a3bc5f84e05f0cd89ee34/axiom/tests/test_segment_datastore.c#L284), `nr_datastore_instance_create` is used when setting the values. However, they both call `test_segment_datastore_end_and_keep` which eventually calls `create_metrics`, so I can see how this test can be repetitive. I can remove this test if you would like?

zsistla: It's not repetitive; having the datastore set without using `nr_datastore_instance_create` https://github.com/newrelic/newrelic-php-agent/blob/02a11df2a22febaba8cb0ca1378b550ff0b68756/axiom/tests/test_segment_datastore.c#L284
Is testing different code paths since `nr_datastore_instance_create`  automatically sets the unknowns in the params that are passed to end_segment.  This test tests that unknown is still created when not already set in the params.

hahuja2: Yeah that's why I added the test, but was just mentioning how they both eventually call `create_metrics`. Let's keep the test 👍

zsistla (approved):

PR #1159, opened 2026-01-27
fix(agent): fix table naming for queries with `EXTRACT` function

zsistla:
Sometimes you can get a `FROM` in the EXTRACT function. 

https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT The EXTRACT function retrieves subfields such as year or hour from date/time values. source must be a value expression of type timestamp, date, time, or interval 
 Ex: SELECT EXTRACT(CENTURY FROM TIMESTAMP '2000-12-16 12:21:13');
 Result: 20
This FROM is NOT followed by a table name but a timestamp and therefore when we are trying to NR_SQL_PARSE_FROM, the whole EXTRACT(...) clause should be ignored to avoid keying off of the FROM value contained inside.

This PR:
1) updates agent's SQL the parser to skip/parse over `EXTRACT(field FROM source)` function
2) adds unit tests

### DIFF
--- a/axiom/tests/test_sql.c
+++ b/axiom/tests/test_sql.c
@@ -604,7 +604,7 @@ static void test_real_world_things(void) {
       "should return correct table name",
       sql, "select", "baz_table");
 
-  sql = " SELECT foo,bar FROM baz_table WHERE (MONTH FROM event_date) = 10";
+  sql = " SELECT foo,bar FROM baz_table WHERE EXTRACT(MONTH FROM event_date) = 10";
   test_get_operation_and_table(
       "Valid EXTRACT function at the end of the query string should return "
       "correct table name",
--- a/axiom/tests/test_sql.c
+++ b/axiom/tests/test_sql.c
@@ -600,9 +600,14 @@ static void test_real_world_things(void) {
       = " SELECT foo, EXTRACT(EPOCH FROM(content.creation)::timestamptz) as "
         "creation_timestamp,bar FROM baz_table ";
   test_get_operation_and_table(
-      "Valid EXTRACT function with no whitespace before parenthesis in select "
-      "should return correct table name",
+      "Valid EXTRACT function at the end of the query string should return "
+      "correct table name",
       sql, "select", "baz_table");
+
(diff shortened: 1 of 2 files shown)


axiom/tests/test_sql.c
```diff
@@ -573,6 +573,47 @@ static void test_real_world_things(void) {
...
+
+  sql
+      = " SELECT foo, EXTRACT(EPOCH FROM(content.creation)::timestamptz) as "
+        "creation_timestamp,bar FROM baz_table ";
+  test_get_operation_and_table(
+      "Valid EXTRACT function with no whitespace before parenthesis in select "
+      "should return correct table name",
+      sql, "select", "baz_table");
+
+  sql = " SELECT foo,bar FROM baz_table WHERE (MONTH FROM event_date) = 10";
```
lavarou: Judging by the test name, `EXTRACT` keyword is missing from the SQL to be parsed:
```suggestion
  sql = " SELECT foo,bar FROM baz_table WHERE EXTRACT(MONTH FROM event_date) = 10";
```

lavarou (dismissed):
LGTM

lavarou (approved):

Manackkel:
Thanks for this fix, I'm wondering if table detection couldn't be improved with parenthesis detection.

for a basic query 
```
SELECT
    label,
    (SELECT COUNT(1) FROM my_table_event WHERE my_table_id = my_table.my_table_id) as event_count
FROM
    my_table
LIMIT 1;
```

`my_table_event` will be reported to Newrelic instead of `my_table`

could you use the name of the first table matched after a `FROM` not inside a parenthesis ?
it would fix both this `EXTRACT(field FROM source)` and subqueries

zsistla:
@Manackkel ,

Thank you for your feedback!  The below is being tracked internally and will be addressed in another PR.

> for a basic query
> 
> ```
> SELECT
>     label,
>     (SELECT COUNT(1) FROM my_table_event WHERE my_table_id = my_table.my_table_id) as event_count
> FROM
>     my_table
> LIMIT 1;
> ```
> 
> `my_table_event` will be reported to Newrelic instead of `my_table`
>

PR #1164, opened 2026-02-05
fix(agent): improve SQL parsing logic

bduranleau-nr:
### With this change the following problems are addressed: 
1. Broken SQL parsing behavior when attempting to extract table name from statements with nested subqueries in `SELECT` and `DELETE` statements.
#### Example: 
```sql
SELECT (label), (SELECT COUNT(1) FROM my_nested_table WHERE my_table_id = my_table.my_table_id) AS event_count) FROM my_table
```

Should return `my_table`, not `my_nested_table`.

2. Failure to properly handle replication safeguard logic in `CREATE` and `DROP` statements.
#### Example:
```sql
CREATE TABLE IF NOT EXISTS birthdays
```

Should return `birthdays`, not `IF`

```sql
DROP TABLE IF EXISTS birthdays
```

Should return `birthdays`, not `IF`.

### Solutions:
1. Track the nesting level of an SQL query while parsing. Do not attempt to parse the table name of a valid statement unless the nesting level is 0.
2. Add logic to check for and parse over keywords `IF`, `NOT` and `EXISTS` if we know we are parsing a `CREATE` or `DROP` statement

### DIFF
--- a/axiom/util_sql.c
+++ b/axiom/util_sql.c
@@ -6,6 +6,7 @@
 #include "nr_axiom.h"
 
 #include <stddef.h>
+#include <stdio.h>
 
 #include "util_hash.h"
 #include "util_logging.h"
@@ -368,11 +369,25 @@ static char* nr_sql_parse_over(const char* str,
   return cend + 1; /* Character is consumed */
 }
 
+static bool nr_parse_sql_keyword(const char* sql, const char* keyword) {
+  int i;
+  for (i = 0; i < nr_strlen(keyword); i++) {
+    if (keyword[i] != nr_tolower(sql[i])) {
+      return false;
+    }
+  }
+  if (NULL == nr_strchr(NR_SQL_DELIMITER_CHARS, sql[i])) {
+    return false;
+  }
+  return true;
+}
+
 void nr_sql_get_operation_and_table(const char* sql,
                                     const char** operation_ptr,
                                     char** table_ptr,
                                     int show_sql_parsing) {
   int i;
+  bool nested = false;
   const char* start = 0;
   const char* end = 0;
   const char* x;
@@ -475,6 +490,10 @@ void nr_sql_get_operation_and_table(const char* sql,
         continue;
       }
 
+      if ('(' == nr_tolower(x[0])) {
+        nested = true;
+      }
+
       if ((NR_SQL_PARSE_CREATE == operations[i].opflag)
           || (NR_SQL_PARSE_DROP == operations[i].opflag)) {
         /*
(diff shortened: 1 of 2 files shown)


axiom/util_sql.c
```diff
@@ -368,11 +368,35 @@ static char* nr_sql_parse_over(const char* str,
...
+ *
+ * @param sql     The SQL statement to parse
+ * @param keyword The SQL keyword to compare
+ *
+ * @return bool
+ *
+ */
+static bool nr_parse_sql_keyword(const char* sql, const char* keyword) {
+  int i;
+  for (i = 0; i < nr_strlen(keyword); i++) {
```
zsistla: Nice addition of the helper func, but as the parser goes through the string, this adds a strlen op every time this is called which could be a lot depending on the length of the sql statement.

Consider instead passing the strlen with the other parameters.

axiom/util_sql.c
```diff
@@ -475,85 +499,60 @@ void nr_sql_get_operation_and_table(const char* sql,
...
+      if (0 == nest_level) {
+        if (NR_SQL_PARSE_CREATE == operations[i].opflag
+            || NR_SQL_PARSE_DROP == operations[i].opflag) {
+          /*
+           * If this is a `CREATE` or `DROP` statement then it will be followed
+           * by `DATABASE` or `TABLE` (ignoring whitespace and comments). Since
+           * 'create' is the first word, we merely advance to the second word to
+           * see if it is DATABASE OR TABLE then advance to the db/table name.
+           */
+          if (nr_parse_sql_keyword(x, "database")) {
```
zsistla: With the improvement of the new helper func, could now consider making the keywords defined like the following to also get the length precomputed for us:
#define SQL_PARSE_DATABASE "database"
#define SQL_PARSE_DATABASE_LEN sizeof(SQL_PARSE_DATABASE) - 1

Then it could look like
if (nr_parse_sql_keyword(x,SQL_PARSE_DATABASE, SQL_PARSE_DATABASE_LEN)) {

and the 
x += 8; 
could also be x += SQL_PARSE_DATABASE_LEN;

axiom/util_sql.c
mfulb: Hopefully the compiler would at least not recompute the strlen() each iteration but I agree if we have the strlen already this passing it in would be good.

axiom/util_sql.c
lavarou: There already exists [`NR_PSTR`](https://github.com/newrelic/newrelic-php-agent/blob/v12.5.0.30/axiom/nr_axiom.h#L117-L120) macro that could be used here:
```suggestion
          if (nr_parse_sql_keyword(x, NR_PSTR("database"))) {
```

axiom/util_sql.c
bduranleau-nr: Implemented here: 45436039873db13e1bdd6e9c09c02bc756393e4a

axiom/util_sql.c
bduranleau-nr: I've opted for @lavarou's suggestion to use `NR_PSTR` in these function calls.

mfulb (approved):

zsistla (approved):

PR #1167, opened 2026-02-12
feat(agent): Configurable ability to disable instrumentation on fiber detection

zsistla:
By default, disables instrumentation per request when detecting PHP Fibers activity.

Has configurable INI setting to toggle the behavior.

Integration tests added to demonstrate behavior.

### DIFF
--- a/agent/php_observer.c
+++ b/agent/php_observer.c
@@ -124,11 +124,32 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
   return handlers;
 }
 
+#if ZEND_MODULE_API_NO > ZEND_8_0_X_API_NO /* PHP8.1+ */
+static void nr_fiber_disable(zend_fiber_context* fiber_context) {
+  if (NULL != NRPRG(txn)) {
+    /* Fiber init/desetroy detected, end and keep the transaction. */
+    nrl_warning(NRL_INSTRUMENT,
+                "Transaction is truncated because PHP Fiber use is detected.");
+    nr_php_txn_end(0, 0 TSRMLS_CC);
+  }
+}
+
+static void nr_fiber_switch_disable(zend_fiber_context* from,
+                                    zend_fiber_context* to) {
+  if (NULL != NRPRG(txn)) {
+    /* Fiber switch detected, end and keep the transaction. */
+    nrl_warning(NRL_INSTRUMENT,
+                "Transaction is truncated because PHP Fiber use is detected.");
+    nr_php_txn_end(0, 0 TSRMLS_CC);
+  }
+}
+#endif /* PHP 8.1+ */
+
 void nr_php_observer_no_op(zend_execute_data* execute_data NRUNUSED){};
 
 void nr_php_observer_minit() {
   /*
-   * Register the Observer API handlers.
+   * Register the Observer API function handlers.
    */
   zend_observer_fcall_register(nr_php_fcall_register_handlers);
 
@@ -138,6 +159,23 @@ void nr_php_observer_minit() {
    * turn it into a no_op when using OAPI.
    */
   NR_PHP_PROCESS_GLOBALS(orig_execute) = nr_php_observer_no_op;
+
+
+#if ZEND_MODULE_API_NO > ZEND_8_0_X_API_NO /* PHP 8.1+ */
+
+  /*
+   * Register the Observer API fiber handlers.
+   */
+
+   /* Currently only register if we are disabling instrumentation. */
+   if (NRINI(fibers_disabled)) {
+    zend_observer_fiber_init_register(nr_fiber_disable);
+		zend_observer_fiber_switch_register(nr_fiber_switch_disable);
+		zend_observer_fiber_destroy_register(nr_fiber_disable);
+   }
+
+#endif /* PHP 8.1+ */
+
 }
 
 #endif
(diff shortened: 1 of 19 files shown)


agent/php_observer.c
```diff
@@ -124,11 +124,34 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
 }
 
+#if ZEND_MODULE_API_NO > ZEND_8_0_X_API_NO /* PHP8.1+ */
+static void nr_fiber_disable(zend_fiber_context* fiber_context) {
+  if (NULL != NRPRG(txn)) {
+    /* Fiber init/desetroy detected, end and keep the transaction. */
+    nrl_warning(NRL_INSTRUMENT,
+                "Transaction is truncated because PHP Fiber use is detected.");
+    nrm_force_add(NRPRG(txn)->unscoped_metrics, "Supportability/PHP/Fiber/used", 0);
+    nr_php_txn_end(0, 0 TSRMLS_CC);
```
mfulb, line 134: Should this also explicitly turn off recording?

What will happen if the start txn API is called after this function?

Should the hook into OAPI be removed now we've determined we're in a fiber env and we disabled the agent recording?

agent/php_observer.c
```diff
@@ -124,11 +124,34 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
+}
+
+static void nr_fiber_switch_disable(zend_fiber_context* from,
+                                    zend_fiber_context* to) {
+  if (NULL != NRPRG(txn)) {
+    /* Fiber switch detected, end and keep the transaction. */
+    nrl_warning(NRL_INSTRUMENT,
+                "Transaction is truncated because PHP Fiber use is detected.");
+    nrm_force_add(NRPRG(txn)->unscoped_metrics, "Supportability/PHP/Fiber/used", 0);
+    nr_php_txn_end(0, 0 TSRMLS_CC);
```
mfulb, line 145: Same questions as for `nr_fiber_disable()`

agent/php_observer.c
zsistla: To answer the questions in order:
1) No, we don't need to explicitly turn off recording since txn_end already does that: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_txn.c#L1286

2) The txn will start again.  As per current behavior the new children will be parented to main.  Once fiber activity is detected again, the txn will again be ended.  Added new test cases to demonstrate this behavior here: [24544ae](https://github.com/newrelic/newrelic-php-agent/pull/1167/commits/24544aed692f4c915829c303220a7d3cbcd2490a)

3) No, the hook (set during `minit`) should not be removed as we want it to be in place for:
a) the next request
b) to protect us if start_txn is called in the customer's application so we can toggle off again if there's fiber action detected again.

agent/php_observer.c
zsistla: To answer the questions in order:
1) No, we don't need to explicitly turn off recording since txn_end already does that: https://github.com/newrelic/newrelic-php-agent/blob/main/agent/php_txn.c#L1286

2) The txn will start again.  As per current behavior the new children will be parented to main.  Once fiber activity is detected again, the txn will again be ended.  Added new test cases to demonstrate this behavior here: [24544ae](https://github.com/newrelic/newrelic-php-agent/pull/1167/commits/24544aed692f4c915829c303220a7d3cbcd2490a)

3) No, the hook (set during `minit`) should not be removed as we want it to be in place for:
a) the next request
b) to protect us if start_txn is called in the customer's application so we can toggle off again if there's fiber action detected again.

tests/integration/span_events/test_nested_span_endstarttxn.php
```diff
@@ -0,0 +1,213 @@
...
+
+env_var_for_expects("GUID_ROOT", newrelic_get_linking_metadata()['span.id']);
+
+function c()
+{
+    echo "Starting Func 'c'\n";
+    env_var_for_expects("GUID_C", newrelic_get_linking_metadata()['span.id']);
+    time_nanosleep(0, 100000000);
+    newrelic_end_transaction();
+    newrelic_start_transaction(ini_get("newrelic.appname"));
```
ZNeumann, line 174: From what I can tell, there are no tests for starting/ending a transaction within fibers.

zsistla: Actually end/start of txns in fibers will be tested thoroughly in PRs for the support Fiber code.

This PR/tests only ends a txn if Fiber activity is detected.
Since the txn would already be stopped by the time we are in a Fiber, stopping it again wouldn't do anything.
If a txn is started within a fiber, this PR will detect that and auto-end the txn again.

Starting the txn again is tested in test_fiber_disabled_nested4/5/6/7/.php
These test:
1) starting a txn in the calling function after a fiber->start
2) starting a txn in the fiber before a Fiber::suspend()
3) starting a txn in the fiber after a Fiber::suspend()
4) starting a txn in the calling function after a fiber->start where the Fiber has been suspended but never resumed.

ZNeumann: Got it, thanks. I was also accidentally looking at old code prior to https://github.com/newrelic/newrelic-php-agent/pull/1167/changes/24544aed692f4c915829c303220a7d3cbcd2490a which addresses my concerns

ZNeumann (approved):

mfulb (approved):

PR #1174, opened 2026-03-11
fix(daemon): correct daemon seen/sent metric generation

bduranleau-nr:
Refactor and improve accuracy of metric harvests.
This PR fixes several bugs related to metrics harvest in the daemon:

### Misreporting of Seen/Sent/Failure metrics for Span Events, Log Events, Transaction Events, Error Events, and Custom Events. 

***Explanation***: 
- Metric harvest occurs during the Default Data harvest. If this harvest type occurs after any of the above event type harvests, the data will be reported as zero with no capability to recover on subsequent harvests. This is because each harvest wipes the data for that event type prior to the call to `considerHarvestPayload`. Seen & Sent metrics rely on the `events` heap for each event type to be populated with the data being harvested, which was calculated via calls to `NumSeen()` and `NumSaved()` in `createFinalMetrics`. For events on the fast harvest cycle (5 seconds, Log Events for example), this means that the Seen/Sent metrics from first 11 harvests would be guaranteed to be lost. The 12th harvest MAY be recorded accurately if Default Data harvest was triggered before the harvest for that event type. For events on the 1 minute harvest cycle (Span Events, for example) the same rule applied where the odds were ~50/50 whether metrics would be correctly generated.

***Fix***:
- A channel for metric data containing the counts of seen, sent and failures has been created and added to a `MetricController`. Metric harvest has been gated behind a WaitGroup, where each other event type increments the waitgroup by 1 when initiated, and decrements by 1 when the `harvestPayload` call returns (regardless of success or failure). The metric harvest will wait until the waitgroup is zero to attempt to consume all of the data stored in the metric channel and generate the seen/sent/failure metrics. This allows fast harvest cycle events to continually populate the channel, ensuring their data isn't lost, and encourages metric harvest to happen last. Due to "jitter" in the non-deterministic harvest trigger function timings, the count MAY be zero before all event types have been harvested. In cases such as these, we will send the seen/sent/failure metrics in the next harvest cycle, ensuring that over multiple harvests, data should never be lost, only slightly lagging.

***

### Overflow of the Data Usage channel causing dropped data

***Explanation***:
- The Data Usage Channel is only allocated with a maximum of 25 elements. For an example application that leverages multiple event types, especially on the fast harvest cycle of 12 per minute, this was incapable of holding all of the data generated and would result in dropped Data Usage metrics that would not be recovered

***Fix***:
- Expand the Data Usage Channel Size to 64 elements.

***

### Cross Contamination of the Data Usage metrics

***Explanation***:
- The Data Usage Channel is stored in the Processor structure. This structure is not application unique- it services all applications connected to the daemon and reporting data. In a scenario where the agent<->daemon relationship is 1-1, this is not an issue. In a situation where the relationship is many-to-one, this could result in the cross-contamination of Data Usage metrics between applications. For example, an application that only reported Log Events and another application only reporting Custom Events would feed data into the same, shared channel. During harvest, it would then consume that data and report it without consideration for the application that generated it, resulting in the Log Event app reporting Custom Event Data Usage metrics and vice-versa.

***Fix***:
- Move the data usage channel and management into the `AppHarvest` structure. This structure records data specific to an individual application that is actively connected and reporting data to the daemon. By doing this we avoid the possibility of another application writing to or reading from another applications data usage channel.

***

### Invalid data usage waitgroup logic

***Explanation***:
- The waitgroup for Data usage was reset at the beginning of each call to `harvestByType`. This results in non-default data harvests feeding into a waitgroup that would be reset without any potential for affecting the waitgroup `wait()` call in the `harvestDataUsage()` call. Effectively the only metrics that would impact this call were the other event types included in the Default Data harvest type.

***Fix***:
- Move the waitgroup into the `AppHarvest` structure to allow it to persist through all of the harvest types for an individual reporting application. Set once per app harvest in the `processConnectAttempt` function and do not reset for the life of that reporting application.

### DIFF
--- a/daemon/internal/newrelic/app_harvest.go
+++ b/daemon/internal/newrelic/app_harvest.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/infinite_tracing"
 	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/log"
+	"golang.org/x/sync/semaphore"
 )
 
 // This type takes the HarvestType values sent from an application's harvest
@@ -45,7 +46,7 @@ type MetricsController struct {
 	mc  chan metricsInfo
 	duc chan dataUsageInfo
 	wg  *sync.WaitGroup
-	mu  sync.Mutex
+	sem *semaphore.Weighted
 }
 
 func (ah *AppHarvest) NewProcessorHarvestEvent(id AgentRunID, t HarvestType) ProcessorHarvest {
@@ -122,6 +123,7 @@ func NewAppHarvest(id AgentRunID, app *App, harvest *Harvest, ph chan ProcessorH
 			mc:  make(chan metricsInfo, 64),
 			duc: make(chan dataUsageInfo, 64),
 			wg:  new(sync.WaitGroup),
+			sem: semaphore.NewWeighted(1),
 		},
 	}
 
--- a/daemon/internal/newrelic/app_harvest.go
+++ b/daemon/internal/newrelic/app_harvest.go
@@ -6,9 +6,11 @@
 package newrelic
 
 import (
+	"sync"
 	"time"
 
 	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/infinite_tracing"
+	"github.com/newrelic/newrelic-php-agent/daemon/internal/newrelic/log"
 )
 
 // This type takes the HarvestType values sent from an application's harvest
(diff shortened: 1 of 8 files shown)


daemon/internal/newrelic/app_harvest.go
```diff
@@ -33,13 +56,73 @@ func (ah *AppHarvest) NewProcessorHarvestEvent(id AgentRunID, t HarvestType) Pro
...
+	// aggregate data from metrics channel
+	loop := true
+	for loop {
+		select {
+		case metric, ok := <-m.mc:
+			if ok { // check that channel is not closed
+				metrics := metricsMap[metric.eventType]
+				metrics.seen += metric.seen
+				metrics.sent += metric.sent
+				if metric.failed > metrics.failed {
```
mfulb, line 109: Can you add a comment to explain the basis for this method of getting the largest value from the channel?  I think it would be helpful if someone works on this in the future.

bduranleau-nr: I've attempted to summarize the behavior in a comment in this commit: d813e1392cc63d5b30b812095c39c09732c4a5b5

mfulb (approved):
Nice work Brian - I have checked this using a test app which exercises numerous end points and the channel size changes look good and the mutex for avoiding more than one harvestMetrics sitting in Wait() appear to handle the case where the network goes does an extended period.

ZNeumann (approved):

PR #1177, opened 2026-03-12
build: cleanup release agent targets

lavarou:
Remove recipes used by legacy build systems which are no dead code. Make
`release-agent` compatible with GitHub Actions release workflow. Replace
`agent-for-php-${php_version}` recipe with `agent-for-release`, which
uses PHP's build system installed on host to dynamically determine PHP API
version and PHP ZTS mode. This makes the build system more robust by relying
on information from build system rather than manual user input when preparing
release.


make/release.mk
```diff
@@ -90,87 +78,33 @@ release-scripts: Makefile | releases/$(RELEASE_OS)/scripts/
...
-# Next build ZTS binaries of the PHP versions requested that are supported
-# on this OS.
-#
-	for PHP in $(SUPPORTED_PHP) ; do \
-		$(MAKE) agent-clean; $(MAKE) release-$$PHP-zts; \
-	done
+release-agent: Makefile | releases/$(RELEASE_OS)/agent/$(RELEASE_ARCH)/
+	@$(MAKE) agent-clean && $(MAKE) agent-for-release
 
+# Older versions of GNU Make had a bug where "#" in a function invocation
```
mfulb, line 89: Are we sure this is a bug that was ever fixed or is it just how `make` has always work to break a line on a `#`.  If it is just present in old `make` maybe we don't need this workaround?

lavarou: Look [here](https://cgit.git.savannah.gnu.org/cgit/make.git/plain/NEWS):
> Version 4.3 (19 Jan 2020)
> ...
> * WARNING: Backward-incompatibility!
>  Number signs (#) appearing inside a macro reference or function invocation
>  no longer introduce comments and should not be escaped with backslashes:
>  thus a call such as:
>    foo := $(shell echo '#')
>  is legal.  Previously the number sign needed to be escaped, for example:
>    foo := $(shell echo '\#')
>  Now this latter will resolve to "\#".  If you want to write makefiles
>  portable to both versions, assign the number sign to a variable:
>    H := \#
>    foo := $(shell echo '$H')
>  This was claimed to be fixed in 3.81, but wasn't, for some reason.

mfulb (approved):

mfulb (dismissed):

mfulb (approved):

make/release.mk
```diff
@@ -90,87 +78,33 @@ release-scripts: Makefile | releases/$(RELEASE_OS)/scripts/
...
+# such as $(shell ...) was treated as a make comment. This makefile needs
+# to be compatible with older versions of GNU Make, so we need to use
```
zsistla, line 91: Which versions?  It'd be nice to list the PHP version(s) built with the older version so we know when we can finally take the backward compatibility out.

zsistla (approved):

make/release.mk
lavarou: Any build system that uses GNU Make version < Version 4.3 (19 Jan 2020) needs this workaround. However, the workaround is future compatible - this works just fine for GNU Make version >= 4.3.

PR #1185, opened 2026-03-27
chore(deps): Update golang dependencies

zsistla:
1) Update golang dependencies.

2) Update go to use default codec.

3) All other functionality remains the same between previous behavior and behavior after updates.

4) regenerated v1.pb.go using legacy methods.

Any additional modernization is beyond the scope of this PR.

Suffice it to say many things have changed from 1.61 to 1.79.
https://github.com/grpc/grpc-go/blob/c20fba0fa91403ad530afff515c6e389f8674643/codec.go#L27

However, the issue here is that change in codec from v1 to v2 and the issue described here:
And the breaking change that occurred here: https://github.com/grpc/grpc-go/issues/7569


daemon/internal/newrelic/infinite_tracing/grpc_sender.go
```diff
@@ -17,9 +17,11 @@ import (
 	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
-	"google.golang.org/grpc/encoding"
+	codecproto "google.golang.org/grpc/encoding/proto"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+	_ "google.golang.org/protobuf/protoadapt"
```
bduranleau-nr:
```suggestion
```
Can be removed

daemon/go.sum
```diff
@@ -1,23 +1,40 @@
-github.com/golang/protobuf v1.5.0/go.mod h1:FsONVRAS9T7sI+LIUmWTfcYkHO4aIWwzhcaSAoJOfIk=
```
bduranleau-nr, line 1: Some of the changes in this file seem unrelated to the grpc/protobuf changes in this PR. I would be interested in whether these persist if the go.sum file is deleted and regenerated via `go mod tidy`

zsistla: deleted and regenerated and got the same results.

bduranleau-nr: it appears these have been added to the grpc dependencies in newer versions: https://github.com/grpc/grpc-go/blob/v1.79.3/go.mod

bduranleau-nr (approved):

PR #1186, opened 2026-03-27
Fixing trivy vulnerable version

hmishra12:


zsistla:
Already addressed in https://github.com/newrelic/newrelic-php-agent/pull/1182

PR #1195, opened 2026-04-06
chore(daemon): modernize daemon codebase

bduranleau-nr:
This PR aims to modernize the daemon codebase. It does this in four parts:
1. Run `go fix ./...` to automatically implement the fixes provided by Golang's `fix` tool.
2. Manually update the version of `flatbuffers` to the latest in `go.mod` and `go.sum`, remove legacy github `protobuf` dependency in favor of `google.golang.org/protobuf`.
3. Remove `string_slice.go` and replace the calls to`StringSliceContains` method with direct calls to `slices.Contains`.
4. Remove deprecated `io/ioutil` package (replaced by `io` and `os` packages).


daemon/cmd/stressor/main.go
```diff
@@ -636,9 +631,7 @@ func main() {
 	}
 
 	for i := 0; i < nworkers; i++ {
```
lavarou: Is there any reason this iteration wasn't converted to `for i := range nworkers` idiom?

bduranleau-nr: It's because the `i` variable isn't used within the loop itself, which causes a compiler error in Go. The current way this loop is written technically "uses" the `i` variable in the conditional check and increment of the loop statement.
I confirmed this wasn't accidentally missed by re-running `go fix ./...` and it doesn't get updated. If I update it manually to `for i := range nworkers`, it causes a compiler error.

lavarou: Then the same method from daemon/cmd/integration_runner/main.go can be used: `for range nworkers` - see https://github.com/newrelic/newrelic-php-agent/blob/chore/modernize-daemon/daemon/cmd/integration_runner/main.go#L473.

bduranleau-nr: I can't explain Go's fix tool behavior here- I've made the change manually and pushed

lavarou (approved):

PR #1196, opened 2026-04-09
feat(agent): Base to build upon for fiber implementation

zsistla:
* show_fiber functionality; indentations align with show_execs; the special keyword is "show_fibers"
* basic user function fiber usage - does not include anything with special instrumentation.  Instrumentation of internal functions have not been tested; although the show_fibers properly detects them.
* a few basic tests
* some get_current_segment, start_segment have been updated
* added values to track global asynchronous context (the actual context we are in) and txn context (the context the txn is currently working with which can be modified with set_current_segment)
* modified nr_txn_get_current_span_id for context to be passed
* added nr_txn_get_current_context
* added a few axiom unit tests

This is not the complete implementation, but will provide a base for other PRs to iterate on.

### DIFF
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -1241,9 +1241,15 @@ static nr_status_t nr_php_api_add_custom_span_attribute(const char* keystr,
                                                             TSRMLS_DC) {
   bool rv;
   char* key = NULL;
-  nr_segment_t* current;
+  nr_segment_t* current = NULL;
+  nrtxn_t* txn = NULL;
 
-  current = nr_txn_get_current_segment(NRPRG(txn), NULL);
+  txn = NRPRG(txn);
+  if (NULL == txn) {
+    return NR_FAILURE;
+  }
+
+  current = nr_txn_get_current_segment(txn, nr_txn_get_current_context(txn));
   if (!current) {
     return NR_FAILURE;
   }
@@ -1482,6 +1488,8 @@ PHP_FUNCTION(newrelic_get_linking_metadata) {
   NR_UNUSED_RETURN_VALUE_USED;
   NR_UNUSED_THIS_PTR;
 
+  nrtxn_t* txn = NULL;
+
   nr_php_api_add_supportability_metric("get_linking_metadata" TSRMLS_CC);
 
   array_init(return_value);
@@ -1494,6 +1502,8 @@ PHP_FUNCTION(newrelic_get_linking_metadata) {
     return;
   }
 
+  txn = NRPRG(txn);
+
   if (nrlikely(NRPRG(app))) {
     nr_php_add_assoc_string_const(return_value, "entity.name",
                                   nr_app_get_entity_name(NRPRG(app)));
@@ -1505,10 +1515,9 @@ PHP_FUNCTION(newrelic_get_linking_metadata) {
                                   nr_app_get_host_name(NRPRG(app)));
   }
 
-  if (nrlikely(NRPRG(txn))) {
-    trace_id = nr_txn_get_current_trace_id(NRPRG(txn));
-    span_id = nr_txn_get_current_span_id(NRPRG(txn));
-
+  if (nrlikely(txn)) {
+    trace_id = nr_txn_get_current_trace_id(txn);
+    span_id = nr_txn_get_current_span_id(txn, nr_txn_get_current_context(txn));
     if (trace_id) {
       nr_php_add_assoc_string(return_value, "trace.id", trace_id);
     }
@@ -1535,6 +1544,8 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
   NR_UNUSED_RETURN_VALUE_USED;
   NR_UNUSED_THIS_PTR;
 
+  nrtxn_t* txn = NULL;
+
   nr_php_api_add_supportability_metric("get_trace_metadata" TSRMLS_CC);
 
   array_init(return_value);
@@ -1547,14 +1558,15 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
     return;
   }
 
-  if (nrlikely(NRPRG(txn))) {
-    trace_id = nr_txn_get_current_trace_id(NRPRG(txn));
+  txn = NRPRG(txn);
+
+  if (nrlikely(txn)) {
+    trace_id = nr_txn_get_current_trace_id(txn);
     if (trace_id) {
       nr_php_add_assoc_string(return_value, "trace_id", trace_id);
     }
     nr_free(trace_id);
-
-    span_id = nr_txn_get_current_span_id(NRPRG(txn));
+    span_id = nr_txn_get_current_span_id(txn, nr_txn_get_current_context(txn));
     if (span_id) {
       nr_php_add_assoc_string(return_value, "span_id", span_id);
     }
@@ -1564,7 +1576,7 @@ PHP_FUNCTION(newrelic_get_trace_metadata) {
 
 /*
  * Purpose      Allows a caller to add a user id string to agent attributes in
-                transaction event, transaction trace, error and span.
+ *              transaction event, transaction trace, error and span.
  *
  */
 #ifdef TAGS
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -1244,6 +1244,8 @@ static nr_status_t nr_php_api_add_custom_span_attribute(const char* keystr,
   nr_segment_t* current = NULL;
   nrtxn_t* txn = NULL;
 
+  txn = NRPRG(txn);
+
   current = nr_txn_get_current_segment(txn, nr_txn_get_current_context(txn));
   if (!current) {
     return NR_FAILURE;
--- a/agent/php_api.c
+++ b/agent/php_api.c
@@ -1244,11 +1244,6 @@ static nr_status_t nr_php_api_add_custom_span_attribute(const char* keystr,
   nr_segment_t* current = NULL;
   nrtxn_t* txn = NULL;
 
-  txn = NRPRG(txn);
-  if (NULL == txn) {
-    return NR_FAILURE;
-  }
-
   current = nr_txn_get_current_segment(txn, nr_txn_get_current_context(txn));
   if (!current) {
     return NR_FAILURE;
--- a/agent/php_execute.c
+++ b/agent/php_execute.c
@@ -1714,93 +1714,93 @@ end:
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA /* PHP8+ and OAPI */
 
-#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO    
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
 void nr_fiber_show_fiber(zend_fiber_context* zfc, const char* fiber_action) {
-  if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).show_fibers)) {
-    zend_fiber* zfc_fiber = NULL;
-    char* zfc_status = NULL;
-    char* fiber_destroy_flag = NULL;
-    char* fiber_threw_flag = NULL;
-    char* fiber_bailout_flag = NULL;
-    char* fiber_func_name = NULL;
-    char* label = "  fiber";
-
-    if (NULL == zfc) {
-      nrl_warning(NRL_AGENT,
-                  "PHP Issue: %s Fiber context is unexpectedly NULL.",
-                  fiber_action);
-      return;
-    }
-
-    switch (zfc->status) {
-      case ZEND_FIBER_STATUS_INIT:
-        zfc_status = "ZEND_FIBER_STATUS_INIT";
-        break;
-      case ZEND_FIBER_STATUS_RUNNING:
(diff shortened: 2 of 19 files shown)


agent/php_api.c
```diff
@@ -1241,9 +1241,15 @@ static nr_status_t nr_php_api_add_custom_span_attribute(const char* keystr,
...
   char* key = NULL;
-  nr_segment_t* current;
+  nr_segment_t* current = NULL;
+  nrtxn_t* txn = NULL;
 
-  current = nr_txn_get_current_segment(NRPRG(txn), NULL);
+  txn = NRPRG(txn);
+  if (NULL == txn) {
+    return NR_FAILURE;
+  }
```
bduranleau-nr: `nr_txn_get_current_segment` checks if `txn` is `NULL` before doing anything else, so this conditional could be removed.

agent/php_execute.c
```diff
@@ -634,6 +634,22 @@ static void nr_framework_log(const char* log_prefix,
...
 }
 
+/*
+ * Return the parent segment.
+ * This will only be non-null when creating a segment for a new fiber context
+ * that needs to be a child of another fiber context. If it is non-null, it will
+ * be set to NULL after being used. The helper function will ensure the fiber
+ * parent segment is only used once.
+ */
+static nr_segment_t* nr_use_fiber_parent() {
```
bduranleau-nr: It looks like this function isn't used anywhere in this PR

agent/php_execute.c
```diff
@@ -1599,7 +1615,10 @@ void nr_php_execute_internal(zend_execute_data* execute_data,
   if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).show_executes)) {
     nr_php_show_exec_internal(NR_EXECUTE_ORIG_ARGS_OVERWRITE, func TSRMLS_CC);
   }
-  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  segment = nr_segment_start(NRPRG(txn), NRPRG(fiber_parent_segment),
+                             NRPRG(current_php_context));
+  NRPRG(fiber_parent_segment) = NULL;
```
bduranleau-nr, line 1604: Do we always want to overwrite `fiber_parent_segment` following the call to `nr_segment_start`? If so, does it make sense to move this logic into that function?

agent/php_execute.c
```diff
@@ -2010,7 +2120,8 @@ static void nr_php_instrument_func_end(NR_EXECUTE_PROTO) {
    * start and then stop a segment. If that happened, we want to ensure we
    * get the now-current segment
    */
-  segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
+  //  segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
```
bduranleau-nr:
```suggestion
```

agent/php_execute.c
zsistla: What are your thoughts on this one? I was going back and forth on this one.  There are only 2 cases where this is relevant, and I wasn't sure whether it was worth the extra abstraction of `nr_use_fiber_parent`.

agent/php_newrelic.h
```diff
@@ -589,6 +596,22 @@ nr_hashmap_t* predis_commands;
...
 
+/*
+ * Pointer to the current context string.
+ * If NULL, execution is occuring in the main PHP Process.
+ * Otherwise it will point to the fiber context string.
+ */
+const char* current_php_context;
+
+/* Contains the fiber address as a string to use as the async_context.*/
+char fiber_context_string[32];
```
bduranleau-nr, line 607: These two variables seem to be closely linked in code, to the point that `fiber_context_string` seems to exist only to populate `current_php_context`. What's the purpose/benefit of separating these variables? Could they be combined? Is the size of `fiber_context_string` always guaranteed to be `32`? Where does that value come from?

agent/php_observer.c
```diff
@@ -130,28 +130,113 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
+
 static void nr_fiber_disable(zend_fiber_context* fiber_context) {
+  nr_fiber_show_fiber(fiber_context, "init/destroy");
   if (NULL != NRPRG(txn)) {
     /* Fiber init/destroy detected, end and keep the transaction. */
     nrl_warning(NRL_INSTRUMENT,
                 "Transaction is truncated because PHP Fiber use is detected.");
-    nrm_force_add(NRPRG(txn)->unscoped_metrics, "Supportability/PHP/Fiber/used",
-                  0);
+    NR_FIBER_USED_CREATE_METRIC;
```
bduranleau-nr: Since the macro ends in a `;` character, each invocation of the macro can drop the `;` character, or vice-versa

agent/php_execute.c
```diff
@@ -1695,6 +1714,96 @@ void nr_php_user_instrumentation_from_opcache(TSRMLS_D) {
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA /* PHP8+ and OAPI */
 
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO    
+void nr_fiber_show_fiber(zend_fiber_context* zfc, const char* fiber_action) {
+  if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).show_fibers)) {
```
bduranleau-nr: We can de-nest this whole function by 1 level if we instead change this logic to the inverse and just return early.
```c
if (nrlikely(0 == NR_PHP_PROCESS_GLOBALS(special_flags).show_fibers)) {
  return;
}
```

agent/php_execute.c
zsistla: To clarify the two spots, it only needs to happen [here](https://github.com/newrelic/newrelic-php-agent/pull/1196/changes#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1620) and [here](https://github.com/newrelic/newrelic-php-agent/pull/1196/changes#diff-beedad7bbfc9687348e557b4752fbf6f54e4407a6e3177e8cabaaec7e3ffd625R1930), not after every call to `nr_segment_start`.

agent/php_observer.c
```diff
@@ -130,28 +130,113 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
     nr_php_txn_end(0, 0 TSRMLS_CC);
   }
 }
 
 static void nr_fiber_switch_disable(zend_fiber_context* from,
                                     zend_fiber_context* to) {
+  nr_fiber_show_fiber(from, "switch from");
+  nr_fiber_show_fiber(to, "switch to");
```
bduranleau-nr: Depending on how often we expect these functions (`nr_fiber_disable`, `nr_fiber_switch_disable`, `nr_fiber_init_observe`, `nr_fiber_destroy_observe`, `nr_fiber_set_contexts`, `nr_fiber_switch_observe`) to execute, It might be worth wrapping these calls in the special flags conditional check to avoid an unconditional `CALL` operation to the function just to immediately return if it's unlikely this flag will be set during normal program execution.

agent/php_newrelic.h
zsistla: Since Fiber use is going to be an extremely small to zero portion of PHP code, some coding decisions were based on reducing any additional toil in non-fiber cases.

Some decisions were made to impact non-fiber code paths as little as possible (i.e., by not adding additional logic/checks if possible), and this is one of them.
Instead of modifying all functions that currently take a context pointer and have them do an additional check for both null and empty string, current_php_context exists to point to fiber context if it exists, otherwise, will only ever be NULL for majority of cases.

The fiber context string :
/* Contains the fiber address as a string to use as the async_context.*/
It will not be more than 32 and is guaranteed to be unique for the life of the fiber.

agent/php_observer.c
zsistla: Good catch, I think the VScode autoadded that.

agent/php_observer.c
```diff
@@ -130,28 +130,113 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
+ */
+static inline void nr_fiber_set_fiber_parent_segment(zend_fiber_context* zfc) {
+  if (zfc->kind != zend_ce_fiber) {
+    /* Main process is the parent, set to NULL. */
+    NRPRG(fiber_parent_segment) = NULL;
+  } else {
+    const char* parent_fiber_context_string = nr_formatf("%p", zfc);
+    NRPRG(fiber_parent_segment)
+        = nr_txn_get_current_segment(NRPRG(txn), parent_fiber_context_string);
+    nr_free(parent_fiber_context_string);
```
bduranleau-nr, line 224: This is attempting to `free` a `const char *` by casting to `void *` and implicitly dropping the `const` qualifier. ~If we need to free this value, we should probably explicitly cast it to `(void *)` in the call to `nr_free`~, but I think ideally we'd want to avoid this kind of logic entirely.

Actually, due to the way we call `free`, we can't add a cast to `(void *)` since this would likely cause a compiler error as opposed to a normal `free` call.

zsistla: removed const:
[ba1b2ca](https://github.com/newrelic/newrelic-php-agent/pull/1196/commits/ba1b2ca6364de5632ddac9129527d88646b10c04)

agent/php_execute.c
zsistla: denested [456746d](https://github.com/newrelic/newrelic-php-agent/pull/1196/commits/456746d8b22f08bfc37940566b44eaa07fc062d9)

agent/php_observer.c
zsistla: removed here [92b4684](https://github.com/newrelic/newrelic-php-agent/pull/1196/commits/92b46841f8f53c051acb606a80556f13e78e9307)

agent/php_txn.c
```diff
@@ -1268,11 +1268,7 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
    */
 #if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO \
     && !defined OVERWRITE_ZEND_EXECUTE_DATA
-  nr_segment_t* segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
-  while (NULL != segment && segment != NRTXN(segment_root)) {
-    nr_segment_end(&segment);
-    segment = nr_txn_get_current_segment(NRPRG(txn), NULL);
-  }
+  nr_txn_finalize_parent_stacks(NRPRG(txn));
```
bduranleau-nr, line 1271: Was this removed assuming future work to process segments during txn_end? `nr_txn_finalize_parent_stacks` hasn't been modified in this PR and doesn't replace the logic from the removed block above as far as I can tell

agent/php_api.c
zsistla: removed

agent/php_txn.c
zsistla: This section of the code is to cleanup still open segments (for instance if a txn_end were called before the request ended).
This does replace the logic which was removed which simply iterated through the [default_parent_stack](https://github.com/newrelic/newrelic-php-agent/blob/R12.6/axiom/nr_txn.h#L270) and ended all open segments.  With the move to non-stacked segments, this code block would mostly work, but would leak memory anytime there was a segment in any of the [parent_stacks](https://github.com/newrelic/newrelic-php-agent/blob/R12.6/axiom/nr_txn.h#L273) which is what happens when there is an async context.  nr_txn_finalize_parent_stacks does what we really needed to do here which is go through all the default and parents stacks and end any open segments vs just going through the default parent stack that the removed code does.

agent/php_observer.c
zsistla: wrapped [9af3be0](https://github.com/newrelic/newrelic-php-agent/pull/1196/commits/9af3be00487d89d8efa902093fd8c6fc4021c891)

agent/php_execute.c
bduranleau-nr: In that case I think it's fine to leave it un-abstracted as-is

agent/php_execute.c
zsistla: removed [b634b33](https://github.com/newrelic/newrelic-php-agent/pull/1196/commits/b634b337b792b9b975683828aa71f344c9804d8c)

axiom/tests/test_txn.c
```diff
@@ -7849,6 +7855,68 @@ static void test_get_current_span_id(void) {
...
+  span_id_async = nr_txn_get_current_span_id(txn, async_context);
+  tlib_fail_if_null("span id is created for segment with async context",
+                    span_id_async);
+  nr_free(span_id_async);
+
+  /*
+   * Test : default segment and async segment are different.
+   * And span ids are also different.
+   */
+  //  nr_txn_set_current_segment(txn, segment_async);
```
bduranleau-nr:
```suggestion
```

agent/php_execute.c
bduranleau-nr: I see the function description was removed, but the function itself is still present

zsistla: really gone this time: [8bcf298](https://github.com/newrelic/newrelic-php-agent/pull/1196/commits/8bcf298b7214f9234a28ffc089f84f79e7d79906)

bduranleau-nr (approved):

PR #1198, opened 2026-04-10
feat(agent): Fiber Support

zsistla:


agent/php_observer.h
```diff
@@ -18,6 +18,8 @@
 
 #include "Zend/zend_observer.h"
 
+#define NR_PHP_DEFAULT_SUSPEND_TIMES 10
```
ZNeumann, line 29: Can we have a comment explaining why this magic number is needed and why 10 was chosen?

axiom/nr_segment.h
```diff
@@ -285,17 +285,48 @@ typedef nr_segment_iter_return_t (*nr_segment_iter_t)(nr_segment_t* segment,
...
+ * segment with the current parent context.
+ */
+#define NR_SEGMENT_START_WITH_PARENT_CONTEXT(txn, parent) \
+  nr_segment_start(txn, NULL, nr_segment_get_context(parent))
+/*
+ * Purpose : This is a wrapper to nr_segment_start to automatically start the
+ * segment with the current txn context.
+ */
+#define NR_SEGMENT_START_WITH_TXN_CONTEXT(txn) \
+  nr_segment_start(txn, NULL, nr_txn_get_current_context(txn))
```
ZNeumann, line 328: What is the advantage of using macros here over, say, inline functions?

axiom/nr_txn.h
```diff
@@ -275,7 +275,9 @@ typedef struct _nrtxn_t {
                                       context */
   nr_segment_t* force_current_segment; /* Enforce a current segment for the
                                           default context, overriding the
-                                          default parent stack. */
+                                          default parent stack. Only used with
+                                          PHP < 8.0. Do not use with OAPI; it is
+                                          not setup to utilize this.*/
```
ZNeumann, line 280: Does it break things to put this in a PHP version compile guard?

axiom/nr_txn.h
```diff
@@ -1089,6 +1119,8 @@ extern nr_segment_t* nr_txn_get_current_segment(nrtxn_t* txn,
  *
  *           This function is useful to temporarily inject segments that don't
  *           use the default allocator.
+ * Note    : This is ONLY ever set with non-OAPI and PHPs less than 8.0. Do not
+ * use with OAPI; it is not setup to utilize this.
  */
 inline static void nr_txn_force_current_segment(nrtxn_t* txn,
```
ZNeumann, line 1125: If this is never to be utilized with OAPI, should it be wrapped in a PHP version compile gaurd?

agent/php_observer.h
zsistla: clarifying comment
[01a1e16](https://github.com/newrelic/newrelic-php-agent/pull/1198/commits/01a1e16197c4d1f28a88e6fbc1c6a5d993947ce3)

axiom/nr_segment.h
zsistla: There was a circular dependency preventing inline functions from working.
The macros provide the same logic without the circular dependency limitations.

axiom/nr_txn.h
zsistla: That's beyond the scope of this PR.  The PR simply expanded the comment to be more explanatory of current behavior.

axiom/nr_txn.h
zsistla: That's beyond the scope of this PR.  The PR simply expanded the comment to be more explanatory of current behavior.

axiom/nr_segment.h
ZNeumann: I think the circular dependency can be avoided by moving `nr_segment_start_with_txn_context` to `axiom/nr_txn`. I would argue that location might be a more appropriate place for the function regardless, as it is a txn focused operation. I am of the opinion that this makes more sense than using a macro, though I am open to counter-arguments.

zsistla: I'm not convinced it would solve the circular dependency.

But regardless, it's not a txn focused operation any more than nr_segment_start is since this is just a variant of nr_segment_start. The macro is only for ease of use and to prevent mistypes when wanting to use `nr_segment_start(txn, NULL, nr_txn_get_current_context(txn))`.
It's just an nr_segment_start shortcut and the location in this file with nr_segment_start does seems appropriate.

ZNeumann (dismissed):

ZNeumann (approved):

PR #1199, opened 2026-04-13
feat: add zts build and install support

lavarou:
Allow to build and install for thread safe PHP (8.2+ only). Following changes were required:
- Add `-lrt` to `TEST_LIBS` - ZTS builds have zend-max-execution-timers enabled and POSIX timer symbols (timer_create, timer_settime, timer_delete) used by zend_max_execution_timer from libphp.a for PHP 8.1+ are unresolved when `-lrt` appears in `LDFLAGS` before `libphp.a` on the link line. Moving `-lrt` to `TEST_LIBS` ensures it is processed after `libphp.a`, satisfying the GNU linker's left-to-right symbol resolution.
- Fix fw_support unit test - Underlying functions rely on INI settings being available and that requires Zend Engine to be created.
- Fix ZTS-incompatible INI modify handlers - Make `newrelic.distributed_tracing.sampler.remote_parent_sampled` and `newrelic.framework.wordpress.hooks.options` INI modify handlers use ts_resource() in ZTS builds. Without this, mh_arg2 is used as a raw pointer instead of a TSRM resource ID, corrupting memory and causing a segfault in ts_free_id() during module shutdown.
- Fix test_redis segfault on ZTS builds - Implement a workaround of loading redis extension via INI at PHP engine startup because dynamic loading with php_load_extension in ZTS doesn't work and causes segfault.

### DIFF
--- a/agent/Makefile.frag
+++ b/agent/Makefile.frag
@@ -279,7 +279,7 @@ tests/test_txn.o: ../VERSION
 #
 # Used when linking test binaries.
 #
-TEST_LIBS := $(PHP_EMBED_LIBRARY) $(shell $(PHP_CONFIG) --libs)
+TEST_LIBS := $(PHP_EMBED_LIBRARY) $(shell $(PHP_CONFIG) --libs) -lrt
 TEST_LDFLAGS := $(shell $(PHP_CONFIG) --ldflags) $(EXPORT_DYNAMIC)
 TEST_LDFLAGS += $(USER_LDFLAGS)
 CROSS_AGENT_DIR := $(CURDIR)/../axiom/tests/cross_agent_tests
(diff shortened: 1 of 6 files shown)


agent/Makefile.frag
```diff
@@ -279,7 +279,7 @@ tests/test_txn.o: ../VERSION
 #
 # Used when linking test binaries.
 #
-TEST_LIBS := $(PHP_EMBED_LIBRARY) $(shell $(PHP_CONFIG) --libs)
+TEST_LIBS := $(PHP_EMBED_LIBRARY) $(shell $(PHP_CONFIG) --libs) -lrt
```
mfulb, line 282: Is this ok if it is not conditional on ZTS?

mfulb: Discussion with PR author reveals this library will be ignored if no symbols are required.

mfulb (approved):

agent/Makefile.frag
lavarou: Yes. For non-ZTS, this is a no-op. Linker has `rt` at its disposal if it needs to resolve symbols, but it doesn’t need any for non-ZTS builds.

Issue #130, opened 2021-04-06
RedisCluster and full Redis support


b-viguier:
## Summary

* NR extension wraps calls to `Redis` extension, but [some functions are missing](https://github.com/newrelic/newrelic-php-agent/blob/3f93ee47f80703d46d8fccd53be7d6b80361a594/agent/php_internal_instrument.c#L3258-L3291) (`mget` in particular)
* What about `RedisCluster`? The exposed interface is quite similar, is it possible to also wrap it?

## Desired Behavior

* `Redis::mget` function is monitored by NR
* `RedisCluster::*` calls are monitored by NR

## Possible Solution

As far as I understand the code, the magic should happen [here](https://github.com/newrelic/newrelic-php-agent/blob/3f93ee47f80703d46d8fccd53be7d6b80361a594/agent/php_internal_instrument.c#L3258-L3291) 😅 

## Additional context

* `Redis::mget` is used by the Redis cache adapter [provided in Symfony applications](https://github.com/symfony/symfony/blob/3ca3de50c336baefa325be3f36231ab7a7795697/src/Symfony/Component/Cache/Traits/RedisTrait.php#L345). Then in our application we see `set` calls, but no `mget`
* Symfony also provides some adapters for `RedisCluster`, but it's hard to bypass the framework to manually corresponding database records 😕 

🙇‍♂️

joshuabenuck:
@b-viguier Thanks for submitting this enhancement request! This is definitely something we are interested in pursuing. We don't currently have a timeline for when we will be able to get to it, but are adding it to our backlog for consideration. We are also open to PRs if you are able to help out in the meantime!

ostrolucky:
In case you use [snc redis bundle](https://github.com/snc/SncRedisBundle), I've made an [extension for it that records redis queries to NR](https://github.com/ostrolucky/newrelic-datastore-redis-cluster-extension)

duartegarin:
Hi NR team!
I would like to bump this.
We just moved to cluster mode and were disappointed to see our redis metrics evaporated.
Running single node Redis only gets you so far and most installations in bigger applications will need sharding in Redis, so this becomes a must have pretty quickly.
Do you have any updates on this? Are there any workarounds?
Thanks!

samuelnogueira:
[...]

PR #1206, opened 2026-04-27
feat: add \RedisCluster instrumentation

shavounet:
The agent already auto-instruments 149 `\Redis::*` methods (`php_internal_instrument.c`). The phpredis `\RedisCluster` class exposes the same operation surface with the same semantics but does not extend `\Redis`, so the existing hooks don't fire on cluster-mode instances. Applications using ElastiCache/Valkey/Redis cluster get effectively zero datastore visibility from this agent — only `redis::connect/select/eval` are auto-recorded (because they're called via the `\Redis` fallback path), all other methods (`get`/`set`/`mget`/`hgetall`/...) are missed.

This change adds 139 parallel `\RedisCluster::*` hooks routing to the existing class-agnostic INNER handlers (`redis_function`, `redis_close`). The handlers use `NR_PHP_INTERNAL_FN_THIS()` and don't depend on the concrete class. `NR_DATASTORE_REDIS` is reused so cluster traffic shows up under the standard `Datastore/Redis/*` metric namespace — matching how `lib_predis.c` already reports cluster-mode Predis traffic.

Excluded methods (10 total, none of which exist on `\RedisCluster`):

- **Connection management** — `connect`/`pconnect`/`open`/`popen`/`select`. `RedisCluster` is constructed via `__construct(seeds[])`, not `connect`, and cluster mode supports only db 0.
- **`\Redis`-only or single-node-only methods** — `move`/`swapdb` (cluster has only db 0), `migrate` (single-node only), `lremove`/`delete` (`\Redis`-only aliases for `lrem`/`del`).

No `__construct` hook is added in this PR, so per-instance datastore metadata (host, port) is never populated — segments record with `peer.address = unknown:unknown`. Timing/throughput aggregates still work; only the per-instance breakdown is missing.

## Validation

End-to-end smoke-tested on a Symfony 7.1 / PHP 8.3 sandbox:

- **Backend**: official `redis:7-alpine`, single-node, cluster-mode enabled with all 16384 slots assigned to itself via `CLUSTER ADDSLOTSRANGE 0 16383` — the minimum topology `\RedisCluster::__construct` accepts.
- **Workload**: a controller that calls `\RedisCluster::set` / `get` / `mget` / `del` on a fresh per-request key, hit repeatedly with `curl`.
- **Comparison**: same image, same sandbox, only `/usr/lib/php/20230831/newrelic.so` swapped between the stock NR agent (12.6.0.34) and a build of this branch. Two distinct `appName`s used to keep the data sets separable in NR.

After the swap, NR receives metrics that didn't exist before — `Datastore/Redis/all`, `Datastore/Redis/allWeb`, and per-operation `Datastore/operation/Redis/<op>` (e.g. `get`, `set`, `mget`, `del`) — and `category=datastore` Span events with `component=Redis`:

| `name` | `category` | `component` |
|---|---|---|
| `Datastore/operation/Redis/get` | `datastore` | `Redis` |
| `Datastore/operation/Redis/set` | `datastore` | `Redis` |
| `Datastore/operation/Redis/mget` | `datastore` | `Redis` |
| `Datastore/operation/Redis/del` | `datastore` | `Redis` |

Refs: #130

### DIFF
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -3290,6 +3290,146 @@ NR_OUTER_WRAPPER(redis_zrevrank)
 NR_OUTER_WRAPPER(redis_zscore)
 NR_OUTER_WRAPPER(redis_zunionstore)
 
+NR_OUTER_WRAPPER(rediscluster_append)
+NR_OUTER_WRAPPER(rediscluster_bitcount)
+NR_OUTER_WRAPPER(rediscluster_bitop)
+NR_OUTER_WRAPPER(rediscluster_bitpos)
+NR_OUTER_WRAPPER(rediscluster_close)
+NR_OUTER_WRAPPER(rediscluster_dbsize)
+NR_OUTER_WRAPPER(rediscluster_decr)
+NR_OUTER_WRAPPER(rediscluster_decrby)
+NR_OUTER_WRAPPER(rediscluster_del)
+NR_OUTER_WRAPPER(rediscluster_eval)
+NR_OUTER_WRAPPER(rediscluster_evalsha)
+NR_OUTER_WRAPPER(rediscluster_exec)
+NR_OUTER_WRAPPER(rediscluster_exists)
+NR_OUTER_WRAPPER(rediscluster_expire)
+NR_OUTER_WRAPPER(rediscluster_expireat)
+NR_OUTER_WRAPPER(rediscluster_flushall)
+NR_OUTER_WRAPPER(rediscluster_flushdb)
+NR_OUTER_WRAPPER(rediscluster_geoadd)
+NR_OUTER_WRAPPER(rediscluster_geodist)
+NR_OUTER_WRAPPER(rediscluster_geohash)
+NR_OUTER_WRAPPER(rediscluster_geopos)
+NR_OUTER_WRAPPER(rediscluster_georadius)
+NR_OUTER_WRAPPER(rediscluster_georadius_ro)
+NR_OUTER_WRAPPER(rediscluster_georadiusbymember)
+NR_OUTER_WRAPPER(rediscluster_georadiusbymember_ro)
+NR_OUTER_WRAPPER(rediscluster_get)
+NR_OUTER_WRAPPER(rediscluster_getbit)
+NR_OUTER_WRAPPER(rediscluster_getrange)
+NR_OUTER_WRAPPER(rediscluster_getset)
+NR_OUTER_WRAPPER(rediscluster_hdel)
+NR_OUTER_WRAPPER(rediscluster_hexists)
+NR_OUTER_WRAPPER(rediscluster_hget)
+NR_OUTER_WRAPPER(rediscluster_hgetall)
+NR_OUTER_WRAPPER(rediscluster_hincrby)
+NR_OUTER_WRAPPER(rediscluster_hincrbyfloat)
+NR_OUTER_WRAPPER(rediscluster_hkeys)
+NR_OUTER_WRAPPER(rediscluster_hlen)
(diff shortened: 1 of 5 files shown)


lavarou:
@shavounet thank you for your contribution! However, in order for your contribution to be complete, integration tests for Redis Cluster should be included. There already exist Redis integration tests that could be used as baseline - please take a look at an example integration test for Redis: [tests/integration/redis/test_basic.php](https://github.com/newrelic/newrelic-php-agent/blob/v12.6.0.34/tests/integration/redis/test_basic.php). 


Existing Redis tests use the `redisdb` service from the docker-compose project as their datastore (see [here](https://github.com/newrelic/newrelic-php-agent/blob/v12.6.0.34/docker-compose.yaml#L25-L27)). For Redis Cluster tests you'll need to add a new cluster-mode service (the existing `redisdb` runs in standalone mode and cannot be used). A single-node cluster works fine — start Redis with `--cluster-enabled yes` and assign all slots via `CLUSTER ADDSLOTSRANGE 0 16383`; the PR description already describes exactly this setup.

I also suggest using `EXPECT_METRICS_EXIST` (see example [here](https://github.com/newrelic/newrelic-php-agent/blob/v12.6.0.34/tests/integration/pdo/mysql/factory/test_query_fetch_into.php#L32-L45)) and `EXPECT_METRICS_DONT_EXIST` (see example [here](https://github.com/newrelic/newrelic-php-agent/blob/v12.6.0.34/tests/integration/frameworks/wordpress/test_wordpress_00_1.php#L39-L42)) directives over the full `EXPECT_METRICS` — they're much less brittle. Importantly, `EXPECT_METRICS_DONT_EXIST` should include `Datastore/operation/Redis/connect`, since the absence of a connect metric is the key correctness assertion: it proves `\RedisCluster` operations are being captured as `Datastore/Redis/*` without the agent treating construction as a connection event.

You can use [containerized development environment](https://github.com/newrelic/newrelic-php-agent/blob/v12.6.0.34/docs/dev_environment.md) to build the agent and run the tests locally, only the New Relic license key is required.

shavounet:
@lavarou thanks for the pointers — pushed 7e44393 with the integration test you asked for, fully mirroring `tests/integration/redis/test_basic.php`.

A few small adaptations from the redis baseline worth flagging (all because cluster client semantics differ from `\Redis`):

| op | adaptation |
|---|---|
| `type()` | uses `Redis::REDIS_STRING` (the `\Redis` class constant) instead of `RedisCluster::REDIS_STRING` — the latter doesn't exist on `\RedisCluster` |
| `msetnx()` | returns an array of per-master results (1-element with same-slot keys), not a bool — assertion adapted accordingly |
| `pfadd()` | returns `bool` on `\RedisCluster` instead of `int` on `\Redis` — `tap_assert`/`tap_refute` instead of `tap_equal(0/1, ...)` |
| `eval`/`evalsha`/`time`/`rawcommand`/`publish` | take a key arg for routing, threaded through accordingly |

Skipped vs the `\Redis` baseline: `connect`/`select` (no equivalent on cluster), `keys $pattern`/`randomkey`/`dbsize` (single-shard semantics in cluster mode — partial result), `pipeline`/`exec` (cluster pipelining requires per-command slot routing — out of scope for a basic test).

Local validation: ran the test PHP directly via `php -dextension=newrelic.so` against the patched agent + a local `redis:7-alpine` running the same `CLUSTER ADDSLOTSRANGE 0 16383` setup — all 37 `tap_assert` calls pass.

lavarou:
@shavounet thank you for your contribution! To let our internal test suites run against it, your work has been incorporated into #1212, which is targeted for an upcoming release. Your original commits are preserved there with attribution. Closing this in favor of #1212 — really appreciate the work here.

PR #1208, opened 2026-04-29
feat(agent): add per-fiber globals

bduranleau-nr:
## Summary

Each PHP Fiber now gets its own snapshot of the request-scoped globals the
agent relies on (`ctx`), so instrumentation running inside
a fiber no longer trips over state owned by the main context (or another
fiber). The fiber observer hooks already in place from the parent branch
are wired up to populate, swap, and tear down these snapshots; the
`NRPRG_CTX` accessor transparently reads from the active
fiber's snapshot when one is set.

All new behavior is gated on PHP 8.1+. 

## What changed

**Per-fiber globals (`agent/php_fibers.{c,h}`, new)**
- New `fiber_globals_t` = `{ ctx_globals_t* }`.
- `nr_fiber_copy_ctx_globals` deep-copy the parts a fiber should own. 
- A per-request `fiber_globals_map` (keyed by `"%p"` of the
  `zend_fiber_context*`) holds these snapshots; it has a destructor so removed
  entries are fully cleaned up. Init is lazy on first fiber; teardown happens
  in `nr_php_post_deactivate`. `php_rshutdown.c` clears `NRPRG(fiber_globals)`
  before normal global teardown so the existing free paths don't reach into a
  snapshot.

**OAPI fiber observer wiring (`agent/php_observer.c`)**
- `nr_fiber_init_observe`: lazy-init `fiber_globals_map`, add a snapshot for
  the new fiber under its `"%p"` key.
- `nr_fiber_destroy_observe`: remove the snapshot for that fiber.
- `nr_fiber_switch_observe`: at the end of the existing switch logic, swap
  `NRPRG(fiber_globals)` to the destination context's snapshot via
  `nr_fiber_switch_global_context`.

**Globals accessors (`agent/php_newrelic.h`)**
- New `fiber_globals_t` struct; `NRPRG(fiber_globals_map)` and
  `NRPRG(fiber_globals)` added to module globals.
- On PHP 8.1+, `NRPRG_CTX(Y)` resolves through
  `NRPRG(fiber_globals)` when it's non-NULL, otherwise read the main-context
  globals exactly as before. Callers don't have to know which context they're
  in. PHP < 8.1 macros are unchanged.

### DIFF
--- a/agent/config.m4
+++ b/agent/config.m4
@@ -213,7 +213,7 @@ if test "$PHP_NEWRELIC" = "yes"; then
   php_api_distributed_trace.c php_api_internal.c php_autorum.c \
   php_call.c php_curl.c php_curl_md.c php_datastore.c php_environment.c \
   php_error.c php_execute.c php_explain.c php_explain_mysqli.c \
-  php_explain_pdo_mysql.c php_extension.c php_file_get_contents.c \
+  php_explain_pdo_mysql.c php_extension.c php_fibers.c php_file_get_contents.c \
   php_globals.c php_hash.c php_header.c php_httprequest_send.c \
   php_internal_instrument.c php_memcached.c php_minit.c php_mshutdown.c php_mysql.c \
   php_mysqli.c php_newrelic.c php_nrini.c php_observer.c php_output.c php_pdo.c \
@@ -221,7 +221,7 @@ if test "$PHP_NEWRELIC" = "yes"; then
   php_rinit.c php_rshutdown.c php_samplers.c php_stack.c \
   php_stacked_segment.c php_txn.c php_user_instrument.c \
   php_user_instrument_wraprec_hashmap.c \
-  php_user_instrument_hashmap.c php_vm.c php_wrapper.c php_fibers.c"
+  php_user_instrument_hashmap.c php_vm.c php_wrapper.c"
   FRAMEWORKS="fw_cakephp.c fw_codeigniter.c fw_drupal8.c \
   fw_drupal.c fw_drupal_common.c fw_joomla.c \
   fw_laminas3.c fw_laravel.c fw_laravel_queue.c \
--- a/agent/config.m4
+++ b/agent/config.m4
@@ -221,7 +221,7 @@ if test "$PHP_NEWRELIC" = "yes"; then
   php_rinit.c php_rshutdown.c php_samplers.c php_stack.c \
   php_stacked_segment.c php_txn.c php_user_instrument.c \
   php_user_instrument_wraprec_hashmap.c \
-  php_user_instrument_hashmap.c php_vm.c php_wrapper.c"
+  php_user_instrument_hashmap.c php_vm.c php_wrapper.c php_fibers.c"
   FRAMEWORKS="fw_cakephp.c fw_codeigniter.c fw_drupal8.c \
   fw_drupal.c fw_drupal_common.c fw_joomla.c \
   fw_laminas3.c fw_laravel.c fw_laravel_queue.c \
--- a/agent/php_fibers.c
+++ b/agent/php_fibers.c
@@ -128,9 +128,11 @@ nr_status_t nrf_fiber_init_global_hashmap(nr_hashmap_t** fiber_globals_map) {
   if (NULL == fiber_globals_map) {
     return NR_FAILURE;
   }
-  if (NULL == *fiber_globals_map) {
-    *fiber_globals_map = nr_hashmap_create(free_fiber_globals);
+  if (NULL != *fiber_globals_map) {
+    return NR_FAILURE;
   }
+
+  *fiber_globals_map = nr_hashmap_create(free_fiber_globals);
   return NR_SUCCESS;
 }
 
--- a/agent/php_fibers.c
+++ b/agent/php_fibers.c
@@ -59,15 +59,12 @@ static void nrf_ctx_global_deep_copy(ctx_globals_t* dest, ctx_globals_t* src) {
   dest->drupal_http_request_depth = 0;
   dest->php_cur_stack_depth = 0;  // PHP allocates a new stack for each fiber
 
-  COPY_BASIC(cufa_callback);
-
   COPY_STRING(mysql_last_conn);
   COPY_STRING(pgsql_last_conn);
 
   COPY_BASIC(datastore_connections);
 
   COPY_BASIC(deprecated_capture_request_parameters);
-  COPY_BASIC(error_group_user_callback);
   COPY_BASIC(check_cufa);
 
   COPY_BASIC(predis_commands);
--- a/agent/php_fibers.c
+++ b/agent/php_fibers.c
@@ -5,11 +5,9 @@
  * This file handles fibers instrumentation.
  */
 
-#include "nr_datastore_instance.h"
 #include "php_includes.h"
 #include "php_compat.h"
 #include "php_fibers.h"
-#include "nr_mysqli_metadata.h"
 #include "php_agent.h"
 #include "php_newrelic.h"
 #include "php_zval.h"
@@ -18,7 +16,6 @@
 #include "util_memory.h"
 #include "util_stack.h"
 #include "util_strings.h"
-#include "util_time.h"
 #include "zend_types.h"
 
 #if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
--- a/agent/php_fibers.c
+++ b/agent/php_fibers.c
@@ -5,22 +5,27 @@
  * This file handles fibers instrumentation.
  */
 
+#include "php_fibers.h"
 #include "nr_mysqli_metadata.h"
 #include "nr_mysqli_metadata_private.h"
 #include "php_agent.h"
-#include "php_newrelic.h"
+#include "php_zval.h"
 #include "util_hashmap.h"
 #include "util_memory.h"
+#include "util_stack.h"
+#include "zend_types.h"
 
 #if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
 
-#define COPY_FIELD(x) (dest->x = src->x)
-#define COPY_HASHMAP(x) (dest->x = nr_hashmap_copy(NRTXNGLOBAL(x)))
+#define COPY_BASIC(x) (dest->x = src->x)
+#define COPY_STRING(x) (nr_strcpy(dest->x, src->x))
+#define COPY_HASHMAP(x) (dest->x = nr_hashmap_copy(src->x))
+#define COPY_STACK(x, y) (nr_stack_copy(&src->x, y))
 
 static void nrf_txn_global_deep_copy(txn_globals_t* dest, txn_globals_t* src) {
-  COPY_FIELD(execute_count);
-  COPY_FIELD(generating_explain_plan);
-  COPY_FIELD(curl_ignore_setopt);
+  COPY_BASIC(execute_count);
+  COPY_BASIC(generating_explain_plan);
+  COPY_BASIC(curl_ignore_setopt);
 
   COPY_HASHMAP(guzzle_objs);
   COPY_HASHMAP(mysqli_queries);
@@ -32,10 +37,55 @@ static void nrf_txn_global_deep_copy(txn_globals_t* dest, txn_globals_t* src) {
   dest->mysqli_links = nr_mysqli_metadata_copy(NRTXNGLOBAL(mysqli_links));
 }
 
-#undef COPY_FIELD
-#undef COPY_HASHMAP
+static void* copy_stack_elem_zval(void* elem) {
+  zval* copy = nr_php_zval_alloc();
+  ZVAL_DUP(copy, elem);
+  return copy;
+}
 
-txn_globals_t* nrf_fiber_init_txn_globals() {
+static void* copy_stack_elem_bool(void* elem) {
+  return elem;
+}
+
+static void* copy_stack_elem_str(void* elem) {
+  return nr_strdup((char*)elem);
+}
+
+static void nrf_ctx_global_deep_copy(ctx_globals_t* dest, ctx_globals_t* src) {
+  COPY_STRING(doctrine_dql);
+
+  COPY_BASIC(drupal_http_request_depth);
+  COPY_BASIC(php_cur_stack_depth);
+
+  COPY_BASIC(cufa_callback);
+
+  COPY_STRING(mysql_last_conn);
+  COPY_STRING(pgsql_last_conn);
+
+  COPY_HASHMAP(datastore_connections);
+
+  COPY_BASIC(deprecated_capture_request_parameters);
+  COPY_BASIC(error_group_user_callback);
+  COPY_BASIC(check_cufa);
+
+  COPY_HASHMAP(predis_commands);
+
+  COPY_STACK(drupal_invoke_all_hooks, copy_stack_elem_zval);
+  COPY_STACK(drupal_invoke_all_states, copy_stack_elem_bool);
+
+  dest->drupal_http_request_segment = NULL;
+
+  COPY_STACK(wordpress_tags, copy_stack_elem_str);
+  COPY_STACK(wordpress_tag_states, copy_stack_elem_bool);
+  COPY_STACK(predis_ctxs, copy_stack_elem_str);
+}
+
+#undef COPY_STACK
+#undef COPY_HASHMAP
+#undef COPY_STRING
+#undef COPY_BASIC
+
+txn_globals_t* nrf_fiber_copy_txn_globals() {
   txn_globals_t* fiber_txn_globals = NULL;
 
   fiber_txn_globals = nr_malloc(sizeof(txn_globals_t));
@@ -43,6 +93,16 @@ txn_globals_t* nrf_fiber_init_txn_globals() {
   nrf_txn_global_deep_copy(fiber_txn_globals, &NRPRG(txn_globals));
 
   return fiber_txn_globals;
+}
+
+ctx_globals_t* nrf_fiber_copy_ctx_globals() {
+  ctx_globals_t* fiber_ctx_globals = NULL;
+
+  fiber_ctx_globals = nr_malloc(sizeof(ctx_globals_t));
+
+  nrf_ctx_global_deep_copy(fiber_ctx_globals, &NRPRG(ctx));
+
+  return fiber_ctx_globals;
 }
 
 void free_fiber_globals(void* fiber_globals) {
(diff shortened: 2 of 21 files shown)


agent/php_observer.c
```diff
@@ -247,6 +283,14 @@ static void nr_fiber_switch_observe(zend_fiber_context* from,
   if (ZEND_FIBER_STATUS_INIT == to->status) {
     nr_fiber_set_fiber_parent_segment(from);
   }
+
+  if (NR_FAILURE
+      == nrf_fiber_switch_global_context(NRPRG(fiber_globals_map),
+                                         &NRPRG(fiber_globals),
+                                         NRPRG_SHARED(current_php_context))) {
+    nrl_warning(NRL_AGENT, "Failed to switch fiber context to %s",
+                NRPRG_SHARED(current_php_context));
```
zsistla, line 301: Should it at least revert to main context in this case instead of letting the previous context be left to be used with the new fiber? Or what should the behavior be? Should we end the txn for the remaining of the request since we won't have the globals we need anyway?

agent/php_observer.c
```diff
@@ -159,25 +160,60 @@ static void nr_fiber_switch_disable(zend_fiber_context* from,
...
 static void nr_fiber_init_observe(zend_fiber_context* zfc) {
+  char zfc_key[32];
   NR_FIBER_USED_CREATE_METRIC
   if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).show_fibers)) {
     nr_fiber_show_fiber(zfc, "init");
   }
+  if (NULL == NRPRG(fiber_globals_map)) {
+    // initialize the fiber global hashmap if it does not already exist
+    nrf_fiber_init_global_hashmap(&NRPRG(fiber_globals_map));
+  }
```
zsistla, line 183: Would this check make more sense to include in the `nrf_add_fiber_context_to_global_hashmap` function?

agent/php_observer.c
```diff
@@ -159,25 +160,60 @@ static void nr_fiber_switch_disable(zend_fiber_context* from,
...
+
+  snprintf(zfc_key, sizeof(zfc_key), "%p", zfc);
+
+  // Add the current context to the global hashmap for the new fiber
```
zsistla, line 187: Nit about commenting style consistency.
When we open sourced, we made all the commenting consistent with the `/* */` style, but I know `//` has been making it's way in since then.

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+
+  nr_hashmap_update(fiber_globals_map, key, nr_strlen(key), fg);
+
+  return NR_SUCCESS;
+}
+
+nr_status_t nrf_remove_fiber_context_from_global_hashmap(
+    nr_hashmap_t* fiber_globals_map,
+    const char* key) {
+  if (NULL == key || nr_strlen(key) < 1) {
```
zsistla: Could consider replacing `(NULL == key || nr_strlen(key) < 1)` 
with `(nr_strempty(key))` which is a little faster because it doesn't use strlen.

agent/php_newrelic.h
```diff
@@ -587,6 +593,10 @@ nrtxn_t* txn;  // The all-important transaction pointer
 
 txn_globals_t txn_globals;  // Transaction Globals
 
+nr_hashmap_t* fiber_globals_map;
+
+fiber_globals_t* fiber_globals;
```
zsistla: Small comments giving context?

axiom/util_stack.c
```diff
@@ -55,3 +55,17 @@ bool nr_stack_remove_topmost(nr_stack_t* s, const void* element) {
 
   return false;
 }
+
+nr_stack_t nr_stack_copy(nr_stack_t* src, nr_stack_clone_elem_t clone) {
```
zsistla: Needs a null check for `src` and a test case added to test_stack.c that passes null.

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+#include "zend_types.h"
+
+#if ZEND_MODULE_API_NO >= ZEND_8_1_X_API_NO
+
+#define COPY_BASIC(x) (dest->x = src->x)
+#define COPY_STRING(x) (dest->x = nr_strdup(src->x))
+#define COPY_HASHMAP(x, y) (dest->x = nr_hashmap_copy(src->x, y))
+#define COPY_STACK(x, y) (dest->x = nr_stack_copy(&src->x, y))
+
+static void* copy_elem_zval(void* elem) {
```
zsistla, line 28: Needs a null check for elem since ZVAL_DUP doesn't check.

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+
+static void* copy_elem_ident(void* elem) {
+  return elem;
+}
+
+static void* copy_elem_str(void* elem) {
+  return nr_strdup((char*)elem);
+}
+
+static void* copy_elem_time(void* elem) {
```
zsistla: Null check?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  nrtime_t* t = NULL;
+  nrtime_t* src = (nrtime_t*)elem;
+  t = (nrtime_t*)nr_malloc(sizeof(nrtime_t));
+  *t = *src;
+  return t;
+}
+
+static void* copy_datastore_instance(void* elem) {
```
zsistla: `elem` Null check?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  nr_datastore_instance_t* d = NULL;
+  nr_datastore_instance_t* src = (nr_datastore_instance_t*)elem;
+  d = nr_malloc(sizeof(nr_datastore_instance_t));
+  d->host = nr_strdup(src->host);
+  d->port_path_or_id = nr_strdup(src->port_path_or_id);
+  d->database_name = nr_strdup(src->database_name);
+  return d;
+}
+
+static void nrf_txn_global_deep_copy(txn_globals_t* dest, txn_globals_t* src) {
```
zsistla: Needs null check for src,dest.
Is there unit test that verifies this case?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  COPY_HASHMAP(pdo_link_options, copy_elem_zval);
+  COPY_BASIC(curl_metadata);
+  COPY_BASIC(curl_multi_metadata);
+  COPY_HASHMAP(prepared_statements, copy_elem_str);
+
+  dest->mysqli_links = nr_mysqli_metadata_copy(src->mysqli_links);
+}
+
+static void nrf_ctx_global_deep_copy(ctx_globals_t* dest, ctx_globals_t* src) {
+  COPY_STRING(doctrine_dql);
```
zsistla, line 54: dest,src null checks.  Are there unit tests for null cases?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  fiber_ctx_globals = nr_malloc(sizeof(ctx_globals_t));
+
+  nrf_ctx_global_deep_copy(fiber_ctx_globals, src);
+
+  return fiber_ctx_globals;
+}
+
+void free_fiber_globals(void* fiber_globals) {
+  fiber_globals_t* f = (fiber_globals_t*)fiber_globals;
+
```
zsistla, line 103: fiber_globals needs a NULL check. Is there unit test for this case?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  if (NULL == f->txn_globals || NULL == f->ctx_globals) {
```
zsistla: Split this?

That way if f->txn_globals isn't NULL but f->ctx_globals is, we can still free f->txn_globals and vice versa. Unit tests?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  nr_free(f->ctx_globals);
+  nr_free(f);
+}
+
+nr_status_t nrf_fiber_init_global_hashmap(nr_hashmap_t** fiber_globals_map) {
+  if (NULL == fiber_globals_map) {
+    return NR_FAILURE;
+  }
+  if (NULL == *fiber_globals_map) {
+    *fiber_globals_map = nr_hashmap_create(free_fiber_globals);
```
zsistla: What does it mean if it's not NULL here?  Is that a case that can happen?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+  }
+  return NR_SUCCESS;
+}
+
+nr_status_t nrf_fiber_destroy_global_hashmap(nr_hashmap_t** fiber_globals_map) {
```
zsistla: Does it need a status if we don't check it?

agent/php_fibers.c
```diff
@@ -0,0 +1,259 @@
...
+
+  return nr_hashmap_delete(fiber_globals_map, key, nr_strlen(key));
+}
+
+nr_status_t nrf_fiber_switch_global_context(nr_hashmap_t* fiber_globals_map,
+                                            fiber_globals_t** fiber_global_ptr,
+                                            const char* key) {
+  fiber_globals_t* fg = NULL;
+
+  if (NULL == fiber_global_ptr) {
```
zsistla, line 197: FAILURE here is very unlikely, but if we encounter failure cases here, what should we do?  Should we stop the txn since the fibers won't have the globals they need?  Or something else?

agent/php_observer.c
bduranleau-nr: I don't think so- I feel like the function to "init" the fiber hashmap belongs explicitly in the fiber_init path. Separating them also allows for more granular testing of the functionality

agent/php_observer.c
bduranleau-nr: I have not come up with a great answer to this question- things will almost certainly break if we hit this case whether we switch to main context or end the txn, although I suppose the "least wrong" or "most safe" approach would be to end the txn. What do you think?

agent/php_fibers.c
bduranleau-nr: added part of d9346281220b61a06b42f678217523fdf4b54d2e

agent/php_fibers.c
bduranleau-nr: added part of d9346281220b61a06b42f678217523fdf4b54d2e

agent/php_fibers.c
bduranleau-nr: A NULL check for this function has been added in the sole caller as part of d9346281220b61a06b42f678217523fdf4b54d2e. Technically, dest probably doesn't need a NULL check since the dest pointer is allocated directly before this function is called by the sole caller. 
A unit test has been added for a NULL src in d9346281220b61a06b42f678217523fdf4b54d2e

agent/php_fibers.c
bduranleau-nr: This function has been removed

agent/php_fibers.c
bduranleau-nr: This function has been removed

agent/php_fibers.c
bduranleau-nr: This function has been removed

agent/php_fibers.c
bduranleau-nr: This has been split and unit tests added in d9346281220b61a06b42f678217523fdf4b54d2e

agent/php_fibers.c
bduranleau-nr: If it's not NULL it means a hashmap already exists and we don't need to do create a new one. I don't expect this to happen since the only caller contains a NULL check for NRPRG(fiber_globals_map) before it calls this function

agent/php_fibers.c
bduranleau-nr: I can remove it if you prefer

agent/php_fibers.c
bduranleau-nr: applied cb6a8ff6ddca9f0f4031b7e12ede646eee26a630

agent/php_fibers.c
bduranleau-nr: Most likely this should result in ending the txn, but that implementation shouldn't come in this function directly- I think it should happen in the caller based on the return value.

agent/php_newrelic.h
bduranleau-nr: added abcc9973128d33d05f5c85f75d18e7dcaa84d736

axiom/util_stack.c
bduranleau-nr: What value should be returned if `src` is NULL? An empty `dest`?

zsistla: NULL dest?

agent/php_fibers.c
zsistla: Doesn't really change things one way or the other. Was just wondering if the intent had been to use it.

agent/php_observer.c
zsistla: Ending the txn sounds safest and cleanest.

agent/php_fibers.c
zsistla: > Most likely this should result in ending the txn, but that implementation shouldn't come in this function directly- I think it should happen in the caller based on the return value.

Makes sense.

axiom/util_stack.c
bduranleau-nr: `dest` cannot be `NULL` / the function cannot return `NULL`- the function returns an `nr_stack_t` value type, not a pointer.

zsistla: Ah right, then empty `dest` makes most sense.

bduranleau-nr: I made 2 commits to try to address this: 828b7de4d12bf5d18c946d42de6a33d5fcb5bc93 implements the change, 1c3435ee7b17c9605dc8f9c086f6bc0df068d7a7 adds unit tests around it

agent/php_observer.c
bduranleau-nr: I added a call to end the txn for this case (where we fail to switch fiber context) as well as the case where we fail to add a fiber context to the hashmap in init: 0261a7cdada71618c61c310b8a075af1ce806935. I opted not to end the txn in the case that we fail to remove a fiber from the hashmap in the destroy function, but I was on the fence as to whether we should include it there or not.

agent/php_fibers.c
bduranleau-nr: The changes made from this comment https://github.com/newrelic/newrelic-php-agent/pull/1208#discussion_r3462764080 should handle ending the txn in this edge case, along with this change 9fbf0db4371d3c9156fff4acb4fdef2246f1d13c which hardens the init function against a non-NULL fiber_globals_map by returning NR_FAILURE and ending the txn

agent/config.m4
```diff
@@ -221,7 +221,7 @@ if test "$PHP_NEWRELIC" = "yes"; then
   php_rinit.c php_rshutdown.c php_samplers.c php_stack.c \
   php_stacked_segment.c php_txn.c php_user_instrument.c \
   php_user_instrument_wraprec_hashmap.c \
-  php_user_instrument_hashmap.c php_vm.c php_wrapper.c"
+  php_user_instrument_hashmap.c php_vm.c php_wrapper.c php_fibers.c"
```
zsistla: Should be alphabetized order.

agent/php_fibers.h
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Params  : 1. A pointer to the source ctx_globals_t to copy from. Must be
+ *              non-NULL. The source is not modified and remains owned by
+ *              the caller; typically this is &NRPRG(ctx).
+ *
+ * Returns : A pointer to a newly allocated ctx_globals_t. Ownership of the
+ *           returned struct is transferred to the caller, which is responsible
+ *           for freeing it.
+ */
+extern ctx_globals_t* nrf_fiber_copy_ctx_globals(ctx_globals_t* src);
```
zsistla: why `nrf` instead of `nr`?
It breaks our convention.  Could we just use `nr` instead?

agent/php_fibers.c
```diff
@@ -0,0 +1,226 @@
...
+
+static void nrf_ctx_global_deep_copy(ctx_globals_t* dest, ctx_globals_t* src) {
+  if (NULL == dest) {
+    return;
+  }
+
+  COPY_STRING(doctrine_dql);
+
+  dest->drupal_http_request_depth = 0;
+  dest->php_cur_stack_depth = 0;  // PHP allocates a new stack for each fiber
```
zsistla: We use this internally to indicate how deep we are in terms of call frames.  

The value here should copy the src value.

So if we are 3 calls in and call a fiber, the php_cur_stack_depth should also start at 3 and it will get incremented when we go through func_begin.

agent/php_observer.c
```diff
@@ -159,25 +160,67 @@ static void nr_fiber_switch_disable(zend_fiber_context* from,
...
+  char zfc_key[32];
   NR_FIBER_USED_CREATE_METRIC
   if (nrunlikely(NR_PHP_PROCESS_GLOBALS(special_flags).show_fibers)) {
     nr_fiber_show_fiber(zfc, "init");
   }
+  if (NULL == NRPRG(fiber_globals_map)) {
+    // initialize the fiber global hashmap if it does not already exist
+    if (NR_FAILURE
+        == nrf_fiber_init_global_hashmap(&NRPRG(fiber_globals_map))) {
+      nrl_warning(NRL_AGENT, "Failed to initialize the fiber global hashmap");
```
zsistla:
```suggestion
      nrl_warning(NRL_AGENT, "Failed to initialize the fiber global hashmap needed for a fiber aware transaction and must therefore end the transaction.");
```

agent/php_observer.c
```diff
@@ -159,25 +160,67 @@ static void nr_fiber_switch_disable(zend_fiber_context* from,
...
+
+  // Add the current context to the global hashmap for the new fiber
+  if (NR_FAILURE
+      == nrf_add_fiber_context_to_global_hashmap(
+          NRPRG(fiber_globals_map),
+          NRPRG(fiber_globals) ? NRPRG(fiber_globals)->ctx_globals
+                               : &NRPRG(ctx),
+          zfc_key)) {
+    nrl_warning(NRL_AGENT,
+                "Failed to add fiber context to global hashmap for fiber %s",
```
zsistla:
```suggestion
                "Failed to add fiber context to global hashmap for fiber %s needed for a fiber aware transaction and must therefore end the transaction.",
```

agent/php_observer.c
```diff
@@ -247,6 +290,15 @@ static void nr_fiber_switch_observe(zend_fiber_context* from,
   if (ZEND_FIBER_STATUS_INIT == to->status) {
     nr_fiber_set_fiber_parent_segment(from);
   }
+
+  if (NR_FAILURE
+      == nrf_fiber_switch_global_context(NRPRG(fiber_globals_map),
+                                         &NRPRG(fiber_globals),
+                                         NRPRG_SHARED(current_php_context))) {
+    nrl_warning(NRL_AGENT, "Failed to switch fiber context to %s",
```
zsistla:
```suggestion
    nrl_warning(NRL_AGENT, "Failed to switch fiber context to %s needed for a fiber aware transaction and must therefore end the transaction.",
```

agent/php_fibers.h
bduranleau-nr: The intent was to differentiate it as fibers specific like we do for logging where we use `nrl_`. I've changed  the signatures to just use `nr_` here 2037ce12f975f2ad0455367c6eceba62f2655b74

agent/php_fibers.c
bduranleau-nr: 836356fdbcc1941974156c9f3df7d9623e6fd066

agent/config.m4
bduranleau-nr: f3078235cd0f10233561a5ec3c15c83aeb9e37d8

axiom/util_stack.c
```diff
@@ -55,3 +55,23 @@ bool nr_stack_remove_topmost(nr_stack_t* s, const void* element) {
...
```
zsistla: The name of this didn't make it apparent to me that it was a pointer.

bduranleau-nr: The function signature doc makes it clear that it expects a pointer here:

>  * Params  : 1. A pointer to the source stack, src.

zsistla: The func sig is clear, but I meant the name of the `nr_stack_clone_elem_t` didn't indicate it was a ptr type value.
Not a blocker, but it would have been clearer to me if it were something like: `nr_stack_clone_elem_ptr_t`

bduranleau-nr: Ah I see- I can make that change

bduranleau-nr: bfa7103a0fef2e371da9abc9c86591f0485bf795

zsistla (approved):

PR #1212, opened 2026-04-30
feat(agent): Add Redis Cluster instrumentation

hahuja2:
This PR incorporates #1206 by @shavounet contribution with minor adjustments:
---
The agent already auto-instruments 149 `\Redis::*` methods (`php_internal_instrument.c`). The phpredis `\RedisCluster` class exposes the same operation surface with the same semantics but does not extend `\Redis`, so the existing hooks don't fire on cluster-mode instances which results in effectively zero datastore visibility for applications using ElastiCache/Valkey/Redis cluster — only `redis::connect/select/eval` are auto-recorded (because they're called via the `\Redis` fallback path), all other methods (`get`/`set`/`mget`/`hgetall`/...) are missed.

This change adds 139 parallel `\RedisCluster::*` hooks routing to new `rediscluster_function` which differs from existing `redis_function` by skipping instance retrieval (see more detailed explanation below). `NR_DATASTORE_REDIS` is reused so cluster traffic shows up under the standard `Datastore/Redis/*` metric namespace — matching how `lib_predis.c` already reports cluster-mode Predis traffic with the only difference being per-instance metrics.

Excluded `\Redis` methods (10 total, none of which exist on `\RedisCluster`):

- **Connection management** — `connect`/`pconnect`/`open`/`popen`/`select`. `RedisCluster` is constructed via `__construct(seeds[])`, not `connect`, and cluster mode supports only db 0.
- **`\Redis`-only or single-node-only methods** — `move`/`swapdb` (cluster has only db 0), `migrate` (single-node only), `lremove`/`delete` (`\Redis`-only aliases for `lrem`/`del`).

No `__construct` hook is added in this PR, so per-instance datastore metadata (host, port) is never populated — segments record with `peer.address = unknown:unknown` and metrics record `unknown/unknown`. Timing/throughput aggregates still work; only the per-instance breakdown is missing.

### DIFF
--- a/agent/php_internal_instrument.c
+++ b/agent/php_internal_instrument.c
@@ -3307,6 +3307,146 @@ NR_OUTER_WRAPPER(redis_zrevrank)
 NR_OUTER_WRAPPER(redis_zscore)
 NR_OUTER_WRAPPER(redis_zunionstore)
 
+NR_OUTER_WRAPPER(rediscluster_append)
+NR_OUTER_WRAPPER(rediscluster_bitcount)
+NR_OUTER_WRAPPER(rediscluster_bitop)
+NR_OUTER_WRAPPER(rediscluster_bitpos)
+NR_OUTER_WRAPPER(rediscluster_close)
+NR_OUTER_WRAPPER(rediscluster_dbsize)
+NR_OUTER_WRAPPER(rediscluster_decr)
+NR_OUTER_WRAPPER(rediscluster_decrby)
+NR_OUTER_WRAPPER(rediscluster_del)
+NR_OUTER_WRAPPER(rediscluster_eval)
+NR_OUTER_WRAPPER(rediscluster_evalsha)
+NR_OUTER_WRAPPER(rediscluster_exec)
+NR_OUTER_WRAPPER(rediscluster_exists)
+NR_OUTER_WRAPPER(rediscluster_expire)
+NR_OUTER_WRAPPER(rediscluster_expireat)
+NR_OUTER_WRAPPER(rediscluster_flushall)
+NR_OUTER_WRAPPER(rediscluster_flushdb)
+NR_OUTER_WRAPPER(rediscluster_geoadd)
+NR_OUTER_WRAPPER(rediscluster_geodist)
+NR_OUTER_WRAPPER(rediscluster_geohash)
+NR_OUTER_WRAPPER(rediscluster_geopos)
+NR_OUTER_WRAPPER(rediscluster_georadius)
+NR_OUTER_WRAPPER(rediscluster_georadius_ro)
+NR_OUTER_WRAPPER(rediscluster_georadiusbymember)
+NR_OUTER_WRAPPER(rediscluster_georadiusbymember_ro)
+NR_OUTER_WRAPPER(rediscluster_get)
+NR_OUTER_WRAPPER(rediscluster_getbit)
+NR_OUTER_WRAPPER(rediscluster_getrange)
+NR_OUTER_WRAPPER(rediscluster_getset)
+NR_OUTER_WRAPPER(rediscluster_hdel)
+NR_OUTER_WRAPPER(rediscluster_hexists)
+NR_OUTER_WRAPPER(rediscluster_hget)
+NR_OUTER_WRAPPER(rediscluster_hgetall)
+NR_OUTER_WRAPPER(rediscluster_hincrby)
+NR_OUTER_WRAPPER(rediscluster_hincrbyfloat)
+NR_OUTER_WRAPPER(rediscluster_hkeys)
+NR_OUTER_WRAPPER(rediscluster_hlen)
+NR_OUTER_WRAPPER(rediscluster_hmget)
+NR_OUTER_WRAPPER(rediscluster_hmset)
+NR_OUTER_WRAPPER(rediscluster_hset)
+NR_OUTER_WRAPPER(rediscluster_hsetnx)
+NR_OUTER_WRAPPER(rediscluster_hstrlen)
+NR_OUTER_WRAPPER(rediscluster_hvals)
+NR_OUTER_WRAPPER(rediscluster_incr)
+NR_OUTER_WRAPPER(rediscluster_incrby)
+NR_OUTER_WRAPPER(rediscluster_incrbyfloat)
+NR_OUTER_WRAPPER(rediscluster_keys)
+NR_OUTER_WRAPPER(rediscluster_lget)
+NR_OUTER_WRAPPER(rediscluster_lindex)
+NR_OUTER_WRAPPER(rediscluster_linsert)
+NR_OUTER_WRAPPER(rediscluster_llen)
+NR_OUTER_WRAPPER(rediscluster_lpop)
+NR_OUTER_WRAPPER(rediscluster_lpush)
+NR_OUTER_WRAPPER(rediscluster_lpushx)
+NR_OUTER_WRAPPER(rediscluster_lrange)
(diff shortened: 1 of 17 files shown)


ZNeumann (dismissed):

agent/php_internal_instrument.c
```diff
@@ -1687,6 +1687,24 @@ NR_INNER_WRAPPER(redis_function) {
...
+ */
+NR_INNER_WRAPPER(rediscluster_function) {
+  nr_datastore_instance_t* instance
+      = nr_datastore_instance_create("unknown", "unknown", NULL);
+
+  nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_REDIS,
+                                             nr_wrapper->extra, instance,
+                                             INTERNAL_FUNCTION_PARAM_PASSTHRU);
+  nr_datastore_instance_destroy(&instance);
+}
```
lavarou, line 1708: **suggestion (blocking, performance):** This allocates and destroys an `nr_datastore_instance_t` with two `nr_strdup("unknown")` calls on every `RedisCluster` operation — including the common case where `instance_reporting.enabled = 0`, which later code ([nr_segment_datastore.c:58](https://github.com/newrelic/newrelic-php-agent/blob/add-rediscluster/axiom/nr_segment_datastore.c#L58-L60)) discards via an early return. That's per-call heap churn in a hot path for a value that is always constant or discarded. Since `nr_segment_datastore_end` copies the fields out (via `nr_datastore_instance_set_host`) and does not take ownership, the instance can safely be a single long-lived/cached object rather than re-created each call. Alternatively, guard the allocation on NRPRG(txn)->options.instance_reporting_enabled so the common disabled-path does no allocation at all. (Note: passing NULL is not equivalent — it would suppress the unknown/unknown metric entirely, which is presumably not the intent.)

```suggestion
NR_INNER_WRAPPER(rediscluster_function) {
  nr_datastore_instance_t instance = {
      .host = "unknown",
      .port_path_or_id = "unknown",
      .database_name = NULL,
  };

  nr_php_instrument_datastore_operation_call(nr_wrapper, NR_DATASTORE_REDIS,
                                             nr_wrapper->extra, &instance,
                                             INTERNAL_FUNCTION_PARAM_PASSTHRU);
}
```

agent/php_internal_instrument.c
```diff
@@ -3811,6 +3968,230 @@ void nr_php_generate_internal_wrap_records(void) {
...
+                      rediscluster_function, 0, "zremrangebyscore")
+  NR_INTERNAL_WRAPREC("rediscluster::zrevrange", rediscluster_zrevrange, rediscluster_function, 0,
+                      "zrevrange")
+  NR_INTERNAL_WRAPREC("rediscluster::zrevrangebylex", rediscluster_zrevrangebylex,
+                      rediscluster_function, 0, "zrevrangebylex")
+  NR_INTERNAL_WRAPREC("rediscluster::zrevrangebyscore", rediscluster_zrevrangebyscore,
+                      rediscluster_function, 0, "zrevrangebyscore")
+  NR_INTERNAL_WRAPREC("rediscluster::zrevrank", rediscluster_zrevrank, rediscluster_function, 0,
+                      "zrevrank")
+  NR_INTERNAL_WRAPREC("rediscluster::zscore", rediscluster_zscore, rediscluster_function, 0,
```
lavarou: **nitpick:** Trailing space after `0,`

agent/php_internal_instrument.c
```diff
@@ -3811,6 +3968,230 @@ void nr_php_generate_internal_wrap_records(void) {
...
+  NR_INTERNAL_WRAPREC("rediscluster::ttl", rediscluster_ttl, rediscluster_function, 0, "ttl")
+  NR_INTERNAL_WRAPREC("rediscluster::type", rediscluster_type, rediscluster_function, 0, "type")
+  NR_INTERNAL_WRAPREC("rediscluster::unlink", rediscluster_unlink, rediscluster_function, 0,
+                      "unlink")
+  NR_INTERNAL_WRAPREC("rediscluster::xack", rediscluster_xack, rediscluster_function, 0, "xack")
+  NR_INTERNAL_WRAPREC("rediscluster::xadd", rediscluster_xadd, rediscluster_function, 0, "xadd")
+  NR_INTERNAL_WRAPREC("rediscluster::xclaim", rediscluster_xclaim, rediscluster_function, 0,
+                      "xclaim")
+  NR_INTERNAL_WRAPREC("rediscluster::xdel", rediscluster_xdel, rediscluster_function, 0, "xdel")
+  NR_INTERNAL_WRAPREC("rediscluster::xgroup", rediscluster_xgroup, rediscluster_function, 0,
```
lavarou: **nitpick:** Trailing space after `0,`

agent/php_internal_instrument.c
```diff
@@ -3811,6 +3968,230 @@ void nr_php_generate_internal_wrap_records(void) {
...
+                      rediscluster_function, 0, "zrevrangebylex")
+  NR_INTERNAL_WRAPREC("rediscluster::zrevrangebyscore", rediscluster_zrevrangebyscore,
+                      rediscluster_function, 0, "zrevrangebyscore")
+  NR_INTERNAL_WRAPREC("rediscluster::zrevrank", rediscluster_zrevrank, rediscluster_function, 0,
+                      "zrevrank")
+  NR_INTERNAL_WRAPREC("rediscluster::zscore", rediscluster_zscore, rediscluster_function, 0, 
+                      "zscore")
+  NR_INTERNAL_WRAPREC("rediscluster::zunionstore", rediscluster_zunionstore, rediscluster_function,
+                      0, "zunionstore")
+
```
lavarou: **nitpick:** clang-format was not run on the new `NR_INTERNAL_WRAPREC` block. The repo's `.clang-format` sets `ColumnLimit: 80`, and all 138 new `rediscluster::` wraprec lines run 92–99 columns, whereas every existing `redis::` wraprec wraps to fit 80. Running clang-format will reflow them and likely catch the two items above for free.

tests/integration/rediscluster/skipif.inc
```diff
@@ -0,0 +1,55 @@
+<?php
+/*
+ * Copyright 2024 New Relic Corporation. All rights reserved.
```
lavarou: **nitpick:** It says "Copyright 2024" here, but it's a new file in this PR and the other 13 new test files all say "Copyright 2026".

lavarou (changes requested):

agent/php_internal_instrument.c
hahuja2: Fixed: https://github.com/newrelic/newrelic-php-agent/pull/1212/commits/74f841bc31858f17fdc08efbc737b846e81992d8

agent/php_internal_instrument.c
hahuja2: Fixed: https://github.com/newrelic/newrelic-php-agent/commit/74f841bc31858f17fdc08efbc737b846e81992d8

agent/php_internal_instrument.c
hahuja2: Fixed: https://github.com/newrelic/newrelic-php-agent/commit/74f841bc31858f17fdc08efbc737b846e81992d8

tests/integration/rediscluster/skipif.inc
hahuja2: Fixed: https://github.com/newrelic/newrelic-php-agent/pull/1212/commits/bad29321e285d7fa25bf71579171abae1b224b27

lavarou (approved):

PR #1215, opened 2026-05-01
feat(agent): handle segment starts with fibers

zsistla:
Analyze all segment starts to work with context information.
Will behave as usual for non-fibers.
Will use context information for starting fibers when in a fiber environment.

1) Updated segment start logic across the agent to be context aware
2) Added fiber integration tests to exercise the code paths
3) Incorporated bug fixes that surfaced during increased test coverage runs.

Of the tests:
* The _basic tests all follow the same pattern.
* For the pdo/*/base-class tests, once you've seen one of the variants, the others are basically the same but for different databases.
* the span_events/fibers/*.php  tests going through various scenarios (starting/stopping/resume/noresume/errors/etc/nesting) etc.
* guzzle7 and curl_multi_exec are actually async so there's quite a few more tests in that area.


agent/lib_guzzle_common.c
```diff
@@ -83,16 +83,40 @@ nr_segment_t* nr_guzzle_obj_add(const zval* obj,
...
   nr_segment_t* segment = NULL;
   char* async_context = NULL;
+  int old_context = 0;
 
   /*
    * Create the async context, in case there was parallelism.
    */
   async_context
       = nr_guzzle_create_async_context_name(async_context_prefix, obj);
+  nrl_info(NRL_INSTRUMENT, "amber Guzzle: created async context %s",
```
bduranleau-nr: This log line should probably be removed or modified

agent/lib_guzzle6.c
```diff
@@ -112,12 +112,22 @@ static void nr_guzzle6_requesthandler_handle_response(zval* handler,
...
   }
 
+  /*
+   * This is a special case because guzzle stores all the external segments in
+   * the NRTXN(guzzle_objs) hashmap. When the extracted segment is ended, we
+   * don't want the context of this segment's parent to be set on the txn
+   * because for guzzle that is a special newrelic guzzle middleware segment.
+   * Save it now so we can restore it later.
+   */
+  old_context = NRTXN(current_async_context);
```
bduranleau-nr, line 129: What happens if any of the return cases between here and the restore at the end of the function get triggered?

agent/lib_mongodb.c
```diff
@@ -164,7 +164,7 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
   nr_mongodb_get_host_and_port_path_or_id(server, &instance.host,
                                           &instance.port_path_or_id TSRMLS_CC);
 
-  segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+  segment = NR_SEGMENT_START_WITH_PARENT_CONTEXT(NRPRG(txn), auto_segment);
```
bduranleau-nr: Should we be modifying this call since it's inside a PHP < 8.0 check?

agent/php_api_datastore.c
```diff
@@ -142,7 +142,8 @@ PHP_FUNCTION(newrelic_record_datastore_segment) {
   }
 
   if (instrument) {
-    segment = nr_segment_start(NRPRG(txn), NULL, NULL);
+    segment = NR_SEGMENT_START_WITH_PARENT_CONTEXT(
+        NRPRG(txn), nr_txn_get_current_segment_txn_context(NRPRG(txn)));
```
bduranleau-nr: Couldn't we use the other macro, `NR_SEGMENT_START_WITH_TXN_CONTEXT` here instead?

agent/php_internal_instrument.c
```diff
@@ -2420,7 +2419,9 @@ NR_INNER_WRAPPER(curl_exec) {
     return;
   }
 
-  nr_php_curl_exec_pre(curlres, NULL, NULL TSRMLS_CC);
+  nr_php_curl_exec_pre(curlres,
+                       nr_txn_get_current_segment_txn_context(NRPRG(txn)),
+                       nr_txn_get_current_context(NRPRG(txn)) TSRMLS_CC);
```
bduranleau-nr, line 2450: This change might benefit from a brief comment describing why it went from NULL, NULL to txn context calls

axiom/nr_segment_private.c
```diff
@@ -89,6 +89,7 @@ void nr_segment_destroy_fields(nr_segment_t* segment) {
   nr_segment_destroy_typed_attributes(segment->type,
                                       &segment->typed_attributes);
   nr_segment_error_destroy_fields(segment->error);
+  nr_segment_children_deinit(&segment->children);
```
bduranleau-nr, line 92: Do we need an additional call to `deinit` here? This parent function `nr_segment_destroy_fields` is only called in 2 places, once in `nr_segment_destroy_children_post_callback` and `nr_segment_discard`, both of which explicitly already call `nr_segment_children_deinit` with `&segment->children`

axiom/nr_segment_traces.c
```diff
@@ -550,6 +550,9 @@ void nr_segment_traces_create_data(
                 "generated for this transaction");
     nr_string_pool_destroy(&segment_names);
     nr_buffer_destroy(&buf);
+    if (span_events) {
+          nr_vector_destroy(&span_events);
+    }
```
bduranleau-nr, line 555: Is this just a pre-existing bugfix?

agent/lib_guzzle6.c
zsistla: Nothing. 
old_context is a local variable so it doesn't change NRTXN(current_async_context);
and the only thing that would change it is ending the segment which is why the restore happens right afterwards.
There's no other need to restore it in the function, so even if a return happens the agent is good.

axiom/nr_segment_traces.c
zsistla: Yes, it wasn't cleaning up before a return and one of the new tests exposed the mem leak.

agent/lib_guzzle_common.c
zsistla: Thankyou!!
[78e0d79](https://github.com/newrelic/newrelic-php-agent/pull/1215/commits/78e0d79783d763a3d1e0eb1ab149e335a015ad18)

agent/lib_mongodb.c
zsistla: Please see:
[e6117fd](https://github.com/newrelic/newrelic-php-agent/pull/1215/commits/e6117fd65cf04cf9f629b729d8ebcb0218846ab5)

agent/php_api_datastore.c
zsistla: Good catch
[da38fcc](https://github.com/newrelic/newrelic-php-agent/pull/1215/commits/da38fcc432ed9586fe4386967d18eb7ae0dea3d4)

agent/php_internal_instrument.c
zsistla: Please see: [22f0b8f](https://github.com/newrelic/newrelic-php-agent/pull/1215/commits/22f0b8fd36c7c99506548e4616d8f0c80eb25dfe)

axiom/nr_segment_private.c
zsistla: This was triggering valgrind memleak notifications. The alternative would be to add a suppression for the valgrind notice, but i'd rather have the call here. Most likely case, it's calling on NULL children and quickly resolved, but better to ensure we cover it then not.

bduranleau-nr (approved):

PR #1223, opened 2026-05-11
fix(axiom): mitigate possible snprintf overflow

bduranleau-nr:
Mitigates a possible `snprintf` overflow by checking the return value of the call to `snprintf` and exit with an `NR_FAILURE` status if that value is negative or greater than the length of the buffer provided. The function signature of `nr_rule_replace_string` has been changed to return `nr_status_t` from `void` to signal to callers failures from the `snprintf` calls.

### DIFF
--- a/axiom/nr_rules.c
+++ b/axiom/nr_rules.c
@@ -168,7 +168,7 @@ nr_status_t nr_rule_replace_string(const char* repl,
         if (num > count) {
           *dest = 0;
           len = snprintf(dest, dest_len, "\\%d", num);
-          if (len < 0 || len >= (int)dest_len) {
+          if (len < 0 || (size_t)len >= dest_len) {
             return NR_FAILURE;
           }
           dest += len;
@@ -178,7 +178,7 @@ nr_status_t nr_rule_replace_string(const char* repl,
 
           if (sub) {
             len = snprintf(dest, dest_len, "%s", sub);
-            if (len < 0 || len >= (int)dest_len) {
+            if (len < 0 || (size_t)len >= dest_len) {
               nr_free(sub);
               return NR_FAILURE;
             }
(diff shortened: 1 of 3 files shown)


axiom/nr_rules.c
```diff
@@ -126,22 +126,22 @@ void nr_rules_sort(nrrules_t* rules) {
         qsort_comparator_for_rules);
 }
 
-void nr_rule_replace_string(const char* repl,
-                            char* dest,
-                            size_t dest_len,
-                            const nr_regex_substrings_t* ss) {
+nr_status_t nr_rule_replace_string(const char* repl,
```
ZNeumann, line 129: Can we add a test or two to axiom/tests/test_rules.c?

axiom/nr_rules.c
```diff
@@ -261,7 +269,10 @@ static nr_rules_result_t nr_rule_apply(char* str,
 
       /* Copy the match replacement. */
       if (rule->rflags & NR_RULE_HAS_CAPTURES) {
-        nr_rule_replace_string(rule->replacement, repl, repl_len, ss);
+        if (NR_FAILURE
+            == nr_rule_replace_string(rule->replacement, repl, repl_len, ss)) {
+          return NR_RULES_RESULT_UNCHANGED;
```
hahuja2, line 275: Before returning `NR_RULES_RESULT_UNCHANGED`, `nr_regex_substrings_destroy(&ss);` needs to be called to prevent a memory leak. This applies for all places that return `NR_RULES_RESULT_UNCHANGED` in this PR.

bduranleau-nr: 636e963d95c4e1c4954e0b89789b51daf3e98a02

axiom/nr_rules.c
bduranleau-nr: 6aec6a9da4fba6b1df7060cf5eaa354c3be4ea5e

ZNeumann (approved):

hahuja2 (approved):

PR #1225, opened 2026-05-21
chore(ci): set explicit permissions for GHA workflows

bduranleau-nr:


.github/workflows/security-scan.yml
lavarou, line 15: **note:** the SARIF upload is already broken on main, independent of this PR

**suggestion (non-blocking):** use this PR to fix the long-standing upload failure

`permissions: {}` doesn't make it worse, but it does lock it in — once the scope is explicitly `{}`, a future loosening of org defaults could no longer rescue this workflow. Since you're already touching the file to set explicit permissions, granting the one scope the job actually needs would fix the broken weekly upload in the same change:

```suggestion
    runs-on: ubuntu-latest
    permissions:
      contents: read          # checkout (harmless to be explicit)
      security-events: write  # upload-sarif → code scanning API
```

lavarou (dismissed):

lavarou (approved):

PR #1228, opened 2026-06-01
Bump phpunit/phpunit to ^8.5.52 (CVE-2026-24765)

mfulb:
## Summary

Resolves Dependabot alert #7: CVE-2026-24765 (GHSA-vvj3-c3rp-c85p) — unsafe deserialization in PHPUnit's PHPT code coverage handling.

Changes:
- Bumps `phpunit/phpunit` from `^4.8` to `^8.5.52` in `tests/include/newrelic-integration/composer.json`
- Removes `sami/sami` (abandoned; its transitive dep `twig/twig ~2.0` carries multiple CVEs that block `composer install` on modern Composer) and associated `sami.php` config
- Updates all 6 test files to use `\PHPUnit\Framework\TestCase` (the `PHPUnit_Framework_TestCase` alias was removed in PHPUnit 6)
- Converts `@expectedException` annotations to `expectException()` calls (annotation removed in PHPUnit 9)
- Updates README to reflect PHP 7.2+ requirement and PHPUnit 8.5; removes the Sami documentation workflow

PHPUnit 8.x is the right target because it requires PHP 7.2+, matching the agent's minimum supported PHP version.

Note: this PHPUnit suite is not run in CI. It is only used when a developer runs `composer install && ./vendor/bin/phpunit` inside `tests/include/newrelic-integration/` to unit test the helper library locally. The integration tests themselves load `src/` files directly and are unaffected by this change.

## Test results

Verified locally with `php:8.3-cli` + clean `composer install` (no special flags):

```
PHPUnit 8.5.52

Tests: 54, Assertions: 97, Skipped: 2.
```

All 52 executed tests pass, zero warnings. 2 skips require the newrelic extension.

## Test plan

- [x] `cd tests/include/newrelic-integration && composer install && ./vendor/bin/phpunit` passes
- [x] Integration tests pass


tests/include/newrelic-integration/composer.json
```diff
@@ -6,9 +6,8 @@
         "php": ">=5.3.2"
```
lavarou: **question (blocking)**: Shouldn't this be bumped to ">=7.2" (minimum PHP version that will run `phpunit/phpunit` version ^8.5.52)?

tests/include/newrelic-integration/README.md
```diff
@@ -54,12 +54,6 @@ if ($metrics['Datastore/operation/SQLite/insert']->count < 4) {
 
 ## Development
 
-This package needs to keep support for PHP 5.3, as
-the PHP agent currently supports PHP 5.3. This is a bit limiting (for 
-example, the filtering in the `NewRelic\Integration\Trace` class would be
-*much* easier with anonymous classes), but it is necessary to maintain the
-backward compatability.
-
```
lavarou, line 62: **suggestion (non-blocking)**: Isn't part of this comment still relevant? Yes, this package no longer needs to keep support for PHP 5.3 (however that's what is still defined in `composer.json` see [here](https://github.com/newrelic/newrelic-php-agent/pull/1228/changes#diff-2a6c8b5fd9667e7d155230827e3f858f4b588fa307c94b7a6e565081c998a8bdL6)) - it needs to be compatible with PHP 7.2. Also the part about filtering in the `NewRelic\Integration\Trace` class is definitely accurate as this PR doesn't modernize that code (which is perfectly fine not to do it in this PR). Consider updating the PHP version in the comment.

mfulb: Addressed in 21c6d79.

tests/include/newrelic-integration/composer.json
mfulb: Fixed in 5a2607c.

lavarou (approved):
Thanks for updating this!

PR #1236, opened 2026-06-17
fix(agent): remove double allocation of request_uri string

bduranleau-nr:


agent/php_error.c
```diff
@@ -76,7 +76,10 @@ static void nr_php_error_call_error_group_callback(nrtxn_t* txn,
   agent_attributes = nr_attributes_agent_to_obj(txn->attributes,
                                                 NR_ATTRIBUTE_DESTINATION_ALL);
 
-  request_uri = nr_strdup(nr_php_get_server_global("REQUEST_URI" TSRMLS_CC));
+  request_uri = nr_php_get_server_global("REQUEST_URI" TSRMLS_CC);
+  if (NULL == request_uri) {
+    request_uri = nr_strdup("");
+  }
```
lavarou: **suggestion (non-blocking):** `nr_strdup("")` isn't required here — the empty string could be handled at the use site instead.

`add_assoc_string` duplicates the string (it doesn't take ownership), and `nr_free(NULL)` is a safe no-op, so this also works and avoids a `malloc`/`free` pair when `REQUEST_URI` is unset (the common CLI case):

```
request_uri = nr_php_get_server_global("REQUEST_URI" TSRMLS_CC);
...
nr_php_add_assoc_string(txn_arr, "request_uri", request_uri ? request_uri : "");
```
Non-blocking, since your version is the more conventional pattern in this file — it keeps `request_uri`/`path`/`method` as a uniform "owned, non-NULL" trio that FREE_MEM treats identically. Just flagging the lighter-weight alternative; fine to keep as-is.

lavarou (dismissed):
**praise:** Excellent test coverage for `nr_php_error_call_error_group_callback` — this helper was previously untested.

The 8 cases map cleanly onto the callback's contract: not-set, valid string, empty string, null, non-string return, the 255-char truncation boundary, the NULL-REQUEST_URI regression this PR fixes, and `$err` field propagation. Testing indirectly through `nr_php_error_record_exception` exercises the real call path rather than the static helper in isolation, and the memory hygiene is careful throughout — the callable_holder lifetime comment and the paired install/uninstall helpers make the intent clear. Nice work backfilling this.

agent/php_error.c
bduranleau-nr: I like this approach- I've implemented that suggestion here 5fa9ca1fba43ed17778d806203a7c59864a2fde3

lavarou (approved):

PR #1237, opened 2026-06-18
feat: make installer compatible with PHP from henderkes

lavarou:
The extension directory reported by PHP from the [henderkes](https://pkgs.henderkes.com/) distribution may not exist: it is created lazily, only when an extension package that needs it is installed. Have the installer create the reported extension dir (mirroring the ini-dir handling) so the agent's installation can continue instead of skipping this PHP.

### DIFF
--- a/agent/newrelic-install.sh
+++ b/agent/newrelic-install.sh
@@ -1236,6 +1236,20 @@ for this copy of PHP. We apologize for the inconvenience.
     return 1
   fi
 
+  #
+  # PHP reports the extension directory via ini_get("extension_dir") even when
+  # that directory does not exist on disk yet. Some PHP packages (e.g. the
+  # henderkes static-php / FrankenPHP php-zts-cli / php-zts-embed packages)
+  # only create the module directory when an extension package is installed,
+  # so a base install reports a directory that is not present. Since this is
+  # exactly where PHP will look for the agent's newrelic.so, create it rather
+  # than skipping this PHP. This mirrors how the ini scan directory is handled
+  # above.
+  #
+  if [ ! -d "${pi_extdir}" ]; then
+    logcmd mkdir -p -m 0755 "${pi_extdir}"
+  fi
+
   if [ ! -d "${pi_extdir}" ]; then
     error "computed PHP extension directory:
     ${pi_extdir}


agent/newrelic-install.sh
```diff
@@ -1236,6 +1236,20 @@ for this copy of PHP. We apologize for the inconvenience.
...
+  # PHP reports the extension directory via ini_get("extension_dir") even when
+  # that directory does not exist on disk yet. Some PHP packages (e.g. the
+  # henderkes static-php / FrankenPHP php-zts-cli / php-zts-embed packages)
+  # only create the module directory when an extension package is installed,
+  # so a base install reports a directory that is not present. Since this is
+  # exactly where PHP will look for the agent's newrelic.so, create it rather
+  # than skipping this PHP. This mirrors how the ini scan directory is handled
+  # above.
+  #
+  if [ ! -d "${pi_extdir}" ]; then
```
mfulb, line 1249: suggestion (non-blocking): consider guarding the `mkdir` with `! -e` rather than `! -d`. If a non-directory file sits at this path, `! -d` still fires a `mkdir` that can only fail. Using `! -e` skips that and matches the existing ini-scan-dir handling at line 1130.

lavarou: I'd lean toward keeping `! -d` here. You're right that `mkdir` can only fail when a non-directory file occupies the path — but that failure isn't silent: `logcmd` records `Executing: mkdir …`, the command's `File exists/Not a directory` output, and `Command returned: 1` into the nrinstall log, which is part of what we collect during troubleshooting. Switching to `! -e` would skip the `mkdir` and leave no trace of the conflict.

The downstream `! -d` check reports "does not exist" in _both_ variants (the code from this PR and your suggestion), so `! -e` wouldn't improve the user-facing message either — the only way to fix that is extra branching, which felt like too much for an already-dense script when the log already carries the signal. Happy to reconsider if you think that log line is too obscure to rely on.

mfulb: It isnt a big deal - I would consider a `-e && ! -d` message to capture the existence of a non-dir path colliding but it is probably a low likelihood in general so `! -d` is probably fine.

mfulb (approved):

PR #1263, opened 2026-07-23
feat(agent): record PHP SAPI name as a supportability metric

lavarou:
## Summary

Adds a new supportability metric, `Supportability/PHP/SAPI/{sapi_name}`, that records the PHP SAPI the agent is running under (e.g. `cli`, `fpm-fcgi`, `cgi-fcgi`, `apache2handler`, `litespeed`). This gives us aggregate visibility into which SAPIs our install base actually uses, so we can prioritize testing and support investment.

There was previously no metric capturing this — `sapi_module.name` is read in a few places for behavioral decisions (CLI detection, background-web-server heuristics, the startup log line) but never surfaced as a metric.

## What changed

- **New metric function** (`agent/php_txn.c`, `agent/php_txn_private.h`): `nr_php_txn_create_sapi_metric(nrtxn_t* txn, const char* sapi_name)` force-adds `Supportability/PHP/SAPI/{sapi_name}` to the transaction's unscoped metrics. It mirrors the existing `nr_php_txn_create_php_version_metric` / `nr_php_txn_create_agent_version_metric` pattern exactly — same guard order (NULL txn, then empty value), same `nrm_force_add(..., 0)`, and it takes the SAPI name as a parameter so the guards are unit-testable.
- **Wired into `nr_php_txn_end`** (`agent/php_txn.c`): called once per transaction, passing `sapi_module.name`, alongside the existing PHP/agent version metrics.
- **Integration-runner scrub list** (`daemon/internal/newrelic/integration/test.go`): added `^Supportability/PHP/SAPI/.*` to `MetricScrubRegexps`. This metric now fires on every transaction, so without scrubbing it (exactly as `Supportability/PHP/Version` and `.../AgentVersion` already are) every existing `EXPECT_METRICS` test with a closed metric list would break. No individual integration test files needed touching.
- **Integration tests** (`tests/integration/supportability/`): two new fixtures using `EXPECT_METRICS_EXIST` (which bypasses the scrub list), verifying the metric under both SAPIs the runner can produce:
  - `test_sapi_metric_cli.php` — no `ENVIRONMENT` block → runs via `php` → asserts `Supportability/PHP/SAPI/cli`
  - `test_sapi_metric_web.php` — `REQUEST_METHOD=GET` forces `Test.IsWeb()` → runs via `php-cgi` → asserts `Supportability/PHP/SAPI/cgi-fcgi`

## Design notes

- **Cardinality:** recorded once per transaction (not once per process), matching the existing version metrics. The count reflects transaction volume per SAPI; distinct app/agent counts are derivable via NRDB faceting. This is an intentional trade-off documented in the spec.
- **No high-security gating and no sanitization** — SAPI name is not sensitive (comparable to PHP version), and PHP's built-in SAPI names are all metric-name-safe as-is.

## Testing

- Agent unit tests (`test_txn`) pass with new negative + happy-path coverage for `nr_php_txn_create_sapi_metric` (NULL txn, NULL name, NULL-txn+valid-name, empty name, and exact metric-name match on the happy path).
- Agent compiles cleanly; daemon Go tests pass.
- Integration fixtures lint clean; CLI SAPI name confirmed as `cli` locally.
- ⚠️ The full `EXPECT_METRICS_EXIST` assertions require a live collector (license key) and run for the first time in CI — the `cgi-fcgi` web-fixture assertion in particular is worth a look on its first CI result.

## Notes for reviewers

- Branch is based on `dev`.
- The `Supportability/PHP/SAPI/` scrub entry is required and intentional — it is not hiding a bug, it's the same mechanism used for the PHP/agent version metrics so per-transaction supportability metrics don't invalidate closed `EXPECT_METRICS` lists.

### DIFF
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -791,6 +791,22 @@ void nr_php_txn_create_php_version_metric(nrtxn_t* txn, const char* version) {
   nr_free(metric_name);
 }
 
+void nr_php_txn_create_sapi_metric(nrtxn_t* txn, const char* sapi_name) {
+  char* metric_name = NULL;
+
+  if (NULL == txn) {
+    return;
+  }
+
+  if (nr_strempty(sapi_name)) {
+    return;
+  }
+
+  metric_name = nr_formatf("Supportability/PHP/SAPI/%s", sapi_name);
+  nrm_force_add(NRTXN(unscoped_metrics), metric_name, 0);
+  nr_free(metric_name);
+}
+
 void nr_php_txn_create_agent_php_version_metrics(nrtxn_t* txn) {
   char* version = NULL;
 
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -1332,6 +1332,9 @@ nr_status_t nr_php_txn_end(int ignoretxn, int in_post_deactivate TSRMLS_DC) {
     /* Agent and PHP version metrics*/
     nr_php_txn_create_agent_php_version_metrics(txn);
 
+    /* SAPI name metric */
+    nr_php_txn_create_sapi_metric(txn, sapi_module.name);
+
     /* PHP packages major version metrics */
     nr_php_txn_create_packages_major_metrics(txn);
 
--- a/agent/php_txn.c
+++ b/agent/php_txn.c
@@ -799,7 +799,7 @@ void nr_php_txn_create_sapi_metric(nrtxn_t* txn, const char* sapi_name) {
   }
 
   if (nr_strempty(sapi_name)) {
-    return;
+    sapi_name = "unknown";
   }
 
   metric_name = nr_formatf("Supportability/PHP/SAPI/%s", sapi_name);
--- a/tests/integration/supportability/test_sapi_metric_cli.php
+++ b/tests/integration/supportability/test_sapi_metric_cli.php
@@ -1,0 +1,20 @@
+<?php
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*DESCRIPTION
+The agent should record a Supportability metric for the PHP SAPI name when
+running under the CLI SAPI.
+*/
+
+/*EXPECT_METRICS_EXIST
+Supportability/PHP/SAPI/cli, 1
+*/
+
+/*EXPECT_TRACED_ERRORS
+null
+*/
+
+echo "SAPI: " . php_sapi_name() . "\n";
--- a/agent/php_txn_private.h
+++ b/agent/php_txn_private.h
@@ -67,6 +67,15 @@ extern void nr_php_txn_create_php_version_metric(nrtxn_t* txn,
                                                  const char* version);
 
 /*
+ * Purpose : Create and record metric for the current PHP SAPI name.
+ *
+ * Params  : 1. The current transaction.
+ *           2. The PHP SAPI name.
+ */
+extern void nr_php_txn_create_sapi_metric(nrtxn_t* txn,
+                                          const char* sapi_name);
+
+/*
  * Purpose : Callback for nr_php_packages_iterate to create major
  *           version metrics.
  *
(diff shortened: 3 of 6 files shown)


tests/integration/supportability/test_sapi_metric_cli.php
```diff
@@ -0,0 +1,20 @@
...
+
+/*EXPECT_METRICS_EXIST
+Supportability/PHP/SAPI/cli, 1
+*/
+
+/*EXPECT_TRACED_ERRORS
+null
+*/
+
+echo "SAPI: " . php_sapi_name() . "\n";
```
bduranleau-nr, line 20: Could we add an `EXPECT` to both of these integration tests with the expectation being the output of the echo statement? Or do you consider that redundant?

agent/php_txn.c
```diff
@@ -791,6 +791,22 @@ void nr_php_txn_create_php_version_metric(nrtxn_t* txn, const char* version) {
...
 }
 
+void nr_php_txn_create_sapi_metric(nrtxn_t* txn, const char* sapi_name) {
+  char* metric_name = NULL;
+
+  if (NULL == txn) {
+    return;
+  }
+
+  if (nr_strempty(sapi_name)) {
```
bduranleau-nr, line 801: Do we know of any potential use-case where the `sapi_name` might be `NULL` or empty? Would it be worth reporting that as `Supportability/PHP/SAPI/unknown`?

tests/integration/supportability/test_sapi_metric_cli.php
lavarou: This `echo` statement is more of a debug message than anything to be verified
by test's expectations.

The integration test's intent is to verify the agent records the correct
*metric name* — not that PHP's own `php_sapi_name()` API works, which is what
`EXPECT` would check. The metric name is already the thing under test, and it's
derived from the exact same `sapi_module.name` value `php_sapi_name()` reads,
so the two are never independent facts to cross-check.

I did look at using `NewRelic\Integration\Metric` (via `Transaction::
getUnscopedMetrics()`)
to assert the metric's name against the script's own `php_sapi_name()` call at
runtime. However, it's not implementable here because:

- `nr_php_txn_create_sapi_metric()` is only reached from `nr_php_txn_end()`,
  which for a normal request runs during RSHUTDOWN (`php_rshutdown.c:102`) —
  after the script's own code has already finished executing. Anything the
  script does with `newrelic_get_metric_table()` runs *before* this metric
  exists.
- I also checked whether either end-transaction API opens a window: `newrelic_
  end_of_transaction()` never reaches `nr_php_txn_end()` at all (it calls the
  lower-level `nr_txn_end()`), so the metric is never created. `newrelic_
  end_transaction()` does call `nr_php_txn_end()`, but that same call
  immediately destroys the transaction (`nr_txn_destroy()`) before returning
  to the script, so `nr_php_recording()` goes false and any subsequent
  `newrelic_get_metric_table()` call returns `FALSE` rather than the table.
  There's no PHP-visible moment where the metric exists and is still
  readable.

Given that, adding an `EXPECT` on the echoed `php_sapi_name()` output —
whether as plain text or via the Metric/Transaction introspection API —
would just be re-verifying PHP's own SAPI detection, not anything the agent
does. It doesn't add coverage beyond what the two SAPI-specific fixtures
(hardcoded per real SAPI) plus `test_txn.c`'s exact-match unit test already
provide. Going to leave this as-is.

agent/php_txn.c
lavarou: `sapi_module.name` is set as a compile-time string literal by every SAPI
implementation (`cli`, `fpm-fcgi`, `apache2handler`, etc.) as part of PHP's
required `sapi_module_struct` contract — it's the same field `php_sapi_name()`
reads, and by the time `nr_php_txn_end()` runs, SAPI startup has long
completed. So a NULL/empty value isn't a case I'd expect to hit in practice.

That said, it costs nothing to guard for it, and it matches the sibling
`nr_php_txn_create_agent_php_version_metrics()` right below this function,
which already falls back to `"unknown"` when `NR_PHP_PROCESS_GLOBALS
(php_version)`
is empty.

lavarou: Updated `nr_php_txn_create_sapi_metric()` to do the same — 
falls back to `Supportability/PHP/SAPI/unknown` instead of silently skipping,
with unit test coverage added in `test_txn.c` - see 8f01e93ddabd916a74ce48c5afbb9532f6c25c2b.

bduranleau-nr (approved):

PR #1266, opened 2026-07-30
feat(agent): Update fiber exclusive time when suspended

zsistla:
1. Ensures we don't count fiber suspend time in the segment's exclusive time duration. In addition to a fiber being able to suspend itself via Fiber::suspend(), since ONLY ONE fiber can run at a time, any fiber with status ZEND_FIBER_STATUS_RUNNING that switches to another fiber will automatically be suspended.

2.  When ZEND_FIBER_STATUS_SUSPENDED is detected in a fiber switch, update the exclusive current segment of a fiber context when a fiber is reanimated (e.g., via resume or caller ending or called fiber ending/suspending) after being suspended.

3. Added tests cases that confirm nested fibers give appropriate exclusive times across various scenarios.


agent/php_observer.c
```diff
@@ -121,8 +121,9 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
         = nr_php_user_instrument_wraprec_hashmap_get(func_name, scope_name);
     // store the wraprec in the op_array extension for the duration of the
     // request for later lookup
-    ZEND_OP_ARRAY_EXTENSION(
-        NR_OP_ARRAY, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)) = wr;
+    ZEND_OP_ARRAY_EXTENSION(NR_OP_ARRAY,
+                            NR_PHP_PROCESS_GLOBALS(op_array_extension_handle))
+        = wr;
```
bduranleau-nr: I don't see this behavior when I format this file. I think the old version where the assignment is on the same line as the global reference is preferable to this change.

agent/php_observer.c
```diff
@@ -246,24 +274,121 @@ static inline void nr_fiber_set_contexts(zend_fiber_context* zfc) {
...
+ * duration. In addition to a fiber being able to suspend itself via
+ * Fiber::suspend(), since ONLY ONE fiber can run at a time, any fiber that
+ * calls another will automatically be suspended.
+ *
+ * Note: We need to handle both types of suspension, but it is possible to
+ * differentiate. To detect that the fiber suspended itself using
+ * Fiber::suspend(), check:
+ *
+ * 1) the "from" fiber context status is ZEND_FIBER_STATUS_RUNNING
+ * 2) the fiber in the "from" fiber context, has a fiber->caller value of
```
bduranleau-nr, line 299: Is this an either/or check, or do we need to explicitly verify that both of these cases are true in order to determine that a fiber suspended itself?

bduranleau-nr: If both, it looks like this function needs an added check for case 2: `fiber->caller != NULL`

agent/php_observer.c
```diff
@@ -246,24 +274,121 @@ static inline void nr_fiber_set_contexts(zend_fiber_context* zfc) {
...
+ * ending/suspending) after being suspended
+ *
+ * Params:  zend_fiber_context* zfc
+ *
+ * Returns : Void
+ * Note: caller is responsible for verifying zfc is not NULL.
+ *
+ * During a fiber switch, update the exclusive time of a the fiber context
+ * being switched "to" if:
+ * 1) The "to" field is ZEND_FIBER_STATUS_SUSPENDED
```
bduranleau-nr: This is a little confusing, since the check for `ZEND_FIBER_STATUS_SUSPENDED` is happening at the level of the parent caller function `nr_fiber_switch_observe`, but the function description implies that this check should be performed as part of the function itself. It might read better if this was described as a prerequisite to calling this function.

agent/php_observer.c
```diff
@@ -246,24 +274,121 @@ static inline void nr_fiber_set_contexts(zend_fiber_context* zfc) {
...
+  if (NULL == NRPRG(txn)) {
+    /* nothing to do if the txn is NULL */
+    return;
+  }
+
+  current_time = nr_txn_now_rel(NRPRG(txn));
+
+  fiber_segment = nr_txn_get_current_segment(NRPRG(txn),
+                                             NRPRG_SHARED(current_php_context));
+  if (NULL != fiber_segment && 0 != fiber_segment->stop_time) {
```
bduranleau-nr: Do we ever expect the stop_time to not be 0 here? If the fiber has to go through the `nr_fiber_handle_fiber_suspend` function prior to this, and have a status of `SUSPENDED`, it should be forced to have a non-zero stop_time, right? Should we add error handling or error logging if this conditional is not true here?

agent/php_observer.h
```diff
@@ -18,6 +18,8 @@
 
 #include "Zend/zend_observer.h"
 
+#define NR_PHP_DEFAULT_SUSPEND_TIMES 10
```
bduranleau-nr, line 21: Is there a particular reason we want to limit the number of child segments to 10? Is this mostly a hedge against resource exhaustion?

tests/integration/span_events/fibers/test_fiber_nested_exclusive_basic.php
```diff
@@ -0,0 +1,98 @@
...
+$fiber1->resume();
+
+$txn = new Transaction;
+$metrics = $txn->getScopedMetrics();
+tap_assert(isset($metrics["Custom/two"]), 'metric for fiber two exists');
+tap_assert(isset($metrics["Custom/one"]), 'metric for fiber one exists');
+
+//echo "one exclusive: " . ($metrics["Custom/one"]->exclusive) . "\n";
+//echo "one total: " . ($metrics["Custom/one"]->total) . "\n";
+//echo "one diff: " . ($metrics["Custom/one"]->total - $metrics["Custom/one"]->exclusive) . "\n";
```
bduranleau-nr: Are these leftover from test debugging?

agent/php_observer.c
zsistla: We don't need to differentiate, we need to handle both types of suspension.

```
* Note: We need to handle both types of suspension, but it is possible to
 * differentiate. To detect that the fiber suspended itself using
 * Fiber::suspend(), check:
 *
 * 1) the "from" fiber context status is ZEND_FIBER_STATUS_RUNNING
 * 2) the fiber in the "from" fiber context, has a fiber->caller value of
 ```

File: tests/integration/span_events/fibers/test_fiber_nested_exclusive_basic.php [hunk 7]
```diff
@@ -0,0 +1,98 @@
...
+    Fiber::suspend();
+    Fiber::suspend();
+    time_nanosleep(0, 100000000);
+    echo "Ending Func 'two'\n";
+}
+
+function one()
+{
+    echo "Starting Func 'one'\n";
+    env_var_for_expects("GUID_ONE", newrelic_get_linking_metadata()['span.id'] ?? '');
```
bduranleau-nr: It looks like we aren't testing for `GUID_ONE` or `GUID_TWO` in these new test files. Suggest adding Span Event Expectations to test for GUID span parentage or removing the `env_var_for_expects` altogether

agent/php_observer.c
bduranleau-nr: It appears that this function is missing a conditional to check that the fiber->caller is not null in that case

agent/php_observer.c
```diff
@@ -246,24 +274,121 @@ static inline void nr_fiber_set_contexts(zend_fiber_context* zfc) {
...
+  nr_segment_t* fiber_segment = NULL;
+  zend_fiber* zfc_fiber = NULL;
+
+  if (NULL == NRPRG(txn)) {
+    /* nothing to do if the txn is NULL */
+    return;
+  }
+
+  if (ZEND_FIBER_STATUS_RUNNING == zfc->status) {
+    zfc_fiber = zend_fiber_from_context(zfc);
```
bduranleau-nr: `zfc_fiber` is unused in this function

agent/php_observer.c
```diff
@@ -246,24 +274,121 @@ static inline void nr_fiber_set_contexts(zend_fiber_context* zfc) {
...
+ * Note: caller is responsible for verifying zfc is not NULL.
+ *
+ * During a fiber switch, update the exclusive time of a the fiber context
+ * being switched "to" if:
+ * 1) The "to" field is ZEND_FIBER_STATUS_SUSPENDED
+ * 2) The segment stop time for the current on that fiber context is not
+ * zero
+ *
+ */
+static inline void nr_fiber_handle_exclusive_time(zend_fiber_context* zfc) {
```
bduranleau-nr: The `zfc` param is unused in this function. Suggest removing it and updating the function description.

agent/php_observer.c
zsistla: I also prefer the other way. I don't know why it changed the format.
I can change it back, but for my formatter, it keeps insisting.
[fb85727](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/fb85727b8b62aea079528014b5d82292b3165485)

agent/php_observer.h
zsistla: Had to pick some number and it seemed reasonable to me.  It doesn't limit the number of exclusive time slots to 10.  If the number of unused slots is less than 10, it will expand the exclusive time.  It's just a starter amount.

agent/php_observer.c
zsistla: updated commentary here: 
[fb85727](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/fb85727b8b62aea079528014b5d82292b3165485)

agent/php_observer.c
zsistla: error msg: [6830a73](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/6830a731b92d3afb4ff1c881bad5239028f34166)

tests/integration/span_events/fibers/test_fiber_nested_exclusive_basic.php
zsistla: removed, this test only focuses on exclusive times
[c19b259](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/c19b25989236103503aac105beda67185b61a3f1)

tests/integration/span_events/fibers/test_fiber_nested_exclusive_basic.php
zsistla: good catch
[c19b259](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/c19b25989236103503aac105beda67185b61a3f1)

agent/php_observer.c
zsistla: We don't need to check fiber->caller for null.  That only differentiates one type of suspension from another.  We need to handle all suspension cases whether fiber->caller is null or not.

agent/php_observer.c
zsistla: removed [8681a89](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/8681a89c36dabc6845d4aa32eb30c21f5a079939)

agent/php_observer.c
bduranleau-nr: I would suggest removing or reworking the comment about checking for the caller being NULL in that case- it implies that this is required in some way in this function

agent/php_observer.c
zsistla: removed [39ae5e6](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/39ae5e6ff6db5e8ee75e407f265bef10ba644f15)

agent/php_observer.c
zsistla: updated here [8681a89](https://github.com/newrelic/newrelic-php-agent/pull/1266/commits/8681a89c36dabc6845d4aa32eb30c21f5a079939)

bduranleau-nr (approved):

PR #1270, opened 2026-08-11
Fix Composer package detection under FrankenPHP's shared thread pool

mfulb:
## Summary

Composer's autoload-based package scan was gated by a single status
flag shared by every thread on an app. Under FrankenPHP's ZTS model,
where many threads run the same app in one long-lived process, only
the first thread to hit that autoload event actually ran the scan and
set the flag — every other thread then saw the flag already set and
skipped scanning entirely, so those threads never captured or reported
any Composer package data.

The fix moves that flag to a per-(app, thread) entry, so each thread
independently runs the scan once for itself instead of once per app.
Keying the entry by app as well as thread was also necessary for
multi-app deployments: FrankenPHP can run multiple apps on the same
shared thread pool, and a flag keyed by thread alone would have let one
app's scan incorrectly suppress scanning for a second app that later
runs on that same thread.

Testing under FrankenPHP's worker mode surfaced a follow-on problem:
a thread's one-time scan can happen during worker bootstrap, before a
transaction exists and before the app has finished registering with
the daemon — so there's nowhere to attach the scanned data at the time
it's collected. To handle that, the per-thread entry now also caches
the scan result independently of any transaction, and each transaction
pulls whatever is cached on its thread when it runs, rather than
requiring a transaction to be live at scan time.

Additional correctness fixes that came out of testing this cache:

- A transaction that ends up discarded (ignored) no longer causes the
  next transaction on that thread to inherit and resend its data.
- Removed a fallback that could re-report already-sent package data on
  every later transaction.
- Restored the `Composer/detected` and `Autoloader/detected` metrics
  for the case where the scan runs with no live transaction (e.g.
  during worker bootstrap) — they were previously lost in that window
  even though the package data itself was recovered by the cache.

### DIFF
--- a/agent/tests/test_lib_composer.c
+++ b/agent/tests/test_lib_composer.c
@@ -1,0 +1,603 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include "tlib_php.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "php_agent.h"
+#include "php_globals.h"
+#include "fw_hooks.h"
+#include "nr_agent.h"
+#include "nr_app.h"
+#include "nr_php_packages.h"
+#include "util_syscalls.h"
+
+tlib_parallel_info_t parallel_info = {.suggested_nthreads = 1, .state_size = 0};
+
+static char* vendor_path = NULL;
+
+/*
+ * Every scenario in this file uses the same default ("PHP Application")
+ * app, since none of them override newrelic.appname/license -- so the same
+ * (app, tid) composer_map entry persists across scenarios unless reset.
+ * Evicts it so each scenario starts from a genuinely fresh entry, matching
+ * how a real worker thread's entry only ever starts fresh once, at thread
+ * creation.
+ */
+static void reset_entry_between_scenarios(void) {
+  nr_app_info_t info;
+  nrapp_t* app;
+
+  nr_memset(&info, 0, sizeof(info));
+  nr_php_txn_populate_app_info_identity(&info, NULL, NULL TSRMLS_CC);
+
+  app = nr_app_find_locked(nr_agent_applist, &info);
+  if (NULL != app) {
+    nr_app_tid_maps_evict(app, (uint64_t)nr_gettid());
+    nrt_mutex_unlock(&app->app_lock);
+  }
+
+  nr_app_info_destroy_fields(&info);
+}
+
+/*
+ * Confirms request start produced a live txn with a real, genuinely
+ * fresh composer entry, asserting whichever precondition failed.
+ * Returns the entry, or NULL if any assertion failed.
+ */
+static nr_composer_thread_entry_t* require_live_entry(TSRMLS_D) {
+  nr_composer_thread_entry_t* entry;
+
+  tlib_pass_if_not_null("request start produced a live txn", NRPRG(txn));
+  if (NULL == NRPRG(txn)) {
+    return NULL;
+  }
+
+  tlib_pass_if_not_null("txn has a real composer entry from RINIT's app",
+                        NRPRG(txn)->composer_info.entry);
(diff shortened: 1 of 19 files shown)


axiom/nr_app.c
```diff
@@ -342,6 +361,46 @@ static int nr_app_info_valid(const nr_app_info_t* info) {
...
+}
+
+nrapp_t* nr_app_find_locked(nrapplist_t* applist, const nr_app_info_t* info) {
+  nrapp_t* app;
+
+  if (0 == nr_app_info_valid(info) || NULL == applist) {
+    return NULL;
+  }
+
+  nrt_mutex_lock(&applist->applist_lock);
```
lavarou: Isn't this mutex already locked on line 415 in `nr_app_find_or_add_app`?

axiom/nr_app.c
```diff
@@ -342,6 +379,55 @@ static int nr_app_info_valid(const nr_app_info_t* info) {
...
+  nrt_mutex_unlock(&applist->applist_lock);
+
+  if (NULL != app && info->high_security != app->info.high_security) {
+    /* Mirrors the check in nr_app_find_or_add_app(): a license+appname
+     * match with a different high_security setting means this is a
+     * different app, not this one. */
+    nr_app_log_high_security_mismatch_on_find(info->appname);
+    nrt_mutex_unlock(&app->app_lock);
+    app = NULL;
+  }
```
lavarou, line 426: **question:** How do we keep `nr_app_find_locked` and `nr_app_find_or_add_app` from
drifting apart on what counts as a "found" app?

Right now both functions run the same extra check after nr_app_find_matching_locked
succeeds — a found app is "unfound" (unlocked, returned as NULL) if
`info->high_security != app->info.high_security`:

  - `nr_app_find_locked`, nr_app.c:419
  - `nr_app_find_or_add_app`, nr_app.c:460

That's a second gate on top of `nr_app_match`, not part of it — `nr_app_match` only
compares license/appname/trace_observer_*, so this check exists specifically to
catch a case `nr_app_match`'s key doesn't cover. Today it's copy-pasted (each with
its own logger and its own rate-limit state, `nr_app_log_high_security_mismatch` vs.
`nr_app_log_high_security_mismatch_on_find`).

The thing I'm unsure how to avoid: if a similar defensive condition is ever added
later — some other field that shouldn't disagree between a match and the info
being searched for — what stops it from landing in one function's copy of the
check and not the other's? At that point `nr_app_find_locked` and
`nr_app_find_or_add_app` would quietly disagree on whether the same info matches
the same app, which seems like a hard bug to notice later.

Would it make sense to pull "is this found app actually a valid match" into one
function both callers go through (the way the search loop itself was just pulled
into `nr_app_find_matching_locked`), so there's exactly one place for that kind of
condition to be added? Or is there a reason the two are meant to be able to
diverge here that I'm missing?

agent/tests/test_lib_composer.c
```diff
@@ -0,0 +1,603 @@
...
+ * txn (the FrankenPHP worker-bootstrap case, where the app hasn't yet been
+ * acknowledged by the daemon) -- every scenario here runs through
+ * tlib_php_request_start()'s live NRPRG(txn). That's adequate today only
+ * because, once the (app, thread) entry is resolved, the function's write
+ * logic never reads NRPRG(txn) again -- the with-txn and no-txn resolution
+ * paths both hand it the same kind of entry pointer, and everything
+ * downstream operates on that pointer alone, not on the txn. If the write
+ * logic ever starts referencing NRPRG(txn) directly, this file's coverage
+ * would no longer stand in for the no-txn case, and a real no-txn scenario
+ * would need to be added.
```
lavarou: **question (non-blocking):** Is `nr_composer_find_app_no_txn` (the no-txn branch with `nr_composer_find_app_no_txn` + `nr_app_get_or_create_thread_composer_entry` calls in `nr_composer_handle_autoload`) tested in any way? No-txn state could be created via `nr_cmd_appinfo_hook` that sets `app->state= NR_APP_UNKNOWN`

lavarou (approved):

PR #1271, opened 2026-08-14
fix(agent): capture Monolog postprocessor context/extra and enriched message

zsistla:
What changed

  - Monolog's Logger::addRecord runs processors after the raw $message/$context/$extra arguments are passed in — processors can sanitize, enrich, or normalize context, add data to extra, and interpolate
  {placeholder} tokens into message. The agent was only ever reading the raw arguments, so it missed all of that post-processing.
  - On PHP 8+, the wrapper for addRecord now reads the compiled $record variable directly from the function's execute data at end of call (via new helper nr_get_func_cv_by_name in agent/php_agent.c/.h), after
  processors have already run. $record is a Monolog\LogRecord object on Monolog 3, or a plain array on Monolog 2 — both are handled.
  - nr_monolog_get_postprocessed_message and nr_monolog_get_postprocessed_attributes (new, in agent/lib_monolog.c) pull message/context/extra from that postprocessed $record. If extraction comes back empty
  (e.g. PHP 7, where the wrapper fires before processors run, or no $record variable found), the agent falls back to the original arg-based extraction — so behavior on PHP 7 is unchanged.
  - nr_monolog_convert_context_data_to_attributes now accepts an existing nr_attributes_t* to add to, so context and extra can be merged into one attribute set instead of only ever capturing context.
  - Array-valued context/extra entries are now flattened to a JSON string attribute instead of being silently dropped.

  Why it matters

  Without this, log context enrichment done by common Monolog processors (e.g. adding request IDs, correlating fields, PII scrubbing) never reached New Relic — the agent reported only what the calling code
  originally passed in, not what was actually logged/handled. This closes that gap for PHP 8+ while preserving existing  behavior via the fallback path.

  Testing

  - additions to agent/tests/test_monolog.c covering both Monolog 2 (array $record) and Monolog 3 (LogRecord object) shapes, processor-modified context/extra, missing/partial $record, and thearray-to-JSON conversion.
  - multiverse tests


agent/lib_monolog.c
```diff
@@ -199,6 +201,20 @@ nrobj_t* nr_monolog_context_data_zval_to_attribute_obj(
       }
       break;
 
+    case IS_ARRAY:
+      /* In this case, the best we can do is flatten it to a json string. */
+      json_zval = nr_php_json_encode((zval*)z TSRMLS_CC);
```
bduranleau-nr, line 206: **nit**: This cast from `const zval*` to `zval*` drops the const qualifier.

agent/lib_monolog.c
```diff
@@ -252,32 +268,35 @@ static zval* nr_monolog_extract_context_data(const size_t argc,
...
- * attributes
+ * attributes. If attributes is NULL, it will create a new one; otherwise, it
+ * will add the context attributes to the existing attributse.
  *
  * Params  : zval* for context array from Monolog
+ *           nr_attributes_t* for attributes to populate
  *
- * Returns : nr_attributes representation of $context on success
+ * Returns : modifies $context on success
  *           NULL otherwise
```
bduranleau-nr: This no longer appears to be true- this function now always returns the passed in `attributes` argument, and never explicitly returns `NULL`.
Stylistically, I'm not sure it's a good idea to simultaneously return `attributes`, which is being used to set local versions of this variable where it's being called, and modifying the reference to this variable. Instead, I think it should be one or the other- in the case of this PR, it looks like modifying the reference in this function is required by the 2 new callers.
I would suggest either having a `void` return type, or a form of basic success/failure return type.

agent/lib_monolog.c
```diff
@@ -252,32 +268,35 @@ static zval* nr_monolog_extract_context_data(const size_t argc,
...
   }
 
-  attributes = nr_attributes_create(NRPRG(txn)->attribute_config);
   if (NULL == attributes) {
-    return NULL;
+    attributes = nr_attributes_create(NRPRG(txn)->attribute_config);
+  }
+  if (NULL == attributes) {
+    return attributes;
   }
```
bduranleau-nr: This check is dead code- it is not possible for `attributes` to be `NULL` at this point after `nr_attributes_create` in the previous conditional- it unconditionally z-allocates memory of size `nr_attributes_t`.

agent/lib_monolog.c
zsistla: `z` comes in as a `const zval*` but nr_php_json_encode expects a `zval*`.  nr_php_json_encode() doesn't mutate its argument, so the cast is safe in practice.

agent/lib_monolog.c
```diff
@@ -340,6 +359,134 @@ static nrtime_t nr_monolog_get_timestamp(const int monolog_api,
...
+      record_message_string
+          = nr_php_get_zval_object_property(record, "message");
+    }
+  }
+
+  /* For monolog 2, $record is a simple array*/
+  if (record_message_string == NULL && nr_php_is_zval_valid_array(record)) {
+    record_message_string
+        = nr_php_zend_hash_find(Z_ARRVAL_P(record), "message");
+  }
```
bduranleau-nr, line 390: This seems better suited as an `if`/`else if`/`else` block, since `record` should not simultaneously be able to be an object and an array, making a second `if` statement always unnecessary when handling Monolog 3 instrumentation.

agent/lib_monolog.c
zsistla: Confirmed — nr_attributes_create() -> nr_zalloc() calls exit(3) on allocation failure rather than returning NULL, so this can't be hit. Removed in abb8b383.

agent/lib_monolog.c
```diff
@@ -340,6 +359,134 @@ static nrtime_t nr_monolog_get_timestamp(const int monolog_api,
...
+
+  if (NULL == record_message_string) {
+    return NULL;
+  }
+
+  if (!(nr_php_is_zval_valid_string(record_message_string))) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: expected $message be a valid string.",
```
bduranleau-nr, line 397: We should explicitly `nr_php_zval_free()` `record_message_string` in this case

agent/lib_monolog.c
```diff
@@ -340,6 +359,134 @@ static nrtime_t nr_monolog_get_timestamp(const int monolog_api,
...
+                     __func__);
+    return NULL;
+  }
+
+  message = nr_strdup(Z_STRVAL_P(record_message_string));
```
bduranleau-nr, line 402: `record_message_string` needs to be `nr_php_zval_free`'d after strdup here to prevent a leak

agent/lib_monolog.c
```diff
@@ -340,6 +356,134 @@ static nrtime_t nr_monolog_get_timestamp(const int monolog_api,
...
+      record_context_array = nr_php_get_zval_object_property(record, "context");
+      record_extra_array = nr_php_get_zval_object_property(record, "extra");
+    }
+  }
+
+  /* For monolog 2, $record is a simple array*/
+  if (record_context_array == NULL && nr_php_is_zval_valid_array(record)) {
+    record_context_array = nr_php_zend_hash_find(Z_ARRVAL_P(record), "context");
+    record_extra_array = nr_php_zend_hash_find(Z_ARRVAL_P(record), "extra");
+  }
```
bduranleau-nr, line 444: Same feedback as [here](https://github.com/newrelic/newrelic-php-agent/pull/1271/changes#r3867115580) in `nr_monolog_get_postprocessed_message`

agent/lib_monolog.c
```diff
@@ -340,6 +356,134 @@ static nrtime_t nr_monolog_get_timestamp(const int monolog_api,
...
+
+  if (!(nr_php_is_zval_valid_array(record_context_array))) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: expected $context be a valid array.",
+                     __func__);
+
```
bduranleau-nr, line 449: Do we want to `return` early if we hit this case, or is a non-valid `record_context_array` something that we expect to happen in certain circumstances (i.e. `"context"` is not guaranteed to be present in a `record`)?

agent/lib_monolog.c
```diff
@@ -340,6 +356,134 @@ static nrtime_t nr_monolog_get_timestamp(const int monolog_api,
...
+  } else {
+    attributes = nr_monolog_convert_context_data_to_attributes(
+        record_context_array, attributes);
+  }
+
+  if (!(nr_php_is_zval_valid_array(record_extra_array))) {
+    nrl_verbosedebug(NRL_INSTRUMENT, "%s: expected $extra be a valid array.",
+                     __func__);
```
bduranleau-nr, line 457: Similar question as [here](https://github.com/newrelic/newrelic-php-agent/pull/1271/changes#r3867172168)

agent/lib_monolog.c
zsistla: Nothing was technically being modified by reference before — the old `nr_attributes_t* attributes` param was passed by value, and every call site had to explicitly recapture the return (`attributes = fn(...)`) to see either the same pointer mutated in place or a freshly nr_attributes_create()'d one. That dual mode (sometimes same pointer, sometimes new) with no type-level signal was problematic.

Changed it to modify by reference for real: signature is now `nr_monolog_convert_context_data_to_attributes(zval* context_data, nr_attributes_t** attributes)`, returns void, callers pass `&attributes`. This matches the nr_attributes_t** idiom already used for nr_attributes_destroy() in this same header, so it's not a new pattern for this type. Doc comment updated to describe the actual contract (including that *attributes can still be NULL afterward if context_data was invalid and *attributes was NULL going in — so the doc's original "NULL otherwise" was directionally right, just no longer expressed as a literal return). Pushed in d8faef08.

agent/lib_monolog.c
zsistla: combined if/else [87e8370](https://github.com/newrelic/newrelic-php-agent/pull/1271/commits/87e83701389079abb72859d8eb8ad3ad2593152c)

agent/lib_monolog.c
zsistla: nr_php_zend_hash_find() (agent/php_hash.c:82) is a thin wrapper over zend_hash_str_find() — a standard Zend API that returns a pointer directly into the HashTable's own internal bucket storage. It's a borrowed reference owned by the array; nothing is allocated on the caller's behalf.

nr_php_get_zval_object_property() (agent/php_agent.c:47) does one of two things depending on type:
- For an object (the Monolog 3 LogRecord case), it calls zend_read_property() (php_agent.c:38), which — for an ordinary declared property, which message/context/extra are — returns a pointer directly into the object's own property table. Also borrowed.
- For an array (the Monolog 2 case), it just calls nr_php_zend_hash_find() itself (php_agent.c:59) — same borrowed-reference situation as above.

So in both branches feeding record_message_string, the pointer is memory owned by the live $record object/array, not something that was allocated for this call and does not need to be freed.

agent/lib_monolog.c
zsistla: See previous comment: https://github.com/newrelic/newrelic-php-agent/pull/1271/changes#r3867434040

agent/lib_monolog.c
zsistla: combined if/else [b37bcfc](https://github.com/newrelic/newrelic-php-agent/pull/1271/commits/b37bcfcf1955347f2da2198be151c33e7310fa4e)

agent/lib_monolog.c
zsistla: I don't think it needs an early return. It is not something that should happen, but if there is some issue with context, but extra might be valid, why not try and check for it at that point?

agent/lib_monolog.c
zsistla: I don't think it needs an early return. It is not something that should happen, but we still need to return attributes and there's not much earlier we could return since this would be the last statement, it would skip the else and return attributes.

bduranleau-nr: In that case, should the log level of the messages from these two cases be increased to something that indicates a warning or error if this is a case that we don't expect an application to hit?

zsistla: for 2.x having a NULL record is theoretically possible: https://github.com/Seldaek/monolog/blob/2.x/src/Monolog/Logger.php#L366
So it's possible and valid from monolog's standpoint to get here with no valid context or extra.

I dont see the benefit of lowering the loglevel since if there ever is any issue, we default back to current behavior. Also, the loglevel matches the loglevel of other messages where things are not as expected (https://github.com/Seldaek/monolog/blob/2.x/src/Monolog/Logger.php#L366).

bduranleau-nr (approved):

PR #1277, opened 2026-08-18
fix(daemon): handle Trace Observer graceful stream close

lavarou:
## Summary

The Trace Observer closes every Infinite Tracing gRPC stream with an `OK` status roughly every 30s as a load-balancing measure. The daemon mishandled this in four ways, all in `grpc_sender.go`'s `connect()`/`send()`:

- **Receive path swallowed the close.** A bare `io.EOF` from `Recv()` (how grpc-go signals a clean `OK` close) was treated as "nothing to report" and discarded silently.
- **Send path misclassified the resulting failure.** The dead stream was only discovered on the next `send()`, where `SendMsg` also returns bare `io.EOF` — misclassified as a generic error, triggering the spec's 15s error backoff for a routine, expected close, and dropping the batch in flight.
- **The `default:` branch rewrote genuine errors to `statusOk`** without returning, so a truly dead stream spun the goroutine and could leak stale statuses into the next stream generation.
- **The receive goroutine read the reassignable `s.stream` field instead of the stream it was spawned for**, letting a goroutine from a stale generation call `Recv()` concurrently with a newer generation's own goroutine — forbidden by grpc-go, and a plain data race.

Net effect: ~33% of wall-clock time not streaming, and one span batch destroyed on every ~45s cycle for as long as spans were being produced.

Fixes, one commit each:
1. Signal `statusImmediateRestart` on a graceful `io.EOF` instead of returning silently.
2. Defer to the receive goroutine's classification on a send-side `EOF`, closing the race where a queued batch could still hit the dead stream first.
3. Drain any `responseError` entry left over from a previous generation before establishing a new stream — closes a narrow leak in fix 2's own channel-reuse design.
4. Stop rewriting genuine receive errors to `statusOk`.
5. Bind the receive goroutine to the stream it was created for — closes the stale-field race and the data race in the same change.

Plus a final review pass (test-gap and invariant hardening) and a small logging improvement that fell out of manual log inspection while verifying the fix.

**Out of scope**: batch retention/retransmission on reconnect, `Span/Sent` over-counting, gzip compression, `ConnectParams.Backoff` being shadowed on a genuinely dropped TCP connection (a different code path than this fix), the `handleSupportability` goroutine leak, and wiring `go test -race` into CI.

## Test plan

- [x] New/updated unit tests in `grpc_sender_test.go` and `trace_observer_test.go`: `TestServerClosesWithOK`, `TestSendAfterServerOkCloseGetsCorrectStatus`, `TestConnectDrainsStaleResponseError`, `TestConcurrentRecvOnReassignedStream`, `TestSendEofGracePeriodTimeout`, `TestOkCloseReconnectsWithoutBackoff`, updated `TestInvalidSpan`, plus `TestSpanBatchSenderCodeString`/`TestSpanBatchSenderStatusString` for the logging change.
- [x] `go test ./...`, `go vet ./...`, `gofmt -l` clean on the `daemon` module.
- [x] `go test -race` clean on `internal/newrelic/infinite_tracing` across repeated runs (not wired into CI — see out-of-scope above).
- [x] End-to-end before/after verification: two runs each of the pre-fix build (version-matched to the customer's reported `12.9.0.38`) and this branch, against a real Trace Observer test harness. Every pre-fix cycle (26/26 across both runs) loses a batch, stalls 15s, and misreports `Supportability/InfiniteTracing/Span/gRPC/UNKNOWN`; every post-fix cycle (40/40 across both runs) reconnects in under 1.2ms with zero loss and zero misclassification.

### DIFF
--- a/daemon/internal/newrelic/infinite_tracing/grpc_sender.go
+++ b/daemon/internal/newrelic/infinite_tracing/grpc_sender.go
@@ -127,7 +127,7 @@ func (s *grpcSpanBatchSender) connect() (error, spanBatchSenderStatus) {
 	// shutdown). A caller that reconnects while a prior stream is still
 	// live could still leave a push arriving after this drain -
 	// TestConcurrentRecvOnReassignedStream forces exactly that
-	// interleaving to prove Fix 4's stream-binding is what makes this
+	// interleaving to prove stream-binding is what makes this
 	// safe in practice.
 	for {
 		select {
--- a/daemon/internal/newrelic/infinite_tracing/grpc_sender.go
+++ b/daemon/internal/newrelic/infinite_tracing/grpc_sender.go
@@ -117,9 +117,18 @@ func (s *grpcSpanBatchSender) connect() (error, spanBatchSenderStatus) {
 	// Drain any status left over from a previous stream generation.
 	// s.responseError is never recreated across statusRestart/
 	// statusImmediateRestart reconnects (only clone(), on statusReconnect,
-	// gets a fresh one), so anything still buffered here can only be a
-	// leftover from a goroutine that has already terminated - the new
-	// generation's goroutine doesn't exist yet.
+	// gets a fresh one), and the new generation's goroutine doesn't exist
+	// yet, so anything still buffered here can only be a leftover push.
+	// That leftover is guaranteed to come from an already-terminated
+	// goroutine only because of a caller-side precondition, not anything
+	// intrinsic to connect(): the only caller, doStreaming() in
+	// trace_observer.go, calls connect() only after the previous stream
+	// has already terminated (a send error, a responseError push, or
+	// shutdown). A caller that reconnects while a prior stream is still
+	// live could still leave a push arriving after this drain -
+	// TestConcurrentRecvOnReassignedStream forces exactly that
+	// interleaving to prove Fix 4's stream-binding is what makes this
+	// safe in practice.
 	for {
 		select {
 		case <-s.responseError:
@@ -143,6 +152,9 @@ func (s *grpcSpanBatchSender) connect() (error, spanBatchSenderStatus) {
 	s.stream = stream
 
 	log.Debugf("connected to grpc endpoint %s", s.Host)
+	// stream is taken as a parameter (not a closure over the local above)
+	// so the binding is structural: nothing inside this goroutine can
+	// accidentally read s.stream, which is reassigned on every reconnect.
 	go func(stream v1.IngestService_RecordSpanBatchClient) {
 		for {
 			in, err := stream.Recv()
--- a/daemon/internal/newrelic/infinite_tracing/grpc_sender.go
+++ b/daemon/internal/newrelic/infinite_tracing/grpc_sender.go
@@ -156,14 +156,8 @@ func (s *grpcSpanBatchSender) connect() (error, spanBatchSenderStatus) {
 				return
 			default:
 				log.Errorf("unexpected error from grpc endpoint:  %v", err)
-				status := newSpanBatchStatusFromGrpcErr(err)
-				if status.code == statusShutdown || status.code == statusReconnect {
-					s.responseError <- status
-					return
-				} else {
-					status.code = statusOk
-					s.responseError <- status
-				}
+				s.responseError <- newSpanBatchStatusFromGrpcErr(err)
+				return
 			}
 		}
 	}()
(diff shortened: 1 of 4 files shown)


daemon/internal/newrelic/infinite_tracing/grpc_sender.go
```diff
@@ -114,6 +114,30 @@ func (s *grpcSpanBatchSender) clone() (spanBatchSender, error) {
...
+	// yet, so anything still buffered here can only be a leftover push.
+	// That leftover is guaranteed to come from an already-terminated
+	// goroutine only because of a caller-side precondition, not anything
+	// intrinsic to connect(): the only caller, doStreaming() in
+	// trace_observer.go, calls connect() only after the previous stream
+	// has already terminated (a send error, a responseError push, or
+	// shutdown). A caller that reconnects while a prior stream is still
+	// live could still leave a push arriving after this drain -
+	// TestConcurrentRecvOnReassignedStream forces exactly that
+	// interleaving to prove Fix 4's stream-binding is what makes this
```
bduranleau-nr: What is "Fix 4" referring to here?

lavarou: Fourth bullet point from pull request description:

> The receive goroutine read the reassignable s.stream field instead of the stream it was spawned for, letting a goroutine from a stale generation call Recv() concurrently with a newer generation's own goroutine — forbidden by grpc-go, and a plain data race.

Code change: bbe9d8a7cd3a14d8b0d873965fab78f7c23aa10e

daemon/internal/newrelic/infinite_tracing/grpc_sender_test.go
```diff
@@ -284,8 +342,20 @@ func TestInvalidSpan(t *testing.T) {
...
+	if status.code != statusRestart {
+		t.Fatalf("expected statusRestart on responseError, got %v", status.code)
+	}
+	if status.metric != "Supportability/InfiniteTracing/Span/gRPC/INTERNAL" {
+		t.Fatalf("expected the INTERNAL supportability metric, got %q", status.metric)
+	}
+
+	// The receive goroutine must push exactly once and return - not loop
+	// back into Recv() and refill the channel. Fix 5's drain at the top of
+	// connect() depends on every exit path behaving this way.
```
bduranleau-nr, line 354:
```suggestion
	// The receive goroutine must push exactly once and return - not loop
	// back into Recv() and refill the channel. The drain at the top of
	// connect() depends on every exit path behaving this way.
```

daemon/internal/newrelic/infinite_tracing/grpc_sender.go
```diff
@@ -114,6 +114,30 @@ func (s *grpcSpanBatchSender) clone() (spanBatchSender, error) {
...
+	// safe in practice.
```
bduranleau-nr, line 131:
```suggestion
	// live could still leave a push arriving after this drain -
	// TestConcurrentRecvOnReassignedStream forces exactly that
	// interleaving to prove stream-binding is what makes this
	// safe in practice.
```

daemon/internal/newrelic/infinite_tracing/grpc_sender_test.go
```diff
@@ -328,3 +398,270 @@ func TestErrToCodeString(t *testing.T) {
...
+	select {
+	case status := <-sender.responseError:
+		t.Fatalf("expected the stale status to have been drained, got %v", status)
+	default:
+	}
+}
+
+// guardedRecordSpanBatchClient wraps a real stream and flags it if Recv()
+// is ever called by more than one goroutine at a time - the exact
+// violation defect 4 makes possible once a stream generation is shared.
```
bduranleau-nr:
```suggestion
// is ever called by more than one goroutine at a time to validate 
// stream generation does not share streams between goroutines.
```

daemon/internal/newrelic/infinite_tracing/grpc_sender_test.go
```diff
@@ -328,3 +398,271 @@ func TestErrToCodeString(t *testing.T) {
...
+	// against the new goroutine.
+	select {
+	case status := <-sender.responseError:
+		t.Fatalf("expected the stale status to have been drained, got %v", status)
+	default:
+	}
+}
+
+// guardedRecordSpanBatchClient wraps a real stream and flags it if Recv()
+// is ever called by more than one goroutine at a time - the exact
```
lavarou:
```suggestion
```

bduranleau-nr (approved):

PR #1279, opened 2026-08-18
feat(agent): Guzzle 8 support

ZNeumann:
Changes the magic file for Guzzle 6+ to be one available in Guzzle 8. This file is also present in Guzzle 5, so the middleware installation has been moved to prevent 5 from installing it.

### DIFF
--- a/agent/lib_guzzle6.c
+++ b/agent/lib_guzzle6.c
@@ -438,75 +438,12 @@ end:
 NR_PHP_WRAPPER_END
 
 void nr_guzzle6_enable(TSRMLS_D) {
-  int retval;
-
   if (0 == NRINI(guzzle_enabled)) {
     return;
   }
 
-  /*
-   * Here's something new: we're going to evaluate PHP code to build our
-   * middleware in PHP, rather than doing it in C. This is mostly because it's
-   * fairly difficult to return a higher-order function from C; while possible,
-   * the code to do so is horrible enough that this actually feels cleaner.
-   *
-   * We'll do it when the library is detected because that should only happen
-   * once, but we'll also be careful to put guards around the function
-   * declaration just in case.
-   *
-   * On the bright side, zend_eval_string() effectively treats the string given
-   * as a standalone file, so we can use a normal namespace declaration to
-   * avoid possible clashes.
-   */
-  retval = zend_eval_string(
-      "namespace newrelic\\Guzzle6;"
-
-      "use Psr\\Http\\Message\\RequestInterface;"
-      "use GuzzleHttp\\Promise\\PromiseInterface;"
-
-      "if (!function_exists('newrelic\\Guzzle6\\middleware')) {"
-      "  function middleware(callable $handler) {"
-      "    return function (RequestInterface $request, array $options) use "
-      "($handler) {"
-
-      /*
-       * Start by adding the outbound CAT/DT/Synthetics headers to the request.
-       */
-      "      foreach (newrelic_get_request_metadata('Guzzle 6') as $k => $v) {"
-      "        $request = $request->withHeader($k, $v);"
-      "      }"
-
-      /*
-       * Set up the RequestHandler object and attach it to the promise so that
-       * we create an external node and deal with the CAT headers coming back
-       * from the far end.
-       */
-      "      $rh = new RequestHandler($request);"
-      "      $promise = $handler($request, $options);"
-      "      if (PromiseInterface::REJECTED == $promise->getState()) {"
-      /*
-                Special case for sync request. When sync requests is rejected,
-                onRejected callback is not called via `PromiseInterface::then`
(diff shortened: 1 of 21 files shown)


lavarou:
**issue (blocking):** `zend_eval_string()` for `newrelic\Guzzle6\middleware()` now runs on every `GuzzleHttp\Client::__construct()` instead of once per library detection

In `agent/lib_guzzle_common.c`, `nr_guzzle_client_construct()` calls `zend_eval_string()` to declare `newrelic\Guzzle6\middleware()`. Previously this eval lived in `nr_guzzle6_enable()` (`agent/lib_guzzle6.c`), which only runs once per library detection via `nr_execute_handle_library()`. Since the construct wrapper fires on every `new GuzzleHttp\Client(...)`, the ~40-line PHP snippet now gets re-lexed/parsed/compiled on **every** client object construction. The `function_exists()` check inside the eval'd string prevents redeclaring the function, but not the re-lex/parse/compile cost. Under any workload that constructs many Client instances — per-extenral-request client construction, clients built in a loop — overhead now scales linearly with instantiations instead of being paid once per process.

The codebase already has the right idiom for reducing this type of overhead this elsewhere: [`agent/lib_monolog.c:404-407`](https://github.com/newrelic/newrelic-php-agent/blob/R12.9/agent/lib_monolog.c#L407-L409) checks `nr_php_find_function(...)` in C before calling `zend_eval_string()`, so the eval only ever happens once. I suggest adding a C-level `nr_php_find_function()` guard around the eval call, matching the monolog pattern.

lavarou:
**issue (non-blocking):** eval failure now logs a warning per construction instead of once

Since the `zend_eval_string()` for `newrelic\Guzzle6\middleware()` now runs on every `GuzzleHttp\Client::__construct()`, if `zend_eval_string()` fails, `nrl_warning()` fires on `new GuzzleHttp\Client()` call rather than once at library-detection time. If `zend_eval_string()` fails once, it is likely that it will fail every subsequent time. This will result in noisy log flooding under high instantiation rates.

agent/php_execute.c
```diff
@@ -434,9 +434,15 @@ static nr_library_table_t libraries[] = {
...
     {"Doctrine 2", NR_PSTR("doctrine/orm/src/query.php"), nr_doctrine2_enable},
 
+    /*
+     * When Guzzle 5 support is removed, Guzzle 6+ magic files should be consolidated
+     * into just using client.php
+     */
     {"Guzzle 4-5", NR_PSTR("hasemitterinterface.php"), nr_guzzle4_enable},
     {"Guzzle 6", NR_PSTR("guzzle/src/functions_include.php"),
      nr_guzzle6_enable},
+    {"Guzzle 8", NR_PSTR("guzzle/src/client.php"),
```
zsistla: Could just have this say  "Guzzle" vs "Guzzle 8".  It'd probably be less confusing in the long run.

agent/lib_guzzle6.c
```diff
@@ -509,6 +513,23 @@ void nr_guzzle6_enable(TSRMLS_D) {
   }
 }
 
+void nr_guzzle8_enable(TSRMLS_D) {
+  /*
+   * The Guzzle 8 magic file is detected on all lower versions of Guzzle.
+   * The PHP version check prevents all Guzzle 5 from running the code,
+   * as Guzzle 8 and 5 are mutually exclusive in supported PHP versions.
+   * The middleware_func check ensures that for Guzzle 6-7 we only install
+   * the middleware a single time.
```
lavarou, line 522: **praise**: Very useful comment with nice explanation!

agent/lib_guzzle_common.c
```diff
@@ -246,6 +246,8 @@ char* nr_guzzle_response_get_header(const char* header,
 
 NR_PHP_WRAPPER_START(nr_guzzle_client_construct) {
   int is_guzzle_45 = 0;
+  int retval;
+
```
lavarou: **nit**: Remove unused variable.

agent/php_execute.c
lavarou: I agree with @zsistla

agent/lib_guzzle_common.c
ZNeumann: removed in [d67fcd9](https://github.com/newrelic/newrelic-php-agent/pull/1279/commits/d67fcd91d70f436ec6f8dfffcf094c8084d9a476)

agent/php_execute.c
ZNeumann: changed in [d67fcd9](https://github.com/newrelic/newrelic-php-agent/pull/1279/commits/d67fcd91d70f436ec6f8dfffcf094c8084d9a476)

lavarou (approved):

bduranleau-nr (approved):

PR #1299, opened 2026-09-16
chore: pin golang version in dev dockerfile

bduranleau-nr:
Pin the docker image tag used in the development Dockerfile to the same version used in the go.mod toolchain.
Prevent usage of unsupported golang version in development from breaking local daemon build & test.


files/Dockerfile
```diff
@@ -67,7 +67,7 @@ RUN apt-get install -y --no-install-recommends libpq-dev
 RUN docker-php-ext-install pgsql
 
 # install latest go to work with the daemon
-COPY --from=golang /usr/local/go /usr/local/go
+COPY --from=golang:1.26.6 /usr/local/go /usr/local/go
```
lavarou, line 70: **question (non-blocking)**: Instead of pinning go version have you considered decreasing severity of `GOLANG_VERIFICATION_FAILURE_SEVERITY` to `info`?  See [this comment](https://github.com/newrelic/newrelic-php-agent/blob/R12.10/Makefile#L170-L174), which explains effects of `GOLANG_VERIFICATION_FAILURE_SEVERITY`.
```suggestion
COPY --from=golang /usr/local/go /usr/local/go
ENV GOLANG_VERIFICATION_FAILURE_SEVERITY=info
```

lavarou (approved):

files/Dockerfile
bduranleau-nr: I wasn't aware we had an option to negate the warning, but I think I still prefer explicitly pinning the golang version to prevent building the daemon on a version of go that has an incompatibility or breaking change, such as the current Go 1.27 release which introduces failures into the daemon `TestCompressEncode` unit test. 
I think the error message & build failure is valuable for that reason and would personally only want to use the `GOLANG_VERIFICATION_FAILURE_SEVERITY=info` setting for special scenarios such as bumping the Go version.

lavarou (approved):

PR #1301, opened 2026-09-18
fix(agent): fall back to splitting $namespace to retrieve databaseName/collectionName for MongoDB

hahuja2:
MongoDB `1.21.4` and `2.4.1` introduced a breaking change where several Operation classes no longer store databaseName/collectionName as their own properties; the constructor only uses them to build a private `$namespace` string. So fall back to splitting that to retrieve databaseName/collectionName since `MongoDB\create_namespace()` forbids `'.'` in database names, so the first `'.'` unambiguously separates the two.


agent/lib_mongodb.c
```diff
@@ -160,6 +162,34 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
   }
 
+  /*
+   * As of mongodb/mongodb 1.21.4 and 2.4.1, several Operation classes no longer
+   * store databaseName/collectionName as their own properties; the constructor
+   * only uses them to build a private $namespace string. Fall back to splitting
+   * that when the direct lookups above come up empty.
+   * MongoDB\create_namespace() forbids '.' in database names, so the first '.'
+   * unambiguously separates the two.
+   */
```
zsistla: Can you add to the comment what the expected format of `namespace` is?

agent/lib_mongodb.c
```diff
@@ -160,6 +162,34 @@ NR_PHP_WRAPPER(nr_mongodb_operation) {
...
+  if ((NULL == params.collection) || (NULL == instance.database_name)) {
+    namespace
+        = nr_php_get_zval_object_property(this_var, "namespace" TSRMLS_CC);
+
+    if (nr_php_is_zval_valid_string(namespace)) {
```
zsistla: This can use `nr_php_is_zval_non_empty_string` since there's no point in extracting anything from an empty string.

agent/lib_mongodb.c
hahuja2: Updated: [5931d65](https://github.com/newrelic/newrelic-php-agent/pull/1301/commits/5931d65ec253cfdbc669cb5909bb01282697577c)

agent/lib_mongodb.c
hahuja2: Good point, updated: [e7cd13a](https://github.com/newrelic/newrelic-php-agent/pull/1301/commits/e7cd13adf89b6db284b5734b506e050d6eea5a4e)

PR #1306, opened 2026-09-22
feat(agent): add support for FrankenPHP worker mode

lavarou:
Execution of a FrankenPHP worker script up to `frankenphp_handle_request` call is recorded as a background transaction. This allows to monitor worker script bootstrap.

Callable passed to `frankenphp_handle_request` is wrapped with custom Zend Observer's fcall_begin and fcall_end callbacks. They're are used to start and end newrelic transactions. This way each request handled by the callable is recorded as a distinct web transaction.

This method of FrankenPHP worker mode instrumentation does not rely on internal implementation of FrankenPHP worker mode but rather relies on its public interface that must be implemented by any application using FrankenPHP worker mode.


agent/php_user_instrument.h
```diff
@@ -110,6 +110,14 @@ typedef struct _nruserfn_t {
 #if ZEND_MODULE_API_NO >= ZEND_7_4_X_API_NO
   char* wordpress_plugin_theme;
 #endif
+#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO
+  // Custom fcall_begin and fcall_end handlers. If set, they will be used
+  // instead of the default ones: nr_php_observer_fcall_begin and
+  // nr_php_observer_fcall_begin. This allows the agent to execute custom
```
lavarou, line 116:
```suggestion
  // nr_php_observer_fcall_end. This allows the agent to execute custom
```

agent/php_observer.c
```diff
@@ -125,22 +126,33 @@ static zend_observer_fcall_handlers nr_php_fcall_register_handlers(
...
-    ZEND_OP_ARRAY_EXTENSION(NR_OP_ARRAY,
-                            NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)) = wr;
+    zend_string* scope_name = OP_ARRAY_IS_A_METHOD(NR_OP_ARRAY) ? NR_OP_ARRAY->scope->name : NULL;
+    wr = nr_php_user_instrument_wraprec_hashmap_get(func_name, scope_name);
+    // store the wraprec in the op_array extension for the duration of the request for later lookup
+    ZEND_OP_ARRAY_EXTENSION(NR_OP_ARRAY, NR_PHP_PROCESS_GLOBALS(op_array_extension_handle)) = wr;
+  } else {
+    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
+      nrl_verbosedebug(NRL_INSTRUMENT, "%s - wraprec already installed", __func__);
+    }
```
lavarou, line 140:
```suggestion
  } else if (NR_USERFN_T_MAGIC == wr->magic) {
    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
      nrl_verbosedebug(NRL_INSTRUMENT, "%s - wraprec already installed", __func__);
    }
   } else {
    if (nrl_should_print(NRL_VERBOSEDEBUG, NRL_INSTRUMENT)) {
      nrl_verbosedebug(NRL_INSTRUMENT, "%s - wraprec is invalid", __func__);
    }
   }
```

agent/php_frankenphp.c
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+#include "php_frankenphp.h"
+#include "php_agent.h"
+#include "php_user_instrument.h"
+#include "php_wrapper.h"
+
+#ifdef ZTS
+
+static void nr_php_frankenphp_request_handler_fcall_begin(
```
zsistla: **non-blocker**
In practice we are only building 8.2+, but ZTS isn't strictly gated by a PHP version and all OAPI calls have been gated by something like `#if ZEND_MODULE_API_NO >= ZEND_8_0_X_API_NO``.

Would be cleaner to have something similar around any OAPI related calls where if ZTS and php >=8.2.

